PR #16, opened 2019-07-09
SC : Support mint/burn mode, approve/transferFrom, ERC721 URI

KimKyungup:
## Proposed changes
This PR enabled mint/burn mode, approve/transferFrom and ERC721 URI.

- Service chain can mint/burn ERC20/721 instead of charge/mint initially.
- Support 2-step deposit(approve, transferFrom) for ERC20/721 token without any modification of token contract.
- Support to migrate ERC721 uri during value transfer.

## Types of changes
- [ ] Bugfix
- [ ] Others

## Checklist


jeonghonah:
The `klaytn-node-tests/klay_bulk_transfer` scenario appears to fail in my local machine. PTAL.

contracts/bridge/Bridge.sol
```diff
@@ -23,12 +33,13 @@ contract Bridge is ITokenReceiver, INFTReceiver, Ownable {
...
-        TOKEN,
-        NFT
+        ERC20,
+        ERC721
     }
 
-    constructor () public payable {
+    constructor (bool _modeMintBurn) public payable {
         isRunning = true;
+        modeMintBurn = _modeMintBurn;
```
jeonghonah: Why don't we set it to the default value instead of the parameter?

contracts/bridge/Bridge.sol
```diff
@@ -99,10 +112,15 @@ contract Bridge is ITokenReceiver, INFTReceiver, Ownable {
     {
         require(handleNonce == _requestNonce, "mismatched handle / request nonce");
 
-        emit HandleValueTransfer(_to, TokenKind.TOKEN, _contractAddress, _amount, handleNonce);
+        emit HandleValueTransfer(_to, TokenKind.ERC20, _contractAddress, _amount, handleNonce);
         lastHandledRequestBlockNumber = _requestBlockNumber;
         handleNonce++;
-        IERC20(_contractAddress).transfer(_to, _amount);
+
+        if (modeMintBurn) {
```
jeonghonah: I think we should actually separate it into parent/child, so why don't we change the variable name to `modeMintBurn --> isParent`?

contracts/bridge/Bridge.sol
```diff
@@ -123,27 +141,33 @@ contract Bridge is ITokenReceiver, INFTReceiver, Ownable {
...
         require(handleNonce == _requestNonce, "mismatched handle / request nonce");
 
-        emit HandleValueTransfer(_to, TokenKind.NFT, _contractAddress, _uid, handleNonce);
+        emit HandleValueTransfer(_to, TokenKind.ERC721, _contractAddress, _uid, handleNonce);
         lastHandledRequestBlockNumber = _requestBlockNumber;
         handleNonce++;
-        IERC721(_contractAddress).safeTransferFrom(address(this), _to, _uid);
+
+        if (modeMintBurn) {
+            ERC721MetadataMintable(_contractAddress).mintWithTokenURI(_to, _uid, _tokenURI);
```
jeonghonah: `modeMintBurn` and `metaMintable` seem to mean different things, but do you have any reasons to treat them with the same variable condition `modeMintBurn`?

contracts/bridge/Bridge.sol
```diff
@@ -156,14 +180,14 @@ contract Bridge is ITokenReceiver, INFTReceiver, Ownable {
         require(allowedTokens[msg.sender] != address(0), "Not a valid token");
         require(_amount > 0, "zero amount");
 
-        emit RequestValueTransfer(TokenKind.TOKEN, _from, _amount, msg.sender, _to, requestNonce);
+        emit RequestValueTransfer(TokenKind.ERC20, _from, _amount, msg.sender, _to, requestNonce,"");
```
jeonghonah: Using `RequestValueTransfer()` in common may increases unnecessary arguement for KLAY and ERC20. How about add another method?

contracts/bridge/Bridge.sol
```diff
@@ -207,7 +235,56 @@ contract Bridge is ITokenReceiver, INFTReceiver, Ownable {
...
+    }
+
+    // requestERC721Transfer requests transfer ERC721 to _to on relative chain.
+    function requestERC721Transfer(address _contractAddress, address _to, uint256 _uid) external {
+        require(isRunning, "stopped bridge");
+        require(allowedTokens[_contractAddress] != address(0), "Not a valid token");
+
+        string memory uri = ERC721Metadata(_contractAddress).tokenURI(_uid);
+
+        IERC721(_contractAddress).transferFrom(msg.sender, address(this), _uid);
```
jeonghonah: The conditions are different from the ERC20, is there any particular reason?

contracts/bridge_test.go
```diff
@@ -123,7 +123,7 @@ func TestBridgeDeployWithKLAY(t *testing.T) {
 
 	chargeAmount := big.NewInt(10000000)
 	bridgeAccount.Value = chargeAmount
-	bridgeAddress, tx, _, err := bridge.DeployBridge(bridgeAccount, backend)
+	bridgeAddress, tx, _, err := bridge.DeployBridge(bridgeAccount, backend, false)
```
jeonghonah: You will also need to test the true case.

node/sc/vt_recovery_test.go
```diff
@@ -518,12 +518,12 @@ func prepare(t *testing.T, vtcallback func(*testInfo)) *testInfo {
...
 	for i := 0; i < testTxCount; i++ {
 		opts := &bind.TransactOpts{From: nodeAuth.From, Signer: nodeAuth.Signer, GasLimit: testGasLimit}
-		_, err = nftLocal.Register(opts, nodeAuth.From, big.NewInt(testNFT+int64(i)))
+		_, err = nftLocal.MintWithTokenURI(opts, nodeAuth.From, big.NewInt(testNFT+int64(i)), "testURI")
 		if err != nil {
 			log.Fatalf("Failed to Register NFT: %v", err)
 		}
 		opts = &bind.TransactOpts{From: chainAuth.From, Signer: chainAuth.Signer, GasLimit: testGasLimit}
-		_, err = nftRemote.Register(opts, remoteAddr, big.NewInt(testNFT+int64(i)))
+		_, err = nftRemote.MintWithTokenURI(opts, remoteAddr, big.NewInt(testNFT+int64(i)), "testURI")
```
jeonghonah: it's a minor comment. Shouldn't we change the URI string for each iteration?

contracts/bridge/Bridge.sol
KimKyungup: I think SP can want mintBurnMode for parent also. So I didn't name it like isChild.

contracts/bridge/Bridge.sol
KimKyungup: To mint token, the ERC721 should be mintable. Also This PR supports URI (metadata) migration. So when token is mint, URI is needed.

contracts/bridge/Bridge.sol
KimKyungup: This event is used in sub-bridge code. We can makde RequestKLAYTransfer(), RequestERC721Transfer(), RequestERC20Transfer(). But In this case, I think they share the same request nonce. So I think it is weird.
If we decide to separate the function, we can do in another PR. :)

contracts/bridge/Bridge.sol
KimKyungup: Right. In ERC721, there is not `burnFrom`. So this code was written.
In wemade code, they also used like this way.

contracts/bridge_test.go
KimKyungup: This TC is not related with the MintBurnMode. So I think it's ok.
But I also agree with your option which mintBurnMode should be tested.
In klaytn-node-test, I added the TC. And In unit test, I also modified deploy-test code. See `func (bm *BridgeManager) DeployBridgeTest(backend *backends.SimulatedBackend, local bool) `. :)

node/sc/vt_recovery_test.go
KimKyungup: Generally, all ERC721 token has different URI.

KimKyungup:
> The `klaytn-node-tests` appears to fail in my local machine. PTAL.

https://github.com/ground-x/klaytn-node-tests/pull/73 can test node-test with this PR. :)

jeonghonah:
> > The `klaytn-node-tests` appears to fail in my local machine. PTAL.
> 
> [ground-x/klaytn-node-tests#73](https://github.com/ground-x/klaytn-node-tests/pull/73) can this PR in klaytn-node-test. :)
> However, caverjs test doesn't work. I need your help for this.

Thanks. It passed on my local machine. I'll fix the caver-js test.

KimKyungup:
@andybclee @jeonghonah PTAL again. I rebased this with master.

KimKyungup:
@andybclee The rebasing without any modification staled your review. So we merge this PR. :-)

ehnuje:
@KimKyungup @jeonghonah @andybclee As we talked f2f, let me merge this PR for now for upcoming PRs.

PR #29, opened 2019-07-16
SC : Add value transfer Fee APIs

KimKyungup:
## Proposed changes
This PR added value transfer fee APIs like below.
- set/get Fee of KLAY/ERC20
- set/get Fee receiver
- `getParentTransactionReceipt` to check if the request was succeeded.

## Types of changes
- [ ] Bugfix
- [ ] Others

## Checklist

- [ x Lint and unit tests pass locally with my changes (`$ make test`)

### DIFF
--- a/node/sc/api_bridge.go
+++ b/node/sc/api_bridge.go
@@ -17,7 +17,9 @@
 package sc
 
 import (
+	"context"
 	"fmt"
+	"github.com/klaytn/klaytn/accounts/abi/bind"
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/contracts/bridge"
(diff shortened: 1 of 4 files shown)


node/sc/api_bridge.go
```diff
@@ -316,6 +319,35 @@ func (sbapi *SubBridgeAPI) RegisterToken(cBridgeAddr, pBridgeAddr, cTokenAddr, p
 	return nil
 }
 
+func (sbapi *SubBridgeAPI) GetParentTransactionReceipt(txHash common.Hash) (*types.Receipt, error) {
```
jeonghonah: It looks like deadcode.

KimKyungup: You are so fast. :) Thanks.

KimKyungup:
> Except for a minor comment, LGTM.
> 
> (Note)
> We've added APIs for go client, but I'm not sure if it's necessary to provide the APIs because we can already use contact call. I think we need to consider the future direction of improvement.

Right. It's good point. When I add the APIs, I think they are need because for HA, some important method should be called by signers. So I think they should be included in subbridge.api to support multple signer.

PR #31, opened 2019-07-17
Revert "Prevent voting except add, remove validator (#3565)"

andybclee:
## Proposed changes

- This PR is to revert the commit `319876f74ab926c780e79114647202a0a9b124d4` which forbid voting.
- This is the part of applying dev-governance branch

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

## Related issues

- https://github.com/ground-x/klaytn/pull/3565

### DIFF
--- a/governance/default.go
+++ b/governance/default.go
@@ -68,21 +68,6 @@ var (
 		"istanbul.policy":               params.Policy,
 		"reward.stakingupdateinterval":  params.StakeUpdateInterval,
 		"reward.proposerupdateinterval": params.ProposerRefreshInterval,
-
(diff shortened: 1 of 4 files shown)


governance/default.go
```diff
@@ -150,8 +135,7 @@ type VoteStatus struct {
 }
 
 type Governance struct {
-	ChainConfig    *params.ChainConfig
-	governanceLock sync.RWMutex
```
kjhman21: Don't we need this lock?

andybclee: This lock was temporarily added. And the data race problem can't be solved in this way after enabling the voting feature. In following PRs, each object which can be accessed from multiple callers will have its own lock.

PR #37, opened 2019-07-18
SC: change the max peer of the main and sub bridges to one

jeonghonah:
## Proposed changes

- When configuring the service chain, main-bridge and sub-bridge are limited to single pair only.
- This is to simplify HA implementation.

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

## Related issues

n/a

## Further comments

n/a

### DIFF
--- a/node/sc/config.go
+++ b/node/sc/config.go
@@ -43,12 +43,6 @@ const (
 
 var logger = log.NewModuleLogger(log.ServiceChain)
 
-// DefaultConfig contains default settings for use on the Klaytn main net.
-var DefaultConfig = SCConfig{
-	NetworkId: 1,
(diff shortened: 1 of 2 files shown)


node/sc/config.go
```diff
@@ -46,7 +46,7 @@ var logger = log.NewModuleLogger(log.ServiceChain)
 // DefaultConfig contains default settings for use on the Klaytn main net.
 var DefaultConfig = SCConfig{
 	NetworkId: 1,
-	MaxPeer:   50,
+	MaxPeer:   1, // Only a single main-bridge and sub-bridge pair is allowed.
```
nogang: If the two config values ​​should always match, it would be better to share one default value.

PR #38, opened 2019-07-18
SC: Support FeeLimit of value transferring.

KimKyungup:
## Proposed changes
This PR enables the FeeLimit for value transferring in Service Chain.
And related TCs are added.

**Before**
If the input fee in request is not equal with designated fee, the request is reverted.

**After**
If the input feeLimit is designated fee or more, the request can be processed.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


contracts/bridge/Bridge.sol
```diff
@@ -172,45 +172,43 @@ contract Bridge is IERC20BridgeReceiver, IERC721BridgeReceiver, Ownable, BridgeF
...
     function () external payable {
-        _requestKLAYTransfer(msg.sender, feeOfKLAY);
+        _requestKLAYTransfer(msg.sender, msg.value);
     }
 
     // requestKLAYTransfer requests transfer KLAY to _to on relative chain.
-    function requestKLAYTransfer(address _to, uint256 _fee) external payable {
-        _requestKLAYTransfer(_to, _fee);
+    function requestKLAYTransfer(address _to, uint256 _feeLimit) external payable {
+        _requestKLAYTransfer(_to, _feeLimit);
```
KimKyungup: Dear reviewer,
I would like to know what do you think if `_amount` is needed in this method?
`_amount` is same with `msg.value` - `_feeLimit`.
`_amount` means how much KLAY will be transfered to `_to`.

In this transaction, user writes msg.value and `_feeLimit`. So `_amount` can be calculated. 
But To user, would it be better UX to have `_amount` written.

kjhman21: Then, what if real fee is smaller than _feeLimit? How do we handle remaining KLAYs?

kjhman21: Is this the case used when we send some KLAYs from the service chain to the main chain?

KimKyungup: This code is used for both ways (SC -> MC, MC -> SC). :)

KimKyungup:
@aidan-kwon @jeonghonah @andybclee PTAL. :)

contracts/bridge/BridgeFee.sol
```diff
@@ -19,15 +19,29 @@ contract BridgeFee {
...
-        if(isValidReceiver() == true && _fee > 0) {
-            receiver.transfer(_fee);
+    function _payKLAYFeeAndRefundChange(uint256 _feeLimit) internal {
+        uint256 fee = feeOfKLAY;
+
+        if(isValidReceiver() == true && fee > 0) {
+            require(_feeLimit >= fee, "invalid feeLimit");
+
+            receiver.transfer(fee);
+            msg.sender.transfer(_feeLimit.sub(fee));
```
jeonghonah: If _feeLimit == fee, this transfer send zero value. Is it O.K.?

contracts/bridge/Bridge.sol
jeonghonah: @KimKyungup I think that `_amount` is not necessary.

contracts/bridge/BridgeFee.sol
```diff
@@ -19,15 +19,29 @@ contract BridgeFee {
...
+        } else {
+            msg.sender.transfer(_feeLimit);
```
jeonghonah: How about simply reverting the tx?

node/sc/bridge_manager_test.go
```diff
@@ -255,15 +255,7 @@ func TestBridgeManager(t *testing.T) {
...
-		// TODO-Klaytn-Servicechain needs to support WaitMined
-		//timeoutContext, cancelTimeout := context.WithTimeout(context.Background(), 10*time.Second)
-		//defer cancelTimeout()
-		//
-		//receipt, err := bind.WaitMined(timeoutContext, sim, tx)
-		//if err != nil {
-		//	log.Fatal("Failed to WithdrawToken.", "err", err, "txHash", tx.Hash().String(),"status",receipt.Status)
-		//}
-		//fmt.Println("WithdrawToken is executed.", "addr", addr.String(), "txHash", tx.Hash().String())
+		CheckReceipt(sim, tx, 1*time.Second, types.ReceiptStatusSuccessful, t)
```
jeonghonah: Is 1 sec enough to wait for the receipt?

aidan-kwon:
I don't understand why the fee-mechanism takes "pay maximum, and refund change" approach. 
Isn't it simple transferring just the amount of fee at the first transfer? 
@KimKyungup @jeonghonah

contracts/bridge/Bridge.sol
kjhman21: For the case of MC -> SC, wouldn't it be double paid due to the transaction fee?

jeonghonah:
@KimKyungup FYI

`SC : `Support FeeLimit of value transferring --> `SC:` Support FeeLimit of value transferring

contracts/bridge/Bridge.sol
KimKyungup: As I told you in offline. It's right. :)
Also, as we talked, I will change _feeLimit -> _amount. Thanks. @kjhman21 @aidan-kwon .

contracts/bridge/BridgeFee.sol
KimKyungup: if isValidReceiver() == falise, it means there is no receiver. So in this case, it should work with zero fee.

node/sc/bridge_manager_test.go
KimKyungup: Yes, it works on the simulated backend. During `sim.Commit()` the receipt can be generated. :)

jeonghonah:
@KimKyungup Does this PR pass klaytn-node-tests now?

contracts/bridge/BridgeFee.sol
KimKyungup: I considered checking zero value. But it can save a little gas.
For simplify code, I would like to stay this code. 
I will compare the used gas, later.

KimKyungup:
@kjhman21 @aidan-kwon @jeonghonah PTAL. :) I reflected your review.

KimKyungup:
> I don't understand why the fee-mechanism takes "pay maximum, and refund change" approach.
> Isn't it simple transferring just the amount of fee at the first transfer?
> @KimKyungup @jeonghonah

As I told you, user can request ERC20 token transfer in token contract.
In this case, in the request method, designated fee should be needed to transfer token to bridge contract. For that ERC20 need the `getFee` interface. 
So I decided pay and refund structure.

contracts/bridge/BridgeFee.sol
```diff
@@ -19,16 +19,36 @@ contract BridgeFee {
...
     }
 
-    function _payKLAYFee(uint256 _fee) internal {
-        if(isValidReceiver() == true && _fee > 0) {
-            receiver.transfer(_fee);
+    function _payKLAYFeeAndRefundChange(uint256 _feeLimit) internal returns(uint256) {
+        uint256 fee = feeOfKLAY;
+
+        if(isValidReceiver() == true && fee > 0) {
+            require(_feeLimit >= fee, "invalid feeLimit");
```
kjhman21:
```suggestion
            require(_feeLimit >= fee, "insufficient feeLimit");
```

contracts/bridge/BridgeFee.sol
```diff
@@ -19,16 +19,36 @@ contract BridgeFee {
...
     }
 
-    function _payERC20Fee(address _token, uint256 _fee) internal {
-        if(isValidReceiver() == true && _fee > 0) {
-            IERC20(_token).transfer(receiver, _fee);
+    function _payERC20FeeAndRefundChange(address from, address _token, uint256 _feeLimit) internal returns(uint256){
+        uint256 fee = feeOfERC20[_token];
+
+        if (isValidReceiver() == true && fee > 0) {
+            require(_feeLimit >= fee, "invalid feeLimit");
```
kjhman21:
```suggestion
            require(_feeLimit >= fee, "insufficient feeLimit");
```

KimKyungup:
> @KimKyungup Does this PR pass klaytn-node-tests now?

Yes. I will make PR for new fee limit feature in klaytn-node-tests.
```
--- PASS: TestERC20TransferWithFee (156.35s)
PASS
ok      github.com/ground-x/klaytn-node-tests/tests     1106.825s
./build/stop.sh
########## Stop Nodes ##########
No matching processes belonging to you were found
./build/clean.sh
########## Clean Node Data ##########
➜  klaytn-node-tests git:(190717a_sc_fee_tc) ✗ ksen version
Klaytn v1.0.0+77da1914e1

```

contracts/bridge_test.go
```diff
@@ -97,7 +97,7 @@ func TransferSignedTx(auth *bind.TransactOpts, backend *backends.SimulatedBacken
 
 // RequestKLAYTransfer send a requestValueTransfer transaction to the bridge contract.
```
jeonghonah: Minor comment:
RequestKLAYTransfer `send` --> RequestKLAYTransfer `sends`

contracts/bridge_test.go
```diff
@@ -97,7 +97,7 @@ func TransferSignedTx(auth *bind.TransactOpts, backend *backends.SimulatedBacken
...
 func RequestKLAYTransfer(b *bridge.Bridge, auth *bind.TransactOpts, to common.Address, value uint64, t *testing.T) {
-	_, err := b.RequestKLAYTransfer(&bind.TransactOpts{From: auth.From, Signer: auth.Signer, GasLimit: gasLimit, Value: big.NewInt(1)}, to, big.NewInt(0))
+	_, err := b.RequestKLAYTransfer(&bind.TransactOpts{From: auth.From, Signer: auth.Signer, GasLimit: gasLimit, Value: new(big.Int).SetUint64(value)}, to, new(big.Int).SetUint64(value))
 	if err != nil {
 		t.Fatalf("fail to DepositKLAY %v", err)
```
jeonghonah: fail to DepositKLAY --> fail to RequestKLAYTransfer

contracts/bridge/Bridge.sol
```diff
@@ -172,45 +172,43 @@ contract Bridge is IERC20BridgeReceiver, IERC721BridgeReceiver, Ownable, BridgeF
...
-            _fee
+            fee
         );
         requestNonce++;
     }
 
     // () requests transfer KLAY to msg.sender address on relative chain.
     function () external payable {
-        _requestKLAYTransfer(msg.sender, feeOfKLAY);
+        _requestKLAYTransfer(msg.sender, msg.value);
```
kjhman21: I'm not sure this is right. If we call like this, msg.value == _feeLimit in _requestKLAYTransfer().
In this case, msg.value.sub(_feeLimit) == 0. Is this intentional?

KimKyungup: In this case, I think designated fee can be input as feeLimit.

contracts/bridge/BridgeFee.sol
```diff
@@ -19,16 +19,36 @@ contract BridgeFee {
...
-    function _payKLAYFee(uint256 _fee) internal {
-        if(isValidReceiver() == true && _fee > 0) {
-            receiver.transfer(_fee);
+    function _payKLAYFeeAndRefundChange(uint256 _feeLimit) internal returns(uint256) {
+        uint256 fee = feeOfKLAY;
+
+        if(isValidReceiver() == true && fee > 0) {
+            require(_feeLimit >= fee, "insufficient feeLimit");
+
+            receiver.transfer(fee);
```
kjhman21: The word `receiver` looks weird to me. If it receives transaction fee, what about using:

- feeReceiver
- bridgeOperator

or something else?

KimKyungup: `feeReceiver` looks good. I will change.

contracts/bridge/BridgeFee.sol
```diff
@@ -19,16 +19,36 @@ contract BridgeFee {
...
+        return 0;
     }
 
-    function _payERC20Fee(address _token, uint256 _fee) internal {
-        if(isValidReceiver() == true && _fee > 0) {
-            IERC20(_token).transfer(receiver, _fee);
+    function _payERC20FeeAndRefundChange(address from, address _token, uint256 _feeLimit) internal returns(uint256){
+        uint256 fee = feeOfERC20[_token];
+
+        if (isValidReceiver() == true && fee > 0) {
```
kjhman21: What happens if the receiver is invalid? It just returns the feeLimit back to the sender?
What is the meaning of the invalid receiver since the receiver should be set in the constructor.?

contracts/bridge/Bridge.sol
jeonghonah: `_requestKLAYTransfer` requires `msg.value > _feeLimit `. Is it always reverted? It will be nice to add a test case for the payable method.

contracts/bridge/BridgeFee.sol
KimKyungup: if there is no feeReceiver, there is no fee. So feeLimit will be refund. 
For now, if isValidReceiver is false, feeReceiver is not set or set zero address.

contracts/bridge/BridgeFee.sol
KimKyungup: reflected :)

contracts/bridge/Bridge.sol
KimKyungup: I reverted it and added a test case for empty payable method. :) Thanks.

contracts/bridge/Bridge.sol
```diff
@@ -193,7 +193,7 @@ contract Bridge is IERC20BridgeReceiver, IERC721BridgeReceiver, Ownable, BridgeF
 
     // () requests transfer KLAY to msg.sender address on relative chain.
     function () external payable {
-        _requestKLAYTransfer(msg.sender, msg.value);
+        _requestKLAYTransfer(msg.sender, feeOfKLAY);
```
jeonghonah: If the user is not aware of fee, it may be confusing why the full amount does not sent. How about removing this payable method?

KimKyungup: () payable method is so simple and easy. But as you mentioned, the transferred amount is msg.value - fee. I would like to keep this easy way to transfer. :)

contracts/bridge/BridgeFee.sol
kjhman21: What is the meaning of "invalid"? only "0" is invalid? What about that case that the user set the wrong fee receiver address? Do we have a way to check "validness" of the address?

It seems that the fee receiver address is set when the constructor is called. This means there is no chance that the fee receiver address is not set. In that case, do we really need to check the "validness" of the address of the fee receiver?

contracts/bridge/Bridge.sol
```diff
@@ -199,18 +197,18 @@ contract Bridge is IERC20BridgeReceiver, IERC721BridgeReceiver, Ownable, BridgeF
     }
 
     // requestKLAYTransfer requests transfer KLAY to _to on relative chain.
-    function requestKLAYTransfer(address _to, uint256 _fee) external payable {
-        _requestKLAYTransfer(_to, _fee);
+    function requestKLAYTransfer(address _to, uint256 _amount) external payable {
+        uint256 feeLimit = msg.value.sub(_amount);
```
aidan-kwon: How about moving this logic to `_requestKLAYTransfer` modifying its parameter type.

KimKyungup: It is just trade off. :) Look 196 line. For () payable function, we need to calculate _amount. T.T

KimKyungup: Anyway, I will change and compare. =)

KimKyungup: If it is applied like below, it need one line more. :)
image

KimKyungup: As we talked, both ways are similar. Let's keep this code. :) Thanks.

contracts/bridge/BridgeFee.sol
KimKyungup: For now, via BridgeFee constructor, Bridge's constructor set the `address(0)`.
After that, user (SP) can set feeReceiver.

How about change code like below?=)
1. `isValidReceiver() == true` ->`feeReceiver != address(0)`

kjhman21: As we discussed, the change is up to you. I just wanted to know the logic and the meaning of "validness". :)

PR #43, opened 2019-07-19
Changed GovernanceVotes GovernanceTallies and GovernanceSet

andybclee:
## Proposed changes

- This PR includes three PRs which were reverted before. Former review and discussions can be found in the following links
- These changes were for removing data race by accessing governance data using locks
- https://github.com/ground-x/klaytn/pull/3502
- https://github.com/ground-x/klaytn/pull/3516
- https://github.com/ground-x/klaytn/pull/3523

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/cmd/utils/nodecmd/chaincmd.go
+++ b/cmd/utils/nodecmd/chaincmd.go
@@ -119,7 +119,7 @@ func initGenesis(ctx *cli.Context) error {
 	}
 
 	data := getGovernanceItemsFromGenesis(genesis)
-	gbytes, err := json.Marshal(data)
+	gbytes, err := json.Marshal(data.Items())
 	if err != nil {
 		logger.Crit("Failed to json marshaling governance data", "err", err)
 	}
@@ -152,7 +152,7 @@ func initGenesis(ctx *cli.Context) error {
 }
 
 func getGovernanceItemsFromGenesis(genesis *blockchain.Genesis) governance.GovernanceSet {
-	g := make(governance.GovernanceSet)
+	g := governance.NewGovernanceSet()
 
 	if genesis.Config.Governance != nil {
 		governance := genesis.Config.Governance
--- a/governance/handler.go
+++ b/governance/handler.go
@@ -242,7 +242,7 @@ func checkAddress(k string, v interface{}) bool {
 	return true
 }
 
-func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes []GovernanceVote, tally []GovernanceTally, header *types.Header, proposer common.Address, self common.Address) (istanbul.ValidatorSet, []GovernanceVote, []GovernanceTally) {
+func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes []GovernanceVote, tally []GovernanceTallyItem, header *types.Header, proposer common.Address, self common.Address) (istanbul.ValidatorSet, []GovernanceVote, []GovernanceTallyItem) {
 	gVote := new(GovernanceVote)
 
 	if len(header.Vote) > 0 {
@@ -294,7 +294,7 @@ func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes 
 			votes = append(votes, *gVote)
 
 			// Tally up the new vote. This will be cleared when Epoch ends.
-			// Add to GovernanceTally if it doesn't exist
+			// Add to GovernanceTallies if it doesn't exist
 			valset, votes, tally = gov.addNewVote(valset, votes, tally, gVote, governanceMode, governingNode, number)
 
 			// If this vote was casted by this node, remove it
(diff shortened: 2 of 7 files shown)


governance/handler.go
```diff
@@ -445,8 +441,7 @@ func (gov *Governance) GetGovernanceItemAtNumber(num uint64, key string) (interf
 	}
 }
 
-func (gov *Governance) GetLatestGovernanceItem(key string) interface{} {
-	gov.currentSetMu.RLock()
-	defer gov.currentSetMu.RUnlock()
-	return gov.currentSet[key]
+func (gov *Governance) GetLatestGovernanceItem(key int) interface{} {
+	ret, _ := gov.currentSet.GetValue(key)
```
jeonghonah: `GetValue` holds the own lock, but the function checks existence in GovernanceKeyMapReverse without a  lock. Is there no race in the method?

```
func (gs *GovernanceSet) GetValue(key int) (interface{}, bool) {
	sKey, ok := GovernanceKeyMapReverse[key]
	if !ok {
		return nil, false
	}
...
```

cmd/utils/nodecmd/chaincmd.go
```diff
@@ -119,7 +119,7 @@ func initGenesis(ctx *cli.Context) error {
 	}
 
 	data := getGovernanceItemsFromGenesis(genesis)
-	gbytes, err := json.Marshal(data)
+	gbytes, err := json.Marshal(data.Items())
```
jeonghonah: I think that this line has made all PRs reverted. How about add a unit test for this modification?

governance/handler.go
andybclee: @jeonghonah GovernanceKeyMapReverse is a map which can't be written online. So all access to the map is "READ" and no race can happen on it.

cmd/utils/nodecmd/chaincmd.go
andybclee: @jeonghonah Actually there are two tests related with this. 
In default_test.go in governance folder, following two tests had added.
These tests will check if the generated genesis hash is same as that of Baobab and Cypress. So if the marshaled string is wrong, its overall hash will also be wrong and won't be able to pass the tests.

```go
func TestBaoBabGenesisHash(t *testing.T) {
	baobabHash := common.HexToHash("0xe33ff05ceec2581ca9496f38a2bf9baad5d4eed629e896ccb33d1dc991bc4b4a")
	genesis := blockchain.DefaultBaobabGenesisBlock()
	genesis.Governance = blockchain.SetGenesisGovernance(genesis)
	blockchain.InitDeriveSha(genesis.Config.DeriveShaImpl)

	db := database.NewMemoryDBManager()
	block, _ := genesis.Commit(db)
	if block.Hash() != baobabHash {
		t.Errorf("Generated hash is not equal to Baobab's hash. Want %v, Have %v", baobabHash.String(), block.Hash().String())
	}
}

func TestCypressGenesisHash(t *testing.T) {
	cypressHash := common.HexToHash("0xc72e5293c3c3ba38ed8ae910f780e4caaa9fb95e79784f7ab74c3c262ea7137e")
	genesis := blockchain.DefaultGenesisBlock()
	genesis.Governance = blockchain.SetGenesisGovernance(genesis)
	blockchain.InitDeriveSha(genesis.Config.DeriveShaImpl)

	db := database.NewMemoryDBManager()
	block, _ := genesis.Commit(db)
	if block.Hash() != cypressHash {
		t.Errorf("Generated hash is not equal to Baobab's hash. Want %v, Have %v", cypressHash.String(), block.Hash().String())
	}
}
```

governance/handler.go
```diff
@@ -305,10 +305,8 @@ func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes
 			logger.Warn("Received Vote was invalid", "number", header.Number, "Validator", gVote.Validator, "key", gVote.Key, "value", gVote.Value)
 		}
 		if number > atomic.LoadUint64(&gov.lastGovernanceStateBlock) {
-			gov.GovernanceVotes = make([]GovernanceVote, len(votes))
-			gov.GovernanceTally = make([]GovernanceTally, len(tally))
-			copy(gov.GovernanceVotes, votes)
-			copy(gov.GovernanceTally, tally)
+			gov.GovernanceVotes.Import(votes)
```
jeonghonah: The `Import()` method might be confusing. How about this change?

- Import() --> Copy()
- Copy() --> MakeCopy()

governance/handler.go
```diff
@@ -341,10 +339,8 @@ func (gov *Governance) removePreviousVote(valset istanbul.ValidatorSet, votes []
 			ret = append(votes[:idx], votes[idx+1:]...)
 			if gov.isGovernanceModeSingleOrNone(governanceMode, governingNode, gVote.Validator) ||
 				(governanceMode == params.GovernanceMode_Ballot && currentVotes <= valset.TotalVotingPower()/2) {
-				if v, ok := gov.changeSet[vote.Key]; ok && v == vote.Value {
-					gov.mu.Lock()
-					delete(gov.changeSet, vote.Key)
-					gov.mu.Unlock()
+				if v, ok := gov.changeSet.GetValue(GovernanceKeyMap[vote.Key]); ok && v == vote.Value {
```
jeonghonah: If there is not a global lock, there might be a race between GetValue() and RemoveItem().

For example,
1) goroutine A: GetValue() is ok and v == vote.Value
2) goroutine B: Updated by SetValue()
3) goroutine A: RemoveItem() --> wrong deletion ( v != vote.Value)

andybclee: @jeonghonah I think this is a good point and it is revealing another issue. Let's have some discussion f2f.

andybclee:
> Except for another concurrency issue, LGTM.

It is recorded on the issue #58 
Will be fixed after current merging process is done.

PR #61, opened 2019-07-24
Make governance variables to be atomic

andybclee:
## Proposed changes
- This PR includes following PRs which were reverted before. To remove data race, several variables in the governance object have been changed to be atomic.
- Add missing atomic operation for lastGovernanceStateBlock https://github.com/ground-x/klaytn/pull/3517
- Make nodeAddress as an atomic.Value https://github.com/ground-x/klaytn/pull/3524
- Add idxCacheLock to protect idxCache https://github.com/ground-x/klaytn/pull/3526

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/api.go
+++ b/governance/api.go
@@ -89,7 +89,7 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 	gMode := api.governance.ChainConfig.Governance.GovernanceMode
 	gNode := api.governance.ChainConfig.Governance.GoverningNode
 
-	if GovernanceModeMap[gMode] == params.GovernanceMode_Single && gNode != api.governance.nodeAddress {
+	if GovernanceModeMap[gMode] == params.GovernanceMode_Single && gNode != api.governance.nodeAddress.Load().(common.Address) {
 		return "", errPermissionDenied
 	}
 	if strings.ToLower(key) == "governance.removevalidator" {
@@ -113,7 +113,7 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 func (api *PublicGovernanceAPI) isRemovingSelf(val interface{}) bool {
 	target := val.(string)
 
-	if common.HexToAddress(target) == api.governance.nodeAddress {
+	if common.HexToAddress(target) == api.governance.nodeAddress.Load().(common.Address) {
 		return true
 	} else {
 		return false
@@ -195,7 +195,7 @@ func (api *PublicGovernanceAPI) ChainConfig() *params.ChainConfig {
 }
 
 func (api *PublicGovernanceAPI) NodeAddress() common.Address {
-	return api.governance.nodeAddress
+	return api.governance.nodeAddress.Load().(common.Address)
 }
 
 func (api *PublicGovernanceAPI) isGovernanceModeBallot() bool {
--- a/governance/default.go
+++ b/governance/default.go
@@ -564,9 +564,7 @@ func (g *Governance) initializeCache() error {
 	if err != nil {
 		return ErrNotInitialized
 	}
-	g.idxCacheLock.Lock()
 	g.idxCache = indices
-	g.idxCacheLock.Unlock()
 
 	// Put governance items into the itemCache
 	for _, v := range indices {
--- a/governance/default.go
+++ b/governance/default.go
@@ -153,16 +153,17 @@ type Governance struct {
 	voteMap     map[string]VoteStatus
 	voteMapLock sync.RWMutex
 
-	nodeAddress      common.Address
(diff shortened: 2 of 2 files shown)


governance/default.go
```diff
@@ -605,13 +617,6 @@ func getGovernanceCacheKey(num uint64) common.GovernanceCacheKey {
 	return common.GovernanceCacheKey(params.GovernanceCachePrefix + "_" + v)
 }
 
-func (g *Governance) addIdxCache(num uint64) {
```
ehnuje: Why did you remove this function?

andybclee: @ehnuje As far as I can remember, there was an issue for adding a cache for test code after idxCacheLock was added. It looks like it could be better. So I revived addIdxCache function and make AddGovernanceCacheForTest to use `addGovernanceCache` while not accessing `addIdxCache` directly.

governance/default.go
```diff
@@ -927,7 +938,5 @@ func AddGovernanceCacheForTest(g *Governance, num uint64, config *params.ChainCo
 	// Don't update cache if num (block number) is smaller than the biggest number of cached block number
 
 	data := getGovernanceItemsFromChainConfig(config)
-	cKey := getGovernanceCacheKey(num)
-	g.itemCache.Add(cKey, data.Items())
```
ehnuje: In the last commit, 930 and 931 has been removed, is it intended?

governance/api.go
```diff
@@ -89,7 +89,7 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 	gMode := api.governance.ChainConfig.Governance.GovernanceMode
 	gNode := api.governance.ChainConfig.Governance.GoverningNode
 
-	if GovernanceModeMap[gMode] == params.GovernanceMode_Single && gNode != api.governance.nodeAddress {
+	if GovernanceModeMap[gMode] == params.GovernanceMode_Single && gNode != api.governance.nodeAddress.Load().(common.Address) {
```
jimni1222: Can 'NodeAddress()' function be used in here? 
like "gNode != api.NodeAddress()"

governance/default.go
```diff
@@ -591,6 +596,9 @@ func (g *Governance) getGovernanceCache(num uint64) (map[string]interface{}, boo
 
 func (g *Governance) addGovernanceCache(num uint64, data GovernanceSet) {
```
jeonghonah: Calling `addIdxCache()` without locking might be dangerous. How about simply embedding the `addIdxCache()` to the call site?

governance/api.go
andybclee: @jimni1222 Yes, but not now. In following PRs, there is a change in the function 'NodeAddress()' to use load(). I'll apply your suggestion after applying it.

governance/default.go
andybclee: @jeonghonah The function was called by two functions before, but after fixing it now only one function is calling it. So as you suggested, it can be merged with the caller.
Thank you for your great feedback. =)

governance/default.go
andybclee: @ehnuje Yes, it was intended. Because now the function calls g.addGovernanceCache(), those two removed lines is being done in the addGovernanceCache().

governance/default.go
```diff
@@ -562,7 +564,10 @@ func (g *Governance) initializeCache() error {
 	if err != nil {
 		return ErrNotInitialized
 	}
+	g.idxCacheLock.Lock()
 	g.idxCache = indices
+	g.idxCacheLock.Unlock()
+
 	// Put governance items into the itemCache
 	for _, v := range indices {
```
laggu: although it is same, would it be better use g.idxCache instead of indices?

andybclee: @laggu Good point. Thanks!

ehnuje: Isn't locking needed here? (at least RLock())

andybclee: @ehnuje initializeCache is called only once when a node starts while calling `NewGovernance()`. So I don't think a locking is needed for `indices`.

andybclee: @ehnuje I removed lock on `g.idxCacheLock` in `initializeCache()` because it is redundant as we discussed. Please take a look.

andybclee:
@KimKyungup @jeonghonah @nogang @ehnuje @laggu @jimni1222 @aidan-kwon Please take a look. =)

governance/default.go
```diff
@@ -563,6 +565,7 @@ func (g *Governance) initializeCache() error {
 		return ErrNotInitialized
 	}
 	g.idxCache = indices
+
```
ehnuje:
```suggestion
```

PR #83, opened 2019-07-29
Refactor voteMap to avoid data race

andybclee:
## Proposed changes

- This PR changes the data structure of voteMap which stores a node's vote status from a normal map into a struct with its own lock
- And this struct has its own methods to Copy, Clear, Get/SetValue, Size and Import. Each methods lock the mutex as it needs
- The original PR is https://github.com/ground-x/klaytn/pull/3525 which was reverted

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/api.go
+++ b/governance/api.go
@@ -167,10 +167,8 @@ type VoteList struct {
 func (api *PublicGovernanceAPI) MyVotes() []*VoteList {
 
 	ret := []*VoteList{}
-	api.governance.voteMapLock.RLock()
-	defer api.governance.voteMapLock.RUnlock()
 
-	for k, v := range api.governance.voteMap {
+	for k, v := range api.governance.voteMap.Copy() {
 		item := &VoteList{
 			Key:      k,
 			Value:    v.Value,
--- a/governance/api.go
+++ b/governance/api.go
@@ -166,7 +166,8 @@ type VoteList struct {
 
 func (api *PublicGovernanceAPI) MyVotes() []*VoteList {
 
-	ret := []*VoteList{}
+	ret := make([]*VoteList, 0, api.governance.voteMap.Size())
+	//ret := []*VoteList{}
 
 	for k, v := range api.governance.voteMap.Copy() {
 		item := &VoteList{
--- a/governance/default.go
+++ b/governance/default.go
@@ -146,7 +146,7 @@ type VoteStatus struct {
 	Num    uint64      `json:"num"`
 }
 
-type VoteSet struct {
+type VoteMap struct {
 	items map[string]VoteStatus
 	mu    *sync.RWMutex
 }
@@ -155,7 +155,7 @@ type Governance struct {
 	ChainConfig *params.ChainConfig
 
 	// Map used to keep multiple types of votes
-	voteMap VoteSet
+	voteMap VoteMap
 
 	nodeAddress      atomic.Value //common.Address
 	totalVotingPower uint64
@@ -183,8 +183,8 @@ type Governance struct {
 	blockChain *blockchain.BlockChain
 }
 
-func NewVoteMap() VoteSet {
-	return VoteSet{
+func NewVoteMap() VoteMap {
+	return VoteMap{
 		items: make(map[string]VoteStatus),
 		mu:    new(sync.RWMutex),
 	}
@@ -211,7 +211,7 @@ func (gt *GovernanceTallyList) Clear() {
 	gt.items = make([]GovernanceTallyItem, 0)
 }
 
-func (vl *VoteSet) Copy() map[string]VoteStatus {
+func (vl *VoteMap) Copy() map[string]VoteStatus {
 	vl.mu.RLock()
 	defer vl.mu.RUnlock()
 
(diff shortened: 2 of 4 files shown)


governance/default.go
```diff
@@ -146,12 +146,16 @@ type VoteStatus struct {
...
+	mu    *sync.RWMutex
+}
+
 type Governance struct {
 	ChainConfig *params.ChainConfig
 
 	// Map used to keep multiple types of votes
-	voteMap     map[string]VoteStatus
-	voteMapLock sync.RWMutex
+	voteMap VoteSet
```
ehnuje: Variable name is still `voteMap` but the type has been changed to `VoteSet`.
It would be better to use `VoteMap` instead of `VoteSet` since it internally manages a map.

governance/api.go
```diff
@@ -167,10 +167,8 @@ type VoteList struct {
 func (api *PublicGovernanceAPI) MyVotes() []*VoteList {
 
 	ret := []*VoteList{}
-	api.governance.voteMapLock.RLock()
-	defer api.governance.voteMapLock.RUnlock()
 
-	for k, v := range api.governance.voteMap {
+	for k, v := range api.governance.voteMap.Copy() {
```
ehnuje: What's the reason of calling `Copy()` and then copy the each element into the new one for `ret`?
How about holding a lock and then copying the item here? This seems to cause unnecessary double copying.

governance/default.go
```diff
@@ -200,6 +211,55 @@ func (gt *GovernanceTallyList) Clear() {
...
+	vl.items = make(map[string]VoteStatus)
+}
+
+func (vl *VoteSet) Size() int {
+	vl.mu.RLock()
+	defer vl.mu.RUnlock()
+
+	return len(vl.items)
+}
+
```
ehnuje: If above operations are frequently called, you can consider removing `defer` for the performance (though I don't  think so..)

governance/default.go
KimKyungup: I think `voteMap` is better than `voteSet`.

governance/api.go
andybclee: @ehnuje I'm not sure on where the double copying happened. AFAIK, in `range` operation `Copy()` will be executed once. Then this loop iterates for converting a map into a slice which is to be shown to an user. If I have a wrong understanding, please let me know.

governance/default.go
andybclee: @ehnuje Is it because `defer` can cause additional overhead?

governance/api.go
ehnuje: First Copy = call `voteMap.Copy()`  
Second Copy = create `VoteList` in the loop (line 172~177), copy `k`, `v.Value`, `v.Casted` and `v.Num`
Hence  `k`, `v.Value`, `v.Casted` and `v.Num` seem to be copied twice, doesn't it?

governance/default.go
ehnuje: @andybclee Yes, that can be an issue if this is called frequently.

ehnuje: If every operations require locking, how about using `sync.Map`?

governance/api.go
```diff
@@ -167,10 +167,8 @@ type VoteList struct {
 func (api *PublicGovernanceAPI) MyVotes() []*VoteList {
 
 	ret := []*VoteList{}
```
ehnuje: Isn't it better to set the size of the slice? 
Like `ret := make([]*VoteList{}, api.governance.voteMap.Size())`

andybclee: @ehnuje It is interesting. `ret := []*VoteList{}` and `ret := make([]*VoteList{}, api.governance.voteMap.Size())` shows different results when it is printed in console.

When your suggestion applied, the result is like below. It adds additional null in slot 0.
```shell
> governance.myVotes
[null, {
    BlockNum: 236,
    Casted: true,
    Key: "reward.ratio",
    Value: "100/0/0"
}]
```

ehnuje: How about `ret := make([]*VoteList{}, 0, api.governance.voteMap.Size())`

ehnuje: https://play.golang.org

ehnuje: This is because the size and the capacity have the same value. As we use append here, 0 should be given for the size.

governance/default.go
andybclee: @ehnuje I'm not sure if this case fits in using `sync.Map`. I read [this article](https://www.reddit.com/r/golang/comments/6vf8bu/syncmap_examples/), but this seems not so good case in using sync.Map.

And all these functions are not called frequently. Only when user voted on something, it can be called. So I think the overhead of using defer is negligible in this case.

governance/default.go
andybclee: @ehnuje @KimKyungup Thanks for your feedback. I changed the name of the type as `VoteMap`. PTAL =)

governance/api.go
andybclee: @ehnuje Done. Thanks!

PR #92, opened 2019-07-30
Read-and-Write-actualGovernanceBlock-atomically

andybclee:
## Proposed changes

- This PR is to fix data race on actualGovernanceBlock in governance object
- Because the variable can be accessed by several goroutines, it have to be treated atomically

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/default.go
+++ b/governance/default.go
@@ -570,14 +570,14 @@ func (g *Governance) initializeCache() error {
 	for _, v := range indices {
 		if num, data, err := g.ReadGovernance(v); err == nil {
 			g.itemCache.Add(getGovernanceCacheKey(num), data)
-			g.actualGovernanceBlock = num
+			atomic.StoreUint64(&g.actualGovernanceBlock, num)
 		} else {
 			logger.Crit("Couldn't read governance cache from database. Check database consistency", "index", v, "err", err)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


governance/default.go
```diff
@@ -570,14 +570,14 @@ func (g *Governance) initializeCache() error {
 	for _, v := range indices {
 		if num, data, err := g.ReadGovernance(v); err == nil {
 			g.itemCache.Add(getGovernanceCacheKey(num), data)
-			g.actualGovernanceBlock = num
+			atomic.StoreUint64(&g.actualGovernanceBlock, num)
```
ehnuje: Is it okay to use `atomic.StoreUint64` and `atomic.LoadUint64`, instead of using `atomic.Value` for `actualGovernanceBlock`?

andybclee: @ehnuje I think the name made you confused. actualGovernanceBlock is the number  of the block and it has uint64 data type.

andybclee:
> As we talked f2f, it seems fine for now. Let's make an issue to track whether to use `atomic.Value` here.

Added #93 and https://groundx.atlassian.net/jira/software/projects/TB/boards/36/backlog?assignee=5ab9f16ca142832a9381c64a&selectedIssue=TB-382

PR #97, opened 2019-07-31
Remove direct access to governance.ChainConfig

andybclee:
## Proposed changes

- This PR has following changes and formerly reviewed at https://github.com/ground-x/klaytn/pull/3682
- Added getters for each filed of chain config
- Replace direct access to chainConfig with newly added getters
- From this PR, ChainConfig is not updated. governance.currentSet holds all current governance information and provides it if requested.
- blockchain has its own chainConfig and It should be updated when there is a change. A getter and setter for `UseGiniCoeff` and `ProposerPolicy` are added and those are called by governance when there is a change

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -101,8 +101,9 @@ type CacheConfig struct {
 // included in the canonical one where as GetBlockByNumber always represents the
 // canonical chain.
 type BlockChain struct {
-	chainConfig *params.ChainConfig // Chain & network configuration
-	cacheConfig *CacheConfig        // stateDB caching and trie caching/pruning configuration
+	chainConfig   *params.ChainConfig // Chain & network configuration
+	chainConfigMu *sync.RWMutex
+	cacheConfig   *CacheConfig // stateDB caching and trie caching/pruning configuration
 
 	db     database.DBManager // Low level persistent database to store final content in
 	triegc *prque.Prque       // Priority queue mapping block numbers to tries to gc
@@ -177,6 +178,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 
 	bc := &BlockChain{
 		chainConfig:     chainConfig,
+		chainConfigMu:   new(sync.RWMutex),
 		cacheConfig:     cacheConfig,
 		db:              db,
 		triegc:          prque.New(),
@@ -226,6 +228,34 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 	// Take ownership of this particular state
 	go bc.update()
 	return bc, nil
+}
+
+func (bc *BlockChain) UseGiniCoeff() bool {
+	bc.chainConfigMu.RLock()
+	defer bc.chainConfigMu.RUnlock()
+
+	return bc.chainConfig.Governance.Reward.UseGiniCoeff
+}
+
+func (bc *BlockChain) SetUseGiniCoeff(val bool) {
+	bc.chainConfigMu.Lock()
+	defer bc.chainConfigMu.Unlock()
+
+	bc.chainConfig.Governance.Reward.UseGiniCoeff = val
+}
+
+func (bc *BlockChain) ProposerPolicy() uint64 {
+	bc.chainConfigMu.RLock()
+	defer bc.chainConfigMu.RUnlock()
+
+	return bc.chainConfig.Istanbul.ProposerPolicy
+}
+
+func (bc *BlockChain) SetProposerPolicy(val uint64) {
+	bc.chainConfigMu.Lock()
+	defer bc.chainConfigMu.Unlock()
+
+	bc.chainConfig.Istanbul.ProposerPolicy = val
 }
 
 func (bc *BlockChain) getProcInterrupt() bool {
@@ -1052,7 +1082,7 @@ func isCommitTrieRequired(bc *BlockChain, blockNum uint64) bool {
 	}
 
 	if bc.chainConfig.Istanbul != nil {
-		return bc.chainConfig.Istanbul.ProposerPolicy == params.WeightedRandom &&
+		return bc.ProposerPolicy() == params.WeightedRandom &&
 			params.IsStakingUpdateInterval(blockNum)
 	}
 	return false
--- a/consensus/istanbul/validator.go
+++ b/consensus/istanbul/validator.go
@@ -22,7 +22,6 @@ package istanbul
 
 import (
 	"github.com/klaytn/klaytn/common"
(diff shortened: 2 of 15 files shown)


consensus/istanbul/validator/weighted.go
```diff
@@ -554,7 +554,7 @@ func (valSet *weightedCouncil) Policy() istanbul.ProposerPolicy { return valSet.
 // It return no error when weightedCouncil
 //   (1) already has up-do-date proposers
 //   (2) successfully calculated up-do-date proposers
-func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64, config *params.ChainConfig) error {
+func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64, chainId uint64) error {
```
jimni1222: When is chainId used?

contracts/reward/reward_test.go
```diff
@@ -149,53 +151,72 @@ func TestUpdateGovernanceParameterByEpoch(t *testing.T) {
...
 		t.Errorf("GovernanceParameter is different")
 	}
 
+	blockNumber = 31
+	currentSet := governance.NewGovernanceSet()
+	delta := governance.NewGovernanceSet()
+	_, current, _ := gov.ReadGovernance(blockNumber)
+	currentSet.Import(current)
+	delta.SetValue(params.Ratio, "40/30/30")
+	gov.WriteGovernance(blockNumber-1, currentSet, delta)
```
jimni1222: This also just question.
Why 'blockNumber-1' as a parameter to gov.WriteGovernance?

consensus/istanbul/validator/weighted.go
laggu: chainId was for hardfork. It was used for checking if the network is baobab.
chainId will be removed in the next reward refactoring pr

governance/handler.go
```diff
@@ -44,67 +44,58 @@ var (
...
-	params.UseGiniCoeff:            {boolT, checkUint64andBool, updateGovernanceConfig},
-	params.DeferredTxFee:           {boolT, checkUint64andBool, updateGovernanceConfig},
-	params.MinimumStake:            {stringT, checkBigInt, updateGovernanceConfig},
-	params.StakeUpdateInterval:     {uint64T, checkUint64andBool, updateGovernanceConfig},
-	params.ProposerRefreshInterval: {uint64T, checkUint64andBool, updateGovernanceConfig},
-	params.Epoch:                   {uint64T, checkUint64andBool, updateGovernanceConfig},
-	params.Policy:                  {uint64T, checkUint64andBool, updateGovernanceConfig},
-	params.CommitteeSize:           {uint64T, checkUint64andBool, updateGovernanceConfig},
-	params.ConstTxGasHumanReadable: {uint64T, checkUint64andBool, updateParams},
+	params.GovernanceMode:          {stringT, checkGovernanceMode, nil},
```
jimni1222: Is there a criterion for when I need to handle a trigger?

blockchain/blockchain.go
```diff
@@ -101,8 +101,9 @@ type CacheConfig struct {
 // included in the canonical one where as GetBlockByNumber always represents the
 // canonical chain.
 type BlockChain struct {
-	chainConfig *params.ChainConfig // Chain & network configuration
-	cacheConfig *CacheConfig        // stateDB caching and trie caching/pruning configuration
+	chainConfig   *params.ChainConfig // Chain & network configuration
+	chainConfigMu *sync.RWMutex
```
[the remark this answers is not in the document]
KimKyungup: `chainConfig` is a pointer. So to protect `chainConfig`, this code is better.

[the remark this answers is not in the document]
KimKyungup: But `chainConfig` didn't change, `chainConfig` can have the Mu. :)

consensus/istanbul/validator.go
```diff
@@ -93,7 +92,7 @@ type ValidatorSet interface {
 	IsSubSet() bool
 
 	// Refreshes a list of candidate proposers with given hash and blockNum
-	Refresh(hash common.Hash, blockNum uint64, config *params.ChainConfig) error
+	Refresh(hash common.Hash, blockNum uint64, chainId uint64) error
```
KimKyungup: Where is the implementation of this method which needs chainID?

KimKyungup: https://github.com/klaytn/klaytn/pull/97/files#r309090594

contracts/reward/reward_test.go
KimKyungup: I also wondering this. :) Could you write the reason? :)

blockchain/blockchain.go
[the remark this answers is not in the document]
andybclee: @ehnuje Changing params.ChainConfig will cause a hard fork. The struct itself is stored in a database, so to change it, all nodes have to be updated. Just for adding a lock, I think it costs too much.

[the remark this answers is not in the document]
ehnuje: @andybclee I understood, thanks for the comment!

governance/handler.go
```diff
@@ -44,67 +44,58 @@ var (
...
+}
+
+func updateStakingUpdateInterval(g *Governance, k string, v interface{}) bool {
+	params.SetStakingUpdateInterval(g.StakingUpdateInterval())
+	return true
+}
+
+func updateProposerUpdateInterval(g *Governance, k string, v interface{}) bool {
+	params.SetProposerUpdateInterval(g.ProposerUpdateInterval())
+	return true
```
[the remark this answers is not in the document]
andybclee: @ehnuje Initially, I thought there can be some error case while reflecting changes. So the struct is defined as below.
```go
type check struct {
	t         reflect.Type
	validator func(k string, v interface{}) bool
	trigger   func(g *Governance, k string, v interface{}) bool
}
```
The above function and others which are used as trigger should have same signature. 

But at the moment, all changes are about setting a value and nothing else. So the return value is practically meaningless.

For now, I think removing return value is okay. I'll commit the change soon.

governance/handler.go
andybclee: @jimni1222 This map defines 
1. Governance Item (as a key of the map)
2. the key's data type
3. the validator function for the type
4. trigger function which to be called when the value of the key changed if it requires more action other than storing it in `governance.currentSet`

So it is triggered when the change have to affect an object other than `governance`

jimni1222: Ah, ! that means, if update need to trigger something more than action in governance.currentSet, then trigger should be defined ! Thank you for explain :)

PR #100, opened 2019-07-31
Fix wrong governance cache behavior

andybclee:
## Proposed changes

- Governance manages its own cache for faster retrieval of recent change history
- There was two issues in managing cache
- 1. Duplicated block number could be written in idxCache if a node stop and run repeatedly. In this case same number was appended at the end of idxCache
- 2. itemCache may have different block's governance information and it makes a node to retrieve database directly though it should be retrieved from the cache
- To solve above issues, check routines were refined. And use `g.db.ReadGovernance` in initializing cache.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/default.go
+++ b/governance/default.go
@@ -608,7 +608,7 @@ func newGovernanceCache() common.Cache {
 
 func (g *Governance) initializeCache() error {
 	// get last n governance change block number
-	indices, err := g.db.ReadRecentGovernanceIdx(params.GovernanceCacheLimit)
+	indices, err := g.db.ReadRecentGovernanceIdx(params.GovernanceIdxCacheLimit)
 	if err != nil {
 		return ErrNotInitialized
 	}
@@ -616,9 +616,10 @@ func (g *Governance) initializeCache() error {
 
 	// Put governance items into the itemCache
 	for _, v := range indices {
-		if num, data, err := g.ReadGovernance(v); err == nil {
-			g.itemCache.Add(getGovernanceCacheKey(num), data)
-			atomic.StoreUint64(&g.actualGovernanceBlock, num)
+		if data, err := g.db.ReadGovernance(v); err == nil {
+			data = adjustDecodedSet(data)
+			g.itemCache.Add(getGovernanceCacheKey(v), data)
+			atomic.StoreUint64(&g.actualGovernanceBlock, v)
 		} else {
 			logger.Crit("Couldn't read governance cache from database. Check database consistency", "index", v, "err", err)
 		}
@@ -665,6 +666,14 @@ func getGovernanceCacheKey(num uint64) common.GovernanceCacheKey {
 
 // Store new governance data on DB. This updates Governance cache too.
 func (g *Governance) WriteGovernance(num uint64, data GovernanceSet, delta GovernanceSet) error {
+	g.idxCacheLock.RLock()
+	idx := make([]uint64, 0, len(g.idxCache))
+	copy(idx, g.idxCache)
+	g.idxCacheLock.RUnlock()
+
+	if len(idx) > 0 && num <= idx[len(idx)-1] {
+		return nil
+	}
 
 	new := NewGovernanceSet()
 	new.Import(data.Items())
@@ -770,7 +779,7 @@ func (gov *Governance) UpdateCurrentGovernance(num uint64) {
 	newNumber, newGovernanceSet, _ := gov.ReadGovernance(num)
 
 	// Do the change only when the governance actually changed
-	if newGovernanceSet != nil && newNumber != atomic.LoadUint64(&gov.actualGovernanceBlock) {
+	if newGovernanceSet != nil && newNumber > atomic.LoadUint64(&gov.actualGovernanceBlock) {
 		atomic.StoreUint64(&gov.actualGovernanceBlock, newNumber)
 		gov.currentSet.Import(newGovernanceSet)
 		gov.triggerChange(newGovernanceSet)
--- a/governance/default.go
+++ b/governance/default.go
@@ -667,11 +667,11 @@ func getGovernanceCacheKey(num uint64) common.GovernanceCacheKey {
 // Store new governance data on DB. This updates Governance cache too.
 func (g *Governance) WriteGovernance(num uint64, data GovernanceSet, delta GovernanceSet) error {
 	g.idxCacheLock.RLock()
-	idx := make([]uint64, 0, len(g.idxCache))
-	copy(idx, g.idxCache)
+	indices := make([]uint64, 0, len(g.idxCache))
+	copy(indices, g.idxCache)
 	g.idxCacheLock.RUnlock()
 
-	if len(idx) > 0 && num <= idx[len(idx)-1] {
+	if len(indices) > 0 && num <= indices[len(indices)-1] {
 		return nil
 	}
 
--- a/params/governance_params.go
+++ b/params/governance_params.go
@@ -54,7 +54,7 @@ const (
 	// Block reward will be separated by three pieces and distributed
 	RewardSliceCount = 3
 	// GovernanceConfig is stored in a cache which has below capacity
-	GovernanceCacheLimit    = 3
+	GovernanceCacheLimit    = 512
 	GovernanceIdxCacheLimit = 1000
 	// The prefix for governance cache
 	GovernanceCachePrefix = "governance"


governance/default.go
```diff
@@ -665,6 +666,14 @@ func getGovernanceCacheKey(num uint64) common.GovernanceCacheKey {
 
 // Store new governance data on DB. This updates Governance cache too.
 func (g *Governance) WriteGovernance(num uint64, data GovernanceSet, delta GovernanceSet) error {
+	g.idxCacheLock.RLock()
+	idx := make([]uint64, 0, len(g.idxCache))
+	copy(idx, g.idxCache)
+	g.idxCacheLock.RUnlock()
+
+	if len(idx) > 0 && num <= idx[len(idx)-1] {
+		return nil
```
laggu: is it okay not to return an error??

governance/default.go
```diff
@@ -608,17 +608,18 @@ func newGovernanceCache() common.Cache {
 
 func (g *Governance) initializeCache() error {
 	// get last n governance change block number
-	indices, err := g.db.ReadRecentGovernanceIdx(params.GovernanceCacheLimit)
+	indices, err := g.db.ReadRecentGovernanceIdx(params.GovernanceIdxCacheLimit)
```
ehnuje: This has been changed from `GovernanceCacheLimit` to `GovernanceIdxCacheLimit`. Is it intended?
Since the one value changed is `GovernanceCacheLimit`

governance/default.go
```diff
@@ -665,6 +666,14 @@ func getGovernanceCacheKey(num uint64) common.GovernanceCacheKey {
 
 // Store new governance data on DB. This updates Governance cache too.
 func (g *Governance) WriteGovernance(num uint64, data GovernanceSet, delta GovernanceSet) error {
+	g.idxCacheLock.RLock()
+	idx := make([]uint64, 0, len(g.idxCache))
```
ehnuje: Using `idx` for a slice looks a bit confusing. How about `indices`?

governance/default.go
```diff
@@ -608,17 +608,18 @@ func newGovernanceCache() common.Cache {
...
 		return ErrNotInitialized
 	}
 	g.idxCache = indices
 
 	// Put governance items into the itemCache
 	for _, v := range indices {
-		if num, data, err := g.ReadGovernance(v); err == nil {
-			g.itemCache.Add(getGovernanceCacheKey(num), data)
-			atomic.StoreUint64(&g.actualGovernanceBlock, num)
+		if data, err := g.db.ReadGovernance(v); err == nil {
```
ehnuje: What is the reason of changing this line?

governance/default.go
```diff
@@ -770,7 +779,7 @@ func (gov *Governance) UpdateCurrentGovernance(num uint64) {
 	newNumber, newGovernanceSet, _ := gov.ReadGovernance(num)
 
 	// Do the change only when the governance actually changed
-	if newGovernanceSet != nil && newNumber != atomic.LoadUint64(&gov.actualGovernanceBlock) {
+	if newGovernanceSet != nil && newNumber > atomic.LoadUint64(&gov.actualGovernanceBlock) {
```
ehnuje: What is the reason of changing this line?

governance/default.go
```diff
@@ -665,6 +666,14 @@ func getGovernanceCacheKey(num uint64) common.GovernanceCacheKey {
 
 // Store new governance data on DB. This updates Governance cache too.
 func (g *Governance) WriteGovernance(num uint64, data GovernanceSet, delta GovernanceSet) error {
+	g.idxCacheLock.RLock()
+	idx := make([]uint64, 0, len(g.idxCache))
+	copy(idx, g.idxCache)
+	g.idxCacheLock.RUnlock()
+
+	if len(idx) > 0 && num <= idx[len(idx)-1] {
```
nogang: If len (idx) is 0, is it not processed?

governance/default.go
```diff
@@ -665,6 +666,14 @@ func getGovernanceCacheKey(num uint64) common.GovernanceCacheKey {
 
 // Store new governance data on DB. This updates Governance cache too.
 func (g *Governance) WriteGovernance(num uint64, data GovernanceSet, delta GovernanceSet) error {
+	g.idxCacheLock.RLock()
+	idx := make([]uint64, 0, len(g.idxCache))
+	copy(idx, g.idxCache)
```
nogang: I do not seem to use idx. Why do I copy?

ehnuje: Just to get the `len(idx)`?

governance/default.go
andybclee: @ehnuje Yes. It is intended. Initially, the number of `GovernanceCacheLimit` was very small. It seemed enough to keep up with block increase, it couldn't respond well for a random search like an API call. Jun insisted to increase to hold all in cache. Keeping all governance configuration in memory is worrying because of memory usage, but keeping just index won't take that amount of memory. So I added a new constant `GovernanceIdxCacheLimit` which has twice bigger than `GovernanceCacheLimit` which limits the number of ItemCache

governance/default.go
andybclee: @ehnuje `g.ReadGovernance` and `g.db.ReadGovernance` works differently. `g.db.ReadGovernance` read the governance information stored at the block `v`.  `g.ReadGovernance` calculates the block number which have to be applied at the block `v`.

So basic design was `idxCache` is to hold the block numbers and `itemCache` is to hold the governance information of matching block number stored in idxCache. But because of the calculation, there was mismatch between idxCache's block numbers and governance information stored in `itemCache`

governance/default.go
andybclee: @nogang @ehnuje It is to get `len(idx)` and `idx[len(idx)-1]`. I think copying can be removed if the checks are done inside the lock.

governance/default.go
andybclee: @nogang If len(idx) is 0, it means that the function is called in making a new governance object. Because `WriteGovernance` is used to store the vote result and also to store the information written in genesis.json, len(idx) can be 0. 

The check routine is to prevent writing already stored governance information again.

governance/default.go
andybclee: Because this is the normal case if snapshot() is called for old blocks, it is not an error.

governance/default.go
andybclee: @ehnuje UpdateCurrentGovernance() is called by snapshot() when blockNumber % Epoch == 0. So while following the incoming blocks, there was no problem. But if istanbul.getSnapshot() is called for the old Epoch block, though `newNumber < actualGovernanceBlock`, the update can be performed. This change is to prevent that current information is overwritten by the old one.

governance/default.go
andybclee: @nogang @ehnuje I think the code can be changed like this
```go
	g.idxCacheLock.RLock()
	defer g.idxCacheLock.RUnlock()

	if len(g.idxCache) > 0 && num <= g.idxCache[len(g.idxCache)-1] {
		return nil
	}
```
But in this case, the lock on idxCache have to last until the writing is done if it is for actually new governance information.

KimKyungup: @andybclee `defer g.idxCacheLock.RUnlock()` can make deadlock with `.addGovernanceCache(num, new)` on line 685.

governance/default.go
KimKyungup: Same routine is in `addGovernanceCache`. 
Is it duplicated? If so, we can refactor them?

andybclee: @KimKyungup That's a good point. Thanks! I added a new commit for this. Please take a look.

governance/default.go
andybclee: @KimKyungup @ehnuje Yes. It makes a problem. I made a separate function which just check the condition and to be called by `WriteGovernance` and `addGovernanceCache`. PTAL.

governance/default.go
```diff
@@ -642,10 +643,7 @@ func (g *Governance) getGovernanceCache(num uint64) (map[string]interface{}, boo
 
 func (g *Governance) addGovernanceCache(num uint64, data GovernanceSet) {
 	// Don't update cache if num (block number) is smaller than the biggest number of cached block number
-	g.idxCacheLock.Lock()
-	defer g.idxCacheLock.Unlock()
```
ehnuje: Removing the lock here would change the original behavior, as line 654-567 won't be protected by the lock anymore. Is it okay?

governance/default.go
andybclee: @ehnuje This part is separated to the new function and the use of idx removed. Please take another look.

governance/default.go
andybclee: @ehnuje Instead, the operation moved to a new function `isNewBlockToAdd`. That part is to check if it is good to proceed or not. And `isNewBlockToAdd` is doing that work. So original behavior was not changed.

andybclee: @ehnuje My bad. It is changed. I think the last commit have to be rolled back.

governance/default.go
andybclee: @KimKyungup I rolled back the last commit. As @ehnuje pointed out, the behavior of `addGovernanceCache` is different from that of `WriteGovernance`, the last commit didn't work.

governance/default.go
ehnuje: @andybclee Could you please check this?

ehnuje: idx -> indices

andybclee: @ehnuje It is changed. Thanks for pointing it out.

PR #110, opened 2019-08-02
Add governance API for debugging

andybclee:
## Proposed changes

- This PR adds APIs to check / debug current status of governance object
- Following APIs are added
- itemCacheFromDb() : Returns governance information stored at given block
- pendingChanges() : Returns `changeSet` of the governance object
- idxCache(): Returns in-memory idxCache that a node has
- idxCacheFromDb(): Returns idxCache which holds all historical change from database
- votes(): Returns a list of governance votes in this epoch

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/api.go
+++ b/governance/api.go
@@ -157,6 +157,33 @@ func (api *PublicGovernanceAPI) ItemsAt(num *rpc.BlockNumber) (map[string]interf
 	}
 }
 
+func (api *PublicGovernanceAPI) PendingChanges() map[string]interface{} {
+	return api.governance.PendingChanges()
+}
+
+func (api *PublicGovernanceAPI) Votes() []GovernanceVote {
+	return api.governance.Votes()
+}
+
+func (api *PublicGovernanceAPI) IdxCache() []uint64 {
+	return api.governance.IdxCache()
+}
+
+func (api *PublicGovernanceAPI) IdxCacheFromDb() []uint64 {
+	return api.governance.IdxCacheFromDb()
+}
+
+func (api *PublicGovernanceAPI) ItemCacheFromDb(num *rpc.BlockNumber) map[string]interface{} {
+	blockNumber := uint64(0)
+	if num == nil || *num == rpc.LatestBlockNumber || *num == rpc.PendingBlockNumber {
+		blockNumber = api.governance.blockChain.CurrentHeader().Number.Uint64()
+	} else {
+		blockNumber = uint64(num.Int64())
+	}
+	ret, _ := api.governance.db.ReadGovernance(blockNumber)
+	return ret
+}
+
 type VoteList struct {
 	Key      string
 	Value    interface{}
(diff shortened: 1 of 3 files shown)


governance/default.go
```diff
@@ -1045,3 +1045,23 @@ func (gov *Governance) UseGiniCoeff() bool {
...
+
+func (gov *Governance) Votes() []GovernanceVote {
+	return gov.GovernanceVotes.Copy()
+}
+
+func (gov *Governance) IdxCache() []uint64 {
+	gov.idxCacheLock.RLock()
+	defer gov.idxCacheLock.RUnlock()
+
+	return gov.idxCache
```
ehnuje: `IdxCache()` should copy `idxCache`, like `PendingChanges()` and `Votes()`

governance/api.go
```diff
@@ -157,6 +157,33 @@ func (api *PublicGovernanceAPI) ItemsAt(num *rpc.BlockNumber) (map[string]interf
...
+	return api.governance.IdxCache()
+}
+
+func (api *PublicGovernanceAPI) IdxCacheFromDb() []uint64 {
+	return api.governance.IdxCacheFromDb()
+}
+
+func (api *PublicGovernanceAPI) ItemCacheFromDb(num *rpc.BlockNumber) map[string]interface{} {
+	blockNumber := uint64(0)
+	if num == nil || *num == rpc.LatestBlockNumber || *num == rpc.PendingBlockNumber {
```
aidan-kwon: Please don't forget describing in KlaytnDocs that the API supports only "lastest" value as a block parameter.
Maybe we need to refine the use of "earliest", "pending" values in the block parameter someday.

governance/default.go
andybclee: @ehnuje I agree that it is the way to go. But it requires the change on the type of idxCache and it can be pretty big. So I'm not sure it have to be done with this PR. I think the change of idxCache have to be in another PR to separate focus of each PR. May I have your opinion?

governance/api.go
andybclee: @aidan-kwon Understood. And API documentation also have to be updated. I will update it considering your advice. Thank you!

governance/default.go
ehnuje: @andybclee Wouldn't it be enough to copy `idxCache` while holding the lock like below? 
Please tell me if I missed something.

```
gov.idxCacheLock.RLock()
defer gov.idxCacheLock.RUnlock()

copiedCache := make([]uint64, 0, len(gov.idxCache))
copy(copiedCache, gov.idxCache)
return copiedCache

```

andybclee: @ehnuje Ah. I think I misunderstood your point. I think your mention is about to make idxCache to have methods such as Copy(), Size() and so on.

ehnuje: @andybclee My bad =( I also agree with that my comment seems a bit confusing.

PR #125, opened 2019-08-06
Add IstanbulMsg related metrics

andybclee:
## Proposed changes

- This PR adds metrics for consensus messages' traffic count and bytes
- klay/prop/consensus/in/packets, klay/prop/consensus/in/traffic for received consensus messages
- klay/prop/consensus/out/packets, klay/prop/consensus/out/traffic for sent consensus messages
- Renamed the message code of `istanbulMsg` to 'IstanbulMsg` to export the code to other packages

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- These metrics were added to evaluate the effect of changing the behavior of committee nodes in sending messages


node/cn/metrics.go
```diff
@@ -21,51 +21,56 @@
...
+	miscOutPacketsMeter          = metrics.NewRegisteredMeter("klay/misc/out/packets", nil)
+	miscOutTrafficMeter          = metrics.NewRegisteredMeter("klay/misc/out/traffic", nil)
+	txReceiveCounter             = metrics.NewRegisteredCounter("klay/tx/recv/counter", nil)
+	txResendCounter              = metrics.NewRegisteredCounter("klay/tx/resend/counter", nil)
+	txSendCounter                = metrics.NewRegisteredCounter("klay/tx/send/counter", nil)
+	txResendRoutineGauge         = metrics.NewRegisteredGauge("klay/tx/resend/routine/gauge", nil)
+	cnPeerCountGauge             = metrics.NewRegisteredGauge("p2p/CNPeerCountGauge", nil)
+	pnPeerCountGauge             = metrics.NewRegisteredGauge("p2p/PNPeerCountGauge", nil)
+	enPeerCountGauge             = metrics.NewRegisteredGauge("p2p/ENPeerCountGauge", nil)
+	propConsensusInPacketsMeter  = metrics.NewRegisteredMeter("klay/prop/consensus/in/packets", nil)
```
KimKyungup: How about this? This packet is istanbulMsg. So I suggest like this.

```suggestion
	propConsensusIstanbulInPacketsMeter  = metrics.NewRegisteredMeter("klay/prop/consensus/istanbul/in/packets", nil)
```

node/cn/metrics.go
```diff
@@ -21,51 +21,56 @@
 package cn
 
 import (
+	"github.com/klaytn/klaytn/consensus/istanbul/backend"
```
KimKyungup: Just my opinion.
In long term, we need a better name of istanbul/backend like below or we can merge istanbul and istanbul/backend.

image

andybclee: @KimKyungup I agree that `backend` is unclear name. Other than this, there are some cases similar to this one. I think it would be better to define a best practice of package naming and apply it considering making a guideline for open source contributors.

node/cn/metrics.go
andybclee: @KimKyungup Got it. The metrics only works for istanbul consensus, it seems better.

PR #142, opened 2019-08-09
Make actualGovernanceBlock from uint64 to atomic.Value

andybclee:
## Proposed changes

- This PR changed the type of `actualGovernanceBlock` into `atomic.Value` as discussed [here](https://github.com/klaytn/klaytn/pull/92#pullrequestreview-268249821)
- This change is to reduce potential error in accessing the variable.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/default.go
+++ b/governance/default.go
@@ -171,7 +171,7 @@ type Governance struct {
 	idxCacheLock *sync.RWMutex
 
 	// The block number when current governance information was changed
-	actualGovernanceBlock uint64
+	actualGovernanceBlock atomic.Value //uint64
 
 	// The last block number at governance state was stored (used not to replay old votes)
 	lastGovernanceStateBlock uint64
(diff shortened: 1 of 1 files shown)


aidan-kwon:
AS-IS code looks fine for me. 
Even though this change can prevent other developers to forget locking the variable, it may cause another mistake storing other types of variable in the atomic value. The changed code also looks okay, but I am not sure the change is necessary.

ehnuje:
@aidan-kwon @andybclee Whether to use `atomic.Value` depends on the usages of the variable. If the current function call places are the only places that the variable is concurrently accessed and other places are not accessed during concurrent calls, it is better to use `atomic.StoreUint64` for better performance in non-concurrent calls. If it's not, we should use `atomic.Value` not to make any unwanted mistakes. This case is the latter one and in that sense, it is better to use `atomic.Value`

KimKyungup:
Why don't we consider to use https://github.com/uber-go/atomic?
```
var atom atomic.Uint32
atom.Store(42)
atom.Sub(2)
atom.CAS(40, 11)
```

PR #146, opened 2019-08-09
Fix bug in making slice. It blocked copying the slice properly

andybclee:
## Proposed changes

- This PR is to fix a bug in writing governance change.
- While transferring commits from the former repository, the code changed as below
```go
idx := []uint64{}
```
to
```go
indices := make([]uint64, 0,  len(g.idxCache))
```
- But because indices has zero length, nothing could be copied to indices so it passed the check routine because its size is zero
```go
	if len(indices) > 0 && num <= indices[len(indices)-1] {
		return nil
	}
```
- It made idxCache have duplicated block numbers in it when node stopped and restarted

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/default_test.go
+++ b/governance/default_test.go
@@ -547,11 +547,13 @@ func TestWriteGovernance_idxCache(t *testing.T) {
 	delta := NewGovernanceSet()
 	src.Import(tstMap)
 
-	gov.WriteGovernance(30, src, delta)
-	gov.WriteGovernance(30, src, delta)
-	gov.WriteGovernance(60, src, delta)
-	gov.WriteGovernance(60, src, delta)
-	gov.WriteGovernance(50, src, delta)
+	blockNum := []uint64{30, 30, 60, 60, 50}
+
+	for _, num := range blockNum {
+		if ret := gov.WriteGovernance(num, src, delta); ret != nil {
+			t.Errorf("Error in testing WriteGovernance, %v", ret)
+		}
+	}
 
 	// idxCache should have 0, 30 and 60
 	if len(gov.idxCache) != 3 && gov.idxCache[len(gov.idxCache)-1] != 60 {
--- a/governance/default_test.go
+++ b/governance/default_test.go
@@ -537,3 +537,24 @@ func TestCypressGenesisHash(t *testing.T) {
 		t.Errorf("Generated hash is not equal to Cypress's hash. Want %v, Have %v", cypressHash.String(), block.Hash().String())
 	}
 }
+
+func TestWriteGovernance_idxCache(t *testing.T) {
+	gov := getGovernance()
+
+	tstMap := copyMap(testGovernanceMap)
+
+	src := NewGovernanceSet()
+	delta := NewGovernanceSet()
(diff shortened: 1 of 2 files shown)


ehnuje:
This one seems to be the one I suggested. My bad =( How about adding a test case for this?

ehnuje:
This one seems to be the one I suggested. My bad =( How about adding a test case for this?

andybclee:
> This one seems to be the one I suggested. My bad =( How about adding a test case for this?

No problem. I remember that I knew how `copy` works but I forgot. :(
Anyway, it is a good idea to add a test case for this. I'll add some test code.

governance/default_test.go
```diff
@@ -537,3 +537,24 @@ func TestCypressGenesisHash(t *testing.T) {
...
+	src.Import(tstMap)
+
+	gov.WriteGovernance(30, src, delta)
+	gov.WriteGovernance(30, src, delta)
+	gov.WriteGovernance(60, src, delta)
+	gov.WriteGovernance(60, src, delta)
+	gov.WriteGovernance(50, src, delta)
+
+	// idxCache should have 0, 30 and 60
+	if len(gov.idxCache) != 3 && gov.idxCache[len(gov.idxCache)-1] != 60 {
```
ehnuje:
```suggestion
	if len(gov.idxCache) != 3 || gov.idxCache[len(gov.idxCache)-1] != 60 {
```
`&&` needs to be replaced by `||`, doesn't it?

governance/default_test.go
```diff
@@ -537,3 +537,24 @@ func TestCypressGenesisHash(t *testing.T) {
...
+func TestWriteGovernance_idxCache(t *testing.T) {
+	gov := getGovernance()
+
+	tstMap := copyMap(testGovernanceMap)
+
+	src := NewGovernanceSet()
+	delta := NewGovernanceSet()
+	src.Import(tstMap)
+
+	gov.WriteGovernance(30, src, delta)
```
KimKyungup: if return err checking is added, it is better. :)

andybclee: @KimKyungup Thanks. I changed the code to check an error in calling WriteGovernance.

governance/default_test.go
```diff
@@ -537,3 +537,26 @@ func TestCypressGenesisHash(t *testing.T) {
...
+	blockNum := []uint64{30, 30, 60, 60, 50}
+
+	for _, num := range blockNum {
+		if ret := gov.WriteGovernance(num, src, delta); ret != nil {
+			t.Errorf("Error in testing WriteGovernance, %v", ret)
+		}
+	}
+
+	// idxCache should have 0, 30 and 60
+	if len(gov.idxCache) != 3 || gov.idxCache[len(gov.idxCache)-1] != 60 {
```
jeonghonah: minor comment:
How about checking 0 and 30 also?

andybclee: @jeonghonah I also thought of it. But I think in this case, checking length of the cache is enough to cover that..

PR #170, opened 2019-08-22
SC: Replace []uint256 to bytes for extraData of Bridge contract

KimKyungup:
## Proposed changes
This PR change the type of extraData in Bridge contracts.

Previous type []uint256 is fixed type and fixed sized.
But bytes type can be casted to various types with `abi.encode/decode` with proper size in solidity.

Therefore extraData types is changed.

This change is expected for Service providers to be able to make their own service with custom bridge contract easily.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


contracts/extbridge/ext_bridge.sol
```diff
@@ -77,12 +113,12 @@ contract ExtBridge is BridgeTransferERC20, BridgeTransferERC721 {
...
         uint64 _requestBlockNumber,
         string memory _tokenURI,
-        uint256[] memory _extraData
+        bytes memory _extraData
     )
         public
     {
         if (_extraData.length > 0) {
-            uint256 offerPrice = _extraData[0];
+            uint256 offerPrice = abi.decode(_extraData, (uint256));
```
aidan-kwon: In my test, the decoder reverted if the size of calldata is less than 32. 
```
The ABI decoder reverts in the beginning of functions and in abi.decode() if passed calldata is too short or points out of bounds. Note that dirty higher order bits are still simply ignored.
```
Refer: https://solidity.readthedocs.io/en/v0.5.11/050-breaking-changes.html?highlight=abi.decode

KimKyungup: Good. :)
So you suggested to change the length check like below? :)
```
f (_extraData.length >= 32) {
```
How about remote the length check? :)

Or Are you thinking that this way doesn't have any benefit for the size?

aidan-kwon: How about explicitly checking the length if `price` is mandatory? 
`require(_extraData.length == 32, "the price should be uint256")`

KimKyungup: How about remove and if encode is fail, the tx revert? :)

KimKyungup: I think in this example, the price is not mandatory. So Could I keep this code? :)

KimKyungup: I reflected your review. :) Thanks.

contracts/extbridge/ext_bridge.sol
```diff
@@ -82,7 +82,7 @@ contract ExtBridge is BridgeTransferERC20, BridgeTransferERC721 {
     )
         public
     {
-        if (_extraData.length > 0) {
+        if (_extraData.length == 32) { // size check of uint256
```
jeonghonah: if this condition fails with (length > 0 && length != 32), how to about reverting this method?

KimKyungup: @aidan-kwon What do you think? :)
In this example, if we want not to allow other extraData (not uint256), original code looks better.

PR #171, opened 2019-08-23
Fix roundchange behavior when the number of nodes are less than 4.

andybclee:
## Proposed changes

- This PR is to solve an issue of block generation when F() is 0 (meaning that the number of nodes are less than 4)
- When a roundchange happened, the required number of messages are calculated differently based on the number of nodes

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/consensus/istanbul/core/roundchange.go
+++ b/consensus/istanbul/core/roundchange.go
@@ -110,7 +110,8 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
 				numCatchUp = n - 1
 	} else {
 		f := int(c.valSet.F())
-		numStartNewRound = int(2*fValue + 1)
+		// 2*f + 1 ROUND CHANGE messages can start new round.	
+		numStartNewRound = 2*f + 1
 		numCatchUp = int(fValue + 1)
 	}
 
--- a/consensus/istanbul/core/roundchange.go
+++ b/consensus/istanbul/core/roundchange.go
@@ -106,7 +106,8 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
 		n := c.valSet.Size()
 		// N ROUND CHANGE messages can start new round.
 		numStartNewRound = n
-		numCatchUp = numStartNewRound - 1
+				// N - 1 ROUND CHANGE messages can catch up the round.
+				numCatchUp = n - 1
 	} else {
 		fValue := c.valSet.F()
 		numStartNewRound = int(2*fValue + 1)
--- a/consensus/istanbul/core/roundchange.go
+++ b/consensus/istanbul/core/roundchange.go
@@ -103,14 +103,14 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
 
 	var numCatchUp, numStartNewRound int
 	if c.valSet.Size() < 4 {
-		n := c.valSet.Size()
+		n := int(c.valSet.Size())
 		// N ROUND CHANGE messages can start new round.
 		numStartNewRound = n
-				// N - 1 ROUND CHANGE messages can catch up the round.
-				numCatchUp = n - 1
+		// N - 1 ROUND CHANGE messages can catch up the round.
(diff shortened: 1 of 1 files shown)


aidan-kwon:
@andybclee It seems that `handleRoundChange` can be triggered by the protocol message from other nodes. When the validator set size is 1, how does it change the round without any received message?

andybclee:
> @andybclee It seems that `handleRoundChange` can be triggered by the protocol message from other nodes. When the validator set size is 1, how does it change the round without any received message?

When starting a new round, each node set a timeout timer using `time.AfterFunc`. In this way, each node set its own round based on a timer.

consensus/istanbul/core/roundchange.go
```diff
@@ -101,20 +101,35 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
...
+	var numCatchUp, numStartNewRound int
+	if c.valSet.Size() < 4 {
+		n := int(c.valSet.Size())
+		// N ROUND CHANGE messages can start new round.
+		numStartNewRound = n
+		// N - 1 ROUND CHANGE messages can catch up the round.
+		numCatchUp = n - 1
+	} else {
+		f := int(c.valSet.F())
+		// 2*f + 1 ROUND CHANGE messages can start new round.
```
KimKyungup:
```suggestion
		// 2*F + 1 ROUND CHANGE messages can start new round.
```

consensus/istanbul/core/roundchange.go
```diff
@@ -101,20 +101,35 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
...
+		numCatchUp = n - 1
+	} else {
+		f := int(c.valSet.F())
+		// 2*F + 1 ROUND CHANGE messages can start new round.
+		numStartNewRound = 2*f + 1
+		// F + 1 ROUND CHANGE messages can start catch up the round.
+		numCatchUp = f + 1
+	}
+
+	if num == numStartNewRound && (c.waitingForRoundChange || cv.Round.Cmp(roundView.Round) < 0) {
```
jeonghonah: In case of single node configuration, which conditions cause to `false` in the if-condition?

aidan-kwon:
@andybclee Thank you for the explanation.
The code looks good to me, but we may choose another approach. 
How do you think about making `valSet.F()` to return 0 when `valSet.Size()` is less than 4.

consensus/istanbul/core/roundchange.go
andybclee: @jeonghonah In this line, all conditions will be true in the single node configuration. So the node can start the new round with a new proposal.

andybclee:
> @andybclee Thank you for the explanation.
> The code looks good to me, but we may choose another approach.
> How do you think about making `valSet.F()` to return 0 when `valSet.Size()` is less than 4.

@aidan-kwon As you can see below, F() returns 0 when the size is smaller than 4.
```go
func (valSet *weightedCouncil) F() int {
	if valSet.Size() > valSet.subSize {
		return int(math.Ceil(float64(valSet.subSize)/3)) - 1
	} else {
		return int(math.Ceil(float64(valSet.Size())/3)) - 1
	}
}
```

aidan-kwon:
@andybclee Oh... I misunderstood about it. Thank you!

consensus/istanbul/core/roundchange.go
```diff
@@ -101,20 +101,35 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
...
+		// We've received enough ROUND CHANGE messages, start a new round immediately.
+		logger.Warn("[RC] Received 2f+1 Round Change Messages. Starting new round")
+		c.startNewRound(roundView.Round)
+		return nil
+	} else if c.waitingForRoundChange && num == numCatchUp {
```
jeonghonah: Is it not necessary to check `cv.Round.Cmp(roundView.Round) < 0` condition here?

PR #174, opened 2019-08-23
Fix protocol message

andybclee:
## Proposed changes

- This PR is to fix a log level and message which may mislead users
- When a consensus message is received while the node's downloader is working for a synchronization, "ProtocolManager failed to handle consensus message" will be logged in `Error` level. But this is the normal case if downloader is working. So the message have to give more detailed information in less scary log level.

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -461,7 +461,7 @@ func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, add
 			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
 			if err != nil {
-				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization", "msg", msg, "err", err)
+				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization.", "msg", msg, "err", err)
 				errCh <- err
 				return
 			}
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -461,7 +461,7 @@ func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, add
 			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
 			if err != nil {
-				p.GetP2PPeer().Log().Error("ProtocolManager failed to handle consensus message", "msg", msg, "err", err)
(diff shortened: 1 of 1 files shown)


node/cn/handler.go
```diff
@@ -461,7 +461,7 @@ func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, add
 			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
 			if err != nil {
-				p.GetP2PPeer().Log().Error("ProtocolManager failed to handle consensus message", "msg", msg, "err", err)
+				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization", "msg", msg, "err", err)
```
KimKyungup: How about checking specific error as you mentioned?
All errors happen caused by that case?

andybclee: @KimKyungup Yes. I checked HandleMsg() and related codes, the only error which can be returned was "ErrStoppedEngine". And the engine could be stopped only by the downloader.

I think the original author might assumed more cases, but no other error is present at the moment in our code base.

andybclee: @KimKyungup FYI, https://groundx.atlassian.net/jira/software/projects/TB/boards/36/backlog?assignee=5ab9f16ca142832a9381c64a&selectedIssue=TPT-597

KimKyungup: Thanks. @andybclee  :) I understand.

kjhman21:
```suggestion
				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization.", "msg", msg, "err", err)
```

node/cn/handler.go
```diff
@@ -461,7 +461,7 @@ func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, add
 			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
 			if err != nil {
-				p.GetP2PPeer().Log().Error("ProtocolManager failed to handle consensus message", "msg", msg, "err", err)
+				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization.", "msg", msg, "err", err)
```
jeonghonah: I'm not sure we can check the block sync here but how about logging error if block sync is not happened?

andybclee: @jeonghonah At the moment, the only error that can be returned by `HandleMsg` is `ErrStoppedEngine` and the error only can be happened when a downloader stopped the consensus engine.

I agree that this can be improved if other error cases are added, but I don't think it can be done at the moment

PR #202, opened 2019-09-03
Add test code to check if all council members are in the recipient list

andybclee:
## Proposed changes

- `backend.GossipSubPeer()` decides who will receive consensus messages from the node. Though the current function signature is set to return an error, but it doesn't. Only returns nil.
- To test the result, I changed `GossipSubPeer` to return a map of recipient addresses
- And test code for the test set is added

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

## Further comments

Added test cases are included in `TC-Consensus` sheet.
Applying connection status can be added in near future by using a mock object.

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -218,7 +218,7 @@ func (sb *backend) Gossip(valSet istanbul.ValidatorSet, payload []byte) error {
 	return nil
 }
 
-// checkInSubList check if the node is in a sublist
+// checkInSubList checks if the node is in a sublist
 func (sb *backend) checkInSubList(prevHash common.Hash, valSet istanbul.ValidatorSet) bool {
 	for _, val := range valSet.SubList(prevHash, sb.currentView.Load().(*istanbul.View)) {
 		if val.Address() == sb.Address() {
@@ -240,7 +240,7 @@ func (sb *backend) getTargetReceivers(valSet istanbul.ValidatorSet) map[common.A
 	return targets
 }
 
-// Broadcast implements istanbul.Backend.Gossip
+// GossipSubPeer implements istanbul.Backend.Gossip
 func (sb *backend) GossipSubPeer(prevHash common.Hash, valSet istanbul.ValidatorSet, payload []byte) map[common.Address]bool {
 	if !sb.checkInSubList(prevHash, valSet) {
 		return nil
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -218,28 +218,38 @@ func (sb *backend) Gossip(valSet istanbul.ValidatorSet, payload []byte) error {
 	return nil
 }
 
-// Broadcast implements istanbul.Backend.Gossip
-func (sb *backend) GossipSubPeer(prevHash common.Hash, valSet istanbul.ValidatorSet, payload []byte) error {
-	isInSubList := false
+// checkInSubList check if the node is in a sublist
+func (sb *backend) checkInSubList(prevHash common.Hash, valSet istanbul.ValidatorSet) bool {
 	for _, val := range valSet.SubList(prevHash, sb.currentView.Load().(*istanbul.View)) {
 		if val.Address() == sb.Address() {
-			isInSubList = true
-			break
+			return true
 		}
 	}
-	if !isInSubList {
+	return false
+}
+
+// getTargetReceivers returns a map of nodes which need to receive a message
+func (sb *backend) getTargetReceivers(valSet istanbul.ValidatorSet) map[common.Address]bool {
+	targets := make(map[common.Address]bool)
+
+	for _, val := range valSet.List() {
+		if val.Address() != sb.Address() {
+			targets[val.Address()] = true
+		}
+	}
+	return targets
+}
+
+// Broadcast implements istanbul.Backend.Gossip
+func (sb *backend) GossipSubPeer(prevHash common.Hash, valSet istanbul.ValidatorSet, payload []byte) map[common.Address]bool {
+	if !sb.checkInSubList(prevHash, valSet) {
 		return nil
 	}
 
 	hash := istanbul.RLPHash(payload)
 	sb.knownMessages.Add(hash, true)
 
-	targets := make(map[common.Address]bool)
-	for _, val := range valSet.List() {
-		if val.Address() != sb.Address() {
-			targets[val.Address()] = true
-		}
-	}
+	targets := sb.getTargetReceivers(valSet)
 
 	if sb.broadcaster != nil && len(targets) > 0 {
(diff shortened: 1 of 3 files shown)


consensus/istanbul/backend/backend.go
```diff
@@ -218,28 +218,38 @@ func (sb *backend) Gossip(valSet istanbul.ValidatorSet, payload []byte) error {
 	return nil
 }
 
-// Broadcast implements istanbul.Backend.Gossip
-func (sb *backend) GossipSubPeer(prevHash common.Hash, valSet istanbul.ValidatorSet, payload []byte) error {
-	isInSubList := false
+// checkInSubList check if the node is in a sublist
```
ehnuje:
```suggestion
// checkInSubList checks if the node is in a sublist
```

consensus/istanbul/backend/backend_test.go
```diff
@@ -0,0 +1,352 @@
...
+	config.Istanbul = governance.GetDefaultIstanbulConfig()
+	return config
+}
+
+func getGovernance(dbm database.DBManager) *governance.Governance {
+	config := getTestConfig()
+	return governance.NewGovernance(config, dbm)
+}
+
+// Test_GossipSubPeerTargets test the gossiping targets are same as council members
```
ehnuje:
```suggestion
// Test_GossipSubPeerTargets checks if the gossiping targets are same as council members
```

consensus/istanbul/backend/backend_test.go
```diff
@@ -0,0 +1,352 @@
...
+	PRIVKEY = "ce7671a2880493dfb8d04218707a16b1532dfcac97f0289d770a919d5ff7b068"
+	// testing node should be in a committee in these blocks
+	committeeBlocks = []int64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22}
+	// Max blockNum
+	maxBlockNum = int64(100)
+)
+
+func getTestCouncil() []common.Address {
+	return []common.Address{
+		common.HexToAddress("0x414790CA82C14A8B975cEBd66098c3dA590bf969"), // Node Address for test
```
ehnuje: Are these pre-generated addresses needed for the test?

consensus/istanbul/backend/backend.go
```diff
@@ -218,28 +218,38 @@ func (sb *backend) Gossip(valSet istanbul.ValidatorSet, payload []byte) error {
...
+
 	for _, val := range valSet.List() {
 		if val.Address() != sb.Address() {
 			targets[val.Address()] = true
 		}
 	}
+	return targets
+}
+
+// Broadcast implements istanbul.Backend.Gossip
```
jeonghonah:
```suggestion
// GossipSubPeer implements istanbul.Backend.Gossip
```

consensus/istanbul/backend/backend_test.go
```diff
@@ -0,0 +1,352 @@
...
+	"github.com/klaytn/klaytn/governance"
+	"github.com/klaytn/klaytn/networks/p2p"
+	"github.com/klaytn/klaytn/params"
+	"github.com/klaytn/klaytn/reward"
+	"github.com/klaytn/klaytn/storage/database"
+	"math/big"
+	"testing"
+)
+
+var (
```
jeonghonah:
```suggestion
const (
```

consensus/istanbul/backend/backend_test.go
```diff
@@ -0,0 +1,352 @@
...
+	// testing node's private key
+	PRIVKEY = "ce7671a2880493dfb8d04218707a16b1532dfcac97f0289d770a919d5ff7b068"
+	// testing node should be in a committee in these blocks
+	committeeBlocks = []int64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22}
```
jeonghonah: Is this array necessary? I think we can simply set to `22` and use it for iteration.

consensus/istanbul/backend/backend_test.go
```diff
@@ -0,0 +1,352 @@
...
+		common.HexToAddress("0xCB1c4478cbE37a4B7a1510065802e0B669979cFD"),
+		common.HexToAddress("0xf7ed2347390b1f742DA9785242dF90BbbC8Af90C"),
+		common.HexToAddress("0x8BC37790112C66DF94cB05dC5168249a7e6d6717"),
+		common.HexToAddress("0x9087C8dd3dE26B6DB3a471D7254d78FB1afCbeF2"),
+		common.HexToAddress("0xcAfaA922f46d0a0FcE13B0f2aA822aF9C4b9a31C"),
+		common.HexToAddress("0xa21b5d69efDFE56a9C272f7957Bd6d4205a1e6Ff"),
+	}
+}
+
+func getTestVotingPowers(num int) []uint64 {
```
jeonghonah: `num` is not used.

consensus/istanbul/backend/backend_test.go
andybclee: @jeonghonah You are right at the moment. But if we change  some logic in the future, the array may not be so linear as shown now. So I want the testing code itself not to be changed.

consensus/istanbul/backend/backend_test.go
andybclee: @ehnuje Actually, it can be generated programmatically. But I worried that some kind of hard fork may happen even though it passed the test. I think testing with a hard coded data based test result would be more helpful to know if it make some hard fork status.

consensus/istanbul/backend/backend_test.go
andybclee: @jeonghonah My bad. I'll fix it.

consensus/istanbul/backend/backend_test.go
ehnuje: I'm sorry but I could hardly think about the situation when this data affects the test when a hard fork happens. Could you please explain in a bit detail?

andybclee: @ehnuje My idea is simple. If the data which passed before can't pass the test now, it is clear that there is some problem. If the test data is generated dynamically and the test would fail, it will be much tough to know what is a problem. And it also make harder to determine the result of `SubList()` is right or not.

ehnuje: Thanks for the explanation. I understood your point.

consensus/istanbul/backend/backend_test.go
andybclee: @jeonghonah Sorry. Slice and arrays can't be a constant in nature in Golang. I'll revert it back to `var`

PR #219, opened 2019-09-10
[SC] set last proposer to a new valset

laggu:
## Proposed changes

This pr is related with round robin policy.

- set last proposer to a new validator set
- only use round to calc next proposer

default validator set for round robin has copy() function.
but in the copy function it makes a new validator set.
Therefore, the proposer of the last block is missed on a new validator set.
the change is to set the last proposer to the new validator set.

also,
To calculate the next proposer in round robin policy, only current round is required.
There is no need to add 1

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/consensus/istanbul/validator/default.go
+++ b/consensus/istanbul/validator/default.go
@@ -385,7 +385,8 @@ func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
 	}
 
 	newValSet := NewSubSet(addresses, valSet.policy, valSet.subSize).(*defaultSet)
-	newValSet.proposer = valSet.proposer
+	_, proposer := newValSet.GetByAddress(valSet.proposer.Address())
+	newValSet.proposer = proposer
 	return newValSet
 }
 
--- a/consensus/istanbul/validator/default.go
+++ b/consensus/istanbul/validator/default.go
@@ -163,7 +163,7 @@ func (valSet *defaultSet) SubList(prevHash common.Hash, view *istanbul.View) []i
 	subset := make([]istanbul.Validator, valSet.subSize)
 	subset[0] = valSet.GetProposer()
 	// next proposer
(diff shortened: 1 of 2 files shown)


consensus/istanbul/validator/default.go
```diff
@@ -222,7 +222,7 @@ func (valSet *defaultSet) SubListWithProposer(prevHash common.Hash, proposer com
 	subset := make([]istanbul.Validator, valSet.subSize)
 	subset[0] = New(proposer)
 	// next proposer
-	subset[1] = valSet.selector(valSet, subset[0].Address(), view.Round.Uint64()+1)
+	subset[1] = valSet.selector(valSet, subset[0].Address(), view.Round.Uint64())
```
jeonghonah: How about updating parameters of the selector method?
`func roundRobinProposer(valSet istanbul.ValidatorSet, curProposer common.Address, curRound uint64)`

consensus/istanbul/validator/default.go
```diff
@@ -383,7 +383,10 @@ func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
 	for _, v := range valSet.validators {
 		addresses = append(addresses, v.Address())
 	}
-	return NewSubSet(addresses, valSet.policy, valSet.subSize)
+
+	newValSet := NewSubSet(addresses, valSet.policy, valSet.subSize).(*defaultSet)
+	newValSet.proposer = valSet.proposer
```
jeonghonah: It will be nice if we add unit tests for `Copy() and SubList()`.

KimKyungup: We don't need to hard copy for proposer?

laggu: reflected. I've added unit tests

laggu: I changed the logic!

PR #221, opened 2019-09-11
sc: Introduce test cases of MainBridge functions

aidan-kwon:
## Proposed changes

`TestMainBridge_removePeer` tests correct removal of a peer from `MainBridge.peers`.
`TestMainBridge_handleMsg` fails when a bridgePeer fails to read a message or reads a too long message.
## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/node/sc/mainbridge_test.go
+++ b/node/sc/mainbridge_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/klaytn/klaytn/blockchain"
 	"github.com/klaytn/klaytn/event"
 	"github.com/klaytn/klaytn/networks/p2p"
+	"github.com/klaytn/klaytn/networks/p2p/discover"
 	"github.com/klaytn/klaytn/networks/rpc"
 	"github.com/klaytn/klaytn/node"
 	"github.com/klaytn/klaytn/node/cn"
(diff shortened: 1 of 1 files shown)


node/sc/mainbridge_test.go
```diff
@@ -116,3 +117,79 @@ func TestMainBridge_basic(t *testing.T) {
...
 }
+
+// TestMainBridge_removePeer tests correct removal of a peer from `MainBridge.peers`.
+func TestMainBridge_removePeer(t *testing.T) {
+	// Create a MainBridge (it may have 0 peers)
+	mBridge := testNewMainBridge(t)
+	defer mBridge.chainDB.Close()
+
+	// Prepare a bridgePeer to be added and removed
+	nodeID := "0x1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d1d"
```
KimKyungup: How about like this?
```
	key1, _ = crypto.GenerateKey()
	addr1 = crypto.PubkeyToAddress(key1.PublicKey)
	nodeID1 = discover.PubkeyID(&key1.PublicKey)
```

node/sc/mainbridge_test.go
```diff
@@ -116,3 +117,79 @@ func TestMainBridge_basic(t *testing.T) {
...
+	pipe1, pipe2 := p2p.MsgPipe()
+
+	// bridgePeer will receive a message through rw1
+	bridgePeer := newBridgePeer(testProtocolVersion, peer, pipe1)
+
+	// Case1. Send a valid message and handle it successfully
+	{
+		data := "valid message"
+		go func() {
+			if err := p2p.Send(pipe2, StatusMsg, data); err != nil {
```
KimKyungup: Should this routine be executed with goroutine?
There is no guarantee that this code will run before `handleMsg` routine caused by goroutine.

aidan-kwon: 1. It needs to be a goroutine because `p2p.Send` is a blocking function.
2. `mBridge.handleMsg` will wait until it receives a channel message from another pipe.

node/sc/mainbridge_test.go
aidan-kwon: Thank you for the suggestion.
Random nodeID may cause a test failure in line 138 `mBridge.removePeer("0x11111111")`(The possiblity is very low though).
So, I didn't change the code here, changed the same code in line 153.

node/sc/mainbridge_test.go
```diff
@@ -116,3 +118,80 @@ func TestMainBridge_basic(t *testing.T) {
...
+				t.Fatal(err)
+			}
+		}()
+
+		if err := mBridge.handleMsg(bridgePeer); err != nil {
+			t.Fatal(err)
```
ehnuje: Are there any reason between using `t.Fatal` and `assert`?

aidan-kwon: I use `t.Fatal` only when some code doesn't expect any specific value except `nil`. It means that the failure of the code is not a part of the test I intended. 

This kind of convention is just my preference. So, I can change it if the change gives more consistency for our code. How do you think @ehnuje ?

ehnuje: I also agree with your point =) Thanks for the explanation.

PR #245, opened 2019-09-23
cn: added node/cn/handler_msg_test.go

ehnuje:
## Proposed changes

- handler_msg_test.go has beed added to update message handling tests

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/cn/handler_msg_test.go
+++ b/node/cn/handler_msg_test.go
@@ -1,0 +1,121 @@
+// Copyright 2019 The klaytn Authors
+// This file is part of the klaytn library.
+//
+// The klaytn library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The klaytn library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+package cn
+
+import (
+	"github.com/golang/mock/gomock"
+	"github.com/klaytn/klaytn/blockchain/types"
+	"github.com/klaytn/klaytn/networks/p2p"
+	mocks2 "github.com/klaytn/klaytn/node/cn/mocks"
+	"github.com/klaytn/klaytn/ser/rlp"
+	"github.com/klaytn/klaytn/work/mocks"
+	"github.com/stretchr/testify/assert"
+	"math/big"
+	"sync/atomic"
+	"testing"
+)
+
+// generateMsg creates a message struct for message handling tests.
+func generateMsg(t *testing.T, msgCode uint64, data interface{}) p2p.Msg {
+	size, r, err := rlp.EncodeToReader(data)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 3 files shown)


node/cn/handler_test.go
```diff
@@ -47,9 +47,13 @@ var addrs []common.Address
 var keys []*ecdsa.PrivateKey
 var nodeids []discover.NodeID
 
-var tx *types.Transaction
+var tx1 *types.Transaction
```
nogang: tx1? Will you have tx2 and tx3 later?

nogang: So don't people usually use arrays in this case? Why do we create a single variable and add a number to the name?

ehnuje: You're right. I will use a slice if needed. In this case, while creating test, I used `tx2` but removed it since it was not needed anymore. I'lll reflect your comment in next PR.

node/cn/handler_test.go
```diff
@@ -61,8 +65,14 @@ func init() {
 		nodeids[i] = discover.PubkeyID(&keys[i].PublicKey)
 	}
 
-	tx = types.NewTransaction(111, addrs[0], big.NewInt(111), 111, big.NewInt(111), nil)
-	txs = types.Transactions{tx}
+	signer := types.MakeSigner(params.BFTTestChainConfig, big.NewInt(2019))
```
KimKyungup: 2nd argument is not used in the function. I was used as a block number.

```suggestion
	signer := types.MakeSigner(params.BFTTestChainConfig, nil)
```

ehnuje: Unless we change the function `types.MakeSigner`, I think there's no need to give `nil` to the second parameter. If so, we'd better create a PR to change the interface and related usages. How do you think?

KimKyungup: Yes, we need to clean the function. Anyway, 2nd parameter is not chain id, it is block number. So I thought the 2019 is weird.

KimKyungup: Anyway. In next PR, let's clean the function. :)

node/cn/handler_msg_test.go
```diff
@@ -0,0 +1,121 @@
...
+	defer mockCtrl.Finish()
+
+	_, msg, mockPeer, mockFetcher := prepareTestHandleNewBlockMsg(t, mockCtrl, blockNum1)
+
+	pm := &ProtocolManager{}
+	pm.fetcher = mockFetcher
+
+	mockPeer.EXPECT().Head().Return(hash1, big.NewInt(blockNum1+1)).AnyTimes()
+
+	assert.NoError(t, handleNewBlockMsg(pm, mockPeer, msg))
```
jeonghonah: Minor comment:
How about adding an error case for `ErrDecode`?

node/cn/handler_msg_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	// If pm.acceptTxs == 1, TxPool.HandleTxMsg is called.
+	{
+		atomic.StoreUint32(&pm.acceptTxs, 1)
+		mockTxPool := mocks.NewMockTxPool(mockCtrl)
+		mockTxPool.EXPECT().HandleTxMsg(gomock.Eq(txs)).AnyTimes()
+		pm.txpool = mockTxPool
+
+		mockPeer.EXPECT().AddToKnownTxs(txs[0].Hash()).Times(1)
+		assert.NoError(t, handleTxMsg(pm, mockPeer, msg))
```
jeonghonah: Minor comment:
How about adding a `tx == nil in txs` case?

ehnuje: I will add more tests in other PRs. Thanks!

PR #253, opened 2019-09-25
SC: Support Token/KLAY locking feature to prevent request value transfer

KimKyungup:
## Proposed changes
This PR introduces the token/KLAY locking to prevent request value transfer.

This enables bridge contract owner to lock each token and KLAY.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/contracts/bridge/BridgeTokens.sol
+++ b/contracts/bridge/BridgeTokens.sol
@@ -1,0 +1,113 @@
+// Copyright 2019 The klaytn Authors
+// This file is part of the klaytn library.
+//
+// The klaytn library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The klaytn library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.5.6;
+
+import "../externals/openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+contract BridgeTokens is Ownable {
+    mapping(address => address) public registeredTokens; // <token, counterpart token>
+    mapping(address => uint) public indexOfTokens; // <token, index>
+    address[] public registeredTokenList;
+    mapping(address => bool) public lockedTokens;
+
+    event TokenRegistered(address indexed token);
+    event TokenDeregistered(address indexed token);
+    event TokenLocked(address indexed token);
+    event TokenUnlocked(address indexed token);
+
+    modifier onlyRegisteredToken(address _token) {
+        require(registeredTokens[_token] != address(0), "not allowed token");
+        _;
+    }
+
+    modifier onlyNotRegisteredToken(address _token) {
+        require(registeredTokens[_token] == address(0), "allowed token");
+        _;
+    }
+
+    modifier onlyLockedToken(address _token) {
+        require(lockedTokens[_token] == true, "unlocked token");
+        _;
+    }
+
+    modifier onlyUnlockedToken(address _token) {
+        require(lockedTokens[_token] == false, "locked token");
+        _;
+    }
+
+    function getRegisteredTokenList() external view returns(address[] memory) {
+        return registeredTokenList;
+    }
+
+    // registerToken can update the allowed token with the counterpart token.
+    function registerToken(address _token, address _cToken)
+        external
+        onlyOwner
+        onlyNotRegisteredToken(_token)
+    {
+        registeredTokens[_token] = _cToken;
+        indexOfTokens[_token] = registeredTokenList.length;
+        registeredTokenList.push(_token);
+
+        emit TokenRegistered(_token);
+    }
+
+    // deregisterToken can remove the token in registeredToken list.
+    function deregisterToken(address _token)
+        external
+        onlyOwner
+        onlyRegisteredToken(_token)
+    {
+        delete registeredTokens[_token];
+        delete lockedTokens[_token];
+
+        uint idx = indexOfTokens[_token];
+        delete indexOfTokens[_token];
+
+        if (idx < registeredTokenList.length-1) {
+            registeredTokenList[idx] = registeredTokenList[registeredTokenList.length-1];
+            indexOfTokens[registeredTokenList[idx]] = idx;
+        }
+        registeredTokenList.length--;
+
+        emit TokenDeregistered(_token);
+    }
+
+    // lockToken can lock the token to prevent request token transferring.
+    function lockToken(address _token)
+        external
+        onlyOwner
+        onlyRegisteredToken(_token)
+        onlyUnlockedToken(_token)
+    {
+        lockedTokens[_token] = true;
+
+        emit TokenLocked(_token);
+    }
+
+    // unlockToken can unlock the token to  request token transferring.
+    function unlockToken(address _token)
+        external
+        onlyOwner
+        onlyRegisteredToken(_token)
+        onlyLockedToken(_token)
+    {
+        delete lockedTokens[_token];
+
+        emit TokenUnlocked(_token);
+    }
+}
--- a/contracts/bridge/BridgeTransferKLAY.sol
+++ b/contracts/bridge/BridgeTransferKLAY.sol
@@ -20,6 +20,43 @@ import "./BridgeTransferCommon.sol";
 
 
 contract BridgeTransferKLAY is BridgeTransfer {
+    bool public isLockedKLAY;
+
+    event KLAYLocked();
+    event KLAYUnlocked();
+
+    modifier lockedKLAY {
+        require(isLockedKLAY == true, "unlocked");
+        _;
+    }
+
+    modifier unlockedKLAY {
+        require(isLockedKLAY == false, "locked");
+        _;
+    }
+
+    // lockKLAY can to prevent request KLAY transferring.
+    function lockKLAY()
+        external
+        onlyOwner
+        unlockedKLAY
+    {
+        isLockedKLAY = true;
+
+        emit KLAYLocked();
+    }
+
+    // unlockToken can allow request KLAY transferring.
+    function unlockKLAY()
+        external
+        onlyOwner
+        lockedKLAY
+    {
+        isLockedKLAY = false;
+
+        emit KLAYUnlocked();
+    }
+
     // handleKLAYTransfer sends the KLAY by the request.
     function handleKLAYTransfer(
         bytes32 _requestTxHash,
@@ -56,7 +93,10 @@ contract BridgeTransferKLAY is BridgeTransfer {
     }
 
     // _requestKLAYTransfer requests transfer KLAY to _to on relative chain.
-    function _requestKLAYTransfer(address _to, uint256 _feeLimit,  bytes memory _extraData) internal {
+    function _requestKLAYTransfer(address _to, uint256 _feeLimit,  bytes memory _extraData)
+        internal
+        unlockedKLAY
+    {
         require(isRunning, "stopped bridge");
         require(msg.value > _feeLimit, "insufficient amount");
 
--- a/node/sc/bridge_test.go
+++ b/node/sc/bridge_test.go
@@ -38,8 +38,7 @@ import (
 )
 
 const (
-	gasLimit uint64 = 3000000         // gasLimit for contract transaction.
-	timeOut         = 3 * time.Second // timeout of context and event loop for simulated backend.
+	timeOut = 3 * time.Second // timeout of context and event loop for simulated backend.
 )
 
 // WaitMined waits the tx receipt until the timeout.
@@ -81,7 +80,7 @@ func TransferSignedTx(auth *bind.TransactOpts, backend *backends.SimulatedBacken
 		nonce,
 		to,
 		value,
-		gasLimit,
+		DefaultBridgeTxGasLimit,
 		gasPrice,
 		nil)
 
@@ -109,7 +108,7 @@ func TransferSignedTx(auth *bind.TransactOpts, backend *backends.SimulatedBacken
 
 // RequestKLAYTransfer sends a requestValueTransfer transaction to the bridge contract.
 func RequestKLAYTransfer(b *bridge.Bridge, auth *bind.TransactOpts, to common.Address, value uint64, t *testing.T) {
-	_, err := b.RequestKLAYTransfer(&bind.TransactOpts{From: auth.From, Signer: auth.Signer, GasLimit: gasLimit, Value: new(big.Int).SetUint64(value)}, to, new(big.Int).SetUint64(value), nil)
+	_, err := b.RequestKLAYTransfer(&bind.TransactOpts{From: auth.From, Signer: auth.Signer, GasLimit: DefaultBridgeTxGasLimit, Value: new(big.Int).SetUint64(value)}, to, new(big.Int).SetUint64(value), nil)
 	if err != nil {
 		t.Fatalf("fail to RequestKLAYTransfer %v", err)
 	}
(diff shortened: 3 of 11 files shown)


KimKyungup:
@jeonghonah @nogang @aidan-kwon PTAL. :) cc @laggu

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,103 @@
...
+        emit TokenLocked(_token);
+    }
+
+    // unlockToken can unlock the token to  request token transferring.
+    function unlockToken(address _token)
+        external
+        onlyOwner
+        lockedToken(_token)
+    {
+        require(allowedTokens[_token] != address(0));
```
aidan-kwon: It is already checked in the modifier `lockedToken`. 
By the way, I cannot sure we should check `allowedTokens` inside of the modifier.

node/sc/bridge_test.go
```diff
@@ -665,3 +665,310 @@ func TestExtendedBridgeAndCallbackERC721(t *testing.T) {
...
+
+	// check value after unlocking
+	isLocked, err := b.LockedTokens(nil, erc20Addr)
+	assert.NoError(t, err)
+	assert.Equal(t, false, isLocked)
+
+	isLocked, err = b.LockedTokens(nil, erc721Addr)
+	assert.NoError(t, err)
+	assert.Equal(t, false, isLocked)
+
```
aidan-kwon: Checking for KLAY is missing.
"""
	isLocked, err = b.IsLockedKLAY(nil)
	assert.NoError(t, err)
	assert.Equal(t, false, isLocked)
"""

node/sc/bridge_test.go
```diff
@@ -665,3 +665,310 @@ func TestExtendedBridgeAndCallbackERC721(t *testing.T) {
...
+	tx, err = b.UnlockKLAY(operator)
+	assert.NoError(t, err)
+	backend.Commit()
+	CheckReceipt(backend, tx, 1*time.Second, types.ReceiptStatusErrExecutionReverted, t)
+}
+
+// TestBridgeContract_CheckValueTransferAfterUnLock checks the following:
+// - the token can be unlock to allow value transfer requests.
+func TestBridgeContract_CheckValueTransferAfterUnLock(t *testing.T) {
+	//backend, operator, tester, b, erc20, erc721, erc20Addr, erc721Addr := generateBridgeTokenTestEnv(t)
```
aidan-kwon:
```suggestion
```

node/sc/bridge_test.go
```diff
@@ -665,3 +665,310 @@ func TestExtendedBridgeAndCallbackERC721(t *testing.T) {
...
+
+	isLocked, err = b.IsLockedKLAY(nil)
+	assert.NoError(t, err)
+	assert.Equal(t, true, isLocked)
+
+	// check to prevent value transfer
+	tx, err = erc20.RequestValueTransfer(tester, big.NewInt(1), operator.From, big.NewInt(0), nil)
+	assert.NoError(t, err)
+	backend.Commit()
+	assert.NotNil(t, bind.CheckWaitMined(backend, tx))
```
aidan-kwon: How can we know whether the transfer is prevented?

KimKyungup: This `CheckWaitMined` will return error when the tx is reverted.

contracts/bridge/BridgeTokens.sol
```diff
@@ -65,6 +68,7 @@ contract BridgeTokens is Ownable {
         delete allowedTokens[_token];
         delete lockedTokens[_token];
 
+        // This iteration consumes 818 gas. 5,000,000 gas limit is enough for the iteration of maxTokenCnt.
```
aidan-kwon: It would be better if you specify the value of `allowedTokenList.length` consuming 818 gas.

node/sc/bridge_test.go
```diff
@@ -665,3 +665,313 @@ func TestExtendedBridgeAndCallbackERC721(t *testing.T) {
...
+	tx, err = erc721.RequestValueTransfer(tester, big.NewInt(1), operator.From, nil)
+	assert.NoError(t, err)
+	backend.Commit()
+	assert.Nil(t, bind.CheckWaitMined(backend, tx))
+
+	tester.Value = big.NewInt(1)
+	tx, err = b.RequestKLAYTransfer(tester, tester.From, big.NewInt(0), nil)
+	assert.NoError(t, err)
+	backend.Commit()
+	assert.NotNil(t, bind.CheckWaitMined(backend, tx))
```
aidan-kwon: Why it doesn't return an error?

KimKyungup: Receipt status can be several value. But now there are not all errors corresponded with all receipt status. So it just return success or not like below.

```
// CheckWaitMined returns an error if the transaction is not found or not successful status.
func CheckWaitMined(b DeployBackend, tx *types.Transaction) error {
	timeoutContext, cancelTimeout := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancelTimeout()

	receipt, err := WaitMined(timeoutContext, b, tx)
	if err != nil {
		return err
	}

	if receipt == nil {
		return errors.New("receipt not found")
	}

	if receipt.Status != types.ReceiptStatusSuccessful {
		return errors.New("not successful receipt")
	}
	return nil
}
```

contracts/bridge/BridgeTokens.sol
KimKyungup: Got it. How about this?
```suggestion
        // This iteration consumes 818 gas. 81800 gas is needed for the maxTokenCnt iteration.
```

aidan-kwon: Looks good. Thank you.

KimKyungup: reflected : ) Thanks.

node/sc/bridge_test.go
aidan-kwon: I means.... Why it is not `assert.Nil(t, bind.CheckWaitMined(backend, tx))`?
Since the KLAY transfer is unlocked now, the backend should process the transfer succesfully.

KimKyungup: It's my mistake. :) Thank you very much. I fixed. HAHAHA

KimKyungup:
@aidan-kwon I squashed all commit for cherry-picking. :-)

contracts/bridge/BridgeTransferKLAY.sol
```diff
@@ -20,6 +20,43 @@ import "./BridgeTransferCommon.sol";
 
 
 contract BridgeTransferKLAY is BridgeTransfer {
+    bool public isLockedKLAY;
+
+    event KLAYLocked();
+    event KLAYUnlocked();
+
+    modifier lockedKLAY {
+        require(isLockedKLAY == true, "already unlocked");
```
jeonghonah: Minor comment: how about this?
```suggestion
        require(isLockedKLAY == true, "unlocked");
```

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,104 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.5.6;
+
+import "../externals/openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+contract BridgeTokens is Ownable {
+    uint constant public maxTokenCnt = 100; // This count was derived by loop iteration gas consumption.
```
jeonghonah:
```suggestion
    uint constant private maxTokenCnt = 100; // This count was derived by loop iteration gas consumption.
```

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,104 @@
...
+    address[] public allowedTokenList;
+    mapping(address => bool) public lockedTokens;
+
+    event TokenRegistered(address indexed token);
+    event TokenDeregistered(address indexed token);
+    event TokenLocked(address indexed token);
+    event TokenUnlocked(address indexed token);
+
+    modifier lockedToken(address _token) {
+        require(allowedTokens[_token] != address(0), "invalid token");
```
jeonghonah: Minor comment:
How about making a new modifier?

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.5.6;
+
+import "../externals/openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+contract BridgeTokens is Ownable {
+    uint constant private maxTokenCnt = 1000; // This count was derived by loop iteration gas consumption.
+    mapping(address => address) public allowedTokens; // <token, counterpart token>
```
jeonghonah: How about `registeredTokens`?

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+    address[] public allowedTokenList;
+    mapping(address => bool) public lockedTokens;
+
+    event TokenRegistered(address indexed token);
+    event TokenDeregistered(address indexed token);
+    event TokenLocked(address indexed token);
+    event TokenUnlocked(address indexed token);
+
+    modifier allowedToken(address _token) {
```
jeonghonah: How about `registeredToken`()?

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+        require(allowedTokens[_token] != address(0), "not allowed token");
+        _;
+    }
+
+    modifier notAllowedToken(address _token) {
```
jeonghonah: How about `notRegisteredToken()`?

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+        emit TokenLocked(_token);
+    }
+
+    // unlockToken can unlock the token to  request token transferring.
+    function unlockToken(address _token)
+        external
+        onlyOwner
+        lockedToken(_token)
+    {
+        lockedTokens[_token] = false;
```
jeonghonah:
```suggestion
        delete lockedTokens[_token];
```

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+        require(allowedTokens[_token] != address(0), "not allowed token");
+        _;
+    }
+
+    modifier notAllowedToken(address _token) {
+        require(allowedTokens[_token] == address(0), "allowed token");
+        _;
+    }
+
+    modifier lockedToken(address _token) {
```
jeonghonah:
```suggestion
    modifier onlyLockedToken(address _token) {
```

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+        require(lockedTokens[_token] == true, "unlocked token");
+        _;
+    }
+
+    modifier unlockedToken(address _token) {
```
jeonghonah:
```suggestion
    modifier onlyUnlockedToken(address _token) {
```

contracts/bridge/BridgeTransferKLAY.sol
```diff
@@ -20,6 +20,43 @@ import "./BridgeTransferCommon.sol";
 
 
 contract BridgeTransferKLAY is BridgeTransfer {
+    bool public isLockedKLAY;
+
+    event KLAYLocked();
+    event KLAYUnlocked();
+
+    modifier lockedKLAY {
```
jeonghonah:
```suggestion
    modifier onlyLockedKLAY {
```

contracts/bridge/BridgeTransferKLAY.sol
```diff
@@ -20,6 +20,43 @@ import "./BridgeTransferCommon.sol";
...
+        require(isLockedKLAY == true, "unlocked");
+        _;
+    }
+
+    modifier unlockedKLAY {
```
jeonghonah:
```suggestion
    modifier onlyUnlockedKLAY {
```

node/sc/bridge_test.go
```diff
@@ -38,8 +38,8 @@ import (
 )
 
 const (
-	gasLimit uint64 = 3000000         // gasLimit for contract transaction.
-	timeOut         = 3 * time.Second // timeout of context and event loop for simulated backend.
+	gasLimit uint64 = DefaultBridgeTxGasLimit // gasLimit for contract transaction.
```
jeonghonah: How about just using `DefaultBridgeTxGasLimit`?

node/sc/bridge_test.go
```diff
@@ -665,3 +665,313 @@ func TestExtendedBridgeAndCallbackERC721(t *testing.T) {
 		t.Fatalf("registerOfferEvent was not found.")
 	}
 }
+
+func generateBridgeTokenTestEnv(t *testing.T) (*backends.SimulatedBackend, *bind.TransactOpts, *bind.TransactOpts, *bridge.Bridge, *sctoken.ServiceChainToken, *scnft.ServiceChainNFT, common.Address, common.Address) {
```
jeonghonah: How about simply returning struct?

node/sc/bridge_test.go
```diff
@@ -665,3 +665,313 @@ func TestExtendedBridgeAndCallbackERC721(t *testing.T) {
...
+	CheckReceipt(backend, tx, 1*time.Second, types.ReceiptStatusErrExecutionReverted, t)
+
+	// fail locking unregistered token.
+	testAddr := common.BytesToAddress([]byte("unregistered"))
+	tx, err = b.LockToken(operator, testAddr)
+	assert.NoError(t, err)
+	backend.Commit()
+	CheckReceipt(backend, tx, 1*time.Second, types.ReceiptStatusErrExecutionReverted, t)
+
+	// fail locking token if it is already locked.
```
laggu: Are those tokens and klay already locked??

node/sc/bridge_test.go
```diff
@@ -665,3 +665,313 @@ func TestExtendedBridgeAndCallbackERC721(t *testing.T) {
...
+
+	return backend, operator, tester, b, erc20, erc721, erc20Addr, erc721Addr
+}
+
+// TestBridgeContract_InitStatus checks initial lock status.
+func TestBridgeContract_InitStatus(t *testing.T) {
+	_, _, _, b, _, _, erc20Addr, erc721Addr := generateBridgeTokenTestEnv(t)
+
+	// initial value check
+	isLocked, err := b.LockedTokens(nil, erc20Addr)
```
laggu: How about changing the name to IsLockedTokens to match the name with IsLockedKLAY?

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+
+pragma solidity ^0.5.6;
+
+import "../externals/openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+contract BridgeTokens is Ownable {
+    uint constant private maxTokenCnt = 1000; // This count was derived by loop iteration gas consumption.
+    mapping(address => address) public allowedTokens; // <token, counterpart token>
+    address[] public allowedTokenList;
+    mapping(address => bool) public lockedTokens;
```
laggu: Is using 2 mapping cheaper than using struct???
How about using struct like this?
struct token {
  tokenAddr
  counterPartAddr
  lock
}
mapping(address => token)

KimKyungup: It is good idea. But it can't return the list of the token. So I think this way is better than your suggestion.
What do you think? :)

contracts/bridge/BridgeTransferKLAY.sol
KimKyungup: I think KLAY is one. So I think I don't need to write `only` for this.
What do you think? @jeonghonah

contracts/bridge/BridgeTransferKLAY.sol
KimKyungup: Same here.

node/sc/bridge_test.go
KimKyungup: There are many tokens. So I wrote. What do you think? :)

KimKyungup:
@jeonghonah @aidan-kwon @laggu I reflect your review. PTAL. :)
Especially, iteration code was removed by `tokenIndex`. Please review this code more. Thanks.

contracts/bridge/BridgeTransferKLAY.sol
jeonghonah: It's okay :)

contracts/bridge/BridgeTransferKLAY.sol
jeonghonah: It's okay :)

contracts/bridge/BridgeTokens.sol
```diff
@@ -0,0 +1,113 @@
...
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.5.6;
+
+import "../externals/openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+contract BridgeTokens is Ownable {
+    mapping(address => address) public registeredTokens; // <token, counterpart token>
+    mapping(address => uint) public indexOfTokens; // <token, index>
+    address[] public registeredTokenList;
```
ElliotHan: Why do you need `registeredTokenList` at all?
It seems like its used only for read-only purpose, via function `getRegisteredTokenList`.
If that is the only case, you can make recover the full list without `getRegisteredTokenList` by reading trough the events. 
Using two types of storage (index mapping&list) for a read-only purpose seems a bit costly

KimKyungup: A service provider requested the feature to get the full list of registered token.
Therefore I added this feature. Also, without this list, it is hard to get the list.
As you said, through events, we can get it. But we should track all blocks. I think it is also costly.

ElliotHan: Alright. but FYR, if you use events in linked list form, you don't have to sweep the whole blocks. Since using storage is **the most costly** operation on blockchain, for most cases you better avoid using it, unless you need on chain calculation.

KimKyungup: As we talked in slack, the token `registeredTokenList` is for getting the list easily not saving storage. :) I will keep your idea for other storage issue. Thanks. =)

PR #258, opened 2019-09-26
Fix sendRoundChange to store correct prevHash when sending a message

andybclee:
## Proposed changes

- This PR is to fix the wrong behavior of `sendRoundChange`
- When sending a consensus related message, each node adds `prevHash` in the message. This prevHash is the parent hash of current head block that a node has. Whiling making a consensus, nodes are discussing the next block to be stored, so its prevHash have to be a hash of the previous block.
- But when sending a `roundchange` message, it loads previous block's `parent hash` not like other messages like `preprepare` and `prepare`. 
- So it makes nodes to have difficulties in making a consensus. Because the wrong hash will result in a wrong committee composition. 
- This bug was hidden because we are sending messages to all council member at the moment. 

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/consensus/istanbul/core/roundchange.go
+++ b/consensus/istanbul/core/roundchange.go
@@ -60,7 +60,7 @@ func (c *core) sendRoundChange(round *big.Int) {
 	rc := &istanbul.Subject{
 		View:     cv,
 		Digest:   common.Hash{},
-		PrevHash: lastProposal.ParentHash(),
+		PrevHash: lastProposal.Hash(),
 	}
 
 	payload, err := Encode(rc)


kjhman21:
@andybclee Could you attach the experimental results so that this change won't make a bad effect?

andybclee:
@kjhman21 The attached file is filtered log by using grep. The log is the trace log of `SubListWithProposer` which is called to calculate a committee.

If you look into logs which have `valSet.Number=0` meaning it is trying to build the block 1, you can see two values of `prevHash`. 

Because the `prevHash` is used to decide the message's recipients, it means the message recipients can be different between messages though it is for the same block and round. 
[check2.txt](https://github.com/klaytn/klaytn/files/3668807/check2.txt)

In `sendPreprepare`, the message is composed of as follows. In this case, request.Proposal is the new block which requires to get a consensus to be stored in the chain, so using `ParentHash()` is right.
https://github.com/klaytn/klaytn/blob/381c97fcdd5f35679d6ec0ec0545dc41f0ffb325/consensus/istanbul/core/preprepare.go#L48-L52

In `sendPrepare`, the situation is similar with that of `sendPreprepare`. It is using the `Proposal().ParentHash` and it is right. If we are building a block 100, we need the hash of the block 99. 
https://github.com/klaytn/klaytn/blob/381c97fcdd5f35679d6ec0ec0545dc41f0ffb325/consensus/istanbul/core/prepare.go#L38-L42

In `sendCommit`, same thing happens. It calls `Subject()` and it returns following hash.
https://github.com/klaytn/klaytn/blob/381c97fcdd5f35679d6ec0ec0545dc41f0ffb325/consensus/istanbul/core/roundstate.go#L86-L93

But in `sendRoundChange`, it is like this. Though it seems similar to previous cases, but it is not. 
https://github.com/klaytn/klaytn/blob/381c97fcdd5f35679d6ec0ec0545dc41f0ffb325/consensus/istanbul/core/roundchange.go#L56-L64
The `lastProposal()` returns the head block of the blockchain not like previous cases, the `prevHash` have to be `lastProposal.Hash()` because we are making consensus for the next block of the lastProposal.

The test result after the change from `lastBlock.ParentHash()` to `lastBlock.Hash()` is in https://docs.google.com/presentation/d/1P_vljxCCMcj2XkTL4J41OJ4HH90b84gbAtTCzuWPV50/edit?usp=sharing

PR #272, opened 2019-10-01
Enable EN to send transactions to CN

KimKyungup:
## Proposed changes
This PR enable EN to send transactions to CN directly.
As below description, Service Chain may not have PN. So this PR was introduced.

~~I would like to discuss about this PR.~~

~~We don't allow EN to send tx to CN even if EN connected with CN.~~
~~How about allowing this?~~

In service chain, SP can want to use CN + EN topology. So I think this feature is needed.
However, In cypress, it is impossible CN connects EN cased by firewall. Therefore, this code doesn't make side effect in Cypress.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/cn/handler_test.go
+++ b/node/cn/handler_test.go
@@ -319,7 +319,7 @@ func TestBroadcastTxsFromPN_PN_Exists(t *testing.T) {
 	pm.BroadcastTxs(txs)
 }
 
-func TestBroadcastTxsFromEN_EN_NotExists(t *testing.T) {
+func TestBroadcastTxsFromEN_ALL_NotExists(t *testing.T) {
 	pm := &ProtocolManager{}
 	pm.nodetype = common.ENDPOINTNODE
 	mockCtrl := gomock.NewController(t)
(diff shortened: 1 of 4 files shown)


KimKyungup:
Dear reviews.
I fixed the unit test code. Please take a look. :)

node/cn/peer_set.go
```diff
@@ -440,9 +440,12 @@ func (peers *peerSet) SampleResendPeersByType(nodeType common.ConnType) []Peer {
...
 	switch nodeType {
 	case common.ENDPOINTNODE:
-		sampledPeers = peers.typePeers(common.PROXYNODE)
-		if len(sampledPeers) == 0 {
-			sampledPeers = peers.typePeers(common.ENDPOINTNODE)
+		sampledPeers = peers.typePeers(common.CONSENSUSNODE)
+		if len(sampledPeers) < 2 {
+			sampledPeers = append(sampledPeers, samplingPeers(peers.typePeers(common.PROXYNODE), 2-len(sampledPeers))...)
+		}
+		if len(sampledPeers) < 2 {
```
jeonghonah: Sampling will need to take into account the priorities? If starting with sampling PN, it seems likely that the PN will be excluded from the sample because of the ENs below.

KimKyungup: If there are more than 2 PN, we don't add EN anymore. So PN has more priority than EN.

ehnuje: Same if statement as above?

node/cn/handler_test.go
```diff
@@ -329,21 +329,22 @@ func TestBroadcastTxsFromEN_EN_NotExists(t *testing.T) {
 	pm.peers = peers
 	cnPeer, pnPeer, enPeer := createAndRegisterPeers(mockCtrl, peers)
 
-	cnPeer.EXPECT().ConnType().Return(common.CONSENSUSNODE).Times(2)
-	pnPeer.EXPECT().ConnType().Return(common.PROXYNODE).Times(2)
-	enPeer.EXPECT().ConnType().Return(common.ENDPOINTNODE).Times(2)
+	cnPeer.EXPECT().ConnType().Return(common.CONSENSUSNODE).Times(3)
```
ehnuje: `ConnType().Return()` is better to be changed into `AnyTimes()` (not necessarily, though)

node/cn/peer_set.go
KimKyungup: Yes it is same condition.

ehnuje: Oh, now I understood your intention. Thanks for the comment!

KimKyungup:
> LGTM. BTW, how about changing the title of PR and description?

Good. I will change the title/content. :) Thanks.

KimKyungup:
@kjhman21 PTAL, again :)

PR #280, opened 2019-10-04
cn: added more tests to handler_msg_test.go

ehnuje:
## Proposed changes

- added more tests to handler_msg_test.go

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/cn/handler_msg_test.go
+++ b/node/cn/handler_msg_test.go
@@ -119,3 +119,90 @@ func TestHandleTxMsg(t *testing.T) {
 		assert.NoError(t, handleTxMsg(pm, mockPeer, msg))
 	}
 }
+
+func prepareTestHandleBlockHeaderFetchRequestMsg(t *testing.T) (*gomock.Controller, *MockPeer, *mocks.MockBlockChain, *ProtocolManager) {
+	mockCtrl := gomock.NewController(t)
+	mockPeer := NewMockPeer(mockCtrl)
+	mockBlockChain := mocks.NewMockBlockChain(mockCtrl)
+
+	return mockCtrl, mockPeer, mockBlockChain, &ProtocolManager{blockchain: mockBlockChain}
+}
+
+func TestHandleBlockHeaderFetchRequestMsg(t *testing.T) {
+	// Decoding the message failed, an error is returned.
+	{
+		mockCtrl, mockPeer, _, pm := prepareTestHandleBlockHeaderFetchRequestMsg(t)
+
+		msg := generateMsg(t, BlockHeaderFetchRequestMsg, newBlock(blockNum1)) // use message data as a block, not a hash
+
+		assert.Error(t, handleBlockHeaderFetchRequestMsg(pm, mockPeer, msg))
+		mockCtrl.Finish()
+	}
+	// GetHeaderByHash returns nil, an error is returned.
+	{
+		mockCtrl, mockPeer, mockBlockChain, pm := prepareTestHandleBlockHeaderFetchRequestMsg(t)
+		mockBlockChain.EXPECT().GetHeaderByHash(hash1).Return(nil).Times(1)
+		mockPeer.EXPECT().GetID().Return(nodeids[0].String()).Times(1)
+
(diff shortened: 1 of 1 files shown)


ehnuje:
@jeonghonah @nogang @KimKyungup Dear reviewers, could you please review this change?

node/cn/handler_msg_test.go
```diff
@@ -119,3 +119,90 @@ func TestHandleTxMsg(t *testing.T) {
...
+		mockPeer := NewMockPeer(mockCtrl)
+		pm := &ProtocolManager{}
+		msg := generateMsg(t, BlockHeaderFetchResponseMsg, newBlock(blockNum1)) // use message data as a block, not a header
+		assert.Error(t, handleBlockHeaderFetchResponseMsg(pm, mockPeer, msg))
+		mockCtrl.Finish()
+	}
+	// FilterHeaders returns nil, error is not returned.
+	{
+		mockCtrl, mockPeer, mockFetcher, pm := prepareTestHandleBlockHeaderFetchResponseMsg(t)
+		mockPeer.EXPECT().GetID().Return(nodeids[0].String()).Times(1)
```
nogang: It is true that handleBlockHeaderFetchResponseMsg calls GetID() twice in this situation. But I don't think that this is functionally important for handleBlockHeaderFetchResponseMsg. How about handling these things with Any() instead of Times()? This test code makes it difficult to refactor or modify the handleBlockHeaderFetchResponseMsg.

ehnuje: @nogang I agree with your opinion. However, for `len(headers) == 0` situation, a log is written but there's no way to distinguish it from the case of `len(headers) != 0`. If I change it to throw an error, it can be distinguished but then there should be additional error handling from the caller of `handleBlockHeaderFetchResponseMsg`. How do you think?

nogang: @ehnuje I know what you mean. There seems to be two choices. For the (len (headers)! = 0) condition, i think there is no branch of logic in this case. It just prints the log. You do not need test code for this branch. However, if it was important for this condition and you needed a logic branch, you should add an error and check it. The choice is up to you.

ehnuje: @ I agree with your point. Let's not limit the number of function calls related to the characteristics of the variable, which are not expected to be changed. (e.g., `Peer.GetID()`)

node/cn/handler_msg_test.go
```diff
@@ -119,3 +119,90 @@ func TestHandleTxMsg(t *testing.T) {
...
+}
+
+func TestHandleBlockHeaderFetchRequestMsg(t *testing.T) {
+	// Decoding the message failed, an error is returned.
+	{
+		mockCtrl, mockPeer, _, pm := prepareTestHandleBlockHeaderFetchRequestMsg(t)
+
+		msg := generateMsg(t, BlockHeaderFetchRequestMsg, newBlock(blockNum1)) // use message data as a block, not a hash
+
+		assert.Error(t, handleBlockHeaderFetchRequestMsg(pm, mockPeer, msg))
```
jeonghonah: minor comment:
How about checking an exact error type?

node/cn/handler_msg_test.go
```diff
@@ -119,3 +119,90 @@ func TestHandleTxMsg(t *testing.T) {
...
+		mockCtrl.Finish()
+	}
+	// GetHeaderByHash returns a header, p.SendFetchedBlockHeader(header) should be called.
+	{
+		mockCtrl, mockPeer, mockBlockChain, pm := prepareTestHandleBlockHeaderFetchRequestMsg(t)
+
+		header := newBlock(blockNum1).Header()
+
+		mockBlockChain.EXPECT().GetHeaderByHash(hash1).Return(header).AnyTimes()
+		mockPeer.EXPECT().SendFetchedBlockHeader(header).AnyTimes()
```
jeonghonah: Question:
AnyTimes() checks if SendFetchedBlockHeader() is not called at all?

PR #304, opened 2019-10-15
Increase FD limit if the server has a small limit

aidan-kwon:
## Proposed changes

**Problem**
- If a server/process running a node exhausted all file descriptors (FD), it can cause the following problems. 
  - **halt a node** if it fails to write data to level DB
  - **can't receive blocks** if all P2P connections are closed 
  - **Bad Block**
- RPC requests can exhaust thousands of FD in a few seconds. 
  - Remote users can trigger the above problems

**Mitigation**
- Hard to mitigate all problems originated from FD exhaust, but easy to prevent FD exhaust from remote users.
  - RPC servers deny connections if the number of connection reaches more than the concurrency limit. 
- Increase the FD limit more than the maximum number of RPC requests allowed to prohibit remote users from exhausting all FD.
  - Partial but simple solution
  - Prevent DoS attack from remote users
  - fd limit > maximum concurrency connection of RPC (http, ws) servers 

**Change**
- Change a number to the const variable `concurrencyLimit`
- Increase FD limit if the server has a small limit

**Test with Scenario**
1. If a node having a small FD limit starts an RPC server, the FD limit is automatically updated with the following log. 
    ```
    WARN[10/15,11:05:48 +09] [39] increase the file descriptor limit of this process for RPC servers  oldLimit=2048 newLimit=8048
    ```
2. A remote user cannot exhaust EN's FD since EN only allows `concurrencyLimit` connections which are less than FD limit. 
    - The remote user will get the following message if he/she tried to connect more than 
    `concurrencyLimit` connections.
      ```
      The connection cannot be served because Server.Concurrency limit exceeded
      ```

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/networks/rpc/server.go
+++ b/networks/rpc/server.go
@@ -62,12 +62,12 @@ func NewServer() *Server {
 	// increase the file descriptor limit if the process has a small limit
 	limit, err := fdlimit.Current()
 	if err != nil {
-		logger.Debug("fail to read fd limit", "err", err)
+		logger.Warn("fail to read fd limit. you may suffer fd exhaustion", "err", err)
 	} else {
 		if limit < minFDLimit {
 			err := fdlimit.Raise(uint64(limit + minFDLimit))
 			if err != nil {
-				logger.Debug("fail to set fd limit", "err", err)
+				logger.Warn("fail to increase fd limit. you may suffer fd exhaustion", "err", err)
 			}
 			logger.Warn("Increase the file descriptor limit of the process for RPC servers", "oldLimit", limit, "newLimit", limit+minFDLimit)
 		}
--- a/networks/rpc/server.go
+++ b/networks/rpc/server.go
@@ -45,6 +45,9 @@ const (
 
 	// pendingRequestLimit is a limit for concurrent RPC method calls
 	pendingRequestLimit = 200000
+
+	// concurrencyLimit is a limit for the number of concurrency connection for RPC servers
+	concurrencyLimit = 3000
 )
 
 // pendingRequestCount is a total number of concurrent RPC method calls
--- a/networks/rpc/server.go
+++ b/networks/rpc/server.go
@@ -23,6 +23,7 @@ package rpc
 import (
 	"context"
 	"fmt"
+	"github.com/klaytn/klaytn/common/fdlimit"
 	"gopkg.in/fatih/set.v0"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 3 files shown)


networks/rpc/server.go
```diff
@@ -45,13 +46,33 @@ const (
...
+	// minFDLimit is the minimum file descriptor limit for RPC servers (heuristic value. concurrency limit of RPC and WS servers)
+	minFDLimit = concurrencyLimit * 2
 )
 
 // pendingRequestCount is a total number of concurrent RPC method calls
 var pendingRequestCount int64 = 0
 
 // NewServer will create a new server instance with no registered handlers.
 func NewServer() *Server {
+	// increase the file descriptor limit if the process has a small limit
```
aidan-kwon: `NewServer()` is used for a handler of HTTP/WS server. 
It means that this function is called when HTTP/WS server is started.

networks/rpc/server.go
```diff
@@ -45,13 +46,33 @@ const (
...
+	limit, err := fdlimit.Current()
+	if err != nil {
+		logger.Debug("fail to read fd limit", "err", err)
+	} else {
+		if limit < minFDLimit {
+			err := fdlimit.Raise(uint64(limit + minFDLimit))
+			if err != nil {
+				logger.Debug("fail to set fd limit", "err", err)
```
jeonghonah: How about `logger.Info()`?

aidan-kwon: Failure of `fdlimit.Current()` or `fdlimit.Raise()` doesn't result in any change/problem to the system. "Increasing fd limit" is an additional feature to prevent the fd exhaust problem that may happen from user's misconfiguration. 
That's why I used `Debug` for the log. @jeonghonah Is that understanble enough?

KimKyungup: minFDLimit is suggestion value as I understood. So I think if increasing is failed, warning log is necessary.

networks/rpc/server.go
```diff
@@ -45,13 +46,33 @@ const (
...
 
 // NewServer will create a new server instance with no registered handlers.
 func NewServer() *Server {
+	// increase the file descriptor limit if the process has a small limit
+	limit, err := fdlimit.Current()
+	if err != nil {
+		logger.Debug("fail to read fd limit", "err", err)
+	} else {
+		if limit < minFDLimit {
+			err := fdlimit.Raise(uint64(limit + minFDLimit))
```
KimKyungup: To raise FD, is there any permission is needed?

aidan-kwon: Not additional permission is needed. It changes fd limit of the process but doesn't change fd limit of the operating system.

networks/rpc/server.go
aidan-kwon: @jeonghonah @KimKyungup I followed your suggestion. Please check this again.

PR #316, opened 2019-10-18
database: introduce state migration database to DBManager

ehnuje:
## Proposed changes

- New member variables `newStateTrieDB` and `inMigration` have been added to `databaseManager` struct for data archiving.
- **All write operations** of state trie database are processed by a batch and the batch will be created from `newStateTrieDB` if `inMigration` flag is `true`.
- **All read operations** will search `newStateTrieDB` first and then search so called `oldStateTrieDB` next if not found. This behavior can be changed to use go-routines.
- **Read operations with suffix** `FromOld` are used to retrieve the data only from the old state trie database. This will be used for data archiving, which migrates the state data from old state trie database to the new one.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -1226,8 +1226,7 @@ func (dbm *databaseManager) ReadStateTrieNodeFromOld(key []byte) ([]byte, error)
 }
 
 func (dbm *databaseManager) HasStateTrieNodeFromOld(key []byte) (bool, error) {
-	db := dbm.getDatabase(StateTrieDB)
-	val, err := db.Get(key)
+	val, err := dbm.ReadStateTrieNodeFromOld(key)
 	if val == nil || err != nil {
 		return false, err
 	}
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -444,7 +444,7 @@ func (dbm *databaseManager) SetStateTrieMigrationDB(blockNum uint64) {
 		return
 	}
 
-	logger.Info("Start setting a new database for state trie migration", blockNum, blockNum)
+	logger.Info("Start setting a new database for state trie migration", "blockNum", blockNum)
 
 	// Store the directory of the new database first to avoid missing directory name situation
 	newDBDir := dbm.config.Dir + strconv.FormatUint(blockNum, 10)
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -27,6 +27,7 @@ import (
 	"github.com/klaytn/klaytn/ser/rlp"
 	"github.com/pkg/errors"
 	"math/big"
+	"os"
 	"path/filepath"
 	"strconv"
 )
@@ -35,6 +36,7 @@ var logger = log.NewModuleLogger(log.StorageDatabase)
 
 type DBManager interface {
 	IsParallelDBWrite() bool
+	IsPartitioned() bool
 	InMigration() bool
 
 	Close()
@@ -199,6 +201,9 @@ const (
 	// databaseEntryTypeSize should be the last item in this list!!
 	databaseEntryTypeSize
 )
+
+const notInMigration = 0
+const inMigration = 1
 
 var dbDirs = [databaseEntryTypeSize]string{
 	"header",
@@ -387,6 +392,10 @@ func (dbm *databaseManager) IsParallelDBWrite() bool {
 	return dbm.config.ParallelDBWrite
 }
 
+func (dbm *databaseManager) IsPartitioned() bool {
+	return dbm.config.Partitioned
+}
+
 func (dbm *databaseManager) InMigration() bool {
 	return dbm.inMigration
 }
@@ -400,18 +409,11 @@ func (dbm *databaseManager) NewBatch(dbEntryType DBEntryType) Batch {
 
 func (dbm *databaseManager) getStateTrieMigrationInfo() (bool, string) {
 	miscDB := dbm.getDatabase(MiscDB)
-
 	enc, _ := miscDB.Get(migrationStatusKey)
-	if len(enc) == 0 {
+	if len(enc) != 8 {
 		return false, ""
 	}
-
-	var status uint64
-	if err := rlp.DecodeBytes(enc, &status); err != nil {
-		logger.Error("Failed to decode migration status", "err", err)
-		return false, ""
-	}
-
+	status := binary.BigEndian.Uint64(enc)
 	if status == 0 {
 		return false, ""
 	}
@@ -457,12 +459,19 @@ func (dbm *databaseManager) SetStateTrieMigrationDB(blockNum uint64) {
 	newDB := createStateTrieMigrationDB(dbm.config, newDBDir)
 
 	// After creating the database, store migration status to the database.
+	// If it fails to store the migration status, remove the new database and kill the process.
 	dbm.newStateTrieDB = newDB
 	dbm.inMigration = true
-	if err := miscDB.Put(migrationStatusKey, encodeUint64(1)); err != nil {
-		logger.Crit("Failed to store the migration status", "status", 1, "err", err)
+	if err := miscDB.Put(migrationStatusKey, encodeUint64(inMigration)); err != nil {
+		logger.Error("Failed to store the migration status, trying to remove the new database for state trie migration", "err", err)
+		newDB.Close()
+		if err := os.RemoveAll(newDBDir); err != nil {
+			logger.Error("Failed to remove the state trie migration database due to an error", "err", err, "dir", newDBDir)
+		} else {
+			logger.Error("Successfully removed the state trie migration database", "dir", newDBDir)
+		}
+		logger.Crit("Killed the process due to the failure of storing the migration status")
 	}
-
 	logger.Info("Finished setting a new database for state trie migration")
 }
 
@@ -1227,11 +1236,6 @@ func (dbm *databaseManager) HasStateTrieNodeFromOld(key []byte) (bool, error) {
 
 // ReadPreimage retrieves a single preimage of the provided hash.
 func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
-	if dbm.inMigration {
-		if val, err := dbm.newStateTrieDB.Get(preimageKey(hash)); err == nil {
-			return val
-		}
-	}
 	db := dbm.getDatabase(StateTrieDB)
 	data, _ := db.Get(preimageKey(hash))
 	return data
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -28,17 +28,20 @@ import (
 	"github.com/pkg/errors"
 	"math/big"
 	"path/filepath"
+	"strconv"
 )
 
 var logger = log.NewModuleLogger(log.StorageDatabase)
 
 type DBManager interface {
 	IsParallelDBWrite() bool
+	InMigration() bool
 
 	Close()
 	NewBatch(dbType DBEntryType) Batch
 	GetMemDB() *MemDB
 	GetDBConfig() *DBConfig
+	SetStateTrieMigrationDB(blockNum uint64)
 
 	// from accessors_chain.go
 	ReadCanonicalHash(number uint64) common.Hash
@@ -99,11 +102,20 @@ type DBManager interface {
 
 	WriteMerkleProof(key, value []byte)
 
+	// State Trie Database related operations
 	ReadCachedTrieNode(hash common.Hash) ([]byte, error)
 	ReadCachedTrieNodePreimage(secureKey []byte) ([]byte, error)
-
 	ReadStateTrieNode(key []byte) ([]byte, error)
 	HasStateTrieNode(key []byte) (bool, error)
+	ReadPreimage(hash common.Hash) []byte
+
+	ReadCachedTrieNodeFromOld(hash common.Hash) ([]byte, error)
+	ReadCachedTrieNodePreimageFromOld(secureKey []byte) ([]byte, error)
+	ReadStateTrieNodeFromOld(key []byte) ([]byte, error)
+	HasStateTrieNodeFromOld(key []byte) (bool, error)
+	ReadPreimageFromOld(hash common.Hash) []byte
+
+	WritePreimages(number uint64, preimages map[common.Hash][]byte)
 
 	// from accessors_indexes.go
 	ReadTxLookupEntry(hash common.Hash) (common.Hash, uint64, uint64)
@@ -134,9 +146,6 @@ type DBManager interface {
 
 	ReadChainConfig(hash common.Hash) *params.ChainConfig
 	WriteChainConfig(hash common.Hash, cfg *params.ChainConfig)
-
-	ReadPreimage(hash common.Hash) []byte
-	WritePreimages(number uint64, preimages map[common.Hash][]byte)
 
 	// below operations are used in parent chain side, not child chain side.
 	WriteChildChainTxHash(ccBlockHash common.Hash, ccTxHash common.Hash)
@@ -244,6 +253,9 @@ type databaseManager struct {
 	config *DBConfig
 	dbs    []Database
 	cm     *cacheManager
+
+	inMigration    bool
+	newStateTrieDB Database
 }
 
 func NewMemoryDBManager() DBManager {
@@ -295,7 +307,7 @@ func singleDatabaseDBManager(dbc *DBConfig) (DBManager, error) {
 
 // partitionedDatabaseDBManager returns DBManager which handles partitioned Database.
 // Each Database will have its own separated Database.
-func partitionedDatabaseDBManager(dbc *DBConfig) (DBManager, error) {
+func partitionedDatabaseDBManager(dbc *DBConfig) (*databaseManager, error) {
 	dbm := newDatabaseManager(dbc)
 	var db Database
 	var err error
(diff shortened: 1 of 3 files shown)


storage/database/db_manager.go
```diff
@@ -371,10 +387,85 @@ func (dbm *databaseManager) IsParallelDBWrite() bool {
...
+	if err != nil {
+		logger.Crit("Failed to create a new database for state trie migration", "err", err)
+	}
+	logger.Info("Created a new database for state trie migration", "newStateTrieDB", newDBConfig.Dir)
+	return newDB
+}
+
+func (dbm *databaseManager) SetStateTrieMigrationDB(blockNum uint64) {
+	if !dbm.config.Partitioned {
+		logger.Warn("Setting a new database for state trie migration is allowed for partitioned database only")
```
KimKyungup: It doesn't seems that only partitioned database is allow in createStateTrieMigrationDB.
Why this warn and return are needed?

storage/database/db_manager.go
```diff
@@ -371,10 +387,85 @@ func (dbm *databaseManager) IsParallelDBWrite() bool {
...
+		return
+	}
+
+	logger.Info("Start setting a new database for state trie migration", blockNum, blockNum)
```
KimKyungup:
```suggestion
	logger.Info("Start setting a new database for state trie migration", "blockNum", blockNum)
```

storage/database/db_manager.go
```diff
@@ -1057,30 +1153,106 @@ func (dbm *databaseManager) WriteMerkleProof(key, value []byte) {
 
 // Cached Trie Node operation.
 func (dbm *databaseManager) ReadCachedTrieNode(hash common.Hash) ([]byte, error) {
+	if dbm.inMigration {
```
KimKyungup: Why don't you separate Read function in migration and Read function not in migration?

storage/database/schema.go
```diff
@@ -185,3 +188,7 @@ func governanceKey(num uint64) []byte {
 	binary.LittleEndian.PutUint64(b, num)
 	return append(governancePrefix[:], b[:]...)
 }
+
+func databaseDirKey(dbEntryType uint64) []byte {
```
KimKyungup: For dbEntryType, uint64 is needed? It can be covered by small uint?

storage/database/schema.go
```diff
@@ -94,16 +97,16 @@ type TxLookupEntry struct {
 	Index      uint64
 }
 
-// encodeBlockNumber encodes a block number as big endian uint64
-func encodeBlockNumber(number uint64) []byte {
+// encodeUint64 encodes a block number as big endian uint64
```
KimKyungup: For dbEntryType, did you change this name?
uint64 is necessary for dbEntryType?

storage/database/db_manager.go
```diff
@@ -295,7 +307,7 @@ func singleDatabaseDBManager(dbc *DBConfig) (DBManager, error) {
 
 // partitionedDatabaseDBManager returns DBManager which handles partitioned Database.
 // Each Database will have its own separated Database.
-func partitionedDatabaseDBManager(dbc *DBConfig) (DBManager, error) {
+func partitionedDatabaseDBManager(dbc *DBConfig) (*databaseManager, error) {
```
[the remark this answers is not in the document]
ehnuje: It is to make line377 and 388 directly access the member variable of `databaseManager`. If it is an interface, we can't access the member variable of it, unless we add additional setters.

storage/database/db_manager.go
ehnuje: Migration of non-partitioned database does not seem needed and helpful. Maybe we can do that but I don't think it is needed. Since in that case, as there is only one LevelDB instance, read-request from migration goroutine will continuously block read/write operations from on-going side.

storage/database/db_manager.go
ehnuje: If they are separated, the caller of read-request function should know whether the node is in migration process or not and `if-else` statements are needed for every functions. Instead, this way doesn't change the way it is used for now and only the migrator will use the newly added functions with the suffix `...FromOld`

storage/database/schema.go
ehnuje: It can be, something like `uint16`, however, maximum number of entries stored by this function is the number of total database partitions, which means less than 10. Maybe we can use `uint16` but what I wanted to use already existing number-encoding logic.

storage/database/db_manager.go
ehnuje: Thanks! I will reflect this.

storage/database/schema.go
ehnuje: Same answer from line#192

storage/database/db_manager.go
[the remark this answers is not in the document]
KimKyungup: It's a internal function. Got it. Thanks. :)

storage/database/schema.go
KimKyungup:
```suggestion
// encodeUint64 encodes a number as big endian uint64
```

ehnuje:
@KimKyungup I really appreciate for your comments =)
@jeonghonah @kjhman21 @nogang Dear reviewers, could you please take a look into this change?

kjhman21:
Can we have the test case that DB is closed during migration and it is opened again?

ehnuje:
@kjhman21 What you mean is the test like below steps?
1. call `NewDBManager` to create a new database
2. call `SetStateTrieMigrationDB` to setting a state trie migration database
3. close the database 
4. call `NewDBManager` and check if relevant information is saved or not

kjhman21:
That could be one test case. What I was thinking is the case:
1. migration is turned on
2. kcnd stop during migration
3. kcnd start

ehnuje:
Does `kcnd` indicate running a whole node?

kjhman21:
> Does kcnd indicate running a whole node?

It could be, but simulation is also okay. I just want to make sure that our `kcn` works very well even if the case I described happens.

kjhman21:
Does this PR include the migration of state trie data from the old DB to new DB?

storage/database/db_manager.go
```diff
@@ -371,10 +387,85 @@ func (dbm *databaseManager) IsParallelDBWrite() bool {
...
+		newDB, err = newDatabase(newDBConfig, StateTrieDB)
+	}
+	if err != nil {
+		logger.Crit("Failed to create a new database for state trie migration", "err", err)
+	}
+	logger.Info("Created a new database for state trie migration", "newStateTrieDB", newDBConfig.Dir)
+	return newDB
+}
+
+func (dbm *databaseManager) SetStateTrieMigrationDB(blockNum uint64) {
```
kjhman21: Is it okay even if the process is terminated during the execution of this function?

storage/database/db_manager.go
```diff
@@ -371,10 +387,85 @@ func (dbm *databaseManager) IsParallelDBWrite() bool {
...
+	}
+
+	// After storing the directory, create a new database for migration process.
+	newDB := createStateTrieMigrationDB(dbm.config, newDBDir)
+
+	// After creating the database, store migration status to the database.
+	dbm.newStateTrieDB = newDB
+	dbm.inMigration = true
+	if err := miscDB.Put(migrationStatusKey, encodeUint64(1)); err != nil {
+		logger.Crit("Failed to store the migration status", "status", 1, "err", err)
```
kjhman21: If we failed here, it would be better to remove the created database `newDBDir` for the next trial.

ehnuje:
> Does this PR include the migration of state trie data from the old DB to new DB?

@kjhman21 This PR does not include anything related to migration operations.

storage/database/db_manager_test.go
```diff
@@ -387,6 +387,34 @@ func TestDBManager_TrieNode(t *testing.T) {
...
+		}
+
+		cachedNode, _ = dbm.ReadCachedTrieNode(hash2)
+		assert.Equal(t, hash2[:], cachedNode)
+
+		stateTrieNode, _ = dbm.ReadStateTrieNode(hash2[:])
+		assert.Equal(t, hash2[:], stateTrieNode)
+
+		hasStateTrieNode, _ = dbm.HasStateTrieNode(hash2[:])
+		assert.True(t, hasStateTrieNode)
```
kjhman21: Could we call ReadStateTrieNodeFromOld(hash1) and ReadStateTrieNodeFromOld(hash2) here?

ehnuje: That would be better. I will update the test.

storage/database/db_manager.go
```diff
@@ -1057,30 +1153,106 @@ func (dbm *databaseManager) WriteMerkleProof(key, value []byte) {
...
+	val, err := db.Get(key)
 	if val == nil || err != nil {
 		return false, err
 	}
 	return true, nil
 }
 
+// ReadPreimage retrieves a single preimage of the provided hash.
+func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
+	if dbm.inMigration {
```
nogang: The code under this condition is not necessary because it runs in `ReadPreimage`.

ehnuje: Right! I forgot to remove this lines. Thanks!

storage/database/db_manager.go
```diff
@@ -1057,30 +1153,106 @@ func (dbm *databaseManager) WriteMerkleProof(key, value []byte) {
...
-func (dbm *databaseManager) ReadStateTrieNode(key []byte) ([]byte, error) {
+func (dbm *databaseManager) ReadStateTrieNodeFromOld(key []byte) ([]byte, error) {
 	db := dbm.getDatabase(StateTrieDB)
 	return db.Get(key)
 }
 
-func (dbm *databaseManager) HasStateTrieNode(key []byte) (bool, error) {
-	val, err := dbm.ReadStateTrieNode(key)
+func (dbm *databaseManager) HasStateTrieNodeFromOld(key []byte) (bool, error) {
+	db := dbm.getDatabase(StateTrieDB)
```
nogang: How about use `ReadStateTrieNodeOle` instead

storage/database/db_manager.go
ehnuje: Updated!

storage/database/db_manager.go
ehnuje: "okay" can be understood two different ways. If terminated during `SetStateTrieMigrationDB`,
1. It does not make problem, but doesn't start migration process automatically after re-run neither
2. It does not make problem, and start migration process after re-run

We'll take the first approach with the assumption there is only little possibility that the node is terminated during the execution of `SetStateTrieMigrationDB`. **Therefore, if it is failed during the execution of the function, migration process should be re-triggered by outer process, not by itself.**

storage/database/db_manager.go
```diff
@@ -1057,30 +1153,106 @@ func (dbm *databaseManager) WriteMerkleProof(key, value []byte) {
...
+}
+
+// State Trie Related operations.
+func (dbm *databaseManager) ReadStateTrieNode(key []byte) ([]byte, error) {
+	if dbm.inMigration {
+		if val, err := dbm.newStateTrieDB.Get(key); err == nil {
+			return val, nil
+		}
+	}
+	return dbm.ReadStateTrieNodeFromOld(key)
```
nogang: How about use `HasStateTrieNodeFromOld` instead. The processing for the value returned from the DB has disappeared. At `line 1078` in the `HasStateTrieNode` method.

storage/database/db_manager.go
```diff
@@ -244,6 +253,9 @@ type databaseManager struct {
 	config *DBConfig
 	dbs    []Database
 	cm     *cacheManager
+
+	inMigration    bool
```
nogang: There is no code that makes imMigration false. Is there a concurrency problem in reading and writing this variable?

ehnuje: > There is no code that makes imMigration false. 

As the default value of Boolean variable is `false`, only setting it `true` is needed.

> Is there a concurrency problem in reading and writing this variable?

Currently, not. Since `partitionedDatabaseDBManager` and `SetStateTrieMigrationDB` are the only two places where the value is set. As `partitionedDatabaseDBManager` is called during the start of the node and `SetStateTrieMigrationDB` is called after the start of the node, there shouldn't be a concurrency issue here.

storage/database/db_manager.go
```diff
@@ -191,6 +202,9 @@ const (
 	databaseEntryTypeSize
 )
 
+const notInMigration = 0
+const inMigration = 1
```
kjhman21: What about using `iota`?
https://yourbasic.org/golang/iota/

kjhman21: Hmm? It would be better to use different names since `dbm.inMigration` is used. It is confusing.

storage/database/db_manager.go
```diff
@@ -357,11 +374,15 @@ func NewDBManager(dbc *DBConfig) DBManager {
...
-		if dbm, err := partitionedDatabaseDBManager(dbc); err != nil {
+		dbm, err := partitionedDatabaseDBManager(dbc)
+		if err != nil {
 			logger.Crit("Failed to partitioned database", "DBType", dbc.DBType, "err", err)
-		} else {
-			return dbm
 		}
+		if inMigration, migrationDir := dbm.getStateTrieMigrationInfo(); inMigration {
+			dbm.newStateTrieDB = createStateTrieMigrationDB(dbm.config, migrationDir)
+			dbm.inMigration = true
```
jeonghonah: Is there any risk of race on those variables?

storage/database/db_manager.go
```diff
@@ -371,10 +392,89 @@ func (dbm *databaseManager) IsParallelDBWrite() bool {
...
+
+	// Store the directory of the new database first to avoid missing directory name situation
+	newDBDir := dbm.config.Dir + strconv.FormatUint(blockNum, 10)
+	miscDB := dbm.getDatabase(MiscDB)
+	if err := miscDB.Put(databaseDirKey(uint64(StateTrieDB)), []byte(newDBDir)); err != nil {
+		logger.Crit("Failed to store a new database directory", "newDBDir", newDBDir, "err", err)
+	}
+
+	// After storing the directory, create a new database for migration process.
+	newDB := createStateTrieMigrationDB(dbm.config, newDBDir)
```
jeonghonah: Is there a possibility that this function will be called more than twice for the same block?

storage/database/db_manager.go
```diff
@@ -371,10 +392,89 @@ func (dbm *databaseManager) IsParallelDBWrite() bool {
...
+
+	// After creating the database, store migration status to the database.
+	// If it fails to store the migration status, remove the new database and kill the process.
+	dbm.newStateTrieDB = newDB
+	dbm.inMigration = true
+	if err := miscDB.Put(migrationStatusKey, encodeUint64(inMigration)); err != nil {
+		logger.Error("Failed to store the migration status, trying to remove the new database for state trie migration", "err", err)
```
jeonghonah: If the node shuts down abnormally before it finishes cleaning, can unnecessary files remain?

storage/database/db_manager.go
```diff
@@ -1057,30 +1162,100 @@ func (dbm *databaseManager) WriteMerkleProof(key, value []byte) {
 
 // Cached Trie Node operation.
 func (dbm *databaseManager) ReadCachedTrieNode(hash common.Hash) ([]byte, error) {
+	if dbm.inMigration {
+		if val, err := dbm.newStateTrieDB.Get(hash[:]); err == nil {
```
jeonghonah: What happened if `err != nil`?

storage/database/db_manager_test.go
```diff
@@ -387,6 +387,49 @@ func TestDBManager_TrieNode(t *testing.T) {
 
 		hasStateTrieNode, _ = dbm.HasStateTrieNode(hash1[:])
 		assert.True(t, hasStateTrieNode)
+
+		if !dbm.IsPartitioned() {
+			continue
+		}
+		dbm.SetStateTrieMigrationDB(123)
+
+		cachedNode, _ = dbm.ReadCachedTrieNode(hash1)
```
jeonghonah: It's not directly related to this PR, but why use both array or slice as a function parameter? I think that slice can be used for both cases.

storage/database/db_manager.go
ehnuje: @jeonghonah Then it will go to line 1170 and return the result from `ReadCachedTrieNodeFromOld`.

ehnuje: This is because when there is no value found, LevelDB returns nil value and `NotFound` error.

storage/database/db_manager.go
ehnuje: That's what something can be happened, however, if we take that possibility into our account, we also should change other database creation codes. I'm not sure if it is really needed. How do you think?

storage/database/db_manager_test.go
ehnuje: Right. Just used the same function signature from go-ethereum. I will change this in a separated PR.

storage/database/db_manager.go
ehnuje: Currently, this is called only when the node starts by `partitionedDatabaseDBManager`. Call `SetStateTrieMigrationDB` from `Database.Commit`, won't be called if it is called from   `partitionedDatabaseDBManager`.

storage/database/db_manager.go
ehnuje: This process will be encapsulated by `Database.Commit` and `Database.Commit` holds `Database.lock` during its operation. Therefore, there seems to be no race for these variables.

storage/database/db_manager.go
ehnuje: > What about using iota?

As there are only two variables, I thought it would be an overkill.

> Hmm? It would be better to use different names since dbm.inMigration is used. It is confusing.

I will change these names with the suffix `Flag`

ehnuje:
@jeonghonah @nogang @kjhman21 @KimKyungup Dear reviewers, I've written answers your comments or reflected them. Please have a look =)

storage/database/db_manager.go
jeonghonah: I agree with you. We don't have to worry about this PR.

ehnuje:
@nogang As the reviewers have approved, let me merge this PR. Please let me know if there's any remaining issue.

PR #336, opened 2019-10-24
all: enable data archiving preparation

ehnuje:
## Proposed changes

- Please note that this is in-developing and the way node sets the new database and subsequent operations will be changed in the future.
- With the flag `dataarchiving.blocknumber`, the node will set-up a new database partition for data archiving of state trie partition.
- Right before finishing `Database.Commit` operation, it will check  if current block number is equal to or greater than the given setting and it will set-up a new database.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- Below is the list of related changes of data archiving feature.
- [database: introduce state migration database to DBManager](https://github.com/klaytn/klaytn/pull/316)

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -78,13 +78,14 @@ const (
 // 2) trie caching/pruning resident in a blockchain.
 type CacheConfig struct {
 	// TODO-Klaytn-Issue1666 Need to check the benefit of trie caching.
-	StateDBCaching       bool // Enables caching of state objects in stateDB.
-	TxPoolStateCache     bool // Enables caching of nonce and balance for txpool.
-	ArchiveMode          bool // If true, state trie is not pruned and always written to database.
-	CacheSize            int  // Size of in-memory cache of a trie (MiB) to flush matured singleton trie nodes to disk
-	BlockInterval        uint // Block interval to flush the trie. Each interval state trie will be flushed into disk.
-	TrieCacheLimit       int  // Memory allowance (MB) to use for caching trie nodes in memory
-	SenderTxHashIndexing bool // Enables saving senderTxHash to txHash mapping information to database and cache.
+	StateDBCaching        bool   // Enables caching of state objects in stateDB.
+	TxPoolStateCache      bool   // Enables caching of nonce and balance for txpool.
+	ArchiveMode           bool   // If true, state trie is not pruned and always written to database.
+	CacheSize             int    // Size of in-memory cache of a trie (MiB) to flush matured singleton trie nodes to disk
+	BlockInterval         uint   // Block interval to flush the trie. Each interval state trie will be flushed into disk.
+	TrieCacheLimit        int    // Memory allowance (MB) to use for caching trie nodes in memory
+	SenderTxHashIndexing  bool   // Enables saving senderTxHash to txHash mapping information to database and cache.
+	DataArchivingBlockNum uint64 // If not zero, the number indicates the starting point of data archiving operation.
 }
 
 // BlockChain represents the canonical chain given a database with a genesis
@@ -183,7 +184,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 		cacheConfig:     cacheConfig,
 		db:              db,
 		triegc:          prque.New(),
-		stateCache:      state.NewDatabaseWithCache(db, cacheConfig.TrieCacheLimit),
+		stateCache:      state.NewDatabaseWithCache(db, cacheConfig.TrieCacheLimit, cacheConfig.DataArchivingBlockNum),
 		quit:            make(chan struct{}),
 		futureBlocks:    futureBlocks,
 		engine:          engine,
@@ -772,7 +773,7 @@ func (bc *BlockChain) Stop() {
 				}
 
 				logger.Info("Writing cached state to disk", "block", recent.Number(), "hash", recent.Hash(), "root", recent.Root())
-				if err := triedb.Commit(recent.Root(), true); err != nil {
+				if err := triedb.Commit(recent.Root(), true, statedb.NoDataArchivingPreparation); err != nil {
 					logger.Error("Failed to commit recent state trie", "err", err)
 				}
 			}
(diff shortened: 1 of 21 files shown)


cmd/utils/flags.go
```diff
@@ -657,6 +657,13 @@ var (
 		Usage: "The maximum difference between current block and event block. 0 means off",
 		Value: 0,
 	}
+	// Data Archiving
+	// TODO-Klaytn-DataArchiving Please note that DataArchivingBlockNumFlag is just for development purpose.
+	DataArchivingBlockNumFlag = cli.Uint64Flag{
+		Name:  "dataarchiving.blocknumber",
+		Usage: "The point when the data archiving starts from. 0 means off",
```
kjhman21:
```suggestion
		Usage: "The block number when the data archiving starts from. 0 means off",
```

storage/statedb/database.go
```diff
@@ -822,9 +829,22 @@ func (db *Database) Commit(node common.Hash, report bool) error {
...
+		db.diskDB.SetStateTrieMigrationDB(blockNum)
+	}
 	return nil
 }
 
+func (db *Database) settingMigrationDBRequired(blockNum uint64) bool {
+	if blockNum == 0 {
+		return false
+	}
+	if blockNum >= db.dataArchivingBlockNum && !db.diskDB.InMigration() {
```
kjhman21: What if we want to archive blockchain data periodically? This seems that it archives data one time.

ehnuje: Yes, as the policy has not been determined, this PR is to test if [1] creating a NEW_DB while the node is running and [2] using both OLD_DB and NEW_DB does not make any problem.

ehnuje:
@jeonghonah @nogang @kjhman21 @KimKyungup Dear reviewers, could you please review this change?

blockchain/blockchain.go
```diff
@@ -1012,7 +1013,7 @@ func (bc *BlockChain) writeStateTrie(block *types.Block, state *state.StateDB) e
 
 	// If we're running an archive node, always flush
 	if bc.isArchiveMode() {
-		if err := trieDB.Commit(root, false); err != nil {
+		if err := trieDB.Commit(root, false, block.NumberU64()); err != nil {
```
kjhman21: I think we should not perform blockchain data archiving if "archive" mode is on.
What happens if we pass the current block number to trieDB.Commit()? Does it perform new db creation?

ehnuje: @kjhman21 Yes, it will. Instead, we can put `NoDataArchivingPreparation` to indicate that no data archiving feature for an archiving node.

kjhman21: @ehnuje So, what is your plan? Will this be changed to `NoDataArchivingPreparation` or something else?

ehnuje: @kjhman21 I didn't consider the situation where the data archiving feature is enabled on the archiving node. (BTW, changing the names seems to be needed 😅 ) Let me update the PR.

storage/statedb/database.go
```diff
@@ -822,9 +829,22 @@ func (db *Database) Commit(node common.Hash, report bool) error {
...
 	db.flushnodes, db.flushsize, db.flushtime = 0, 0, 0
 
+	if db.settingMigrationDBRequired(blockNum) {
+		db.diskDB.SetStateTrieMigrationDB(blockNum)
+	}
 	return nil
 }
 
+func (db *Database) settingMigrationDBRequired(blockNum uint64) bool {
+	if blockNum == 0 {
```
nogang: Shouldn't `dataArchivingBlockNum` be compared to zero instead `blockNum`?

node/cn/backend.go
```diff
@@ -243,7 +243,8 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 		vmConfig    = vm.Config{EnablePreimageRecording: config.EnablePreimageRecording}
 		cacheConfig = &blockchain.CacheConfig{StateDBCaching: config.StateDBCaching,
 			ArchiveMode: config.NoPruning, CacheSize: config.TrieCacheSize, BlockInterval: config.TrieBlockInterval,
-			TxPoolStateCache: config.TxPoolStateCache, TrieCacheLimit: config.TrieCacheLimit, SenderTxHashIndexing: config.SenderTxHashIndexing}
+			TxPoolStateCache: config.TxPoolStateCache, TrieCacheLimit: config.TrieCacheLimit,
+			SenderTxHashIndexing: config.SenderTxHashIndexing, DataArchivingBlockNum: config.DataArchivingBlockNum}
```
jeonghonah: How about updating `node/cn/sc_backend` also?

tags-for-todo-and-note.md
```diff
@@ -55,3 +55,4 @@ Please note that the tag name always
 | gRPC                                           | Related to gRPC.                                          |
 | HF                                             | Related to HardFork.                                      |
 | StateDB                                        | Related to StateDB and stateObject.                       |
+| DataArchiving                                  | Related to Data Archiving feature                         |
```
jeonghonah:
```suggestion
| DataArchiving                                  | Related to Data Archiving feature.                        |
```

storage/statedb/database.go
nogang: I need to make sure `blockNum` is the same as `NoDataArchivingPreparation`.

node/cn/backend.go
ehnuje: @jeonghonah After discussed with @KimKyungup , we don't need to maintain `sc_backend.go` as it is related to  the clique, which we do not support or maintain anymore.

storage/statedb/database.go
ehnuje: Updated!

tags-for-todo-and-note.md
ehnuje: Updated!

storage/statedb/database.go
```diff
@@ -822,9 +829,22 @@ func (db *Database) Commit(node common.Hash, report bool) error {
...
+		db.diskDB.SetStateTrieMigrationDB(blockNum)
+	}
 	return nil
 }
 
+func (db *Database) settingMigrationDBRequired(blockNum uint64) bool {
+	if blockNum == NoDataArchivingPreparation {
+		return false
+	}
+	if blockNum >= db.dataArchivingBlockNum && !db.diskDB.InMigration() {
```
nogang: Shouldn't it work if DataArchivingBlockNum is 0?
 because of comment `DataArchivingBlockNum uint64 // If not zero, the number indicates the starting point of data archiving operation.`

ehnuje: @nogang You're right. I missed that point. Please check the last commit =) Thanks!

PR #363, opened 2019-11-05
[Draft] Resolve weird balance of msg.sender in contract call

KimKyungup:
## Proposed changes

To resovle https://github.com/klaytn/klaytn/issues/359.

This PR is just draft version. Let's discuss about this.
This PR added the code to skip to subtract the balance of msg.sender during contract call.

In ethereum, https://github.com/ethereum/go-ethereum/pull/2799 introduced `CheckNonce` flag which can skip the nonce check for contract call.
I think we can skip checking nonce and balance of msg.from for contract call.

As @kjhman21 said, we need to test more with this PR.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [ ] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

Issue #359, opened 2019-11-04
msg.sender.balance gives me an unexpected value.


scalalang2:
**Describe the bug**
If you query the balance property on an address typed variable
It returns an unexpected value.

**How to reproduce**

1. Paste following code and copy it to Klaytn IDE.
```
pragma solidity 0.5.6; 

contract Count {
    address payable david = 0x8aB3A34499F160e2E1cCfe3CaF0C256804a6b84b;
    
    function _balance() public view returns (uint256) {
        return david.balance;
    }
}
```

2. Deploy and call the _balance() function.

**Expected behavior**
It should returns current balance of address.

**Attachments**
![스크린샷 2019-11-04 오후 1 38 49](https://user-images.githubusercontent.com/2603286/68099908-7b99a980-ff08-11e9-84ec-0ba23a3bd1eb.png)

kjhman21:
Thanks for the reporting the bug.

I just copied your contract code, deployed it, and executed it as what you did.

However, I think I got a right result. Please check out the screenshot below.

![image](https://user-images.githubusercontent.com/20692739/68103080-c1f70480-ff18-11e9-9abc-6c9eb349a580.png)

scalalang2:
Here is the contract address I've tried.
0xc48ADBFebD0C2753a9883f648349AB0cA8DcDe13

Can you try again with above contract address?

kjhman21:
@scalalang2 Also works well....

![image](https://user-images.githubusercontent.com/20692739/68104622-21581300-ff1f-11e9-9ee5-66a32bfb384e.png)

scalalang2:
I think the problem is occurred when the sender and address is equal.
I've changed the 'david' variable to another account, then it works well.

Please try below code.
```
pragma solidity 0.5.6; 

contract Count {
    function _balance() public view returns (uint256) {
        return msg.sender.balance;
    }
}
```

scalalang2:
And I've found that this problem was revealed at Ethereum.
It's related with

```go
// klaytn/node/cn/api_backend.go

func (b *CNAPIBackend) GetEVM(ctx context.Context, msg blockchain.Message, state *state.StateDB, header *types.Header, vmCfg vm.Config) (*vm.EVM, func() error, error) {
	state.SetBalance(msg.ValidatedSender(), math.MaxBig256)
	vmError := func() error { return nil }

	context := blockchain.NewEVMContext(msg, header, b.cn.BlockChain(), nil)
[...]

PR #372, opened 2019-11-08
Resolve weird balance of msg.sender in contract call

KimKyungup:
## Proposed changes
I already suggested https://github.com/klaytn/klaytn/pull/363. But it was too big change so there are much thing have to be tested. Therefore I made this PR simpler and better than before.

To resovle #359. 

This PR removed the root cause code to set very big balance to `from` for zero balance account.
~~Instead of the removed code, this PR added the code to set zero gas price if the gas price is not set in `klay_call` API.
`klay_call` is usually for just reading block chain without writing the state.
Therefore without gas fee, user can read the block chain. So I think it makes sense.
However, even though after this PR, if user want to call with non-zero gasPrice, it works well like before.~~

Instead of the removed code, this PR added `addBalance` for adding gas fee to sender's balance.
This means `msg.balance` in EVM will be same with the balance of the sender in the state :)
 
## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/cn/api_backend.go
+++ b/node/cn/api_backend.go
@@ -160,7 +160,7 @@ func (b *CNAPIBackend) GetTd(blockHash common.Hash) *big.Int {
 }
 
 func (b *CNAPIBackend) GetEVM(ctx context.Context, msg blockchain.Message, state *state.StateDB, header *types.Header, vmCfg vm.Config) (*vm.EVM, func() error, error) {
-	// Add gas fee to sender's balance for estimating gasLimit.
+	// Add gas fee to sender for estimating gasLimit or calling a function by insufficient balance sender.
 	state.AddBalance(msg.ValidatedSender(), new(big.Int).Mul(new(big.Int).SetUint64(msg.Gas()), msg.GasPrice()))
 	vmError := func() error { return nil }
 
--- a/node/cn/api_backend.go
+++ b/node/cn/api_backend.go
@@ -31,7 +31,6 @@ import (
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/blockchain/vm"
 	"github.com/klaytn/klaytn/common"
-	"github.com/klaytn/klaytn/common/math"
 	"github.com/klaytn/klaytn/event"
 	"github.com/klaytn/klaytn/networks/rpc"
 	"github.com/klaytn/klaytn/node/cn/gasprice"
@@ -161,7 +160,8 @@ func (b *CNAPIBackend) GetTd(blockHash common.Hash) *big.Int {
 }
 
 func (b *CNAPIBackend) GetEVM(ctx context.Context, msg blockchain.Message, state *state.StateDB, header *types.Header, vmCfg vm.Config) (*vm.EVM, func() error, error) {
-	state.SetBalance(msg.ValidatedSender(), math.MaxBig256)
+	// Add gas fee to sender's balance for estimating gasLimit.
+	state.AddBalance(msg.ValidatedSender(), new(big.Int).Mul(new(big.Int).SetUint64(msg.Gas()), msg.GasPrice()))
 	vmError := func() error { return nil }
 
 	context := blockchain.NewEVMContext(msg, header, b.cn.BlockChain(), nil)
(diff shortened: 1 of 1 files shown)


kjhman21:
@KimKyungup Please check the below cases.
1. Is it possible to call a function having a call of transfer()?
2. If possible, what happens if the sender's balance is negative?

KimKyungup:
> @KimKyungup Please check the below cases.
> 
> 1. Is it possible to call a function having a call of transfer()?
> 2. If possible, what happens if the sender's balance is negative?

@kjhman21 
If `transfer/send` KLAY in solidity means sending KLAY from the contract to a receiver, So it doesn't related with sender's balance.
I tested calling a function having a call of transfer() worked well even if the msg.sender doesn't have KLAY with the contract has some balance.

And I also tested a call a payable function, It worked well with 0 msg.value. But In call argument, there is no value argument, So I couldn't test with not zero msg.value.

api/api_public_blockchain.go
```diff
@@ -253,12 +253,14 @@ func (s *PublicBlockChainAPI) doCall(ctx context.Context, args CallArgs, blockNr
...
-	gas, gasPrice := uint64(args.Gas), args.GasPrice.ToInt()
+	gas := uint64(args.Gas)
 	if gas == 0 {
 		gas = math.MaxUint64 / 2
 	}
-	if gasPrice.Sign() == 0 {
-		gasPrice = new(big.Int).SetUint64(defaultGasPrice)
+
+	gasPrice := new(big.Int).SetUint64(defaultGasPrice)
+	if args.GasPrice != nil {
```
jeonghonah: Before this, we used default gas price if gasPrice was zero. It's going to change the meaning of the API. Wouldn't that be a problem?

aidan-kwon:
Why the zero gas price is assigned only for `Call` function? 
Is there any side effects for `EstimateGas` or `EstimateComputationCost`?

api/api_public_blockchain.go
KimKyungup: Before this PR the given gas Price zero means 2 cases like below.
- user gave zero gas price.
- user didn't set the gas price.

But almost client usually doesn't set the gasPrice for `Call` API. Because gasPrice is not important for `Call`.

And because of defaultGasPrice, `state.SetBalance(msg.ValidatedSender(), math.MaxBig256)` code is needed. This made weird balance.

KimKyungup:
> @KimKyungup Why the zero gas price is assigned only for `Call` function?
> Is there any side effects for `EstimateGas` or `EstimateComputationCost`?

I think that people use `Call` for only read the contract.
But `EstimateGas`/`EstimateComputationCost` are usually used for writing.

So for executing real transaction, I think zero gasPrice is useless. :)

However, I also think there is no side effect to add same code to both APIs.
Do you want to add the code to them also?
This issue is only related with `Call`, So I wanted add this code to this only. :)

aidan-kwon:
> I think that people use `Call` for only read the contract.
> But `EstimateGas`/`EstimateComputationCost` are usually used for writing.
> 
> So for executing real transaction, I think zero gasPrice is useless. :)
> 
> However, I also think there is no side effect to add same code to both APIs.
> Do you want to add the code to them also?
> This issue is only related with `Call`, So I wanted add this code to this only. :)

This change limits the use of `EstimateGas`/`EstimateComputationCost`.
A user doesn't have enough KLAY cannot execute those APIs properly. 

I think it is better to assign the zero gas price in `_doCall` function.

api/api_public_blockchain.go
jeonghonah: If gas price is not meaningful for the `Call`, why do we have to keep the value set by the user? How about simply use `zero` always?

KimKyungup: Becase `gasPrice` is one parameter for EVM. 
So if user want to set it, we need to give a way to use their `gasPrice`.
We don't need to simplify the optional feature.

KimKyungup:
Dear reviewers,
let me think about how to solve estimateGas API without the SetBalance code. 
I will let you know again. Thanks. :)

KimKyungup:
@aidan-kwon @jeonghonah PTAL. :) I pushed simple code.

KimKyungup:
@jeonghonah @nogang @kjhman21 PTAL. :)

PR #394, opened 2019-11-22
SC: Add handledEvent Map for prevent duplicated tx

KimKyungup:
## Proposed changes

This PR prevents the duplicated transaction for already handled event.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/sc/bridge_manager.go
+++ b/node/sc/bridge_manager.go
@@ -246,7 +246,7 @@ func (bi *BridgeInfo) UpdateInfo() error {
 	if err != nil {
 		return err
 	}
-	bi.UpdateRequestNonce(rn)
+	bi.SetRequestNonce(rn)
 
 	hn, err := bi.bridge.LowerHandleNonce(nil)
 	if err != nil {
@@ -255,8 +255,8 @@ func (bi *BridgeInfo) UpdateInfo() error {
 
 	bi.lowerHandleNonce = hn
 
-	bi.UpdateHandledNonce(hn)
-	bi.UpdateRequestNonceFromCounterpart(hn)
+	bi.SetHandleNonce(hn)
+	bi.SetRequestNonceFromCounterpart(hn)
 
 	isRunning, err := bi.bridge.IsRunning(nil)
 	if err != nil {
@@ -346,29 +346,33 @@ func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEv
 	return nil
 }
 
-// UpdateRequestNonceFromCounterpart updates the request nonce from counterpart bridge.
-func (bi *BridgeInfo) UpdateRequestNonceFromCounterpart(nonce uint64) {
+// SetRequestNonceFromCounterpart sets the request nonce from counterpart bridge.
+func (bi *BridgeInfo) SetRequestNonceFromCounterpart(nonce uint64) {
 	if bi.requestNonceFromCounterPart < nonce {
 		vtRequestNonceCount.Inc(int64(nonce - bi.requestNonceFromCounterPart))
 		bi.requestNonceFromCounterPart = nonce
 	}
 }
 
-// UpdateRequestNonce updates the request nonce of the bridge.
-func (bi *BridgeInfo) UpdateRequestNonce(nonce uint64) {
+// SetRequestNonce sets the request nonce of the bridge.
+func (bi *BridgeInfo) SetRequestNonce(nonce uint64) {
 	if bi.requestNonce < nonce {
 		bi.requestNonce = nonce
 	}
 }
 
-// UpdateHandledNonce updates the handled nonce with new nonce.
-func (bi *BridgeInfo) UpdateHandledNonce(nonce uint64) {
+// MarkHandledNonce marks the handled nonce and set the handle nonce value.
+func (bi *BridgeInfo) MarkHandledNonce(nonce uint64) {
+	bi.SetHandleNonce(nonce)
+	bi.handledEvent.PutWithLimit(requestEvent{nonce}, 10000000)
+}
+
+// SetHandleNonce sets the handled nonce with new nonce.
+func (bi *BridgeInfo) SetHandleNonce(nonce uint64) {
 	if bi.handleNonce < nonce {
 		vtHandleNonceCount.Inc(int64(nonce - bi.handleNonce))
 		bi.handleNonce = nonce
 	}
-
-	bi.handledEvent.PutWithLimit(requestEvent{nonce}, 10000000)
 }
 
 // UpdateLowerHandleNonce updates the lower handle nonce.
@@ -395,7 +399,7 @@ func (bi *BridgeInfo) AddRequestValueTransferEvents(evs []*RequestValueTransferE
 			logger.Trace("List is full but add requestValueTransfer ", "newNonce", ev.Nonce(), "removedNonce", maxNonce)
 		}
 
-		bi.UpdateRequestNonceFromCounterpart(ev.RequestNonce + 1)
+		bi.SetRequestNonceFromCounterpart(ev.RequestNonce + 1)
 		bi.pendingRequestEvent.Put(ev)
 		vtPendingRequestEventCounter.Inc(1)
 	}
--- a/node/sc/bridge_manager.go
+++ b/node/sc/bridge_manager.go
@@ -363,7 +363,7 @@ func (bi *BridgeInfo) SetRequestNonce(nonce uint64) {
 
 // MarkHandledNonce marks the handled nonce and set the handle nonce value.
 func (bi *BridgeInfo) MarkHandledNonce(nonce uint64) {
-	bi.SetHandleNonce(nonce)
+	bi.SetHandleNonce(nonce + 1)
 	bi.handledEvent.PutWithLimit(requestEvent{nonce}, 10000000)
 }
 
--- a/node/sc/bridge_manager.go
+++ b/node/sc/bridge_manager.go
@@ -108,6 +108,16 @@ type BridgeInfo struct {
 
 	newEvent chan struct{}
 	closed   chan struct{}
+
+	handledEvent *bridgepool.ItemSortedMap
+}
+
+type requestEvent struct {
+	nonce uint64
+}
+
+func (ev requestEvent) Nonce() uint64 {
+	return ev.nonce
 }
 
 func NewBridgeInfo(sb *SubBridge, addr common.Address, bridge *bridgecontract.Bridge, cpAddr common.Address, cpBridge *bridgecontract.Bridge, account *accountInfo, local, subscribed bool, cpBackend Backend) (*BridgeInfo, error) {
@@ -131,6 +141,7 @@ func NewBridgeInfo(sb *SubBridge, addr common.Address, bridge *bridgecontract.Br
 		0,
 		make(chan struct{}),
 		make(chan struct{}),
+		bridgepool.NewItemSortedMap(),
 	}
 
 	if err := bi.UpdateInfo(); err != nil {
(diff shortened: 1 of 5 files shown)


nogang:
please check the checklist

node/sc/bridgepool/sorted_map_list.go
```diff
@@ -100,6 +100,23 @@ func (m *ItemSortedMap) Put(event itemWithNonce) {
...
+	m.put(event)
+}
+
+// PutWithLimit inserts a new item into the map with size limit of the map.
+func (m *ItemSortedMap) PutWithLimit(event itemWithNonce, limit int) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if len(m.items) >= limit {
+		return
```
nogang: Does the caller need to know that the item insertion failed?

node/sc/bridgepool/sorted_map_list.go
```diff
@@ -100,6 +100,23 @@ func (m *ItemSortedMap) Put(event itemWithNonce) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	m.put(event)
+}
+
+// PutWithLimit inserts a new item into the map with size limit of the map.
+func (m *ItemSortedMap) PutWithLimit(event itemWithNonce, limit int) {
```
nogang: If the new value is more suitable, do you have to delete the existing value and keep the new one?

node/sc/bridge_manager.go
```diff
@@ -211,7 +222,7 @@ func (bi *BridgeInfo) processingPendingRequestEvents() error {
 	logger.Trace("Get ready request value transfer event", "len(readyEvent)", len(ReadyEvent), "len(pendingEvent)", bi.pendingRequestEvent.Len())
 
 	for idx, ev := range ReadyEvent {
-		if ev.RequestNonce < bi.lowerHandleNonce {
+		if ev.RequestNonce < bi.lowerHandleNonce || bi.handledEvent.Exist(ev.RequestNonce) {
```
nogang: This seems to be less computational cost
```suggestion
		if ev.RequestNonce <= bi.lowerHandleNonce || bi.handledEvent.Exist(ev.RequestNonce) {
```

node/sc/bridge_manager.go
```diff
@@ -335,23 +346,29 @@ func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEv
...
 		bi.requestNonce = nonce
 	}
 }
 
-// UpdateHandledNonce updates the handled nonce with new nonce.
-func (bi *BridgeInfo) UpdateHandledNonce(nonce uint64) {
+// MarkHandledNonce marks the handled nonce and set the handle nonce value.
+func (bi *BridgeInfo) MarkHandledNonce(nonce uint64) {
+	bi.SetHandleNonce(nonce)
+	bi.handledEvent.PutWithLimit(requestEvent{nonce}, 10000000)
```
nogang: Is there a problem that it may be added to handle nonce but not put in event because of limit?

node/sc/bridge_manager.go
KimKyungup: `lowerHandleNonce` may not be handled. So we should not ignore it.

node/sc/bridge_manager.go
KimKyungup: No, there is no problem. 
If a transaction tries to handle the nonce, it will be prevented in the bridge contract.

node/sc/bridgepool/sorted_map_list.go
KimKyungup: Old one is already in the map is more important than the new one.

node/sc/bridgepool/sorted_map_list.go
KimKyungup: For now, there is nothing to do for that case. Therefore I didn't add a routine to return the error.

node/sc/bridge_manager.go
```diff
@@ -335,23 +346,29 @@ func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEv
...
+// SetRequestNonce sets the request nonce of the bridge.
+func (bi *BridgeInfo) SetRequestNonce(nonce uint64) {
 	if bi.requestNonce < nonce {
 		bi.requestNonce = nonce
 	}
 }
 
-// UpdateHandledNonce updates the handled nonce with new nonce.
-func (bi *BridgeInfo) UpdateHandledNonce(nonce uint64) {
+// MarkHandledNonce marks the handled nonce and set the handle nonce value.
```
jeonghonah:
```suggestion
// MarkHandledNonce marks the handled nonce and sets the handle nonce value.
```

node/sc/bridge_manager.go
```diff
@@ -335,23 +346,29 @@ func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEv
...
+func (bi *BridgeInfo) MarkHandledNonce(nonce uint64) {
+	bi.SetHandleNonce(nonce + 1)
+	bi.handledEvent.PutWithLimit(requestEvent{nonce}, 10000000)
```
jeonghonah: How about defining a const for the `10000000`?

node/sc/bridge_manager.go
```diff
@@ -335,23 +346,29 @@ func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEv
...
+}
+
+// SetHandleNonce sets the handled nonce with new nonce.
```
jeonghonah:
```suggestion
// SetHandleNonce sets the handled nonce with a new nonce.
```

node/sc/bridgepool/sorted_map_list.go
jeonghonah: How about simply setting the `limit` in the ItemSortedMap itself?

node/sc/bridgepool/sorted_map_list.go
```diff
@@ -203,6 +220,27 @@ func (m *ItemSortedMap) Remove(nonce uint64) bool {
...
+// maintenance.
+func (m *ItemSortedMap) Forward(threshold uint64) items {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	var removed items
+
+	// Pop off heap items until the threshold is reached
+	for m.index.Len() > 0 && (*m.index)[0] < threshold {
+		nonce := heap.Pop(m.index).(uint64)
+		removed = append(removed, m.items[nonce])
```
jeonghonah: How about removing the `items` which is not used currently?

node/sc/bridgepool/sorted_map_list.go
```diff
@@ -203,6 +220,27 @@ func (m *ItemSortedMap) Remove(nonce uint64) bool {
 	return true
 }
 
+// Forward removes all items from the map with a nonce lower than the
+// provided threshold. Every removed transaction is returned for any post-removal
+// maintenance.
+func (m *ItemSortedMap) Forward(threshold uint64) items {
```
jeonghonah: How about this?
```suggestion
func (m *ItemSortedMap) Truncate(threshold uint64) items {
```

KimKyungup: This name is from `txSortedMap`. So I think original name is better.

node/sc/bridgepool/sorted_map_list.go
KimKyungup: This `ItemSortedMap` is used for this PR and BridgeTxPool. I don't want to modify the construct(`NewItemSortedMap`). What do you think?

node/sc/bridgepool/sorted_map_list.go
KimKyungup: Returned items can be used. So not to remove is better as we talked.

node/sc/bridgepool/sorted_map_list.go
ehnuje: If we need to limit the size of `ItemSortedMap` in general, I think it is better to set the limit in `ItemSortedMap`.

node/sc/bridge_manager.go
```diff
@@ -108,6 +110,16 @@ type BridgeInfo struct {
 
 	newEvent chan struct{}
 	closed   chan struct{}
+
+	handledEvent *bridgepool.ItemSortedMap
+}
+
+type requestEvent struct {
```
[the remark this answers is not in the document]
KimKyungup: `ItemSortedMap` is support the item interface which has `Nonce() uint64` method.
So I created this.

node/sc/bridgepool/sorted_map_list.go
KimKyungup: I accepted you guys opinion. I appreciate your good opinion. = )
@jeonghonah @ehnuje @nogang

node/sc/bridgepool/sorted_map_list.go
KimKyungup: I added an error. :) Thank you so much.

node/sc/bridgepool/sorted_map_list.go
```diff
@@ -96,15 +103,21 @@ func (m *ItemSortedMap) Exist(nonce uint64) bool {
 
 // Put inserts a new item into the map, also updating the map's nonce
 // index. If a item already exists with the same nonce, it's overwritten.
-func (m *ItemSortedMap) Put(event itemWithNonce) {
+func (m *ItemSortedMap) Put(event itemWithNonce) error {
```
ehnuje: This `error` is not handled.

node/sc/bridgepool/sorted_map_list.go
```diff
@@ -96,15 +103,21 @@ func (m *ItemSortedMap) Exist(nonce uint64) bool {
...
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	if m.sizeLimit != 0 && len(m.items) >= m.sizeLimit {
+		return ErrSizeLimit
```
ehnuje: If it is hard to handle from the caller of `Put` function, it is better to leave a log here instead of returning the error.

ehnuje: Also, regarding the log, it would be better to leave a log with the information about the limit and `itemWithNonce`.

KimKyungup: In this PR, it is not hard to handle. There is no need to handle the error.
And If the log is needed, I think this level log is not proper, it is better for caller to make a log.
Because, this ItemSortedMap is for several purpose like bridge tx pool and handledEventMap.

node/sc/bridgepool/sorted_map_list.go
KimKyungup: Do you mean the caller doesn't handle the error?
I thought we don't need to handle it. :)

ehnuje: Thanks for the explanation! I understood your intention.

KimKyungup:
@ehnuje @jeonghonah @nogang I reflected your review. :) PTAL.

node/sc/bridgepool/bridge_tx_pool.go
```diff
@@ -329,7 +329,7 @@ func (pool *BridgeTxPool) add(tx *types.Transaction) error {
 	}
 
 	if pool.queue[from] == nil {
-		pool.queue[from] = NewItemSortedMap()
+		pool.queue[from] = NewItemSortedMap(0)
```
ehnuje: Is it okay to call `NewItemSortedMap(0)`?

KimKyungup: Yes. It means there is no limitation.
Before this PR, the itemSortedMap had no limitation. There is additional code to limitate for bridge tx pool.
The bridge tx pool is not related with this PR. So I wanted to retain the code.

ehnuje: Using `0` to indicate limitlessness looks a bit confusing to me. 
How about using `math.MaxInt` or declaring a const like `const UnlimitedItemSortedMap = -1`?

KimKyungup: `UnlimitedItemSortedMap` looks good. :) I will reflect.

PR #397, opened 2019-11-25
SC: Add a routine to update nonces during VTR

KimKyungup:
## Proposed changes

This PR added a routine to update nonces during VTR.

This PR was introduced for updating nonces periodically.
Before this PR, there is no periodic updating nonces.
Therefore the nonces can be updated by event.
If there is no anymore event or in abnormal situation, the nonce can't not be updated.

So this periodic updating is applied.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

This PR will be rebased after https://github.com/klaytn/klaytn/pull/394 is merged.

### DIFF
--- a/node/sc/vt_recovery.go
+++ b/node/sc/vt_recovery.go
@@ -205,12 +205,16 @@ func updateRecoveryHintFromTo(prevHint *valueTransferHint, from, to *BridgeInfo)
 	if err != nil {
 		return nil, err
 	}
+	from.SetRequestNonce(requestNonce)
+	to.SetRequestNonceFromCounterpart(requestNonce)
 	hint.requestNonce = requestNonce
 
 	handleNonce, err := to.bridge.LowerHandleNonce(nil)
 	if err != nil {
 		return nil, err
 	}
+	to.UpdateLowerHandleNonce(handleNonce)
+
 	if prevHint != nil {
 		hint.prevHandleNonce = prevHint.handleNonce
 		hint.candidate = prevHint.candidate


KimKyungup:
Dear reviewers, 
This PR will be rebased after https://github.com/klaytn/klaytn/pull/394.

node/sc/vt_recovery.go
```diff
@@ -205,12 +205,16 @@ func updateRecoveryHintFromTo(prevHint *valueTransferHint, from, to *BridgeInfo)
 	if err != nil {
 		return nil, err
 	}
+	from.SetRequestNonce(requestNonce)
```
jeonghonah: `vtrecovery` is optional. Should we update the nonce-related information here?

KimKyungup: You are right. First time I tried to make another routine to retrieve and update. 
But it was a duplicated with this code.
So I changed it.

I think VT recovery is almost default feature. 
So It's time to set vtrecovery as an the default option. :) 
What do you think?

jeonghonah: I agree with you. `vtrecovery` is actually not an option anymore. Let's make it a default.

KimKyungup:
I rebased this PR. PTAL.

KimKyungup:
> Could you please provide more background regarding the change?

I updated the content. Please take a look. :) Thanks.

KimKyungup:
> @KimKyungup Thanks for the explanation!
> So the periodic update is guaranteed by the go routine in `valueTransferRecovery.Start()`, right?

Right. First time, I tried to add new go routine. 
But This recovery routine was already there. 
So I added this code in the routine.  Thanks. :)

Issue #432, opened 2020-01-02
Ken panic during syncing


Jongsic:
```
INFO[12/31,11:05:09 Z] [5] Imported new chain segment                number=16322964 hash=b6a8c7…874fcd blocks=1   txs=0   elapsed=10.655ms      trieDBSize=1.13mB   mgas=0.000  mgasps=0.000
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0x8f5016]

goroutine 6447039 [running]:
github.com/klaytn/klaytn/blockchain/types.(*Transaction).ValidatedSender(0x0, 0x0, 0x0, 0xc000000000)
        /ext-go/1/src/github.com/klaytn/klaytn/blockchain/types/transaction.go:225 +0x16
github.com/klaytn/klaytn/blockchain.NewEVMContext(0x1645500, 0x0, 0xc10d6aab40, 0x161ab20, 0xc001832000, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /ext-go/1/src/github.com/klaytn/klaytn/blockchain/evm.go:54 +0xe5
github.com/klaytn/klaytn/node/cn.(*PrivateDebugAPI).traceBlock.func1(0xc15622e540, 0xc0df45dd40, 0xc0c0b30680, 0x19, 0x19, 0x163ad20, 0xc0f69ecff0, 0xc0e9456e70, 0xc010957440, 0x1630d60, ...)
        /ext-go/1/src/github.com/klaytn/klaytn/node/cn/api_tracer.go:483 +0x1d7
created by github.com/klaytn/klaytn/node/cn.(*PrivateDebugAPI).traceBlock
        /ext-go/1/src/github.com/klaytn/klaytn/node/cn/api_tracer.go:477 +0x45e
```

```
INFO[12/31,23:52:47 Z] [5] Imported new chain segment                number=16369022 hash=238a9f…1db20e blocks=1  txs=4   elapsed=29.954ms     trieDBSize=5.76mB   mgas=0.051   mgasps=1.712    ignored=1
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0x8f5016]

goroutine 2081661 [running]:
github.com/klaytn/klaytn/blockchain/types.(*Transaction).ValidatedSender(0x0, 0x0, 0x0, 0xc100000000)
[...]

PR #448, opened 2020-02-14
Introduce GCLock for cachedNode

KimKyungup:
## Proposed changes

This PR resolves https://github.com/klaytn/klaytn/issues/432 without increasing `triesInMemory`.
This added a lock for GC cachedNode.

During `traceBlock`, this GCLock can prevent to gc the cachedNode in the stateDB.

Also this added some exception code for error case of `AsMessageWithAccountKeyPicker` during `traceBlock`, `traceXXXX`.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

- https://github.com/klaytn/klaytn/issues/432

## Further comments

After this PR, I will rename caches related with state trie later.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -457,17 +457,17 @@ func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
 
 // StateAtWithGCLock returns a new mutable state based on a particular point in time with read lock of the state nodes.
 func (bc *BlockChain) StateAtWithGCLock(root common.Hash) (*state.StateDB, error) {
-	bc.LockGCCachedNode()
+	bc.RLockGCCachedNode()
 
 	exist := bc.stateCache.TrieDB().DoesExistCachedNode(root)
 	if !exist {
-		bc.UnlockGCCachedNode()
+		bc.RUnlockGCCachedNode()
 		return nil, ErrNotExistNode
 	}
 
 	stateDB, err := state.New(root, bc.stateCache)
 	if err != nil {
-		bc.UnlockGCCachedNode()
+		bc.RUnlockGCCachedNode()
 		return nil, err
 	}
 
@@ -1055,7 +1055,6 @@ func (bc *BlockChain) writeStateTrie(block *types.Block, state *state.StateDB) e
 	} else {
 		// Full but not archive node, do proper garbage collection
 		trieDB.Reference(root, common.Hash{}) // metadata reference to keep trie alive
-		bc.chPushBlockGCPrque <- gcBlock{root, block.NumberU64()}
 
 		// If we exceeded our memory allowance, flush matured singleton nodes to disk
 		var (
@@ -1074,61 +1073,40 @@ func (bc *BlockChain) writeStateTrie(block *types.Block, state *state.StateDB) e
 			logger.Trace("Commit the state trie into the disk", "blocknum", block.NumberU64())
 			trieDB.Commit(block.Header().Root, true, block.NumberU64())
 		}
+
+		bc.chPushBlockGCPrque <- gcBlock{root, block.NumberU64()}
 	}
 	return nil
 }
 
-// LockGCCachedNode locks the GC lock of CachedNode.
-func (bc *BlockChain) LockGCCachedNode() {
-	bc.stateCache.LockGCCachedNode()
+// RLockGCCachedNode locks the GC lock of CachedNode.
+func (bc *BlockChain) RLockGCCachedNode() {
+	bc.stateCache.RLockGCCachedNode()
 }
 
-// UnlockGCCachedNode unlocks the GC lock of CachedNode.
-func (bc *BlockChain) UnlockGCCachedNode() {
-	bc.stateCache.UnlockGCCachedNode()
+// RUnlockGCCachedNode unlocks the GC lock of CachedNode.
+func (bc *BlockChain) RUnlockGCCachedNode() {
+	bc.stateCache.RUnlockGCCachedNode()
 }
 
 // gcCachedNodeLoop runs a loop to gc.
 func (bc *BlockChain) gcCachedNodeLoop() {
-	var prqueLock sync.Mutex
-
-	gcTrigger := make(chan uint64)
 	trieDB := bc.stateCache.TrieDB()
 
-	bc.wg.Add(2)
+	bc.wg.Add(1)
 	go func() {
 		defer bc.wg.Done()
 		for {
 			select {
 			case block := <-bc.chPushBlockGCPrque:
-				prqueLock.Lock()
 				bc.triegc.Push(block.root, -float32(block.blockNum))
-				prqueLock.Unlock()
 				logger.Trace("Push GC block", "blkNum", block.blockNum, "hash", block.root.String())
-
-				select {
-				case gcTrigger <- block.blockNum:
-				default:
-					logger.Warn("Skip to trigger GC cached node", "blkNum", block.blockNum)
-				}
-			case <-bc.quit:
-				return
-			}
-		}
-	}()
-
-	go func() {
-		defer bc.wg.Done()
-		for {
-			select {
-			case blkNum := <-gcTrigger:
-				logger.Trace("trigger GC cached node by", "block", blkNum)
+				blkNum := block.blockNum
 				if blkNum > triesInMemory {
 					chosen := blkNum - triesInMemory
 
 					// Garbage collect anything below our required write retention
 					cnt := 0
-					prqueLock.Lock()
 					for !bc.triegc.Empty() {
 						root, number := bc.triegc.Pop()
 						if uint64(-number) > chosen {
@@ -1138,7 +1116,6 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 						trieDB.Dereference(root.(common.Hash))
 						cnt++
 					}
-					prqueLock.Unlock()
 
 					logger.Debug("GC cached node", "currentBlk", blkNum, "chosenBlk", chosen, "deferenceCnt", cnt)
 				}
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1095,8 +1095,8 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 	gcTrigger := make(chan uint64)
 	trieDB := bc.stateCache.TrieDB()
 
+	bc.wg.Add(2)
 	go func() {
-		bc.wg.Add(1)
 		defer bc.wg.Done()
 		for {
 			select {
@@ -1118,7 +1118,6 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 	}()
 
 	go func() {
-		bc.wg.Add(1)
 		defer bc.wg.Done()
 		for {
 			select {
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1101,24 +1101,25 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 			case block := <-bc.chBlock:
 				bc.triegc.Push(block.root, -float32(block.blockNum))
 				logger.Trace("Push GC block", "blkNum", block.blockNum, "hash", block.root.String())
+
 				blkNum := block.blockNum
-				if blkNum > triesInMemory {
-					chosen := blkNum - triesInMemory
-
-					// Garbage collect anything below our required write retention
-					cnt := 0
-					for !bc.triegc.Empty() {
-						root, number := bc.triegc.Pop()
-						if uint64(-number) > chosen {
-							bc.triegc.Push(root, number)
-							break
-						}
-						trieDB.Dereference(root.(common.Hash))
-						cnt++
-					}
-
-					logger.Debug("GC cached node", "currentBlk", blkNum, "chosenBlk", chosen, "deferenceCnt", cnt)
+				if blkNum <= triesInMemory {
+					continue
 				}
+
+				// Garbage collect anything below our required write retention
+				chosen := blkNum - triesInMemory
+				cnt := 0
+				for !bc.triegc.Empty() {
+					root, number := bc.triegc.Pop()
+					if uint64(-number) > chosen {
+						bc.triegc.Push(root, number)
+						break
+					}
+					trieDB.Dereference(root.(common.Hash))
+					cnt++
+				}
+				logger.Debug("GC cached node", "currentBlk", blkNum, "chosenBlk", chosen, "deferenceCnt", cnt)
 			case <-bc.quit:
 				return
 			}
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -107,8 +107,9 @@ type BlockChain struct {
 	chainConfigMu *sync.RWMutex
 	cacheConfig   *CacheConfig // stateDB caching and trie caching/pruning configuration
 
-	db     database.DBManager // Low level persistent database to store final content in
-	triegc *prque.Prque       // Priority queue mapping block numbers to tries to gc
+	db        database.DBManager // Low level persistent database to store final content in
+	triegc    *prque.Prque       // Priority queue mapping block numbers to tries to gc
+	gcTrigger chan uint64        // This delivers latest block number to gc loop
 
 	hc            *HeaderChain
 	rmLogsFeed    event.Feed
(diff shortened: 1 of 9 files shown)


storage/statedb/database.go
```diff
@@ -339,6 +340,16 @@ func (db *Database) DiskDB() database.DBManager {
 	return db.diskDB
 }
 
+// LockGCCachedNode locks the GC lock of CachedNode.
+func (db *Database) LockGCCachedNode() {
+	db.gcLock.Lock()
+}
+
+// UnLockGCCachedNode unlocks the GC lock of CachedNode.
+func (db *Database) UnLockGCCachedNode() {
```
kjhman21: We usually use 'Unlock' rather than 'UnLock'. Please check out the below call.

kjhman21:
1. Can we make a test case for this? e.g., make a fault when statedb is garbage collected?
2. I'm not sure a better solution, but using a lock in this way can cause unintended errors when others change this implementation. Is there any better way of using gc lock..?

node/cn/api_tracer.go
```diff
@@ -454,10 +460,23 @@ func (api *PrivateDebugAPI) traceBlock(ctx context.Context, block *types.Block,
...
 	if err != nil {
-		return nil, err
+		// If no state is locally available, the desired state will be generated.
+		statedb, err = api.computeStateDB(parent, reexec)
+		if err != nil {
+			return nil, err
+		}
+		logger.Debug("Get stateDB by computeStateDB", "block", block.NumberU64())
+	} else {
+		logger.Debug("Get stateDB from stateCache", "block", block.NumberU64(), "&stateDB", fmt.Sprintf("%p", statedb))
```
ehnuje: Is it helpful to leave the pointer value of stateDB?

blockchain/blockchain.go
```diff
@@ -1045,23 +1059,51 @@ func (bc *BlockChain) writeStateTrie(block *types.Block, state *state.StateDB) e
...
 		}
 
-		if current := block.NumberU64(); current > triesInMemory {
-			// Find the next state trie we need to commit
-			header := bc.GetHeaderByNumber(current - triesInMemory)
-			chosen := header.Number.Uint64()
+		// trigger to GC stateDB cache
+		select {
+		case bc.gcTrigger <- block.NumberU64():
+		default:
```
ehnuje: If block number is not able to be passed to the `bc.gcTrigger` channel, it does nothing. Is it what you expected?

node/cn/api_tracer.go
KimKyungup: I used this during debugging. Now I don't need. I will remove. Thanks.

blockchain/blockchain.go
KimKyungup: Yes it is. This routine is main routine. 
So if this trigger can't pass the signal and it waits, the main routine has some problem. 

I don't want this situation. Therefore this signal is not buffered. 
But, if a trigger is ignored, it will try again in next block.

KimKyungup:
> 2\. I'm not sure a better solution, but using a lock in this way can cause unintended errors when others change this implementation. Is there any better way of using gc lock..?

KimKyungup:
> 1. Can we make a test case for this? e.g., make a fault when statedb is garbage collected?
- If a stateDB has a fault caused by any reason, getObject() will try to find the object from TrieNodeCahce (BigCache) or LevelDB. If it is impossible to get the object from them, it will return nil and it can make some same situation of the issue. Let's consider more test cases.


> 2. I'm not sure a better solution, but using a lock in this way can cause unintended errors when others change this implementation. Is there any better way of using gc lock..?

Before this solution, I thought some solution like below.
- TrieInMemory 4 -> 128 : It will increase memory and increase `Cap()` also. More `Cap()` can decrease performance.
- Every `traceBlock` or `traceSomething` always regenerate the state before the snapshot. 
: This makes some decrease the performance of traceBlock.

I tried some other solution but I found the key of this problem is GC.
This change doesn't make some effect for main routine.  Only 1 lock/unlock routine is added in Dereference(). There is no other routine related with the lock.

But this change make an effect when trace APIs are called.

blockchain/blockchain.go
kjhman21: Hmm... but you need to take care of this because it may change original behavior. Please add some logs for the unwanted case.

ehnuje: I also agree with the @kjhman21 's point that we need to leave a log here. 
Log would be needed for the both cases.

blockchain/blockchain.go
```diff
@@ -444,6 +447,17 @@ func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
...
 
+// StateAtWithGCLock returns a new mutable state based on a particular point in time with read lock of the state nodes.
+func (bc *BlockChain) StateAtWithGCLock(root common.Hash) (*state.StateDB, error) {
+	bc.LockGCCachedNode()
+	exist := bc.stateCache.TrieDB().DoesExistCachedNode(root)
+	if exist {
+		return state.New(root, bc.stateCache)
+	}
+	bc.UnlockGCCachedNode()
+	return nil, errors.New("the node does not exist in state cache")
```
ehnuje: Please define a variable for this error.

blockchain/blockchain.go
```diff
@@ -444,6 +447,17 @@ func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
 	return state.New(root, bc.stateCache)
 }
 
+// StateAtWithGCLock returns a new mutable state based on a particular point in time with read lock of the state nodes.
+func (bc *BlockChain) StateAtWithGCLock(root common.Hash) (*state.StateDB, error) {
```
ehnuje: I'm a bit worried about this function since the caller of the function should take care of unlocking.

blockchain/blockchain.go
```diff
@@ -444,6 +447,17 @@ func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
...
+	bc.LockGCCachedNode()
+	exist := bc.stateCache.TrieDB().DoesExistCachedNode(root)
+	if exist {
+		return state.New(root, bc.stateCache)
```
ehnuje: What if `state.New` returns an error? We don't unlock here, but the caller also doesn't unlock the lock.

blockchain/blockchain.go
KimKyungup: Reflected. :)

blockchain/blockchain.go
KimKyungup: Thanks. Reflected. :)

blockchain/blockchain.go
KimKyungup: Yes. You should pay attention to the use of this feature.
If you are too worried about this, we have the option of moving this lock / error handling code outward.

blockchain/blockchain.go
KimKyungup: I added a log here and inside of the loop. Thanks.

kjhman21: What about adding a log for `default:` case? I think this is unusual and in that case one block number can be missing. Can we use `logger.Error()` for `default:` case?

blockchain/blockchain.go
kjhman21: > Yes. You should pay attention to the use of this feature.

Actually, this is what I addressed before. Taking care of unlocking outside of this function is very error-prone.

blockchain/blockchain.go
KimKyungup: That default situation is not error case. 
It is intended case.
So How about `logger.Warn()`?

ehnuje: logger.Warn() seems to be enough for this case (unless it skips too much)

KimKyungup: I added warn log. Thanks. :)

KimKyungup:
@kjhman21 @ehnuje @winnie-byun 
I added 6699fa9 for preventing data race of `triegc             *prque.Prque`.

As you know, I changed the gc routine as a goroutine.
This routine makes a data race with `bc.triegc.Push(root, -float32(block.NumberU64()))` in `writeStateTrie` and another GC logic in `blockchain.Stop()`.
So I moved the push logic into the loop routine and applied wait group to the loop.

But, I think it is a bit complex. PTAL.

blockchain/blockchain.go
```diff
@@ -179,20 +188,21 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig
...
-		badBlocks:       badBlocks,
-		parallelDBWrite: db.IsParallelDBWrite(),
-		nonceCache:      nonceCache,
-		balanceCache:    balanceCache,
+		chainConfig:        chainConfig,
+		chainConfigMu:      new(sync.RWMutex),
+		cacheConfig:        cacheConfig,
+		db:                 db,
+		triegc:             prque.New(),
+		chPushBlockGCPrque: make(chan gcBlock, 1000),
```
ehnuje: How about using simpler name like below?
```suggestion
		gcBlockCh: make(chan gcBlock, 1000),
```

storage/statedb/database.go
```diff
@@ -98,6 +98,7 @@ type Database struct {
 	gctime  time.Duration      // Time spent on garbage collection since last commit
 	gcnodes uint64             // Nodes garbage collected since last commit
 	gcsize  common.StorageSize // Data storage garbage collected since last commit
+	gcLock  sync.RWMutex       // Lock for preventing to garbage collect cachedNode without flushing.
```
ehnuje: Minor suggestion, but just for consistency.
```suggestion
	gcLock  sync.RWMutex       // Lock for preventing to garbage collect cachedNode without flushing
```

storage/statedb/database.go
```diff
@@ -431,13 +442,15 @@ func (db *Database) node(hash common.Hash) node {
...
+
+	// Retrieve the node from the state cache if available
 	db.lock.RLock()
 	node := db.nodes[hash]
 	db.lock.RUnlock()
-
 	if node != nil {
 		return node.obj(hash)
 	}
+
```
ehnuje: Changes in L#437 and L#453 seems unnecessary.

storage/statedb/database.go
```diff
@@ -474,6 +487,13 @@ func (db *Database) Node(hash common.Hash) ([]byte, error) {
 	return enc, err
 }
 
+// DoesExistCachedNode return if the noce exist from cached trie node in memory.
```
ehnuje:
```suggestion
// DoesExistCachedNode returns if the node exists on the cached trie node in memory.
```

storage/statedb/database.go
```diff
@@ -474,6 +487,13 @@ func (db *Database) Node(hash common.Hash) ([]byte, error) {
...
+func (db *Database) DoesExistCachedNode(hash common.Hash) bool {
```
ehnuje: Don't we need to look for the node in the trie node cache?
Also, it seems that you need to lock the `db.lock.RLock()` according to the line 475~477.

node/cn/handler_test.go
```diff
@@ -112,6 +112,24 @@ func newBlock(blockNum int) *types.Block {
 	return block
 }
 
+func newBlockWithParentHash(blockNum int, parentHash common.Hash) *types.Block {
```
ehnuje: Why is this function needed?

node/cn/api_tracer.go
```diff
@@ -780,16 +826,24 @@ func (api *PrivateDebugAPI) computeTxEnv(blockHash common.Hash, txIndex int, ree
...
 		return nil, vm.Context{}, nil, fmt.Errorf("parent %#x not found", block.ParentHash())
 	}
-	statedb, err := api.computeStateDB(parent, reexec)
+
+	statedb, deferFn, err := api.stateAt(parent, reexec)
 	if err != nil {
-		return nil, vm.Context{}, nil, err
+		return nil, vm.Context{}, nil, fmt.Errorf("can not get the state of block %#x: %v", parent.Root(), err)
 	}
+	defer deferFn()
```
ehnuje: `deferFn()` is not always called.

node/cn/api_tracer.go
```diff
@@ -769,6 +795,26 @@ func (api *PrivateDebugAPI) traceTx(ctx context.Context, message blockchain.Mess
...
+	if err != nil {
+		// If no state is locally available, the desired state will be generated.
+		stateDB, err = api.computeStateDB(block, reexec)
+		if err != nil {
+			return nil, func() {}, err
+		}
+		logger.Debug("Get stateDB by computeStateDB", "block", block.NumberU64())
+		return stateDB, func() {}, nil
+	}
+	logger.Debug("Get stateDB from stateCache", "block", block.NumberU64())
```
ehnuje: How about including `reexec` to the log?

node/cn/handler_test.go
KimKyungup: In `TestPrivateDebugAPI_computeTxEnv` unit test, there is a logic to check the parent hash of the block.

storage/statedb/database.go
KimKyungup: I added lines for readability.

storage/statedb/database.go
KimKyungup: I added Lock. Thanks. :)

blockchain/blockchain.go
```diff
@@ -1045,23 +1074,56 @@ func (bc *BlockChain) writeStateTrie(block *types.Block, state *state.StateDB) e
...
+	bc.wg.Add(1)
+	go func() {
+		defer bc.wg.Done()
+		for {
+			select {
+			case block := <-bc.chBlock:
+				bc.triegc.Push(block.root, -float32(block.blockNum))
+				logger.Trace("Push GC block", "blkNum", block.blockNum, "hash", block.root.String())
+				blkNum := block.blockNum
+				if blkNum > triesInMemory {
```
ehnuje:
```suggestion
				if blkNum <= triesInMemory {
				    continue
				}
```
This will reduce an indentation 😅

node/cn/api_tracer.go
```diff
@@ -769,6 +795,26 @@ func (api *PrivateDebugAPI) traceTx(ctx context.Context, message blockchain.Mess
...
 }
 
+// stateAt returns the given block's state from cached stateDB with the GC lock or by regenerating.
+func (api *PrivateDebugAPI) stateAt(block *types.Block, reexec uint64) (*state.StateDB, func(), error) {
+	var stateDB *state.StateDB
+
+	// If we have the state fully available, use that.
+	stateDB, err := api.cn.blockchain.StateAtWithGCLock(block.Root())
+	if err != nil {
+		// If no state is locally available, the desired state will be generated.
```
kjhman21: What about defining a function variable here?  e.g.,
```
var f = func() {}

...

return nil, f, err

...

return stateDB, f, nil
```

storage/statedb/database.go
```diff
@@ -541,6 +563,9 @@ func (db *Database) reference(child common.Hash, parent common.Hash) {
 
 // Dereference removes an existing reference from a root node.
 func (db *Database) Dereference(root common.Hash) {
+	db.gcLock.Lock()
```
kjhman21: hmm.. I'm not 100% sure, but it seems that `bd.lock` is for `db.nodes`.

Should we have a different lock for gc?

blockchain/blockchain.go
KimKyungup: I updated. :)

node/cn/api_tracer.go
KimKyungup: I changed it. :)

storage/statedb/database.go
KimKyungup: `database.lock` is for reading/writing only `database.nodes`.
But `gcLock` is for preventing to gc trie node in `database.nodes` without writing to persistent database.

In `StateAtWithGCLock`, with `RGCLock`, it returns the stateDB.
During the process, `DoesExistCachedNode` check if the state root is available with `database.lock`. 

After the function returns, the root can be removed by `Cap` or ` Commit`.
But in this cases, the root already was written on persistent database during `Cap`, `Commit`.

So, the stateDB is valid.

KimKyungup:
@kjhman21 @ehnuje @winnie-byun PTAL. :)

kjhman21:
@KimKyungup Please add test cases and results for this PR.

KimKyungup:
I attached the test code in RPC console.
This code can try call traceBlock API repeatedly.

Before this PR, it always makes the issued situation.
But after this PR, the issued situation is resolved.

```
for (i=0;i<10;i++) {
	debug.traceBlockByNumber(klay.blockNumber,{tracer:"callTracer"})
	debug.traceBlockByNumber(klay.blockNumber-1,{tracer:"callTracer"})
	debug.traceBlockByNumber(klay.blockNumber-2,{tracer:"callTracer"})
	debug.traceBlockByNumber(klay.blockNumber-3,{tracer:"callTracer"})
	debug.traceBlockByNumber(klay.blockNumber-4,{tracer:"callTracer"})
}
```

PR #476, opened 2020-04-06
Introduce auto restart feature

KimKyungup:
## Proposed changes
This PR introduce auto restart node feature.
If there is no chain head event during timeout (15 mins), the watchdog triggers to restart the node.
This feature can resolve the delay situation like consensus delay or any problem of sync.

This feature is introduced to resolve the consensus delay.
Therefore, the auto restart feature is enabled by default for CN.
However, for EN/PN, the feature is disable by default
If user wants to use this feature, he/she can use it.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/build/packaging/linux/bin/kbnd
+++ b/build/packaging/linux/bin/kbnd
@@ -79,6 +79,10 @@ start() {
         OPTIONS="$OPTIONS --datadir $DATA_DIR"
     fi
 
+    BASEDIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+    CURRENTFILE=`basename "$0"`
+    OPTIONS="$OPTIONS --autorestart.daemon.path $BASEDIR/$CURRENTFILE"
+
     $BIN/kbn ${OPTIONS} >> ${LOG_DIR}/kbnd.out 2>&1 &
     RETVAL=$?
     PIDNUM=$!
@@ -110,6 +114,7 @@ status() {
 
 restart() {
     stop
+    sleep 3
     start
 }
 
--- a/build/packaging/linux/conf/kcnd_baobab.conf
+++ b/build/packaging/linux/conf/kcnd_baobab.conf
@@ -37,6 +37,7 @@ WS_PORT=8552
 WS_ORIGINS="*"
 
 # Setting 1 is to enable options, otherwise disabled.
+AUTO_RESTART=1
 METRICS=1
 PROMETHEUS=1
 DB_NO_PARALLEL_WRITE=0
--- a/cmd/utils/nodecmd/flags_test.go
+++ b/cmd/utils/nodecmd/flags_test.go
@@ -766,6 +766,17 @@ var flagsWithValues = []struct {
 		wrongValues: commonTwoErrors,
 		errors:      []int{ErrorInvalidValue, ErrorInvalidValue},
 	},
+	{
+		flag:        "--autorestart.daemon.path",
+		flagType:    FlagTypeArgument,
+		values:      []string{"~/klaytn/bin/kcnd", "~/klaytn/bin/kpnd", "~/klaytn/bin/kend"},
+		wrongValues: commonThreeErrors,
+		errors:      []int{NonError, NonError, NonError},
+	},
+	{
+		flag:     "--autorestart.enable",
+		flagType: FlagTypeBoolean,
+	},
 }
 
 func testFlags(t *testing.T, flag string, value string, idx int) {
--- a/cmd/utils/nodecmd/flags_test.go
+++ b/cmd/utils/nodecmd/flags_test.go
@@ -777,6 +777,13 @@ var flagsWithValues = []struct {
 		flag:     "--autorestart.enable",
 		flagType: FlagTypeBoolean,
 	},
+	{
+		flag:        "--autorestart.timeout",
+		flagType:    FlagTypeArgument,
+		values:      []string{"10m", "60s", "1h"},
+		wrongValues: commonThreeErrors,
+		errors:      []int{ErrorInvalidValue, ErrorInvalidValue, ErrorInvalidValue},
+	},
 }
 
 func testFlags(t *testing.T, flag string, value string, idx int) {
--- a/node/cn/backend.go
+++ b/node/cn/backend.go
@@ -321,9 +321,9 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 		}
 	}
 
-	var resetFn func()
+	var restartFn func()
 	if config.AutoRestartFlag {
-		resetFn = func() {
+		restartFn = func() {
 			daemonPath := config.DaemonPathFlag
 			logger.Warn("Restart node", "command", daemonPath+" restart")
 			cmd := exec.Command(daemonPath, "restart")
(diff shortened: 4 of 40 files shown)


node/cn/backend.go
```diff
@@ -320,8 +321,18 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 		}
 	}
 
+	var resetFn func()
```
kjhman21, line 324: This could lead to nil pointer dereference. You need to set empty function to this variable to avoid that.

build/packaging/linux/conf/kcnd_baobab.conf
```diff
@@ -37,6 +37,7 @@ WS_PORT=8552
 WS_ORIGINS="*"
 
 # Setting 1 is to enable options, otherwise disabled.
+AUTO_RESTART=1
```
ehnuje, line 40: It seems that CNs are set to 1 and ENs are set to 0, right? 
If so, could you please explain the default value by node type at the PR description?
Also, is there any reason not to restart the EN by default?

Additionally, there seems to be no `.conf` for PNs. Could you please check?

work/worker.go
```diff
@@ -289,11 +295,27 @@ func (self *worker) update() {
 	quitByErr := make(chan bool, 1)
 	go self.handleTxsCh(quitByErr)
 
+	// Initialize reset watchdog
+	if self.nodetype == common.CONSENSUSNODE && self.resetFn != nil && self.resetTimeOut > 0 {
+		callback := func() {
+			logger.Warn("Consensus timeout")
+			self.resetFn()
+		}
+		logger.Info("Initialize auto reset watchdog", "timeout", self.resetTimeOut.String())
```
ehnuje, line 304: "auto reset" or "auto restart"? It seems that the usages of "reset" and "restart" are mixed. 
"restart" seems to be more appropriate.

work/worker.go
```diff
@@ -289,11 +295,27 @@ func (self *worker) update() {
...
+		self.resetWatchdog = time.AfterFunc(self.resetTimeOut, callback)
+		defer self.resetWatchdog.Stop()
+	}
+
 	for {
 		// A real event arrived, process interesting content
 		select {
 		// Handle ChainHeadEvent
 		case <-self.chainHeadCh:
+			// Refresh reset watchdog
```
ehnuje, line 314: How about adding a new `case` with `time.Timer` instead of using `time.AfterFunc`?

ehnuje:
```
case <-resetTimer.C:
    self.restartFn() // It does nothing if an empty function is set
```

work/worker.go
KimKyungup: I changed `reset` to `restart`. Thanks.

work/worker.go
KimKyungup: I thought this routine is simple.
Also if I add the watchdog reset routine in case statements, the watchdog can't work properly when the loop has a problem.

work/worker.go
```diff
@@ -289,11 +295,27 @@ func (self *worker) update() {
...
 	go self.handleTxsCh(quitByErr)
 
+	// Initialize restart watchdog
+	if self.nodetype == common.CONSENSUSNODE && self.restartFn != nil && self.restartTimeOut > 0 {
+		callback := func() {
+			logger.Warn("Consensus timeout")
+			self.restartFn()
+		}
+		logger.Info("Initialize auto restart watchdog", "timeout", self.restartTimeOut.String())
+		self.restartWatchdog = time.AfterFunc(self.restartTimeOut, callback)
```
ehnuje, line 305:
```suggestion
		self.restartWatchdog = time.AfterFunc(self.restartTimeOut, self.restartFn)
```

work/worker.go
```diff
@@ -289,11 +295,27 @@ func (self *worker) update() {
 	quitByErr := make(chan bool, 1)
 	go self.handleTxsCh(quitByErr)
 
+	// Initialize restart watchdog
+	if self.nodetype == common.CONSENSUSNODE && self.restartFn != nil && self.restartTimeOut > 0 {
+		callback := func() {
+			logger.Warn("Consensus timeout")
+			self.restartFn()
+		}
+		logger.Info("Initialize auto restart watchdog", "timeout", self.restartTimeOut.String())
```
ehnuje, line 304:
```suggestion
		logger.Info("Initialize auto restart watchdog", "timeout", self.restartTimeOut)
```
`time.Duration` will be printed as expected without calling `String()`. Could you please check?

work/worker.go
```diff
@@ -289,11 +295,27 @@ func (self *worker) update() {
 	quitByErr := make(chan bool, 1)
 	go self.handleTxsCh(quitByErr)
 
+	// Initialize restart watchdog
+	if self.nodetype == common.CONSENSUSNODE && self.restartFn != nil && self.restartTimeOut > 0 {
+		callback := func() {
```
ehnuje, line 300: Do we need to wrap `restartFn` here? How about directly pass `self.restartFn` to `time.AfterFunc`?

work/worker.go
KimKyungup: Good. I reflected.

work/worker.go
KimKyungup: Good. I reflected. I checked the same result without "String()".

work/worker.go
KimKyungup: You are right. I updated the code like you want.

build/packaging/linux/conf/kcnd_baobab.conf
KimKyungup: Right. This is enabled by default in CN only. I added the description.

I added `.conf` file of PNs. Thanks.

KimKyungup:
This feature was tested. 
Below shows the block number / current round of CNs.


The consensus delay was intentionally triggered every 150, 250, 350 ... blocks.
And The restart feature resolved the consensus delay.

Some consensus delay was resolved by committee member change without restarting.

image

node/cn/backend.go
```diff
@@ -320,8 +321,18 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 		}
 	}
 
+	var restartFn func()
```
KimKyungup, line 324: @kjhman21 This `restartFn` will be used if it is not nil as below code.
So It is ok. 

```
if self.restartFn != nil && self.restartTimeOut > 0 {
```

node/cn/backend.go
KimKyungup: @kjhman21 This `restartFn` will be used if it is not nil as below code.
So It is ok. 

```
if self.restartFn != nil && self.restartTimeOut > 0 {
```

node/cn/backend.go
kjhman21: Okay. I think it would be better to set an empty function to this variable and remove the condition in the `if`.
I don't mind anyway. It's your call. :)

kjhman21:
@KimKyungup Is it possible to check what happens if it is triggered on normal status?

node/cn/config.go
```diff
@@ -139,6 +139,11 @@ type Config struct {
 
 	// use separate network different from baobab or cypress
 	IsPrivate bool
+
+	// Restart
+	AutoRestartFlag    bool
+	RestartTimeOutFlag time.Duration
+	DaemonPathFlag     string
```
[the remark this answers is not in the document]
KimKyungup: Done. Thanks.

KimKyungup:
PTAL. :-)

build/packaging/linux/bin/kbnd
```diff
@@ -110,6 +114,7 @@ status() {
 
 restart() {
     stop
+    sleep 3
```
KimKyungup, line 117: test

cmd/utils/nodecmd/flags_test.go
```diff
@@ -766,6 +766,17 @@ var flagsWithValues = []struct {
...
 	},
+	{
+		flag:        "--autorestart.daemon.path",
+		flagType:    FlagTypeArgument,
+		values:      []string{"~/klaytn/bin/kcnd", "~/klaytn/bin/kpnd", "~/klaytn/bin/kend"},
+		wrongValues: commonThreeErrors,
+		errors:      []int{NonError, NonError, NonError},
+	},
+	{
+		flag:     "--autorestart.enable",
```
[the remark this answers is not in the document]
KimKyungup: Thankas. I added. :)

KimKyungup:
image

PR #532, opened 2020-06-02
Support for state migration to checks state storage and new dump API

KimKyungup:
## Proposed changes

This PR adds the feature to check state/storage trie and adds new dump API.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/state/iterator.go
+++ b/blockchain/state/iterator.go
@@ -39,7 +39,6 @@ type NodeIterator struct {
 
 	accountHash common.Hash // Hash of the node containing the account
 	codeHash    common.Hash // Hash of the contract source code
-	addrHash    common.Hash // Hash of address of the account
 	Code        []byte      // Source code associated with a contract
 
 	Type   string
--- a/blockchain/state/iterator.go
+++ b/blockchain/state/iterator.go
@@ -42,6 +42,7 @@ type NodeIterator struct {
 
 	Hash   common.Hash // Hash of the current entry being iterated (nil if not standalone)
 	Parent common.Hash // Hash of the first full ancestor node (nil if current is the root)
+	Path   []byte      // the hex-encoded path to the current node.
 
 	Error error // Failure set in case of an internal error in the iteratord
 }
@@ -148,14 +149,14 @@ func (it *NodeIterator) retrieve() bool {
 	// Otherwise retrieve the current entry
 	switch {
 	case it.dataIt != nil:
-		it.Hash, it.Parent = it.dataIt.Hash(), it.dataIt.Parent()
+		it.Hash, it.Parent, it.Path = it.dataIt.Hash(), it.dataIt.Parent(), it.dataIt.Path()
 		if it.Parent == (common.Hash{}) {
 			it.Parent = it.accountHash
 		}
 	case it.Code != nil:
 		it.Hash, it.Parent = it.codeHash, it.accountHash
 	case it.stateIt != nil:
-		it.Hash, it.Parent = it.stateIt.Hash(), it.stateIt.Parent()
+		it.Hash, it.Parent, it.Path = it.stateIt.Hash(), it.stateIt.Parent(), it.stateIt.Path()
 	}
 	return true
 }
--- a/blockchain/state/iterator.go
+++ b/blockchain/state/iterator.go
@@ -25,6 +25,7 @@ import (
 	"github.com/klaytn/klaytn/blockchain/types/account"
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/ser/rlp"
+	"github.com/klaytn/klaytn/storage/database"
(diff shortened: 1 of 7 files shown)


blockchain/state/iterator.go
```diff
@@ -38,10 +39,13 @@ type NodeIterator struct {
 
 	accountHash common.Hash // Hash of the node containing the account
 	codeHash    common.Hash // Hash of the contract source code
+	addrHash    common.Hash // Hash of address of the account
```
winnie-byun, line 42: What is `addrHash` for? It is not used.

blockchain/state/iterator.go
```diff
@@ -148,14 +153,98 @@ func (it *NodeIterator) retrieve() bool {
...
+			it.Type = "state_leaf"
+		}
+
+		it.Hash, it.Parent, it.Path = it.stateIt.Hash(), it.stateIt.Parent(), it.stateIt.Path()
 	}
 	return true
 }
+
+// CheckStateConsistency checks the consistency of all state/storage trie of given two state database.
+func CheckStateConsistency(oldDB database.DBManager, newDB database.DBManager, root common.Hash) error {
```
winnie-byun, line 181: What do you think of using`reflect.DeepEqual` for checking?

blockchain/state/iterator.go
```diff
@@ -148,14 +153,98 @@ func (it *NodeIterator) retrieve() bool {
 	// Otherwise retrieve the current entry
 	switch {
 	case it.dataIt != nil:
-		it.Hash, it.Parent = it.dataIt.Hash(), it.dataIt.Parent()
+		it.Type = "storage"
```
winnie-byun, line 156: How about using `it.Type`'s own enum type?

blockchain/state/iterator.go
KimKyungup: It can be removed. I will. :)

node/cn/api.go
```diff
@@ -240,6 +240,49 @@ func (api *PublicDebugAPI) DumpBlock(blockNr rpc.BlockNumber) (state.Dump, error
...
+}
+
+type DumpStateTrieResult struct {
+	Root  string `json:"root"`
+	Tries []Trie `json:"tries"`
+}
+
+// DumpStateTrie retrieves all state/storage tries of the given state root.
+func (api *PublicDebugAPI) DumpStateTrie(blockNr uint64) (DumpStateTrieResult, error) {
+	//func (api *PublicDebugAPI) DumpStateTrie(blockNr rpc.BlockNumber) error {
```
winnie-byun, line 257: Any reason why there is a comment? Also, `rpc.BlockNumber` might be better idea then `uint64`.

blockchain/state/iterator.go
KimKyungup: This `type` will be printed out in API.
So if we can use enum type, we need additional code to convert string.
I think this is enough. :)

node/cn/api.go
KimKyungup: I removed.
`rpc.BlockNumber` doesn't support decimal, supports only hex.
So I changed it.

blockchain/state/iterator.go
KimKyungup: `reflect.DeepEqual` is better? :)

winnie-byun: If you are checking all of the variables, it is could be better. It makes sure there is no error in our code.

winnie-byun: `state.NodeIterator` has pointers which leads to unequality. I will take my suggestion back. Thank you for considering.

storage/statedb/encoding.go
```diff
@@ -119,3 +123,13 @@ func prefixLen(a, b []byte) int {
 func hasTerm(s []byte) bool {
 	return len(s) > 0 && s[len(s)-1] == 16
 }
+
+// HexKeyPathToHashString turns hex nibbles of trie path key into hash string of the trie path key.
+func HexKeyPathToHashString(path []byte) string {
```
winnie-byun, line 128: Why do you turn `Path` to `Hash`? Isn't it better if we can how the nodes are connected?

KimKyungup: Path is bytes. Hash is ascii. 
So hash is more short and understandable. :)
This is an example. 
`10101010 -> 0xAAAA`

KimKyungup:
PTAL. :)

storage/statedb/encoding.go
winnie-byun: To me, the word `Hash` sounds like you hashed the path.
Could you change the function name to `HexPathToString` or `HexKeyPathToEncodedString`?
```suggestion
// HexPathToString turns hex nibbles of trie path key into string of the trie path key.
func HexPathToString(path []byte) string {
```

PR #545, opened 2020-06-16
Istanbul message handler checks whether the sender is a member of the committee

aidan-kwon:
## Proposed changes

**AS-IS** 
Istanbul message handler checks whether the sender is a validator.
However, it doesn't check whether the sender is a member of the committee. 

*Committee is a subset of validators and composed by the block number and the round number.

**TO-BE**
Istanbul message handler checks whether the sender is a member of the committee. 

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

If a validator is not a member of the committee, he/she doesn't propagate prepare/commit/roundChange messages, but he/she sends them to eventMux of its Istanbul core. It may result in the mismatch of message count among validators and prevent achieving a consensus.

This issue can be mitigated with this PR by validating the sender in the receiver side. Moreover, we can validate the sender in the sender side also before it sends messages. It will be handled in the following PR.


consensus/istanbul/core/commit.go
```diff
@@ -71,6 +71,9 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 	}
 
 	//logger.Error("receive handle commit","num", commit.View.Sequence)
+	if !c.valSet.CheckInSubList(msg.Hash, commit.View, src.Address()) {
+		return errNotFromCommittee
+	}
 
 	if err := c.checkMessage(msgCommit, commit.View); err != nil {
```
ehnuje, line 78: Is it possible to call `CheckInSubList` inside `checkMessage`? Don't we need to check if the sender is in the committee for `preprepare` or `processBacklog`?

consensus/istanbul/core/commit.go
```diff
@@ -71,6 +71,9 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 	}
 
 	//logger.Error("receive handle commit","num", commit.View.Sequence)
+	if !c.valSet.CheckInSubList(msg.Hash, commit.View, src.Address()) {
+		return errNotFromCommittee
+	}
```
KimKyungup, line 76: How about add this routine in `handleCheckedMsg` to prevent duplicated code?

consensus/istanbul/backend/backend.go
```diff
@@ -202,12 +202,7 @@ func (sb *backend) Gossip(valSet istanbul.ValidatorSet, payload []byte) error {
 
 // checkInSubList checks if the node is in a sublist
 func (sb *backend) checkInSubList(prevHash common.Hash, valSet istanbul.ValidatorSet) bool {
```
KimKyungup, line 204:
```suggestion
func (sb *backend) checkInCommittee(prevHash common.Hash, valSet istanbul.ValidatorSet) bool {
```

aidan-kwon: "SubList" is commonly used in overall code instead of "Committee".
I will consider changing all of them later.

consensus/istanbul/core/commit.go
aidan-kwon: 1. The message needs to be decoded in each format to get the `View` used in calculating the sublist. If this routine is moved to `handleCheckedMsg`, the message should be decoded twice. (decoded structures are not an interface)
2.`preprepare` messages don't need to be checked whether in the sublist. If we apply different routines for each message type, `handleCheckedMsg` will be complicated.

consensus/istanbul/core/commit.go
aidan-kwon: We don't need to check `preprepare` message since it only has one valid sender(proposer) at a time and is checked with its own routine.  
Because `CheckInSubList` is not a common routine for all Istanbul message, I think it is better to place `CheckInSubList` in the current position.

consensus/istanbul/core/commit.go
KimKyungup: Why preprepare message does not need to check it?

consensus/istanbul/core/handler_test.go
```diff
@@ -0,0 +1,454 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if err := eventMux.Post(istanbulMsg); err != nil {
+			t.Fatal(err)
+		}
+
+		time.Sleep(time.Second)
+		assert.Nil(t, istCore.current.Preprepare)
```
KimKyungup, line 257: How can this message be ignored?

consensus/istanbul/core/commit.go
aidan-kwon: It only has one valid sender(proposer) at a time and is checked with its own routine.

KimKyungup:
@andybclee @laggu Please take a look. :)

consensus/istanbul/backend/backend.go
laggu: what about remove `func (sb *backend) checkInSubList` and use valSet.CheckInSubList directly on line 243?

consensus/istanbul/core/roundchange.go
```diff
@@ -85,6 +85,10 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
 		return errInvalidMessage
 	}
 
+	if !c.valSet.CheckInSubList(msg.Hash, rc.View, src.Address()) {
+		return errNotFromCommittee
+	}
```
laggu, line 90: I think a test for round change is required. 
It seems validators excluded in committee don't handle round change.

aidan-kwon: As we discussed offline, it will not exclude any committee member with this change. 
Anyway, we will test round change with "Klaytn-deploy".

consensus/istanbul/backend/backend.go
aidan-kwon: @laggu Good suggestion. I will consider it later.

consensus/istanbul/core/roundchange.go
andybclee: Because it is all about checking senders and only committee members can send a message, it looks okay to me.

PR #569, opened 2020-06-30
istanbul timeoutEvent embed a view of the next round

aidan-kwon:
## Proposed changes

**Problem Situation**
When the round change of istanbul occurred, some node changes its round successively in a very short time.
The phenomenon can happen when `timeoutEven` is posted handling (2f+1)th round change message. 
When (2f+1)th round change message is handled, Istanbul core resets round change timer triggering `timeoutEvent`.
However, there is a small change the timeout can be reached before resetting the timer.

**Solution**
The problem occurred since Istanbul core just increases the current round to 1 when `timeoutEvent` handled.
This PR makes `timeoutEvent` to embed a target round preventing confusing of the next round. 

- Introduce a test case triggering round change bug
- istanbul timeoutEvent embed a view of the next round
- Detail round change logs

You better see this PR commit by commit. 

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -353,6 +353,7 @@ func (c *core) stopTimer() {
 func (c *core) newRoundChangeTimer() {
 	c.stopTimer()
 
+	// TODO-Klaytn-Istanbul: Replace &istanbul.DefaultConfig.Timeout to c.config.Timeout
 	// set timeout based on the round number
 	timeout := time.Duration(atomic.LoadUint64(&istanbul.DefaultConfig.Timeout)) * time.Millisecond
 	round := c.current.Round().Uint64()
@@ -388,7 +389,10 @@ func (c *core) newRoundChangeTimer() {
 			}
 		}
 
-		c.sendEvent(timeoutEvent{})
+		c.sendEvent(timeoutEvent{&istanbul.View{
+			Sequence: c.current.sequence,
+			Round:    new(big.Int).Add(c.current.round, common.Big1),
+		}})
 	}))
 
 	logger.Debug("New RoundChangeTimer Set", "seq", c.current.Sequence(), "round", round, "timeout", timeout)
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -361,37 +361,42 @@ func (c *core) newRoundChangeTimer() {
 		timeout += time.Duration(math.Pow(2, float64(round))) * time.Second
 	}
 
-	c.roundChangeTimer.Store(time.AfterFunc(timeout, func() {
-		var loc, proposer string
+	current := c.current
+	proposer := c.valSet.GetProposer()
 
-		if c.current.round.Cmp(common.Big0) == 0 {
+	c.roundChangeTimer.Store(time.AfterFunc(timeout, func() {
+		var loc, proposerStr string
+
+		if round == 0 {
 			loc = "startNewRound"
 		} else {
 			loc = "catchUpRound"
 		}
-		if c.valSet.GetProposer() == nil {
-			proposer = ""
+		if proposer == nil {
+			proposerStr = ""
 		} else {
-			proposer = c.valSet.GetProposer().String()
+			proposerStr = proposer.String()
 		}
 
 		if c.backend.NodeType() == common.CONSENSUSNODE {
 			// Write log messages for validator activities analysis
-			preparesSize := c.current.Prepares.Size()
-			commitsSize := c.current.Commits.Size()
-			logger.Warn("[RC] timeoutEvent Sent!", "set by", loc, "sequence", c.current.sequence, "round", c.current.round, "proposer", proposer, "preprepare is nil?", c.current.Preprepare == nil, "len(prepares)", preparesSize, "len(commits)", commitsSize)
+			preparesSize := current.Prepares.Size()
+			commitsSize := current.Commits.Size()
+			logger.Warn("[RC] timeoutEvent Sent!", "set by", loc, "sequence",
+				current.sequence, "round", current.round, "proposer", proposerStr, "preprepare is nil?",
+				current.Preprepare == nil, "len(prepares)", preparesSize, "len(commits)", commitsSize)
 
 			if preparesSize > 0 {
-				logger.Warn("[RC] Prepares:", "messages", c.current.Prepares.GetMessages())
+				logger.Warn("[RC] Prepares:", "messages", current.Prepares.GetMessages())
 			}
 			if commitsSize > 0 {
-				logger.Warn("[RC] Commits:", "messages", c.current.Commits.GetMessages())
+				logger.Warn("[RC] Commits:", "messages", current.Commits.GetMessages())
 			}
 		}
 
 		c.sendEvent(timeoutEvent{&istanbul.View{
-			Sequence: c.current.sequence,
-			Round:    new(big.Int).Add(c.current.round, common.Big1),
+			Sequence: current.sequence,
+			Round:    new(big.Int).Add(current.round, common.Big1),
 		}})
 	}))
 
--- a/consensus/istanbul/core/events.go
+++ b/consensus/istanbul/core/events.go
@@ -22,6 +22,7 @@ package core
 
 import (
 	"github.com/klaytn/klaytn/common"
+	"github.com/klaytn/klaytn/consensus/istanbul"
 )
 
 type backlogEvent struct {
@@ -30,4 +31,6 @@ type backlogEvent struct {
 	Hash common.Hash
 }
 
-type timeoutEvent struct{}
+type timeoutEvent struct {
+	view *istanbul.View
+}
--- a/consensus/istanbul/core/events.go
+++ b/consensus/istanbul/core/events.go
@@ -32,5 +32,5 @@ type backlogEvent struct {
 }
 
 type timeoutEvent struct {
-	view *istanbul.View
+	nextView *istanbul.View
 }
--- a/consensus/istanbul/core/handler.go
+++ b/consensus/istanbul/core/handler.go
@@ -121,11 +121,17 @@ func (c *core) handleEvents() {
 					//c.backend.Gossip(c.valSet, p)
 				}
 			}
-		case _, ok := <-c.timeoutSub.Chan():
-			if !ok {
+		case ev, ok := <-c.timeoutSub.Chan():
+			if !ok || ev.Data == nil {
+				logger.Warn("Invalid message type from timeout channel")
 				return
 			}
-			c.handleTimeoutMsg()
+			data, ok := ev.Data.(timeoutEvent)
+			if !ok || data.view == nil {
+				logger.Warn("Invalid message from timeout channel", "msg", ev.Data)
+				return
+			}
+			c.handleTimeoutMsg(data.view)
 		case event, ok := <-c.finalCommittedSub.Chan():
 			if !ok {
 				return
@@ -193,7 +199,26 @@ func (c *core) handleCheckedMsg(msg *message, src istanbul.Validator) error {
 	return errInvalidMessage
 }
 
-func (c *core) handleTimeoutMsg() {
+func (c *core) handleTimeoutMsg(view *istanbul.View) {
+	// TODO-Klaytn-Istanbul: Check whether this statement is required. EN/PN should not handle consensus msgs.
+	if c.backend.NodeType() != common.CONSENSUSNODE {
+		logger.Warn("Received an istanbul timeout message from EN or PN",
+			"nodeType", c.backend.NodeType().String())
+		return
+	}
+
+	lastProposal, _ := c.backend.LastProposal()
+	if lastProposal == nil {
+		logger.Error("Cannot find the last proposal")
+		return
+	}
+
+	if lastProposal.Number().Cmp(view.Sequence) >= 0 {
+		logger.Debug("This timeoutMsg is outdated",
+			"blockNumber", lastProposal.Number().Uint64(), "msgView", view.String())
+		return
+	}
+
 	// If we're not waiting for round change yet, we can try to catch up
 	// the max round with F+1 round change message. We only need to catch up
 	// if the max round is larger than current round.
@@ -206,11 +231,10 @@ func (c *core) handleTimeoutMsg() {
 		}
 	}
 
-	lastProposal, _ := c.backend.LastProposal()
-	if lastProposal != nil && lastProposal.Number().Cmp(c.current.Sequence()) >= 0 {
+	if lastProposal.Number().Cmp(c.current.Sequence()) >= 0 {
 		c.logger.Trace("round change timeout, catch up latest sequence", "number", lastProposal.Number().Uint64())
 		c.startNewRound(common.Big0)
 	} else {
-		c.sendNextRoundChange("handleTimeoutMsg. lastProposal is nil or lastProposal's number is smaller than current sequence")
+		c.sendRoundChange(view.Round)
 	}
 }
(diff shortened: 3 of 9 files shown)


consensus/istanbul/core/handler_test.go
```diff
@@ -452,3 +453,129 @@ func TestCore_handlerMsg(t *testing.T) {
 		assert.Nil(t, err)
 	}
 }
+
+// TestCore_handleTimeoutMsg_race tests a race condition between round change triggers.
+// There should be no race condition when and round change message and timeout event are handled simultaneously.
```
jeongkyun-oh, line 458: `when and` -> `when`?

aidan-kwon: Thanks! I revised it.

consensus/istanbul/core/handler.go
```diff
@@ -193,7 +199,26 @@ func (c *core) handleCheckedMsg(msg *message, src istanbul.Validator) error {
 	return errInvalidMessage
 }
 
-func (c *core) handleTimeoutMsg() {
+func (c *core) handleTimeoutMsg(view *istanbul.View) {
+	// TODO-Klaytn-Istanbul: Check whether this statement is required. EN/PN should not handle consensus msgs.
+	if c.backend.NodeType() != common.CONSENSUSNODE {
+		logger.Warn("Received an istanbul timeout message from EN or PN",
```
winnie-byun, line 205: Isn't this should be like `This is not consensus node, but received an istanbul timeout message`? the log `from EN or PN` doesn't seem quite right.

consensus/istanbul/core/handler.go
```diff
@@ -121,11 +121,17 @@ func (c *core) handleEvents() {
...
-		case _, ok := <-c.timeoutSub.Chan():
-			if !ok {
+		case ev, ok := <-c.timeoutSub.Chan():
+			if !ok || ev.Data == nil {
+				logger.Warn("Invalid message type from timeout channel")
+				return
+			}
+			data, ok := ev.Data.(timeoutEvent)
+			if !ok || data.view == nil {
+				logger.Warn("Invalid message from timeout channel", "msg", ev.Data)
```
winnie-byun, line 131: Maybe the logs are written backward?
```suggestion
				logger.Warn("Invalid message from timeout channel")
				return
			}
			data, ok := ev.Data.(timeoutEvent)
			if !ok || data.view == nil {
				logger.Warn("Invalid message type from timeout channel", "msg", ev.Data)
```

consensus/istanbul/core/handler.go
aidan-kwon: You are right. Thanks!

consensus/istanbul/core/handler.go
```diff
@@ -206,11 +231,10 @@ func (c *core) handleTimeoutMsg() {
...
 	}
 
-	lastProposal, _ := c.backend.LastProposal()
-	if lastProposal != nil && lastProposal.Number().Cmp(c.current.Sequence()) >= 0 {
+	if lastProposal.Number().Cmp(c.current.Sequence()) >= 0 {
 		c.logger.Trace("round change timeout, catch up latest sequence", "number", lastProposal.Number().Uint64())
 		c.startNewRound(common.Big0)
 	} else {
-		c.sendNextRoundChange("handleTimeoutMsg. lastProposal is nil or lastProposal's number is smaller than current sequence")
+		c.sendRoundChange(view.Round)
```
winnie-byun, line 238: What do you think about replacing all `sendNextRoundChange` calls to `sendRoundChange`? `sendNextRoundChange` is still called in `core.commit()' and  `core.handlePreprepare()`. Using `sendNextRoundChange` may cause the same problem that is fixed here.

consensus/istanbul/core/roundchange.go
```diff
@@ -122,21 +123,24 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
...
-			logger.Warn("[RC] Send round change because we have F+1 roundchange messages")
+			logger.Warn("[RC] Send round change because we have F+1 round change messages",
+				"currentRound", cv.Round.Uint64(), "msgRound", roundView.Round.Uint64())
 			c.sendRoundChange(roundView.Round)
 		}
 		return nil
 	} else if cv.Round.Cmp(roundView.Round) < 0 {
 		// Only gossip the message with current round to other validators.
-		logger.Warn("[RC] Received round is bigger but not enough number of messages. Message ignored")
+		logger.Warn("[RC] Received round is bigger but not enough number of messages. Message ignored",
```
winnie-byun, line 142: The round change messages are stored and the log `Message ignored` seems confusing. Also, `Warn` looks like it is pretty high because a round change can happen frequently when in a network disorder.
```suggestion
		logger.Info("[RC] Received round is bigger but not enough number of messages. Nothing happens",
```

consensus/istanbul/core/handler.go
winnie-byun, line 205: This is the right log if I understand right?
```suggestion
		logger.Warn("This is not a consensus node, but received an istanbul timeout message",
```

consensus/istanbul/core/handler.go
aidan-kwon: The round change timer is the only routine related to `round` executing concurrently. After this change, other routines calling `sendNextRoundChange` will not have a concurrent issue since they doesn't access some values concurrently. 
Anyway, removing `sendNextRoundChange` sounds good for me. I will consider it later.

consensus/istanbul/core/roundchange.go
aidan-kwon: Thanks. I changed them

consensus/istanbul/core/handler.go
aidan-kwon: Thanks. I changed it

consensus/istanbul/core/handler.go
```diff
@@ -193,7 +199,26 @@ func (c *core) handleCheckedMsg(msg *message, src istanbul.Validator) error {
 	return errInvalidMessage
 }
 
-func (c *core) handleTimeoutMsg() {
+func (c *core) handleTimeoutMsg(view *istanbul.View) {
+	// TODO-Klaytn-Istanbul: Check whether this statement is required. EN/PN should not handle consensus msgs.
+	if c.backend.NodeType() != common.CONSENSUSNODE {
```
KimKyungup, line 204: @aidan-kwon 
Please check if this log is printed in EN/PN and remove this if it is not printed.

consensus/istanbul/core/roundchange.go
```diff
@@ -122,21 +123,24 @@ func (c *core) handleRoundChange(msg *message, src istanbul.Validator) error {
...
+			"currentRound", cv.Round.Uint64(), "newRound", roundView.Round.Uint64())
 		c.startNewRound(roundView.Round)
 		return nil
 	} else if c.waitingForRoundChange && num == numCatchUp {
 		// Once we received enough ROUND CHANGE messages, those messages form a weak certificate.
 		// If our round number is smaller than the certificate's round number, we would
 		// try to catch up the round number.
 		if cv.Round.Cmp(roundView.Round) < 0 {
-			logger.Warn("[RC] Send round change because we have F+1 roundchange messages")
+			logger.Info("[RC] Send round change because we have F+1 round change messages",
```
ehnuje, line 135:
```suggestion
			logger.Info("[RC] Send round change because we have f+1 round change messages",
```

consensus/istanbul/core/handler.go
```diff
@@ -193,7 +199,26 @@ func (c *core) handleCheckedMsg(msg *message, src istanbul.Validator) error {
...
+		logger.Warn("This is not consensus node, but received an istanbul timeout message",
+			"nodeType", c.backend.NodeType().String())
+		return
+	}
+
+	lastProposal, _ := c.backend.LastProposal()
+	if lastProposal == nil {
+		logger.Error("Cannot find the last proposal")
```
ehnuje, line 212: Don't we need to leave additional information regarding the received timeout message? (maybe something from `view` variable?)

ehnuje: Also "Cannot find the last proposal" seems a bit general. 
How about changing it to "Received timeout message but can't find the last proposal"?

consensus/istanbul/core/handler.go
```diff
@@ -121,11 +121,17 @@ func (c *core) handleEvents() {
 					//c.backend.Gossip(c.valSet, p)
 				}
 			}
-		case _, ok := <-c.timeoutSub.Chan():
-			if !ok {
+		case ev, ok := <-c.timeoutSub.Chan():
+			if !ok || ev.Data == nil {
+				logger.Warn("Invalid message from timeout channel")
```
ehnuje, line 126: Is receiving a `nil` message possible? If not how about leaving the log as `Error`?

consensus/istanbul/core/events.go
```diff
@@ -30,4 +31,6 @@ type backlogEvent struct {
 	Hash common.Hash
 }
 
-type timeoutEvent struct{}
+type timeoutEvent struct {
+	view *istanbul.View
```
KimKyungup, line 35: To make this clearer.
```suggestion
	nextView *istanbul.View
```

consensus/istanbul/core/handler.go
```diff
@@ -193,7 +199,26 @@ func (c *core) handleCheckedMsg(msg *message, src istanbul.Validator) error {
 	return errInvalidMessage
 }
 
-func (c *core) handleTimeoutMsg() {
+func (c *core) handleTimeoutMsg(view *istanbul.View) {
```
KimKyungup, line 202: `nextView` or `targetView`?
```suggestion
func (c *core) handleTimeoutMsg(nextView *istanbul.View) {
```

consensus/istanbul/core/handler.go
aidan-kwon: EN/PN can reach the log.

consensus/istanbul/core/roundchange.go
aidan-kwon: Applied, thanks.

consensus/istanbul/core/handler.go
aidan-kwon: Applied, thanks.

consensus/istanbul/core/events.go
aidan-kwon: Applied, thanks.

consensus/istanbul/core/handler.go
aidan-kwon: Applied, thanks.

consensus/istanbul/core/handler.go
aidan-kwon: Applied, thanks.

ehnuje (dismissed):

consensus/istanbul/core/core.go
```diff
@@ -353,42 +353,51 @@ func (c *core) stopTimer() {
 func (c *core) newRoundChangeTimer() {
 	c.stopTimer()
 
+	// TODO-Klaytn-Istanbul: Replace &istanbul.DefaultConfig.Timeout to c.config.Timeout
```
KimKyungup, line 356: For now, it is impossible to change this value by governacne? @aidan-kwon

consensus/istanbul/core/handler.go
```diff
@@ -121,11 +121,17 @@ func (c *core) handleEvents() {
...
-		case _, ok := <-c.timeoutSub.Chan():
-			if !ok {
+		case ev, ok := <-c.timeoutSub.Chan():
+			if !ok || ev.Data == nil {
+				logger.Error("Drop an empty message from timeout channel")
+				return
+			}
+			data, ok := ev.Data.(timeoutEvent)
+			if !ok || data.nextView == nil {
+				logger.Error("Invalid message from timeout channel", "msg", ev.Data)
```
KimKyungup, line 131: Does printing `ev.Data` works well? It prints bytes?
```suggestion
				logger.Error("Invalid message from timeout channel", "msg", ev.Data)
```

consensus/istanbul/core/handler.go
```diff
@@ -193,7 +199,26 @@ func (c *core) handleCheckedMsg(msg *message, src istanbul.Validator) error {
 	return errInvalidMessage
 }
 
-func (c *core) handleTimeoutMsg() {
+func (c *core) handleTimeoutMsg(nextView *istanbul.View) {
+	// TODO-Klaytn-Istanbul: EN/PN should not handle consensus msgs.
```
KimKyungup, line 203: Is this TODO already done by below condition ?

consensus/istanbul/core/core.go
aidan-kwon: A governance item for Istanbul timeout is prepared now, but I think it is not working now. 
We should finish the implementation later.

consensus/istanbul/core/handler.go
aidan-kwon: The following logic prevents EN/PN to handle `timeout` message.
The TODO comment means that EN/PN should stop hadlers for all consensus messages including `prepare`, `commit`.

consensus/istanbul/core/handler.go
aidan-kwon: @KimKyungup If `ev.Data` is string, data is printed as a string. If `ev.Data` is bytes, data is printed as bytes.
Actually, this code should not be reached by any context since `c.timeoutSub.Chan` only subscribes `timeoutEvent` and `ev.Data` would be `nil` if `data.nextView` is nil.

jeongkyun-oh (dismissed):

KimKyungup (approved):

jeongkyun-oh (dismissed):

KimKyungup (approved):

jeongkyun-oh (approved):

consensus/istanbul/core/core.go
```diff
@@ -353,42 +353,51 @@ func (c *core) stopTimer() {
...
 	// set timeout based on the round number
 	timeout := time.Duration(atomic.LoadUint64(&istanbul.DefaultConfig.Timeout)) * time.Millisecond
 	round := c.current.Round().Uint64()
 	if round > 0 {
 		timeout += time.Duration(math.Pow(2, float64(round))) * time.Second
 	}
 
+	current := c.current
+	proposer := c.valSet.GetProposer()
```
yoomee1313, line 365: In previous code, the value is fetched when `time.AfterFunc` is executed. Can I ask why you change like this?

yoomee1313 (approved):

consensus/istanbul/core/core.go
aidan-kwon: @yoomee1313 `c.current` and `c.valSet.GetProposer()` is updatable in the consensus algorithm. It probably not be happened during the timer is running in the current code, but I want to make sure that timeout-related logs use the original values when it declared.

consensus/istanbul/core/handler.go
```diff
@@ -121,11 +121,17 @@ func (c *core) handleEvents() {
...
+		case ev, ok := <-c.timeoutSub.Chan():
+			if !ok || ev.Data == nil {
+				logger.Warn("Invalid message type from timeout channel")
+				return
+			}
+			data, ok := ev.Data.(timeoutEvent)
+			if !ok || data.view == nil {
+				logger.Warn("Invalid message from timeout channel", "msg", ev.Data)
 				return
 			}
```
winnie-byun: Due to this code, we should upgrade all CN at the same time. Could we just accept `type timeoutEvent struct {
	nextView *istanbul.View
}` and `type timeoutEvent struct {}` at the same time for now? After all CN has upgraded and started sending the new struct of `timeoutEvnet`, we can change this logic to handle only the new struct

aidan-kwon: @winnie-byun TimeoutEvent is not propagated to other nodes. It is a local event and is sent to the node itself. Therefore, we don't need to consider version matching among CNs.

ehnuje (approved):

winnie-byun (approved):

PR #580, opened 2020-07-30
Add fastCallTracer for trace API

jeongkyun-oh:
## Proposed changes

- This PR substitute JS call tracer with go version for traceAPI.
- The result may not be the same as JS call tracer which is different from the empty values.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/node/cn/api_tracer.go
+++ b/node/cn/api_tracer.go
@@ -29,6 +29,7 @@ import (
 	"github.com/klaytn/klaytn/log"
 	"io/ioutil"
 	"os"
+	"reflect"
 	"runtime"
 	"sync"
 	"time"
@@ -55,6 +56,8 @@ const (
 	// and reexecute to produce missing historical state necessary to run a specific
 	// trace.
 	defaultTraceReexec = uint64(128)
+
+	fastCallTracer = "fastCallTracer"
 )
 
 // TraceConfig holds extra parameters to trace functions.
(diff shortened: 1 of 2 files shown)


node/cn/api_tracer.go
```diff
@@ -755,15 +758,27 @@ func (api *PrivateDebugAPI) traceTx(ctx context.Context, message blockchain.Mess
...
+			if tracer, err = tracers.New(*config.Tracer); err != nil {
+				return nil, err
+			}
 		}
 		// Handle timeouts and RPC cancellations
 		deadlineCtx, cancel := context.WithTimeout(ctx, timeout)
 		go func() {
 			<-deadlineCtx.Done()
-			tracer.(*tracers.Tracer).Stop(errors.New("execution timeout"))
+			switch t := tracer.(type) {
```
KimKyungup, line 774:
```suggestion
			tracer.Stop(errors.New("execution timeout"))
```
Does all tracers support `Stop()`?

node/cn/api_tracer.go
```diff
@@ -55,6 +56,8 @@ const (
 	// and reexecute to produce missing historical state necessary to run a specific
 	// trace.
 	defaultTraceReexec = uint64(128)
+
+	fastCallTracer = "fastCallTracer"
```
KimKyungup, line 60: Please add some comment about this fast Call Tracer. :)

node/cn/api_tracer.go
jeongkyun-oh: Unfortunately, no. Tracer interface doesn't have `Stop()` method.

node/cn/api_tracer.go
jeongkyun-oh: Applied. 64c4e21

node/cn/api_tracer.go
ehnuje: How about adding Stop to the Tracer interface?

jeongkyun-oh: @ehnuje As you know, `Tracer` interface has only Capture series. For consistency, adding `Stop` method looks not good. Also, there is another tracer which does not have `Stop` although it can be newly added.

jeongkyun-oh: How about converting it to `Stoppable` which has `Stop` method?

ehnuje: If the function is not added to the `Tracer` interface, we need to cast the `Tracer` interface to the `Stoppable` interface, which is being done in this code. Let's think about this later if we have additional tracers in the future.

PR #589, opened 2020-08-05
Introduce KASAnchor package

KimKyungup:
## Proposed changes

This PR introduce KAS package with `sendRequest`.
- `sendRequest` can anchor the block to KAS Server.
-  including test code for `sendRequest`
- add mocks for `Anchor`

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


kas/anchor.go
```diff
@@ -0,0 +1,126 @@
...
+	"bytes"
+	"encoding/json"
+	"errors"
+	"github.com/klaytn/klaytn/blockchain/types"
+	"github.com/klaytn/klaytn/common"
+	"net/http"
+)
+
+var (
+	codeOK                = 0
```
jeongkyun-oh, line 29: How about making this as `const`?

kas/anchor_test.go
```diff
@@ -0,0 +1,86 @@
...
+	// OK case
+	{
+		expectedRes := http.Response{Status: strconv.Itoa(http.StatusOK)}
+		expectedRespBody := respBody{
+			Code: 0,
+		}
+		bodyBytes, _ := json.Marshal(expectedRespBody)
+		expectedRes.Body = ioutil.NopCloser(bytes.NewReader(bodyBytes))
+
+		m.EXPECT().Do(gomock.Any()).Times(1).Return(&expectedRes, nil)
```
jeongkyun-oh, line 71: How about using `gomock.Eq()` instead of `gomock.Any()` to check actual param?

kas/anchor.go
```diff
@@ -0,0 +1,126 @@
...
+	req.SetBasicAuth(anchor.kasConfig.User, anchor.kasConfig.Pwd)
+	for k, v := range header {
+		req.Header.Set(k, v)
+	}
+
+	resp, err := anchor.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
```
jeongkyun-oh, line 120: no need to check `resp.Body != nil`?

KimKyungup: It is not necessary.
See here.

https://golang.org/src/net/http/response.go
```
	// The http Client and Transport guarantee that Body is always
	// non-nil, even on responses without a body or responses with
	// a zero-length body. It is the caller's responsibility to
	// close Body. The default HTTP client's Transport may not
	// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
	// not read to completion and closed.
```

kas/anchor_test.go
KimKyungup: I think it is difficult to make `*Request`.

kas/anchor.go
```diff
@@ -0,0 +1,129 @@
...
+//
+// The klaytn library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+package kas
```
ehnuje, line 17: How about using `anchor` or `anchoring` instead of `kas`? As this feature will be expanded to support various types of anchoring, not KAS.

KimKyungup: As we talked, let's think the better name and path. :)

PR #591, opened 2020-08-06
Implementation of anchoring via KAS API

KimKyungup:
## Proposed changes

This PR complete `kas` package.
- `AnchorPeriodicBlock` will be called by all chain event with a block to anchor periocially.
- `AnchorByBlockNum` will be called by API to anchor specific block.
- Add test codes to convert a block to a payload of KAS anchor API.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/kas/anchor.go
+++ b/kas/anchor.go
@@ -84,7 +84,9 @@ func (anchor *Anchor) AnchorPeriodicBlock(block *types.Block) {
 		return
 	}
 
-	anchor.AnchorBlock(block)
+	if err := anchor.AnchorBlock(block); err != nil {
+		logger.Warn("Failed to anchor a block via KAS", "blkNum", block.NumberU64())
+	}
 }
 
 // blockToAnchoringDataInternalType0 makes AnchoringDataInternalType0 from the given block.
@@ -130,7 +132,7 @@ func (anchor *Anchor) AnchorBlock(block *types.Block) error {
 
 	if res.Code != codeOK {
 		result, _ := json.Marshal(res)
-		logger.Error("Failed to anchor a block via KAS", "blkNum", block.NumberU64(), "result", string(result))
+		logger.Debug("Failed to anchor a block via KAS", "blkNum", block.NumberU64(), "result", string(result))
 		return fmt.Errorf("error code %v", res.Code)
 	}
 
--- a/kas/anchor.go
+++ b/kas/anchor.go
@@ -76,7 +76,7 @@ func (anchor *Anchor) AnchorPeriodicBlock(block *types.Block) {
 	}
 
 	if block == nil {
-		logger.Error("KAS Anchor : failed to get a block form the event")
+		logger.Error("KAS Anchor : can not anchor nil block")
 		return
 	}
 
@@ -84,21 +84,11 @@ func (anchor *Anchor) AnchorPeriodicBlock(block *types.Block) {
 		return
 	}
 
-	anchor.anchorBlock(block)
-}
-
-// AnchorByBlockNum anchors the block of given block number to KAS by API.
-func (anchor *Anchor) AnchorByBlockNum(blkNum uint64) error {
-	block := anchor.bc.GetBlockByNumber(blkNum)
-	if block == nil {
-		return errNotFoundBlock
-	}
-
-	return anchor.anchorBlock(block)
+	anchor.AnchorBlock(block)
 }
 
 // blockToAnchoringDataInternalType0 makes AnchoringDataInternalType0 from the given block.
-// TxCount is the total number of transactions during N-P+1 ~ N blocks. (N: given block, P: anchoring period)
+// TxCount is the number of transactions of the last N blocks. (N is a anchor period.)
 func (anchor *Anchor) blockToAnchoringDataInternalType0(block *types.Block) *types.AnchoringDataInternalType0 {
 	start := uint64(0)
 	if block.NumberU64() >= anchor.kasConfig.AnchorPeriod {
@@ -127,8 +117,8 @@ func (anchor *Anchor) blockToAnchoringDataInternalType0(block *types.Block) *typ
 	}
 }
 
-// anchorBlock converts given block to payload and anchor the payload via KAS anchor API.
-func (anchor *Anchor) anchorBlock(block *types.Block) error {
+// AnchorBlock converts given block to payload and anchor the payload via KAS anchor API.
+func (anchor *Anchor) AnchorBlock(block *types.Block) error {
 	anchorData := anchor.blockToAnchoringDataInternalType0(block)
 
 	payload := dataToPayload(anchorData)
@@ -140,7 +130,7 @@ func (anchor *Anchor) anchorBlock(block *types.Block) error {
 
 	if res.Code != codeOK {
 		result, _ := json.Marshal(res)
-		logger.Debug("Failed to anchor a block via KAS", "blkNum", block.NumberU64(), "result", string(result))
+		logger.Error("Failed to anchor a block via KAS", "blkNum", block.NumberU64(), "result", string(result))
 		return fmt.Errorf("error code %v", res.Code)
 	}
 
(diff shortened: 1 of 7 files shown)


kas/anchor.go
```diff
@@ -66,6 +68,86 @@ func NewKASAnchor(kasConfig *KASConfig, db AnchorDB, bc BlockChain) *Anchor {
...
+	if block == nil {
+		logger.Error("KAS Anchor : failed to get a block form the event")
+		return
+	}
+
+	if block.NumberU64()%anchor.kasConfig.AnchorPeriod != 0 {
+		return
+	}
+
+	anchor.anchorBlock(block)
```
jeongkyun-oh, line 87: Doesn't it need to handle error?

kas/anchor.go
```diff
@@ -66,6 +68,86 @@ func NewKASAnchor(kasConfig *KASConfig, db AnchorDB, bc BlockChain) *Anchor {
...
+}
+
+// AnchorByBlockNum anchors the block of given block number to KAS by API.
+func (anchor *Anchor) AnchorByBlockNum(blkNum uint64) error {
+	block := anchor.bc.GetBlockByNumber(blkNum)
```
aidan-kwon, line 92: As far as I know, most of Klaytn APIs get `blockNumber` in methods of API packages considering other keywords like "latest". 
I think `anchor.anchorBlock` is enough for this package.

kas/anchor.go
KimKyungup: Yes, it is. If there is an error, `anchorBlock` will log.
In next stage, I will add more feature to retry anchoring in `anchorBlcok`.

kas/anchor.go
KimKyungup: It makes sense. I will remove this function and use `anchorBlock` directly in API.

kas/anchor.go
```diff
@@ -66,6 +68,86 @@ func NewKASAnchor(kasConfig *KASConfig, db AnchorDB, bc BlockChain) *Anchor {
...
+	if block == nil {
+		return errNotFoundBlock
+	}
+
+	return anchor.anchorBlock(block)
+}
+
+// blockToAnchoringDataInternalType0 makes AnchoringDataInternalType0 from the given block.
+// TxCount is the total number of transactions during N-P+1 ~ N blocks. (N: given block, P: anchoring period)
+func (anchor *Anchor) blockToAnchoringDataInternalType0(block *types.Block) *types.AnchoringDataInternalType0 {
```
aidan-kwon, line 102: Is this function called only when `block.NumberU64() % anchor.kasConfig.AnchorPeriod == 0`?

KimKyungup: `chainEvent` calls this func with the right periodic block number.
But API can calls this func with any block number.

aidan-kwon: As we discussed off-line, consider changing the semantic of this function to returning last N block number.

kas/anchor.go
```diff
@@ -66,6 +68,86 @@ func NewKASAnchor(kasConfig *KASConfig, db AnchorDB, bc BlockChain) *Anchor {
...
+	payload := dataToPayload(anchorData)
+
+	res, err := anchor.sendRequest(payload)
+	if err != nil {
+		return err
+	}
+
+	if res.Code != codeOK {
+		result, _ := json.Marshal(res)
+		logger.Debug("Failed to anchor a block via KAS", "blkNum", block.NumberU64(), "result", string(result))
```
ehnuje, line 143:
```suggestion
		logger.Error("Failed to anchor a block via KAS", "blkNum", block.NumberU64(), "result", string(result))
```

kas/anchor.go
```diff
@@ -66,6 +68,86 @@ func NewKASAnchor(kasConfig *KASConfig, db AnchorDB, bc BlockChain) *Anchor {
...
 
+// AnchorPeriodicBlock periodically anchor blocks to KAS.
+// if given block is invalid, it does nothing.
+func (anchor *Anchor) AnchorPeriodicBlock(block *types.Block) {
+	if !anchor.kasConfig.Anchor {
+		return
+	}
+
+	if block == nil {
+		logger.Error("KAS Anchor : failed to get a block form the event")
```
ehnuje, line 79:
```suggestion
		logger.Error("KAS Anchor : failed to get a block from the event")
```

ehnuje: Log message is a bit weird since there is no event in the current function call.

kas/anchor.go
ehnuje: @KimKyungup Sorry for reopening the conversation, however, in that case I think `anchorBlock` should not return an error.

kas/anchor.go
```diff
@@ -66,6 +68,86 @@ func NewKASAnchor(kasConfig *KASConfig, db AnchorDB, bc BlockChain) *Anchor {
...
+}
+
+// AnchorByBlockNum anchors the block of given block number to KAS by API.
+func (anchor *Anchor) AnchorByBlockNum(blkNum uint64) error {
+	block := anchor.bc.GetBlockByNumber(blkNum)
+	if block == nil {
+		return errNotFoundBlock
+	}
+
+	return anchor.anchorBlock(block)
```
ehnuje, line 97: Some errors from `anchorBlock` are handled and some are not.

kas/anchor.go
KimKyungup: reflected

kas/anchor.go
KimKyungup: reflected

kas/anchor.go
KimKyungup: reflected

kas/anchor.go
KimKyungup: reflected

kas/anchor.go
KimKyungup: But for API, the function should return an error.
API will be added in next PR.

kas/anchor.go
```diff
@@ -66,6 +68,76 @@ func NewKASAnchor(kasConfig *KASConfig, db AnchorDB, bc BlockChain) *Anchor {
...
+	if block == nil {
+		logger.Error("KAS Anchor : can not anchor nil block")
+		return
+	}
+
+	if block.NumberU64()%anchor.kasConfig.AnchorPeriod != 0 {
+		return
+	}
+
+	anchor.AnchorBlock(block)
```
ehnuje, line 87: If the returned error won't be handled, `AnchorBlock` should not return an error.

KimKyungup: This case doesn't not need to handle. But anchor API will handle the error.

ehnuje: If the error from `AnchorPeriodicBlock` does not need to be handled, it is not natural that an error is logged inside `AnchorBlock`. How about moving out log outside `AnchorBlock` and leave the log `KASAnchor` in the API?

KimKyungup: I think they are different between not need to be handled and need to be logged.
Both `AnchorPeriodicBlock` and `API` need to log the error.
But for now `AnchorPeriodicBlock` doesn't need to be handled.

KimKyungup:
> Please, consider re-locate `kas` package.
> This package is only for the service chain.

@aidan-kwon Thanks. After all implementation PR, I will move this package into the `service chain` package.

kas/anchor.go
ehnuje: The point I want to emphasize is that if you don't need to handle the error, that is not an error. 
As you don't handle the error from `AnchorPeriodicBlock`, for me, it should be a warning, not an error.

KimKyungup: OK. I will change the caller chose log or not.

KimKyungup: @ehnuje PTAL. I reflected. :)

KimKyungup: 17be7acc984ed74970560ce1895972f97c3c0da7

KimKyungup:
@jeongkyun-oh @aidan-kwon PTAL  :)

PR #660, opened 2020-09-07
Introduce default flag group for help which is used by various binaries 

aidan-kwon:
## Proposed changes

Even though node binaries have the same flag group, each binary defines its own flag group used for help. 
Because of that, it is easy to have a mistake when a new flag is added. 
(Currenlty, some flags don't have help descriptions even though they are used)

- Move `ConfigFileFlag` declare position from `nodeflag` pakcage to `utils` package.
- Introduce default flaggroup which is used by various binaries
- Categorize help flag groups at one point

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist


cmd/kcn/main.go
```diff
@@ -241,7 +80,7 @@ func init() {
 	app.Flags = append(app.Flags, nodecmd.DBMigrationFlags...)
 
 	cli.AppHelpTemplate = utils.GlobalAppHelpTemplate
-	cli.HelpPrinter = utils.NewHelpPrinter(cnHelpFlagGroups)
+	cli.HelpPrinter = utils.NewHelpPrinter(utils.CategorizeFlags(app.Flags))
```
winnie-byun, line 83: This is both suggestion and question. I think it is a good idea to apply this to kbn, too. kbn needs help print for flags and it would be great if kbn works like other binaries. However, even though `cli.HelpPrinter` is not defined in kbn, kbn has its own help print out. Do you know how the print works in kbn?

winnie-byun (changes requested):

cmd/utils/flaggroup.go
```diff
@@ -0,0 +1,319 @@
...
+			AutoRestartFlag,
+			RestartTimeOutFlag,
+			DaemonPathFlag,
+		},
+	},
+}
+
+// CategorizeFlags classified each flag into pre-defined flagGroups.
+func CategorizeFlags(flags []cli.Flag) []FlagGroup {
+	flagGroupsMap := make(map[string][]cli.Flag)
```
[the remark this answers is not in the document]
aidan-kwon: I could do that, but I afraid repeating too many `for` loops. If `flagGroups` is directly used, a flag should iterate `flagGroups` slice to find its group the flag is classified as. Since `CategorizeFlags` already has a loop repeating `for` 3 times, I introduced `flagGroupsMap` converting it to `flagGroups` outside of the repeated `for` loop.

cmd/kcn/main.go
aidan-kwon: As you said, we have other binaries such `kbn`, `homi` and `kgen`. They also have a help description for flags but don't have flag groups because they only have few flags.

cmd/utils/flaggroup.go
```diff
@@ -0,0 +1,319 @@
+// Copyright 2020 The klaytn Authors
```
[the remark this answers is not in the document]
aidan-kwon: I added. Thnks!

cmd/utils/flaggroup_test.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+func TestFlagGroups_Duplication(t *testing.T) {
+	testGroups := FlagGroups
+	exist := make(map[string]bool)
+
+	for _, group := range testGroups {
+		for _, flag := range group.Flags {
+			if exist[flag.GetName()] == true {
+				t.Error("a flag belong to more than 2 groups", "flag", flag.GetName())
```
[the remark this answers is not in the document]
aidan-kwon: I changed it. Thanks!

cmd/kcn/main.go
aidan-kwon: Oh, I understand now why you specified `kbn`. I found out that `kbn` has more flags than I expected. I think some flags are not necessary for `kbn`. Let's handle this later.

cmd/utils/doc.go
```diff
@@ -31,6 +31,7 @@ Each file contains the following contents
  - customflags.go	: Provides `DirectoryString`, `DirectoryFlags` and marshaling functions to support custom flags
  - files.go		: Provides `WriteFile` function to store contents in a given file
  - flags.go		: Defines various flags which can be used in running a node
+ - flaggroup.go : Categorizes flags into groups to print structured help descriptions.
```
jeongkyun-oh, line 34: It would be great if the indentation is aligned with other comments.

cmd/utils/flaggroup.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+// CategorizeFlags classified each flag into pre-defined flagGroups.
+func CategorizeFlags(flags []cli.Flag) []FlagGroup {
+	flagGroupsMap := make(map[string][]cli.Flag)
+	isFlagAdded := make(map[string]bool) // Check duplicated flags
+
+	// Find its group for each flag
+	for _, flag := range flags {
+		if isFlagAdded[flag.GetName()] {
```
ehnuje, line 265: Isn't it an error case if `isFlagAdded` returns true?

cmd/utils/flaggroup.go
```diff
@@ -0,0 +1,319 @@
...
+	"sort"
+
+	"github.com/klaytn/klaytn/api/debug"
+	"gopkg.in/urfave/cli.v1"
+)
+
+const uncategorized = "MISC" // Uncategorized flags will belong to this group
+
+// FlagGroups categorizes flags into groups to print structured help.
+var FlagGroups = []FlagGroup{
```
ehnuje, line 29: How about changing this to a map with key as flag name and value as flag group name?

aidan-kwon: Good idea. I added a `TODO` comment about it.

cmd/utils/flaggroup.go
aidan-kwon: I added a debug log. I think it is not an error since it doesn't affect any semantics of Klaytn node and a minor issue.

cmd/utils/flaggroup.go
```diff
@@ -0,0 +1,319 @@
...
+			GenKeyFlag,
+			WriteAddressFlag,
+			AutoRestartFlag,
+			RestartTimeOutFlag,
+			DaemonPathFlag,
+		},
+	},
+}
+
+// CategorizeFlags classified each flag into pre-defined flagGroups.
```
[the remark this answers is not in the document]
aidan-kwon: changed. Thanks.

cmd/utils/doc.go
aidan-kwon: Changed. Thanks

jeongkyun-oh (approved):

winnie-byun (approved):

cmd/utils/flaggroup.go
ehnuje: I mean that's something should be checked during the test so that we can avoid that situation, though it does not make any functional or performance issue.

aidan-kwon: I agreed with you. So, I made `TestFlagGroups_Duplication` to avoid the case a flag belongs to more than one group. 

Maybe you also consider the case one flag is added twice with the same values. However, I am considering the case is a normal. Because the input flag slice is a combination of other flag groups having different classification standards(`CommonNodeFlags`, `CommonRPCFlags` and others defined in `cmd/utils/nodecmd/nodeflags.go`. In those standards, a flag can belong to multiple groups.

ehnuje: Thanks for the explanation. I understand your point. 😄

ehnuje (approved):

KimKyungup (approved):

PR #673, opened 2020-09-15
Initial commit for kafka client

jeongkyun-oh:
## Proposed changes

- Added timeout for CI build test
- `KafkaBroker` is implemented for kafka client.
- The implementation is derived from `wallet-api` and modified.
- The unit-tests can be tested with local kafka servers which is not yet ready.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+	r.producer.Input() <- item
+
+	return nil
+}
+
+func (r *KafkaBroker) Subscribe(topic string, handler interface{}) error {
+	r.CreateTopic(topic)
+	h, ok := handler.(func(*sarama.ConsumerMessage) error)
+	if !ok {
+		return errors.New("unsupported type")
```
ehnuje, line 91: If it fails, how about adding information of the type of given handler by `reflect.TypeOf`?

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"context"
+
+	"github.com/Shopify/sarama"
+)
+
+type Consumer struct {
+	cancel   chan bool
+	handler  map[string]func(*sarama.ConsumerMessage) error
+	consumer sarama.ConsumerGroup
```
ehnuje, line 28: `consumer` in `Consumer` looks confusing.

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+package kafka
+
+import (
+	"context"
+
+	"github.com/Shopify/sarama"
+)
+
+type Consumer struct {
+	cancel   chan bool
```
ehnuje, line 26:
```suggestion
	cancelCh   chan bool
```

ehnuje: Why do we need boolean value for this channel?

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (r *Consumer) Setup(sess sarama.ConsumerGroupSession) error {
+	logger.Info("consumer was initialized", "id", sess.MemberID())
+	r.isActive = true
+	return nil
+}
+
+func (r *Consumer) Cleanup(sess sarama.ConsumerGroupSession) error {
+	logger.Info("consumer was cleaned up", "id", sess.MemberID())
+	r.isActive = false
```
ehnuje, line 88: Is there any possibility of concurrent access to `isActive`?

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}()
+
+	return nil
+}
+
+func (r *Consumer) Setup(sess sarama.ConsumerGroupSession) error {
+	logger.Info("consumer was initialized", "id", sess.MemberID())
+	r.isActive = true
+	return nil
```
ehnuje, line 83: If it always returns nil, why do we need return value?

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+			case <-r.ctx.Done():
+				return
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (r *Consumer) Setup(sess sarama.ConsumerGroupSession) error {
```
ehnuje, line 80: When do `Setup` and `Cleanup` are called?

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+	if r.isActive {
+		r.cancel <- true
+	}
+	go func() {
+		defer r.consumer.Close()
+		var topics []string
+		for topic := range r.handler {
+			topics = append(topics, topic)
+		}
+		h := func(err chan<- error) {
```
ehnuje, line 58: Why do you declare `h` as a variable here? (Since it is only used in the line 63)

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+	return &Consumer{
+		cancel:   make(chan bool),
+		handler:  map[string]func(*sarama.ConsumerMessage) error{},
+		ctx:      ctx,
+		consumer: consumer,
+		isActive: false,
+	}
+}
+
+func (r *Consumer) Subscribe(topic string, handler func(*sarama.ConsumerMessage) error) error {
```
ehnuje, line 43: `r` for `Consumer` type looks confusing.

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+		kb.newProducer()
+
+		// TODO-ChainDataFetcher context has to be passed by outside.
+		kb.consumer = NewConsumer(context.Background(), kb.newConsumer(groupID))
+	})
+
+	return kb
+}
+
+func (r *KafkaBroker) Publish(topic string, msg interface{}) error {
```
ehnuje, line 67: `r` for `KafkaBroker` looks confusing.

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+	config.Version = sarama.MaxVersion
+	config.Consumer.Group.Session.Timeout = 6 * time.Second
+	config.Consumer.Group.Heartbeat.Interval = 2 * time.Second
+
+	id, _ := uuid.GenerateUUID()
+	config.ClientID = fmt.Sprintf("%s-%s", groupID, id)
+
+	consumer, err := sarama.NewConsumerGroup(r.brokers, groupID, config)
+	if err != nil {
+		logger.Crit("NewConsumerGroup is failed", "err", err)
```
ehnuje, line 149: How about leaving configurations in the log?

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+	return consumer
+}
+
+func (r *KafkaBroker) newClusterAdmin() {
+	config := sarama.NewConfig()
+	config.Version = sarama.MaxVersion
+
+	admin, err := sarama.NewClusterAdmin(r.brokers, config)
+	if err != nil {
+		logger.Crit("NewClusterAdmin is failed", "err", err)
```
ehnuje, line 161: How about leaving configurations in the log?

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (r *KafkaBroker) newProducer() {
+	config := sarama.NewConfig()
+	config.Producer.RequiredAcks = sarama.WaitForLocal
+	config.Producer.Compression = sarama.CompressionSnappy
+	config.Producer.Flush.Frequency = 500 * time.Millisecond
+
+	producer, err := sarama.NewAsyncProducer(r.brokers, config)
+	if err != nil {
+		logger.Crit("Failed to start Sarama producer", "err", err)
```
ehnuje, line 132: How about leaving configurations in the log?

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+			err <- r.consumer.Consume(r.ctx, topics, r)
+		}
+		res := make(chan error, 1)
+		for {
+			go h(res)
+			select {
+			case err := <-res:
+				if err != nil {
+					logger.Error("consumed messages return an error", "topic", topic, "err", err)
+				}
```
ehnuje, line 68: If `err == nil`, what will happen?

ehnuje (commented):
Please leave some comments for the newly added types and functions.

datasync/chaindatafetcher/kafka/kafka_test.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+type kafkaData struct {
+	Name   string `json:"name"`
+	Data   string `json:"data"`
+	Number int    `json:"number"`
+	IsGood bool   `json:"isGood"`
+}
+
+type KafkaSuite struct {
```
winnie-byun, line 40: What is the difference between `KafkaSuite` and `Kafka`? They seem to have common fields.

datasync/chaindatafetcher/kafka/kafka_test.go
```diff
@@ -0,0 +1,156 @@
...
+	suite.Suite
```
winnie-byun, line 41: Does Suite makes it enable to run test codes of singleton pattern?
If so, we should apply them to other place, too. :)

datasync/chaindatafetcher/kafka/kafka_test.go
```diff
@@ -0,0 +1,156 @@
...
+	kfk             types.EventBroker
+	topic           string
+	data            string
+	consumerGroupId string
+	brokers         []string
+	replica         int16
+	partitions      int32
```
winnie-byun, line 48: Why is this defined again, here?
`KafakaBroker` has the duplicated settings.

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/klaytn/klaytn/log"
+)
+
+var (
+	logger = log.NewModuleLogger(log.ChainDataFetcher)
+	kb     *KafkaBroker
+	once   sync.Once
+)
+
+type KafkaBroker struct {
```
winnie-byun, line 39:
```suggestion
type Kafka struct {
```
I think just `Kafka` looks better for abstraction. It is up to you actually. :)

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/klaytn/klaytn/datasync/chaindatafetcher/types"
+	"github.com/klaytn/klaytn/log"
+)
+
+var (
+	logger = log.NewModuleLogger(log.ChainDataFetcher)
+	kb     *KafkaBroker
+	once   sync.Once
+)
+
```
winnie-byun, line 38:
```suggestion

// KafkaBroker connects to kafka brokers in an existing kafka cluster.
```

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+type KafkaBroker struct {
+	producer   sarama.AsyncProducer
+	admin      sarama.ClusterAdmin
```
winnie-byun, line 41: I have a question. Is admin a distinct server, or one of kafka brokers? Or is it like an authority anyone can aquire?
This seems weird, because it looks like anyone who can connect to kafka cluster, has admin access.

winnie-byun (commented):
Could you add a `doc.go` file and some comments for each struct?
It would look much better with the descriptions :)

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (r *Consumer) Subscribe(topic string, handler func(*sarama.ConsumerMessage) error) error {
+	if r.handler[topic] != nil {
+		return nil
+	}
+	r.handler[topic] = handler
+
+	if r.isActive {
+		r.cancel <- true
```
aidan-kwon, line 50: What does this line for? Doesn't it always close `for` loop in the below if there was no other `for` loops?

datasync/chaindatafetcher/kafka/kafka_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	t := time.NewTicker(10 * time.Second)
+	rand.Seed(time.Now().UnixNano())
+	for i := 0; i < numTests; i++ {
+		// generate random kafkadata
+		expected := &kafkaData{
+			Name:   s.topic,
+			Data:   s.data + strconv.Itoa(rand.Int()),
+			Number: rand.Int(),
+			IsGood: rand.Int()%2 == 0,
```
yoomee1313, line 114: Can you explain what "isGood" means and why even is true?

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Added on 1f989f2dcfbb1c9eeed58c88629d79559f246f26.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Renamed it to `members` instead of `consumer`. 1f989f2dcfbb1c9eeed58c88629d79559f246f26.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Not really. Modified `bool` to `struct{}`. 1f989f2dcfbb1c9eeed58c88629d79559f246f26

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: This is a method implementing an interface.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: So, firstly `Setup` is called, `ConsumeClaim` and then `Cleanup` in `Consume` method in the order.  These methods implement `ConsumerGroupHandler` interface.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Renamed the receiver to `c`. 1f989f2dcfbb1c9eeed58c88629d79559f246f26

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Renamed the receiver to `k`. 1f989f2dcfbb1c9eeed58c88629d79559f246f26

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Added on 1f989f2dcfbb1c9eeed58c88629d79559f246f26.

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Added on 1f989f2dcfbb1c9eeed58c88629d79559f246f26.

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Added on 5201a3d0ddf4c006885cf22ce5a838ad2482d49f.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Nothing. Another consume routine will be triggered.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: I made it into the `for` loop. bce4ddda62ea8744a41b83ef6862599afb92ad7e

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Currently, I don't think so assuming that the `Subscribe` method is called only once.

datasync/chaindatafetcher/kafka/kafka_test.go
jeongkyun-oh: Hmm.. `SetupSuite` is run just before each test function, so I created a new kafka struct every test.

datasync/chaindatafetcher/kafka/kafka_test.go
jeongkyun-oh: Right. I used literal values for settings. 114ec1d79d766d757c889feb6740a9745f1969a7

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Agreed! 55a975720c3097a2b2108b666979ae02da2b30a7

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Thanks. 55a975720c3097a2b2108b666979ae02da2b30a7

datasync/chaindatafetcher/kafka/kafka.go
jeongkyun-oh: Right. It is one of the brokers. There are settings for authorization, but let's think about this later on.

datasync/chaindatafetcher/kafka/kafka_test.go
jeongkyun-oh: This is random test value to check if the published data is the same as consumed data.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Initially, the variable is set to `false`, so it won't close the `for` loop. The `Consume` method calls `Setup`, `ConsumeClaim`, and `Cleanup` consecutively. 
- `Setup` sets `isActive` be true.
- `ConsumeClaim` method has a `for` loop for fetching message from a channel. 
- `Cleanup` sets `isActive` be false.

So, when `Subscribe` is called while processing messages, it terminates the loop and create another loop to process messages.

datasync/chaindatafetcher/kafka/kafka_test.go
```diff
@@ -0,0 +1,148 @@
...
+		d := &kafkaData{}
+		s.Nil(json.Unmarshal(msg.Value, d))
+		done <- d
+
+		return nil
+	})
+	s.NoError(err)
+
+	// sleep for a while for the subscription to be set
+	time.Sleep(3 * time.Second)
```
aidan-kwon, line 96: @jeongkyun-oh Could you reduce the sleep time here and other places? It will increase CircleCi testing time.

jeongkyun-oh: I reduced them. But, the more time is taken for the other packages for `test-datasync` for your information.

스크린샷 2020-09-23 오전 11 12 45

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+		handler:  map[string]func(*sarama.ConsumerMessage) error{},
+		ctx:      ctx,
+		members:  members,
+		isActive: false,
+	}
+}
+
+func (c *Consumer) Subscribe(topic string, handler func(*sarama.ConsumerMessage) error) error {
+	if c.handler[topic] != nil {
+		return nil
```
KimKyungup, line 46: Why don't you return an error like `already exist topic?

jeongkyun-oh: The `Subcribe` method should be called only once. If it is called again with the same topic, it means that a goroutine already subscribes the topic. That is why it returns nil instead of error.

jeongkyun-oh:
Dear reviewers, please take another look. @aidan-kwon @ehnuje @yoomee1313 @winnie-byun @KimKyungup

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	c.handler[topic] = handler
+
+	if c.isActive {
+		c.cancelCh <- struct{}{}
```
winnie-byun, line 51: I think `Close()` function is missing? The channel is only closed when a new subsribe is occured.

winnie-byun (changes requested):

datasync/chaindatafetcher/kafka/kafka_test.go
aidan-kwon: thanks!

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,165 @@
...
+	partitions int32                                          // partitions is the number of partitions of a topic.
+}
+
+func New(groupID string, brokerList []string, replicas int16, partitions int32) types.EventBroker {
+	once.Do(func() {
+		kb = &Kafka{
+			brokers:    brokerList,
+			handlers:   map[string]func(*sarama.ConsumerMessage) error{},
+			replicas:   replicas,
+			partitions: partitions,
```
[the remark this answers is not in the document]
jeongkyun-oh: This will be updated when the flags are added.

datasync/chaindatafetcher/kafka/consumer.go
ehnuje: Which interface?

datasync/chaindatafetcher/kafka/kafka_test.go
ehnuje: @jeongkyun-oh @winnie-byun AFAIK, `SetupSuite` and `TearDownSuite` are called only once for the given suite. If you want to call function before/after each test, we should use `SetupTest` or `TearDownTest`. 


> Does `Suite` makes it enable to run test codes of singleton pattern?
> If so, we should apply them to other place, too. :)

@winnie-byun Something similar. `suite` package helps you avoiding repetitive resource allocation/deallocation. (and others, too!)

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: `ConsumerGroupHandler` is an interface for the consumer group handler.
I added some comments. f4179cafed18c871487dfeca5330c4d5b5712096

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,165 @@
...
+
+// Kafka connects to kafka brokers in an existing kafka cluster.
+type Kafka struct {
+	producer   sarama.AsyncProducer                           // producer creates and push a message to kafka brokers.
+	admin      sarama.ClusterAdmin                            // admin operates CRUD jobs for topics.
+	brokers    []string                                       // brokers a list of broker URLs.
+	handlers   map[string]func(*sarama.ConsumerMessage) error // handlers is a key-value map for topic-handler.
+	consumer   *Consumer                                      // consumer can subscribe a topic pushed by producer.
+	replicas   int16                                          // replicas is the replication factor of kafka settings. This is a number of the replicated partitions in the kafka cluster.
+	partitions int32                                          // partitions is the number of partitions of a topic.
```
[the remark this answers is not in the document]
jeongkyun-oh: Thanks. 96d4b144ff489af30dbe1062acdd39131dfe3b17

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,165 @@
...
+	})
+
+	return kb
+}
+
+func (k *Kafka) Publish(topic string, msg interface{}) error {
+	k.CreateTopic(topic)
+	item := &sarama.ProducerMessage{
+		Topic: topic,
+		Key:   sarama.StringEncoder(topic),
```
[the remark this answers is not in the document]
jeongkyun-oh: I thought the number of partition is 1, so it doesn't matter much, but I applied `nil` key for round-robin way.
23dae4af6b67a7033846e8290ea22c5d5174acef

datasync/chaindatafetcher/types/event.go
```diff
@@ -0,0 +1,33 @@
...
+// The klaytn library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+package types
+
```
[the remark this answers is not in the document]
jeongkyun-oh: Thanks. 9c25167823e3092f442e808be453a1859de35679

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: The channel is not closed anywhere. Does it really necessary to close the channel?

datasync/chaindatafetcher/kafka/kafka_test.go
ehnuje: @jeongkyun-oh Isn't it enough to check `Number` as it is also a random value?

datasync/chaindatafetcher/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	k.producer = producer
+}
+
+func (k *Kafka) newConsumerGroup(groupID string) sarama.ConsumerGroup {
+	config := sarama.NewConfig()
+	config.Version = sarama.MaxVersion
+	config.Consumer.Group.Session.Timeout = 6 * time.Second
+	config.Consumer.Group.Heartbeat.Interval = 2 * time.Second
```
ehnuje, line 142: Please use const value or config value instead of directly assigning here.

datasync/chaindatafetcher/kafka/consumer.go
winnie-byun: Sorry, I didn't see `c.ctx.Done()`. It is not necessary to do anything.

datasync/chaindatafetcher/kafka/kafka.go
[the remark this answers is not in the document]
winnie-byun: Sorry, I didn't explain enough. What do you think of code as follow :
```
type KafkaConfig struct {
	groupID    string
	brokers    []string // brokers is a list of broker URLs.
	replicas   int16    // replicas is a replication factor of kafka settings. This is the number of the replicated partitions in the kafka cluster.
	partitions int32    // partitions is the number of partitions of a topic.
}

func GetDefualtConfig() *KafkaConfig {
	return &KafkaConfig{
		groupID:    "",
		brokers:    []string{},
		replicas:   1,
		partitions: 1,
	}
}

// Kafka connects to kafka brokers in an existing kafka cluster.
type Kafka struct {
	KafkaConfig
	producer sarama.AsyncProducer                           // producer creates and push a message to kafka brokers.
	admin    sarama.ClusterAdmin                            // admin operates CRUD jobs for topics.
	handlers map[string]func(*sarama.ConsumerMessage) error // handlers is a key-value map for topic-handler.
	consumer *Consumer                                      // consumer can subscribe a topic pushed by producer.
}
```

KimKyungup (approved):

jeongkyun-oh:
Dear reviewers, I made another PR to make the client code simpler. Please take a look #685 instead of this.

PR #675, opened 2020-09-16
Implemented checkpoint db

jeongkyun-oh:
## Proposed changes

- Kafka is not persistent database, so the chaindatafetcher needs to store the checkpoint.
- The repository interface is divided into three types:
  1. `Repository` which handles `ChainEvent`
  2. `CheckpointDB` which writes/reads a checkpoint
  3. `ComponentSetter` which set a component (e.g. blockchain, debugAPI, blockchainAPI, database)

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


KimKyungup (dismissed):

datasync/chaindatafetcher/api.go
```diff
@@ -47,11 +47,11 @@ func (api *PublicChainDataFetcherAPI) Status() string {
 }
 
 func (api *PublicChainDataFetcherAPI) ReadCheckpoint() (int64, error) {
-	return api.f.repo.ReadCheckpoint()
+	return api.f.checkpointDB.ReadCheckpoint()
 }
 
 func (api *PublicChainDataFetcherAPI) WriteCheckpoint(checkpoint int64) error {
```
KimKyungup, line 53: Consider this next. :)
```suggestion
func (api *PublicChainDataFetcherAPI) WriteCheckpoint(checkpoint uint64) error {
```

jeongkyun-oh: Okay. I will make a ticket for this.

yoomee1313 (approved):

datasync/chaindatafetcher/kafka/checkpoint_db.go
```diff
@@ -0,0 +1,45 @@
...
+func (db *CheckpointDB) ReadCheckpoint() (int64, error) {
+	checkpoint, err := db.manager.ReadChainDataFetcherCheckpoint()
+	return int64(checkpoint), err
+}
+
+func (db *CheckpointDB) WriteCheckpoint(checkpoint int64) error {
+	return db.manager.WriteChainDataFetcherCheckpoint(uint64(checkpoint))
+}
+
+func (db *CheckpointDB) SetComponent(component interface{}) {
```
ehnuje, line 40: Is `SetComponent` an interface method?

jeongkyun-oh: Yes, it is. `CheckpointDB` implements two interfaces which is `CheckpointDB` and `ComponentSetter` to obtain the database manager from CN backend.

ehnuje (dismissed):

storage/database/db_manager.go
```diff
@@ -2027,3 +2031,22 @@ func (dbm *databaseManager) ReadGovernanceState() ([]byte, error) {
...
+	db := dbm.getDatabase(MiscDB)
+	hasCheckpoint, err := db.Has(chaindatafetcherCheckpointKey)
+	if err != nil || !hasCheckpoint {
+		return 0, err
+	}
+
+	data, err := db.Get(chaindatafetcherCheckpointKey)
+	if err != nil {
+		return 0, err
+	}
```
aidan-kwon, line 2050: AFAIK... In most DB, `Has` calls `Get` method internally and then checks its result. 
How about just calling `Get` method only and checking its returns (both data and err)?

jeongkyun-oh: Right, but it was hard to handle the error when key does not exist depending on the database.
This is because the errors are different based on them.

aidan-kwon: I think the current code also cannot distinguish the existence of the item for all databases. 
If you need to distinguish the difference, you better define new errors in this area. Currently, some database return  `0, nil` when the item doesn't exist, but another database returns `0, error`.

jeongkyun-oh: Right... The `Has` method works differently based on the sort of db. I guess I need to handle the existence error db by db.

aidan-kwon: Ok. Let's go back to the original subject. Do we still need to use `Has` method? Each of its returns needs to be handled just like `Get`.

jeongkyun-oh: It is not needed if the error returned by `Get` method is distinguishable. Let me check about this.

jeongkyun-oh: Please take another look.

ehnuje (approved):

yoomee1313 (approved):

aidan-kwon (approved):

aidan-kwon (approved):

KimKyungup (approved):

PR #726, opened 2020-10-23
database: collect database performance metrics by default

ehnuje:
## Proposed changes

- Now database's read/write performance metrics are collected by default.
- `--db.no-perf-metrics` can be used to disable collecting the metrics.
- API to disable/enable collecting the metrics will be done in another PR.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -262,6 +262,10 @@ var (
 		Name:  "db.no-parallel-write",
 		Usage: "Disables parallel writes of block data to persistent database",
 	}
+	DBNoPerformanceMetricsFlag = cli.BoolFlag{
+		Name:  "db.no-perf-metrics",
+		Usage: "Disables performance metrics of database's read and write operations",
+	}
 	TrieMemoryCacheSizeFlag = cli.IntFlag{
 		Name:  "state.cache-size",
 		Usage: "Size of in-memory cache of the global state (in MiB) to flush matured singleton trie nodes to disk",
@@ -1417,6 +1421,7 @@ func SetKlayConfig(ctx *cli.Context, stack *node.Node, cfg *cn.Config) {
 
 	cfg.LevelDBCompression = database.LevelDBCompressionType(ctx.GlobalInt(LevelDBCompressionTypeFlag.Name))
 	cfg.LevelDBBufferPool = !ctx.GlobalIsSet(LevelDBNoBufferPoolFlag.Name)
+	cfg.EnableDBPerfMetrics = !ctx.GlobalIsSet(DBNoPerformanceMetricsFlag.Name)
 	cfg.LevelDBCacheSize = ctx.GlobalInt(LevelDBCacheSizeFlag.Name)
 
 	cfg.DynamoDBConfig.TableName = ctx.GlobalString(DynamoDBTableNameFlag.Name)
--- a/storage/database/dynamodb.go
+++ b/storage/database/dynamodb.go
@@ -48,8 +48,8 @@ import (
 var overSizedDataPrefix = []byte("oversizeditem")
 
 // Performance of batch operations of DynamoDB are collected by default.
-var dynamoBatchWriteTimeMeter metrics.Meter
-var dynamoBatchWriteSizeMeter metrics.Meter
+var dynamoBatchWriteTimeMeter metrics.Meter = &metrics.NilMeter{}
+var dynamoBatchWriteSizeMeter metrics.Meter = &metrics.NilMeter{}
 
 // errors
 var dataNotFoundErr = errors.New("data is not found with the given key")
--- a/storage/database/dynamodb.go
+++ b/storage/database/dynamodb.go
@@ -49,7 +49,6 @@ var overSizedDataPrefix = []byte("oversizeditem")
 
 // Performance of batch operations of DynamoDB are collected by default.
 var dynamoBatchWriteTimeMeter metrics.Meter = &metrics.NilMeter{}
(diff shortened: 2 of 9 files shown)


cmd/utils/flags.go
```diff
@@ -231,6 +231,10 @@ var (
 		Name:  "db.leveldb.no-buffer-pool",
 		Usage: "Disables using buffer pool for LevelDB's block allocation",
 	}
+	LevelDBNoPerformanceCheckFlag = cli.BoolFlag{
+		Name:  "db.leveldb.no-perf-check",
+		Usage: "Disables performance checking of LevelDB's read and write operations",
+	}
```
aidan-kwon, line 237: How about adding a word "Metric" in the expanation of the flag?
```suggestion
	LevelDBNoPerformanceCheckMetricFlag = cli.BoolFlag{
		Name:  "db.leveldb.metrics.no-perf-check",
		Usage: "Disables performance checking metric of LevelDB's read and write operations",
	}
```

KimKyungup (approved):

cmd/utils/flags.go
ehnuje: @aidan-kwon I think "Performance Check" already has meaning of "metrics", how do you think?

aidan-kwon: "Performance Check" reminds me of profiling something. Anyway... it's not a big deal :)

aidan-kwon (approved):

yoomee1313 (approved):

storage/database/leveldb_database.go
```diff
@@ -415,12 +457,13 @@ hasError:
...
 
 func (db *levelDB) NewBatch() Batch {
-	return &ldbBatch{db: db.db, b: new(leveldb.Batch)}
+	return &ldbBatch{db: db.db, b: new(leveldb.Batch), ldb: db}
 }
 
 type ldbBatch struct {
 	db   *leveldb.DB
 	b    *leveldb.Batch
+	ldb  *levelDB
```
jeongkyun-oh, line 466: Dosen't this variable `ldb` have the above variables `db` and `b` as well?

ehnuje: @jeongkyun-oh `ldb` does have `db`, but not `b`. `db` can be removed if we put `*levelDB` as a member variable. Thanks!

jeongkyun-oh (approved):

winnie-byun:
Just two questions.
1. Will there be a performance degradation, due to this PR?
2. This might not be relevant to this PR. What happens if `meter.Mark()` is called before `metrics.NewRegisteredMeter()`? Is this an error? I am curioused because the code always call in sequence like : `newDatabaseManager()` -> `db.Meter()`. I think initializing the meters should occur before `newDatabaseManager` or inside it.

ehnuje:
> Just two questions.
> 
> 1. Will there be a performance degradation, due to this PR?

There is no performance degradation due to this PR. 

> 2. This might not be relevant to this PR. What happens if `meter.Mark()` is called before `metrics.NewRegisteredMeter()`? Is this an error? I am curioused because the code always call in sequence like : `newDatabaseManager()` -> `db.Meter()`. I think initializing the meters should occur before `newDatabaseManager` or inside it.

Panic happens when calling to the nil reference. To avoid that case, `dynamoBatchWriteTimeMeter` and `dynamoBatchWriteSizeMeter`, static variables, are initialized by `NilMeter`, which does nothing when `Mark` is called. As Put/Get/BatchWrite operations can be done after returning `dbManager` object, I think it is okay with the current initialization sequences. How do you think? @winnie-byun

storage/database/dynamodb.go
```diff
@@ -513,6 +537,7 @@ func (batch *dynamoBatch) Put(key, val []byte) error {
 	}
 
 	// If the size of the item is larger than the limit, it should be handled in different way
+	dynamoBatchWriteSizeMeter.Mark(int64(dataSize))
```
aidan-kwon, line 540: `dataSize` denotes the value length of a single item. What about removing the metric or changing it to batch item count?

storage/database/leveldb_database.go
```diff
@@ -99,6 +99,12 @@ type levelDB struct {
 	diskWriteMeter  metrics.Meter // Meter for measuring the effective amount of data written
 	blockCacheGauge metrics.Gauge // Gauge for measuring the current size of block cache
 
+	perfCheck           bool
+	getTimeMeter        metrics.Meter
+	putTimeMeter        metrics.Meter
+	batchWriteTimeMeter metrics.Meter
+	batchWriteSizeMeter metrics.Meter
```
aidan-kwon, line 106: `batchWriteSizeMeter` is declared, but not used anywhere. Is it for matching with metrics of Dynamo?

storage/database/dynamodb.go
ehnuje: You're right. I will remove this metric from both LevelDB and DynamoDB.

winnie-byun:
If there is no performance degradation, why did you define `DBNoPerformanceMetricsFlag`?

aidan-kwon (approved):

yoomee1313 (approved):

ehnuje:
> If there is no performance degradation, why did you define `DBNoPerformanceMetricsFlag`?

@winnie-byun Because I'm uncertain that I've tested all the possible cases. If we won't see any performance degradation for reasonable period of time, we surely can remove the flag.

winnie-byun (approved):

PR #785, opened 2020-11-26
Protocol Manager doesn't handle p2p messages if it has fake modules

aidan-kwon:
## Proposed changes

Protocol Manager doesn't handle p2p messages if it has fake modules to prevent unexpected errors.
To avoid `msgCh` overflows, it discards all received message.  

Even though the node with fake modules doesn't handle messages, it can deliver its txs to other peers. 

```
INFO[11/26,11:13:17 +09] [33] [Dial] Add dial candidate from static nodes  id=c0d515e5f3c6415e NodeType=0 ip=10.63.239.90 port="[32323 32324]"
INFO[11/26,11:13:17 +09] [33] Added a multichannel P2P Peer             id=c0d515e5f3c6415e conn=staticdial peerID=c0d515e5f3c6415e
WARN[11/26,11:13:17 +09] [33] ProtocolManager does not handle p2p messages  id=c0d515e5f3c6415e conn=staticdial fakeFetcher=true fakeDownloader=true
WARN[11/26,11:13:17 +09] [33] ProtocolManager does not handle p2p messages  id=c0d515e5f3c6415e conn=staticdial fakeFetcher=true fakeDownloader=true
WARN[11/26,11:13:17 +09] [33] ProtocolManager does not handle p2p messages  id=c0d515e5f3c6415e conn=staticdial fakeFetcher=true fakeDownloader=true
WARN[11/26,11:13:17 +09] [33] ProtocolManager does not handle p2p messages  id=c0d515e5f3c6415e conn=staticdial fakeFetcher=true fakeDownloader=true
```

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -465,14 +465,24 @@ func (pm *ProtocolManager) handle(p Peer) error {
 }
 
 func (pm *ProtocolManager) processMsg(msgCh <-chan p2p.Msg, p Peer, addr common.Address, errCh chan<- error) {
-	for msg := range msgCh {
-		if err := pm.handleMsg(p, addr, msg); err != nil {
-			p.GetP2PPeer().Log().Error("ProtocolManager failed to handle message", "msg", msg, "err", err)
-			errCh <- err
-			return
+	_, fakeF := pm.fetcher.(*fetcher.FakeFetcher)
+	_, fakeD := pm.downloader.(*downloader.FakeDownloader)
+	if fakeD || fakeF {
+		p.GetP2PPeer().Log().Warn("ProtocolManager does not handle p2p messages", "fakeFetcher", fakeF, "fakeDownloader", fakeD)
+		for msg := range msgCh {
+			msg.Discard()
 		}
-		msg.Discard()
+	} else {
+		for msg := range msgCh {
+			if err := pm.handleMsg(p, addr, msg); err != nil {
+				p.GetP2PPeer().Log().Error("ProtocolManager failed to handle message", "msg", msg, "err", err)
+				errCh <- err
+				return
+			}
+			msg.Discard()
+		}
 	}
+
 	p.GetP2PPeer().Log().Info("ProtocolManager.processMsg closed", "PeerName", p.GetP2PPeer().Name())
 }
 


yoomee1313 (approved):

node/cn/handler.go
```diff
@@ -465,14 +465,24 @@ func (pm *ProtocolManager) handle(p Peer) error {
...
 func (pm *ProtocolManager) processMsg(msgCh <-chan p2p.Msg, p Peer, addr common.Address, errCh chan<- error) {
-	for msg := range msgCh {
-		if err := pm.handleMsg(p, addr, msg); err != nil {
-			p.GetP2PPeer().Log().Error("ProtocolManager failed to handle message", "msg", msg, "err", err)
-			errCh <- err
-			return
+	_, fakeF := pm.fetcher.(*fetcher.FakeFetcher)
+	_, fakeD := pm.downloader.(*downloader.FakeDownloader)
+	if fakeD || fakeF {
+		p.GetP2PPeer().Log().Warn("ProtocolManager does not handle p2p messages", "fakeFetcher", fakeF, "fakeDownloader", fakeD)
```
yoomee1313, line 471: Just one question.. Can I ask why the warning msg is logged when it is a fake module? According to the description, it seems to be a normal operation. Or, is it a wrong behavior?

aidan-kwon: I thought discarding messages without handling is a very unexpected behavior of "handler". Because fake modules are introduced only for KES service nodes, I wanna inform the node is using fake modules and a handler is working in unexpected ways to node operators.

yoomee1313: Oh, using the fake module is not desirable behavior to non KES service nodes. Understood :)

winnie-byun (approved):

node/cn/handler.go
```diff
@@ -465,14 +465,24 @@ func (pm *ProtocolManager) handle(p Peer) error {
...
+		for msg := range msgCh {
+			msg.Discard()
```
KimKyungup, line 473: There is no `return`. Is it ok?

ehnuje (approved):

node/cn/handler.go
```diff
@@ -465,14 +465,24 @@ func (pm *ProtocolManager) handle(p Peer) error {
 }
 
 func (pm *ProtocolManager) processMsg(msgCh <-chan p2p.Msg, p Peer, addr common.Address, errCh chan<- error) {
-	for msg := range msgCh {
-		if err := pm.handleMsg(p, addr, msg); err != nil {
-			p.GetP2PPeer().Log().Error("ProtocolManager failed to handle message", "msg", msg, "err", err)
-			errCh <- err
-			return
+	_, fakeF := pm.fetcher.(*fetcher.FakeFetcher)
+	_, fakeD := pm.downloader.(*downloader.FakeDownloader)
```
KimKyungup, line 469: Instead of changing this, How about modifying FakeFetcer/FakeDownloader.
The root cause is the FakerFetcher/Downloader was not designed well.
What do you think?

node/cn/handler.go
ehnuje: As we don't care what the message is, there's no need to handle the error from `Discard`.

KimKyungup: This function is executed as a goroutine.
If it is the fakeD/F case, there is no condition to return this function.
Am I misunderstanding?

jeongkyun-oh (approved):

node/cn/handler.go
ehnuje: Oh, that can be a problem if the node is disconnected with the peer.
To recognize the error case, it seem that we need to check the error of `Discard`.

jeongkyun-oh: The `processMsg` goroutine can be terminated when `msgCh` is closed, which is in `Handle` method with `defer close(channel)`.

KimKyungup: I missed that point. Thanks @jeongkyun-oh

node/cn/handler.go
```diff
@@ -465,14 +465,24 @@ func (pm *ProtocolManager) handle(p Peer) error {
...
+	if fakeD || fakeF {
```
jeongkyun-oh, line 470: If fetcher or downloader is fake, then is it okay with ignoring all the other messages such as `NewBlockHashes`, `NewBlockMsg`?

node/cn/handler.go
aidan-kwon: The goroutine will be closed when `msgCh` is closed.

node/cn/handler.go
aidan-kwon: I think it's okay. Could you tell me what you are considering?

jeongkyun-oh: Unhandling those messages will not update td for remote peers. Never mind.
BTW, is there any chance to turn on only one of fake fetcher/downloader? Otherwise, how about changing `||` to `&&`?

KimKyungup (approved):

node/cn/handler.go
aidan-kwon: As you said, both of fake modules will be used together in all valid scenarios. However, I wanna warn the node operators using one of fake modules and make them avoid unexpected failures because the received p2p messages maybe not handled properly.

jeongkyun-oh: Understood. Thanks.

PR #804, opened 2020-12-03
Introduce timeout for dynamoDB client and increase maxRetries

aidan-kwon:
## Proposed changes

- Introduce timeout (10 seconds) for dynamoDB client to attempt again early. 
- Increase maxRetries from 3 (default) to 5 

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/storage/database/dynamodb.go
+++ b/storage/database/dynamodb.go
@@ -27,6 +27,7 @@ package database
 
 import (
 	"bytes"
+	"net/http"
 	"strconv"
 	"strings"
 	"sync"
@@ -56,6 +57,7 @@ var noTableNameErr = errors.New("dynamoDB table name not provided")
 const dynamoWriteSizeLimit = 399 * 1024 // The maximum write size is 400KB including attribute names and values
 const dynamoBatchSize = 25
 const dynamoMaxRetry = 5
+const dynamoTimeout = 10 * time.Second
 
 // batch write
 const WorkerNum = 10
@@ -121,24 +123,6 @@ func GetDefaultDynamoDBConfig() *DynamoDBConfig {
 	}
 }
 
-// GetTestDynamoConfig gets dynamo config for local test
-//
-// Please Run DynamoDB local with docker
-//    $ docker run -d -p 4566:4566 localstack/localstack:0.11.5
-func GetTestDynamoConfig() *DynamoDBConfig {
-	return &DynamoDBConfig{
-		Region:             "us-east-1",
-		Endpoint:           "http://localhost:4566",
-		S3Endpoint:         "http://localhost:4566",
-		TableName:          "klaytn-default" + strconv.Itoa(time.Now().Nanosecond()),
-		IsProvisioned:      false,
-		ReadCapacityUnits:  10000,
-		WriteCapacityUnits: 10000,
-		ReadOnly:           false,
-		PerfCheck:          false,
-	}
-}
-
 // NewDynamoDB creates either dynamoDB or dynamoDBReadOnly depending on config.ReadOnly.
 func NewDynamoDB(config *DynamoDBConfig) (Database, error) {
 	if config.ReadOnly {
@@ -176,6 +160,8 @@ func newDynamoDB(config *DynamoDBConfig) (*dynamoDB, error) {
 				Endpoint:         aws.String(config.Endpoint),
 				Region:           aws.String(config.Region),
 				S3ForcePathStyle: aws.Bool(true),
+				MaxRetries:       aws.Int(dynamoMaxRetry),
+				HTTPClient:       &http.Client{Timeout: dynamoTimeout}, // default client is &http.Client{}
 			},
 		})))
 	}
--- a/storage/database/dynamodb_test.go
+++ b/storage/database/dynamodb_test.go
@@ -91,6 +91,8 @@ func (s *SuiteDynamoDB) TestDynamoDB_Put() {
 }
 
 func (s *SuiteDynamoDB) TestDynamoDB_Timeout() {
+	// fakeEndpoint allows TCP handshakes, but doesn't answer anything to client.
+	// The fake server is used to produce a network failure scenario.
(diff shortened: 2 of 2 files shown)


storage/database/dynamodb.go
```diff
@@ -56,6 +57,7 @@ var noTableNameErr = errors.New("dynamoDB table name not provided")
 const dynamoWriteSizeLimit = 399 * 1024 // The maximum write size is 400KB including attribute names and values
 const dynamoBatchSize = 25
 const dynamoMaxRetry = 5
```
KimKyungup, line 59: `Increase maxRetries from 3 (default) to 5`
There is no change, right? @aidan-kwon

aidan-kwon: `dynamoMaxRetry` was used only for batch write. Other dynamoDB client operations had a default maxRetry value, 3. With this change in line 163, `dynamoMaxRetry` is applied for all dynamoDB client operations.

storage/database/dynamodb_test.go
```diff
@@ -85,6 +90,59 @@ func (s *SuiteDynamoDB) TestDynamoDB_Put() {
 	s.NoError(returnedErr)
 }
 
+func (s *SuiteDynamoDB) TestDynamoDB_Timeout() {
```
winnie-byun, line 93: I am confused with this test. Could you add more comments for this test?
1) Instead of using the localstack, the code creates a new temp server `fakeEndpoint`?
2)Why are we using a new temp server instead of local stack?
3) Is there a purpose of setting the regions to `"ap-northeast-2"`? Why did you set different region for dynamoDB client and dynamoDB?

storage/database/dynamodb_test.go
```diff
@@ -85,6 +90,59 @@ func (s *SuiteDynamoDB) TestDynamoDB_Put() {
...
+
+	dynamoDBClient = dynamodb.New(session.Must(session.NewSessionWithOptions(session.Options{
+		Config: aws.Config{
+			Endpoint:   aws.String(conf.Endpoint),
+			Region:     aws.String("ap-northeast-2"),
+			MaxRetries: aws.Int(2),
+			HTTPClient: &http.Client{Timeout: 1 * time.Second},
+		},
+	})))
+	expectedElapsed := time.Duration(*dynamoDBClient.Config.MaxRetries+1) * dynamoDBClient.Config.HTTPClient.Timeout
```
winnie-byun, line 132: Is `+1` for the first try?

storage/database/dynamodb.go
winnie-byun:
```suggestion
const dynamoMaxRetry = 5 // used in DynamoDB client and batchwrite
```

storage/database/dynamodb_test.go
aidan-kwon: Yes

storage/database/dynamodb.go
aidan-kwon: Does it mean everywhere using dynamoDB?

storage/database/dynamodb_test.go
aidan-kwon: I added more comments.
1. I wanna produce a network failure scenario. If the localstack is used, it is hard to drop dynamoDB requests allowing TCP handshake. 
2. I just used a test config already specified. The config is just used to create dynamoDB without any errors, so the actual values are not important. Anyway, I match up the region values as you remarked.

ehnuje (dismissed):
LGTM except for the other reviewers' comments

storage/database/dynamodb_test.go
```diff
@@ -85,6 +90,62 @@ func (s *SuiteDynamoDB) TestDynamoDB_Put() {
 	s.NoError(returnedErr)
 }
```
winnie-byun, line 92:
```suggestion

// TestDynamoDB_Timeout tests if a timeout error occurs.
// When there is no answer from DynamoDB server due to network failure, 
// a timeout error should occur.
// A fake server is setup to simulate a server with a response latency.
```

storage/database/dynamodb_test.go
```diff
@@ -85,6 +90,62 @@ func (s *SuiteDynamoDB) TestDynamoDB_Put() {
...
 
+func (s *SuiteDynamoDB) TestDynamoDB_Timeout() {
+	// fakeEndpoint allows TCP handshakes, but doesn't answer anything to client.
+	// The fake server is used to produce a network failure scenario.
+	fakeEndpoint := "127.0.0.1:14566"
```
winnie-byun, line 96: Could you make a variable for max tries and timeout? It would help a lot figuring out this test.
```suggestion
	fakeEndpoint := "127.0.0.1:14566"
	maxRetries := 2
	timeout = 1 * time.Second
```
```
			if err := listen.SetDeadline(time.Now().Add((maxRetries+1) * timeout * 2)); err != nil {
```

winnie-byun (changes requested):

storage/database/dynamodb_test.go
aidan-kwon: Applied.

storage/database/dynamodb_test.go
```diff
@@ -85,6 +90,68 @@ func (s *SuiteDynamoDB) TestDynamoDB_Put() {
...
+		}
+
+		listen, err := net.ListenTCP("tcp", tcpAddr)
+		if err != nil {
+			s.FailNow(err.Error())
+		}
+		defer listen.Close()
+
+		for {
+			// Deadline prevents infinite waiting of the fake server
```
winnie-byun, line 116: Just one more comment :)
```suggestion
			// Deadline prevents infinite waiting of the fake server
			// Wait longer than (maxRetries+1) * timeout
```

ehnuje (approved):

winnie-byun (approved):

KimKyungup (approved):

PR #815, opened 2020-12-10
abi: import files from latest geth release

yoomee1313:
## Proposed changes
**This PR updates abi package and abigen. It applies all abi related changes up to v1.9.21.**

abi, bind package is identical with geth v1.9.21. However, backends package is not exactly identical because some geth changes are not applied yet. I didn’t applied them or applied part of them because those changes are huge and not only included at abi package. Also, I removed unnecessary codes such as vyper, clef, etc. You can refer the **commit msgs** which geth PR is applied. 

**Below PRs are the PR lists that i didn’t included. If needed, please refer them.**
* internal/ethapi: return revert reason (geth [#21083](https://github.com/ethereum/go-ethereum/pull/21083))
* all: seperate consensus error and evm internal error (geth [#20830](https://github.com/ethereum/go-ethereum/pull/20830))
* eth/filters, interfaces.go: EIP-234 Add blockHash option to eth_getLogs (geth [#16734](https://github.com/ethereum/go-ethereum/pull/16734))
* accounts/abi/bind/backend, internal/ethapi: recap gas limit with balance  (geth [#21043](https://github.com/ethereum/go-ethereum/pull/21043), geth [#21346](https://github.com/ethereum/go-ethereum/pull/21346))

**fix NewFallback TC**
NewFallback TC contains a bug. Fixed TC and fixed NewFallback contract to below. 
Related Geth PR https://github.com/ethereum/go-ethereum/pull/22053
```
pragma solidity ^0.6.2;

contract NewFallbacks {
  event Fallback(bytes data);
  fallback() external { 
      emit Fallback(msg.data);
  }
  event Received(address addr, uint value); 
  receive() external payable {
      emit Received(msg.sender, msg.value);
  }
}
```

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

## Related issues

## Further comments


ehnuje (commented):
It would be appreciated if you describe the which release you used to make this PR. Also, did you check the code diff result of this code and the code in the geth? Are they identical?

yoomee1313:
> It would be appreciated if you describe the which release you used to make this PR. Also, did you check the code diff result of this code and the code in the geth? Are they identical?

This is good idea. I'll make a description about the release information. I checked the diff result, but it's not identical because some PRs are not applied since they are related with other packages. However, at that time, i thought it wasn't necessary to bring those PRs, but now, it would be better to bring them to match the codebase as much as possible. I'll let you know after more work.

common/compiler/solidity.go
```diff
@@ -187,15 +180,3 @@ func (s *Solidity) run(cmd *exec.Cmd, source string) (map[string]*Contract, erro
 	}
 	return contracts, nil
 }
-
```
KimKyungup, line 190: This is moved to common/compiler/helpers.go. :)

KimKyungup (commented):
I compared this abi pkg code with geth reference code with @yoomee1313.
There is some difference. But @yoomee1313 will list up the PRs which is not applied in this PR. They caused some differences.

And I reviewed extra code also. I think it's ok.

However, the new fallback test code doesn't work.
@yoomee1313 Please check the test. 
I think there may be some differences in the transaction call routine.

accounts/abi/bind/bind_test.go
```diff
@@ -950,12 +1066,635 @@ var bindTests = []struct {
...
+			sim.Commit()
+	
+			var gotEvent bool
+			iter, _ := c.FilterReceived(nil)
+			defer iter.Close()
+			for iter.Next() {
+				if iter.Event.Addr != addr {
+					t.Fatal("Msg.sender mismatch")
+				}
+				if iter.Event.Value.Uint64() != 100 {
```
KimKyungup, line 1658: How about theses?
```suggestion
				if iter.Event.Value.Uint64() != opts.Value.Uint64() {
```

```suggestion
				if iter.Event.Value.Cmp(opts.Value) != 0 {
```

common/compiler/solidity.go
yoomee1313: @KimKyungup yes, geth have made helpers.go file and moved slurpFiles to the helpers.go file.

accounts/abi/bind/bind_test.go
yoomee1313: Thank you, i applied it.

KimKyungup (approved):

KimKyungup:
In content, all geth PR link is broken. Could you link them with their PRs?

ehnuje (dismissed):

ehnuje:
@yoomee1313 Could you please share the progress of this PR? Are there any other things left to merge this change?

yoomee1313:
Forgive me for being late. I was delayed trying to find out about the situations below.
(1) NewFallback fix verification has been completed.
-> In the original code, calldata was copied to memory using calldatacopy and then copied to the event through parameters.
-> However, there was a problem in the above process, so the structure has been changed to copy `msg.data` to the event through parameters.
(2) The execution of the OpenZeppelin Test Suite has been completed.
-> In the openzeppelin test, it was checked whether this PR had an effect on the revert reason error. The error is that the revert reason is not returned.
-> The conclusion is that the error is not affected by this PR. Instead, it is related to the PRs (especially eth_call revert reason) which are described in the above description. We will complete the support for this in the future with next `incompatibility change`.

In conclusion, this PR supports `solidity 0.6` including New Fallback and struct type.
Please take a look! @KlaytnDev

KimKyungup (approved):

ehnuje (approved):

KimKyungup:
PTAL @aidan-kwon. Let's merge~!!!

aidan-kwon (approved):

PR #818, opened 2020-12-11
Removed break insertedFailed and sleep a bit when insertTasks is full

jeongkyun-oh:
## Proposed changes

- The insertFailed break point is removed.
  - if a insertion takes a long time, the iterating queue does the same job while doing push and pop the same task.
  - if it breaks, then it can continue the next job.
- If it does not sleep, it leaves the log `Failed to import ...` log too much.
  - the channel is full already, so the number of insertion jobs is enough

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/datasync/fetcher/fetcher.go
+++ b/datasync/fetcher/fetcher.go
@@ -33,9 +33,10 @@ import (
 )
 
 const (
-	arriveTimeout = 500 * time.Millisecond // Time allowance before an announced block is explicitly requested
-	gatherSlack   = 100 * time.Millisecond // Interval used to collate almost-expired announces with fetches
-	fetchTimeout  = 5 * time.Second        // Maximum allotted time to return an explicitly requested block
+	arriveTimeout       = 500 * time.Millisecond // Time allowance before an announced block is explicitly requested
+	gatherSlack         = 100 * time.Millisecond // Interval used to collate almost-expired announces with fetches
+	fetchTimeout        = 5 * time.Second        // Maximum allotted time to return an explicitly requested block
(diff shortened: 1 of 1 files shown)


KimKyungup (dismissed):

datasync/fetcher/fetcher.go
```diff
@@ -342,7 +342,9 @@ func (f *Fetcher) loop() {
 				if f.queueChangeHook != nil {
 					f.queueChangeHook(hash, true)
 				}
-				break insertFailed
+				// insertTasks channel is full, sleep a bit
+				time.Sleep(insertTasksWaitTime)
```
aidan-kwon, line 346: How about setting a timer instead of `default`? Since the task has the exact block number the fetcher needs, it would be better to wait for a while instead of sending the task to the end of the queue. 
```
select {
case f.insertTasks <- insertTask{peer, block}: // do somthing
case <-time.NewTimer(insertTasksWaitTime).C:  // do the same things in `default`
}
```

aidan-kwon (approved):

datasync/fetcher/fetcher.go
aidan-kwon: Don't mind too much because it's a small comment.

ehnuje (dismissed):

datasync/fetcher/fetcher.go
jeongkyun-oh: Looks better. Thanks!

aidan-kwon (approved):

KimKyungup (approved):

winnie-byun (approved):

PR #777, opened 2020-11-24
Increase redis timeout durations and change a log level

aidan-kwon:
## Proposed changes

- Increase redis timeout durations
  - Ping takes 10 ~ 20 ms from my office to EC2 in the best network condition. I think it could take longer than 100 ms easily depends on network congestion. 
- Change a log level

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [ ] New feature or enhancement
- [x] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #819, opened 2020-12-15
Redis cache sets items asynchronously

aidan-kwon:
## Proposed changes

As redis client sets items synchronously, the node will wait for a while when redis client failed/delayed to set items. 
Even though the timeout duration is very short, it may delay block processing because it happens on the critical time path.

This change introduced worker goroutines to set items asynchronously. 
Also, there are minor changes related to redis. 

- Redis client sets items asynchronously
- Minor changes 
  - pubsub will be created in `NewRedisCache` and will be closed when the cache is closed 
  - In test cases, give a sleep time when redis (asynchronously) set items

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/storage/statedb/cache_hybrid_test.go
+++ b/storage/statedb/cache_hybrid_test.go
@@ -11,7 +11,7 @@ import (
 func getTestHybridConfig() *TrieNodeCacheConfig {
 	return &TrieNodeCacheConfig{
 		CacheType:          CacheTypeHybrid,
-		LocalCacheSizeMB:   1024,
+		LocalCacheSizeMB:   100,
 		FastCacheFileDir:   "",
 		RedisEndpoints:     []string{"localhost:6379"},
 		RedisClusterEnable: false,
--- a/storage/statedb/cache_hybrid_test.go
+++ b/storage/statedb/cache_hybrid_test.go
@@ -3,6 +3,7 @@ package statedb
 import (
 	"bytes"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -27,6 +28,7 @@ func TestHybridCache_Set(t *testing.T) {
 	// Set an item
 	key, value := randBytes(32), randBytes(500)
 	cache.Set(key, value)
+	time.Sleep(100 * time.Millisecond)
 
 	// Type assertion to check both of local cache and remote cache
 	hybrid, ok := cache.(*HybridCache)
@@ -71,6 +73,7 @@ func TestHybridCache_Get(t *testing.T) {
 		// Store an item into remote cache
 		key, value := randBytes(32), randBytes(500)
 		remoteCache.Set(key, value)
+		time.Sleep(100 * time.Millisecond)
 
 		// Make sure the item is not stored in the local cache.
 		assert.Equal(t, len(localCache.Get(key)), 0)
@@ -89,6 +92,7 @@ func TestHybridCache_Get(t *testing.T) {
 		key, value := randBytes(32), randBytes(500)
 		localCache.Set(key, value)
 		remoteCache.Set(key, []byte{0x11})
+		time.Sleep(100 * time.Millisecond)
 
 		// Get the item from the hybrid cache and check the validity
 		returnedVal := hybrid.Get(key)
@@ -127,6 +131,7 @@ func TestHybridCache_Has(t *testing.T) {
 		// Store an item into remote cache
 		key, value := randBytes(32), randBytes(500)
 		remoteCache.Set(key, value)
+		time.Sleep(100 * time.Millisecond)
 
 		// Get the item from the hybrid cache and check the validity
 		returnedVal, returnedExist := hybrid.Has(key)
@@ -140,6 +145,7 @@ func TestHybridCache_Has(t *testing.T) {
 		key, value := randBytes(32), randBytes(500)
 		localCache.Set(key, value)
 		remoteCache.Set(key, []byte{0x11})
+		time.Sleep(100 * time.Millisecond)
 
 		// Get the item from the hybrid cache and check the validity
 		returnedVal, returnedExist := hybrid.Has(key)
--- a/storage/statedb/cache_hybrid_test.go
+++ b/storage/statedb/cache_hybrid_test.go
@@ -11,7 +11,7 @@ import (
 func getTestHybridConfig() *TrieNodeCacheConfig {
 	return &TrieNodeCacheConfig{
 		CacheType:          CacheTypeHybrid,
-		LocalCacheSizeMB:   1024 * 1024,
+		LocalCacheSizeMB:   1024,
 		FastCacheFileDir:   "",
 		RedisEndpoints:     []string{"localhost:6379"},
 		RedisClusterEnable: false,
@@ -28,7 +28,7 @@ func TestHybridCache_Set(t *testing.T) {
 	// Set an item
 	key, value := randBytes(32), randBytes(500)
 	cache.Set(key, value)
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(sleepDurationForAsyncBehavior)
 
 	// Type assertion to check both of local cache and remote cache
 	hybrid, ok := cache.(*HybridCache)
@@ -73,7 +73,7 @@ func TestHybridCache_Get(t *testing.T) {
 		// Store an item into remote cache
 		key, value := randBytes(32), randBytes(500)
 		remoteCache.Set(key, value)
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(sleepDurationForAsyncBehavior)
 
 		// Make sure the item is not stored in the local cache.
 		assert.Equal(t, len(localCache.Get(key)), 0)
@@ -92,7 +92,7 @@ func TestHybridCache_Get(t *testing.T) {
 		key, value := randBytes(32), randBytes(500)
 		localCache.Set(key, value)
 		remoteCache.Set(key, []byte{0x11})
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(sleepDurationForAsyncBehavior)
 
 		// Get the item from the hybrid cache and check the validity
 		returnedVal := hybrid.Get(key)
@@ -131,7 +131,7 @@ func TestHybridCache_Has(t *testing.T) {
 		// Store an item into remote cache
 		key, value := randBytes(32), randBytes(500)
 		remoteCache.Set(key, value)
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(sleepDurationForAsyncBehavior)
 
 		// Get the item from the hybrid cache and check the validity
 		returnedVal, returnedExist := hybrid.Has(key)
@@ -145,7 +145,7 @@ func TestHybridCache_Has(t *testing.T) {
 		key, value := randBytes(32), randBytes(500)
 		localCache.Set(key, value)
 		remoteCache.Set(key, []byte{0x11})
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(sleepDurationForAsyncBehavior)
 
 		// Get the item from the hybrid cache and check the validity
 		returnedVal, returnedExist := hybrid.Has(key)
--- a/storage/statedb/cache_redis.go
+++ b/storage/statedb/cache_redis.go
@@ -18,6 +18,7 @@ package statedb
 
 import (
 	"errors"
+	"runtime"
 	"time"
 
 	"github.com/go-redis/redis/v7"
@@ -25,6 +26,7 @@ import (
 )
 
 const (
+	redisSetItemChannelSize       = 1024
 	redisSubscriptionChannelSize  = 100 // default value of redis client
 	redisSubscriptionChannelBlock = "latestBlock"
 )
@@ -37,8 +39,14 @@ var (
 )
 
 type RedisCache struct {
-	client redis.UniversalClient
-	pubSub *redis.PubSub
+	client    redis.UniversalClient
+	setItemCh chan setItem
+	pubSub    *redis.PubSub
+}
+
+type setItem struct {
+	key   []byte
+	value []byte
 }
 
 func newRedisClient(endpoints []string, isCluster bool) (redis.UniversalClient, error) {
@@ -68,6 +76,8 @@ func newRedisClient(endpoints []string, isCluster bool) (redis.UniversalClient, 
 	}), nil
 }
 
+// NewRedisCache creates a redis cache containing redis client, setItemCh and pubSub.
+// It generates worker goroutines to process Set commands asynchronously.
 func NewRedisCache(config *TrieNodeCacheConfig) (*RedisCache, error) {
 	cli, err := newRedisClient(config.RedisEndpoints, config.RedisClusterEnable)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


storage/statedb/cache_redis.go
```diff
@@ -18,13 +18,15 @@ package statedb
...
 	"errors"
+	"runtime"
 	"time"
 
 	"github.com/go-redis/redis/v7"
 	"github.com/klaytn/klaytn/common/hexutil"
 )
 
 const (
+	redisSetItemChannelSize       = 1024
```
winnie-byun, line 29: 1. Do you think this is enough size to handle all items?
  Last week, 1318 WCU was used in DynamoDB. 
2. Does channel size with power of 2 perform better?

storage/statedb/cache_hybrid_test.go
```diff
@@ -27,6 +28,7 @@ func TestHybridCache_Set(t *testing.T) {
 	// Set an item
 	key, value := randBytes(32), randBytes(500)
 	cache.Set(key, value)
+	time.Sleep(100 * time.Millisecond)
```
winnie-byun, line 31: How about setting the sleep time with `redisCacheTimeout`?
```suggestion
	time.Sleep(redisCacheTimeout)
```

storage/statedb/cache_redis_test.go
```diff
@@ -146,13 +191,13 @@ func TestNewRedisCache_Timeout(t *testing.T) {
...
 		MaxRetries:   0,
-	}), nil}
+	}), nil, nil}
 
 	key, value := randBytes(32), randBytes(500)
 
 	start := time.Now()
 	cache.Set(key, value)
-	assert.Equal(t, redisCacheTimeout, time.Since(start).Round(redisCacheTimeout/2))
+	time.Sleep(100 * time.Millisecond) // redis cache set item asynchronously
```
winnie-byun, line 200: How about setting the sleep time with `redisCacheTimeout`?
```suggestion
	time.Sleep(redisCacheTimeout) // redis cache set item asynchronously
```

storage/statedb/cache_redis_test.go
```diff
@@ -98,22 +99,66 @@ func TestNewRedisCache(t *testing.T) {
...
+				i--
+			}
+		}
+		successCh <- struct{}{}
+	}()
+
+	for i := 0; i < itemsLen; i++ {
+		if i == redisSetItemChannelSize {
+			// sleep for a while because Set command can drop an item if setItem channel is full
+			time.Sleep(2 * time.Second)
```
winnie-byun, line 147: Is there a reason you this time to 2 seconds? It is a long time for a test.

storage/statedb/cache_redis_test.go
```diff
@@ -98,22 +99,66 @@ func TestNewRedisCache(t *testing.T) {
...
+	successCh := make(chan struct{})
+	go func() {
+		for i := 0; i < itemsLen; i++ {
+			v := cache.Get(items[i].key)
+			if !bytes.Equal(v, items[i].value) {
+				// if the item is not set yet, wait and retry
+				time.Sleep(100 * time.Millisecond)
+				i--
+			}
+		}
```
winnie-byun, line 140:
```suggestion
		for i := 0; i < itemsLen; i++ {
			MaxRetry := 5
			v := cache.Get(items[i].key)
			if len(v) == 0 && MaxRetry > 0 {
				// if the item is not set yet, wait and retry
				time.Sleep(100 * time.Millisecond)
				i--
				MaxRetry--
			} else if MaxRetry <= 0 {
				t.failNow()
			} else {
				assert.Equal(v, items[i].value)
			}
		}
```
1. I think the comparing should be done with `assert.Equal`.
2. How about replacing the check condition with `if len(items[i].value)` for retrying logic.
3. How about setting `maxRetry` for each item. If there is a problem connecting to a redis server, this code should wait more than 3min for the test comes out.

storage/statedb/cache_redis.go
aidan-kwon: 1. In the test, the size was enough to execute `debug.StartWarmUp()` without any errors. Redis could handle write commands faster than DynamoDB in more cases. Then, the channel will not be fulled.
2. I don't know exactly. Even if it does, the difference is probably not quite big. It just took the number thinking that it might help to allocate memory efficiently. However, the difference is probably not quite big also. If you have any other opinions, please let me know.

storage/statedb/cache_hybrid_test.go
aidan-kwon: Thank you for the interesting suggestion, but I am trying to increase `redisCacheTimeout` in https://github.com/klaytn/klaytn/pull/777. The value could be longer than currently suggested if this async set PR is merged. I think it is a little long to use frequently.

storage/statedb/cache_redis_test.go
aidan-kwon: Thank you for the interesting suggestion, but I am trying to increase `redisCacheTimeout` in https://github.com/klaytn/klaytn/pull/777. The value could be longer than currently suggested if this async set PR is merged. I think it is a little long to use frequently. How do you think?

storage/statedb/cache_redis_test.go
aidan-kwon: Just a heuristic number derived in my test. I gave enough time for the sleep because it only happens once. We could decrease it more tightly also. Do you have any suggestions?

storage/statedb/cache_redis_test.go
aidan-kwon: **Answer for 2, 3**
The main thread waits just 5 seconds for `successCh`. When the "for" loop lasts more than 5 seconds, the test will be finished with `t.Fatal()`.
```
	select {
	case <-successCh:
	case <-time.After(5 * time.Second):
		t.Fatal("timeout")
	}
```

**Answer for 1**
I wanted to make the test code simple, but it would be more precise if the "if" branch in your qute is changed adopting your suggestion.
```
if v == nil {
	time.Sleep(100 * time.Millisecond)
	i--
} else {
	assert.Equal(t, v, items[i].value)
}
```
How do you think about it?

storage/statedb/cache_redis_test.go
aidan-kwon: Oh, `cache.Set` cannot be tested in the test since it operates asynchronously. I am gonna remove the 3 lines using `cache.Set`. You let me check this again. Thanks!

storage/statedb/cache_redis_test.go
aidan-kwon: FYI, `t.FailNow()` in goroutines cannot finish the test. 
```
// FailNow must be called from the goroutine running the
// test or benchmark function, not from other goroutines
// created during the test. Calling FailNow does not stop
// those other goroutines.
```

storage/statedb/cache_redis.go
```diff
@@ -76,22 +86,45 @@ func NewRedisCache(config TrieNodeCacheConfig) (*RedisCache, error) {
...
 	}
 
-	logger.Info("Initialize trie node cache with redis", "endpoint", config.RedisEndpoints,
+	cache := &RedisCache{
+		client:    cli,
+		setItemCh: make(chan setItem, redisSetItemChannelSize),
+		pubSub:    cli.Subscribe(),
+	}
+
+	workerNum := runtime.NumCPU() / 2
```
jeongkyun-oh, line 95: How about adding a guard for single-core machine just in case?

aidan-kwon: Good suggestion. I will add 1. Thanks!

storage/statedb/cache_redis_test.go
winnie-byun: Thank you for considering.

storage/statedb/cache_hybrid_test.go
winnie-byun: 1. `100 * time.Millisecond` looks fine for now.
2. I think there should be a vairable for the time, though. It is used a lot in the redis test codes.

storage/statedb/cache_redis_test.go
winnie-byun: What do you think of switching the go routine part? 'Set` is done in a new go routine, `Get` is done in test goroutine. Because the `Get` part has the checking logics, it should be in the main routine.

storage/statedb/cache_redis_test.go
aidan-kwon: I cached it to test `redisCache.set` instead of `cache.Set`. Because `redisCache.set` writes synchronously, we can use the previous testing logic.

storage/statedb/cache_redis_test.go
aidan-kwon: I switched it. Thanks

jeongkyun-oh (approved):

storage/statedb/cache_redis_test.go
winnie-byun: Thank you for kind explanation :)

storage/statedb/cache_redis.go
```diff
@@ -76,22 +86,45 @@ func NewRedisCache(config TrieNodeCacheConfig) (*RedisCache, error) {
...
 
 func (cache *RedisCache) Get(k []byte) []byte {
 	val, err := cache.client.Get(hexutil.Encode(k)).Bytes()
 	if err != nil {
-		// TODO-Klyatn: Print specific errors if needed
 		logger.Debug("cannot get an item from redis cache", "err", err, "key", hexutil.Encode(k))
 		return nil
 	}
 	return val
 }
```
winnie-byun, line 117:
```suggestion

// Set writes data asynchronously. Not all data is written if a setItemCh is full.
// To write all data, please use set instead.
```

storage/statedb/cache_redis.go
```diff
@@ -76,22 +86,45 @@ func NewRedisCache(config TrieNodeCacheConfig) (*RedisCache, error) {
...
 
 func (cache *RedisCache) Set(k, v []byte) {
+	item := setItem{key: k, value: v}
+	select {
+	case cache.setItemCh <- item:
+	default:
+		logger.Error("redis setItem channel is full")
+	}
+}
+
```
winnie-byun, line 126:
```suggestion

// set writes data synchronously. This can take some time to write all data.
// To write data asynchronously, please use Set instead.
```

storage/statedb/cache_hybrid_test.go
aidan-kwon: Converted it to a variable, `sleepDurationForAsync`

storage/statedb/cache_redis_test.go
```diff
@@ -98,22 +99,68 @@ func TestNewRedisCache(t *testing.T) {
...
 	cache.Set(key, value)
+	time.Sleep(100 * time.Millisecond)
 
 	retValue := cache.Get(key)
 	assert.Equal(t, bytes.Compare(value, retValue), 0)
 }
 
-// TestNewRedisCache_Timeout test timout feature of redis client.
-func TestNewRedisCache_Timeout(t *testing.T) {
+func TestRedisCache_Set_LargeNumberItems(t *testing.T) {
```
winnie-byun, line 119: When the Set channel is full, not writing is expected behavior.
I am not sure what this test is trying to do.
Is this an example code for asynchronous write without failure?

storage/statedb/cache_hybrid_test.go
winnie-byun: Thank you for adopting the suggestion :)

storage/statedb/cache_redis_test.go
aidan-kwon: As you mentioned, it is hard to test channel full because it's expected behavior. The test just ensures that all items passed to the channel are consumed correctly (not remained in the channel) in normal operation.

storage/statedb/cache_redis.go
aidan-kwon: Thanks. I added the comments.

storage/statedb/cache_redis.go
aidan-kwon: Thanks. I added the comments.

winnie-byun (approved):
Thank you for accepting my comments :)

aidan-kwon:
> Thank you for accepting my comments :)

Thank you for your active review!

storage/statedb/cache_redis_test.go
```diff
@@ -98,22 +101,68 @@ func TestNewRedisCache(t *testing.T) {
...
+			// if the item is not set yet, wait and retry
+			time.Sleep(sleepDurationForAsyncBehavior)
+			i--
+		} else {
+			assert.Equal(t, v, items[i].value)
+		}
+	}
+}
+
+// TestRedisCache_Timeout test timout feature of redis client.
```
yoomee1313, line 164:
```suggestion
// TestRedisCache_Timeout tests timout feature of redis client.
```

storage/statedb/cache_redis_test.go
```diff
@@ -146,12 +195,13 @@ func TestNewRedisCache_Timeout(t *testing.T) {
...
 		WriteTimeout: redisCacheTimeout,
 		MaxRetries:   0,
-	}), nil}
+	}), nil, nil}
 
 	key, value := randBytes(32), randBytes(500)
 
 	start := time.Now()
-	cache.Set(key, value)
+	redisCache := cache.(*RedisCache) // Because RedisCache.Set item asynchronously, use RedisCache.set
```
yoomee1313, line 203:
```suggestion
	redisCache := cache.(*RedisCache) // Because RedisCache.Set writes item asynchronously, use RedisCache.set
```

yoomee1313:
It is very clear and it looks good to me except for minor comments.

jeongkyun-oh (approved):

yoomee1313 (approved):

winnie-byun (approved):

KimKyungup (dismissed):

storage/statedb/cache_hybrid_test.go
```diff
@@ -3,14 +3,15 @@ package statedb
...
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func getTestHybridConfig() TrieNodeCacheConfig {
 	return TrieNodeCacheConfig{
 		CacheType:          CacheTypeHybrid,
-		LocalCacheSizeMB:   1024 * 1024,
+		LocalCacheSizeMB:   1024,
```
KimKyungup, line 14: Is this 1GB?? Is it too much for testing?

storage/statedb/cache_redis.go
```diff
@@ -76,22 +86,49 @@ func NewRedisCache(config TrieNodeCacheConfig) (*RedisCache, error) {
...
 }
 
+// Set writes data asynchronously. Not all data is written if a setItemCh is full.
+// To write data synchronously, use set instead.
 func (cache *RedisCache) Set(k, v []byte) {
+	item := setItem{key: k, value: v}
+	select {
+	case cache.setItemCh <- item:
+	default:
+		logger.Error("redis setItem channel is full")
```
ehnuje, line 125: `Warn` would be more appropriate?

ehnuje (commented):
@aidan-kwon Have you checked the performance impact of this change?

storage/statedb/cache_hybrid_test.go
aidan-kwon: Changed it to 100 MB

storage/statedb/cache_redis.go
aidan-kwon: Changed to `Warn`.

aidan-kwon:
> @aidan-kwon Have you checked the performance impact of this change?

@ehnuje I didn't test the full network test with this change but tested `(cache *RedisCache) Set` method and KES service node. When the cache set time asynchronously to the local Redis server, it decreased the writing time compared to the synchronous write (The exact performance gain is different depends on the network latency). 
Could you tell me in detail which point do you worried about?

ehnuje:
> > @aidan-kwon Have you checked the performance impact of this change?
> 
> @ehnuje I didn't test the full network test with this change but tested `(cache *RedisCache) Set` method and KES service node. When the cache set time asynchronously to the local Redis server, it decreased the writing time compared to the synchronous write (The exact performance gain is different depends on the network latency).
> Could you tell me in detail which point do you worried about?

@aidan-kwon I was not worried but wanted to check the performance gain since the purpose of this change is to improve the performance. If it does not show improvement, it might only increase the number of goroutines.

ehnuje (approved):

aidan-kwon:
> @aidan-kwon I was not worried but wanted to check the performance gain since the purpose of this change is to improve the performance. If it does not show improvement, it might only increase the number of goroutines.

Oh, I understand the point. This change is not for the ordinary environment, but for the unstable network environment. Some people worried that synchronous cache writing could delay block processing with an unstable network since the write timeout takes a few seconds. With this change, block processing performance is less affected by the Redis server's network problem.

yoomee1313 (dismissed):

winnie-byun (dismissed):

jeongkyun-oh (approved):

KimKyungup (dismissed):

aidan-kwon:
Reviewers, I resolved conflicts. PTAL

jeongkyun-oh (approved):

KimKyungup (approved):

ehnuje (approved):

yoomee1313 (approved):

PR #830, opened 2020-12-29
blockchain, state: refactored state warm-up related codes

ehnuje:
## Proposed changes

- This PR is to refactor state warm-up related codes before introducing state trie warm-up
- Most of logic `blockchain.GetContractStorageRoot` is moved into `StateDB.GetContractStorageRoot`
- As we don't use hash value from `resultHashCh`, it is changed to `resultCh`, which type is `chan struct{}`
- `finishCh` and `resultErrCh` are used for the same channel, so changed it to `errCh` for consistency
- To clearly represent what it does in the go-routine in  `StartWarmUp` function, codes which does not affect the performance critically are pulled out.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/state_migration.go
+++ b/blockchain/state_migration.go
@@ -24,10 +24,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/klaytn/klaytn/blockchain/types"
-	"github.com/klaytn/klaytn/blockchain/types/account"
-
 	"github.com/VictoriaMetrics/fastcache"
+	"github.com/klaytn/klaytn/blockchain/types"
 
 	"github.com/alecthomas/units"
 	lru "github.com/hashicorp/golang-lru"
@@ -385,9 +383,9 @@ func (bc *BlockChain) StateMigrationStatus() (bool, uint64, int, int, int, float
 	return bc.db.InMigration(), bc.db.MigrationBlockNumber(), bc.readCnt, bc.committedCnt, bc.pendingCnt, bc.progress, bc.migrationErr
 }
 
-func (bc *BlockChain) concurrentIterateTrie(root common.Hash, db state.Database, resultCh chan common.Hash, finishCh chan error) (resultErr error) {
+func (bc *BlockChain) concurrentIterateTrie(root common.Hash, db state.Database, resultCh chan struct{}, errCh chan error) (resultErr error) {
 	defer func() {
-		finishCh <- resultErr
+		errCh <- resultErr
 	}()
 
 	stateDB, err := state.New(root, db)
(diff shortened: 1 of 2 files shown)


blockchain/state_migration.go
```diff
@@ -491,31 +489,24 @@ func (bc *BlockChain) StartWarmUp() error {
...
-
-		resultHashCh := make(chan common.Hash, 10000)
-		resultErrCh := make(chan error)
-
-		for _, child := range children {
-			go bc.concurrentIterateTrie(child, db, resultHashCh, resultErrCh)
-		}
+	resultCh := make(chan struct{}, 10000)
+	errCh := make(chan error)
+	for _, child := range children {
```
KimKyungup, line 502: @ehnuje 
These early children are almost 16. So 16 goroutines are created.
I think this number is quite large. 
So could you consider priority queue in next time like `sync` in sync.go.

```suggestion
	for _, child := range children {
```

ehnuje: @KimKyungup If we're running warming-up routines, doesn't it mean that the performance is heavily degraded by the state trie or the given storage trie? In that case, I think it would be better to concentrate the resources to warming-up routine, though it might affect other operations. How do you think?

KimKyungup: @ehnuje 
When I think about it again, depending on the instance, it seems that 16 goroutines can be too much or too little. It is not easy to change to a queue right away, so let's proceed this way as before.

KimKyungup (approved):

blockchain/state_migration.go
```diff
@@ -491,31 +489,24 @@ func (bc *BlockChain) StartWarmUp() error {
...
+		go bc.concurrentIterateTrie(child, db, resultCh, errCh)
+	}
 
+	go func() {
```
yoomee1313, line 506: `for _, child := range` for loop is taken out from this go func. Is it ok?

winnie-byun (approved):

blockchain/state_migration.go
ehnuje: @yoomee1313 Yes, above code(=L#502~L#504) is to run child iterators with different go-routines and below is to receive result from those child go-routines and check if the task is done or not.

aidan-kwon (approved):

blockchain/state_migration.go
ehnuje: @yoomee1313 As it is approved by 3 reviewers, let me merge this change for now. Please let me know if you have any issues 😄

PR #832, opened 2020-12-30
blockchain: first implementation of state prefetcher

ehnuje:
## Proposed changes

- Applied go-ethereum change https://github.com/ethereum/go-ethereum/pull/19328
- With this PR, when inserting a list of blocks, next block will be executed against current state by Prefetcher

## Cypress sync test results

- EN0 = StatePrefetcher ON / EN1 = StatePrefetcher OFF
- EN0 shows faster progress (15~25%), lower block processing time and lower cache miss

### Head Block Number

screencapture-3-36-12-130-3000-d-KnJhY-iWz-klaytn-main-overview-2021-01-08-13_55_51 (1)

### Block Processing Time

screencapture-3-36-12-130-3000-d-KnJhY-iWz-klaytn-main-overview-2021-01-07-09_24_52 (1)

### Trie Node Cache Miss

screencapture-3-36-12-130-3000-d-BjUoOqkWz-klaytn-data-overview-2021-01-07-09_23_59 (1)

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #837, opened 2021-01-07
[FOR REVIEW, DO NOT MERGE] Parallelized prefetching a list of blocks

ehnuje:
## Proposed changes

- Describe your changes to communicate to the maintainers why we should accept this pull request.
- If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

- This is just a review purpose, DO NOT MERGE THIS
- This PR is made based on https://github.com/klaytn/klaytn/pull/832

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -194,8 +194,13 @@ type BlockChain struct {
 	lastCommittedBlock uint64
 	quitWarmUp         chan struct{}
 
-	prefetchCh        chan *types.Block
+	prefetchCh        chan prefetchItem
 	followupInterrupt uint64
+}
+
+type prefetchItem struct {
+	prefetchBlock *types.Block
+	parentRoot    common.Hash
 }
 
 // NewBlockChain returns a fully initialised block chain using information
@@ -235,7 +240,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 		badBlocks:          badBlocks,
 		parallelDBWrite:    db.IsParallelDBWrite(),
 		stopStateMigration: make(chan struct{}),
-		prefetchCh:         make(chan *types.Block, 2048),
+		prefetchCh:         make(chan prefetchItem, 2048),
 	}
 
 	bc.validator = NewBlockValidator(chainConfig, bc, engine)
@@ -285,10 +290,11 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 }
 
 func (bc *BlockChain) prefetchWorker() {
-	for followup := range bc.prefetchCh {
+	logger.Info("prefetchWorker started")
+	for item := range bc.prefetchCh {
 		start := time.Now()
-		throwaway, _ := state.New(bc.CurrentBlock().Root(), bc.stateCache)
-		bc.prefetcher.Prefetch(followup, throwaway, bc.vmConfig, &bc.followupInterrupt)
+		throwaway, _ := state.New(item.parentRoot, bc.stateCache)
+		bc.prefetcher.Prefetch(item.prefetchBlock, throwaway, bc.vmConfig, &bc.followupInterrupt)
 		blockPrefetchExecuteTimer.Update(time.Since(start))
 	}
 }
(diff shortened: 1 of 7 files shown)


blockchain/blockchain.go
```diff
@@ -1632,31 +1638,47 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
...
+				}
+				logger.Info("sent blocks to prefetch channel", "numBlocks", len(chain)-1)
+			}
+		}
 
 		// Process block using the parent state as reference point.
 		receipts, logs, usedGas, internalTxTraces, procStats, err := bc.processor.Process(block, stateDB, bc.vmConfig)
 		if err != nil {
 			bc.reportBlock(block, receipts, err)
+			atomic.StoreUint64(&bc.followupInterrupt, 1)
```
KimKyungup, line 1658: @ehnuje 
This code stop the prefetching future block after every one block.
I think this makes very frequent context switch.
So I suggest to start prefetching all blocks(chain) before `for i, block := range chain {` like `senderCacher()`.
Also we can stop the fetch goroutine after this loop.

I would like you to test this fetch logic also.

KimKyungup:
@aidan-kwon @winnie-byun @jeongkyun-oh 
If you have time, please review this prefetch logic and share your opinion.

@ehnuje For reviews, please explain shortly about this prefetch.

aidan-kwon:
I prefer pre-fetching just one block because of the following reasons. 
1. Prefetcher goroutines usually work in synchronization mode. When a node has the latest state, the goroutines do nothing in the common environment but the goroutines are maintained. I think it wastes node resources. 
2. It is a trade-off between CPU+Memory and DB read. Prefetcher sacrifices CPU and Memory to read DB data faster. However, I cannot sure Klaytn nodes (especially ENs) has enough computation power and memory to process parallel block insert safely. Nodes already have hundreds of goroutines and large size caches.

blockchain/blockchain.go
ehnuje: As this PR showed worse performance than only fetching next block, I'll just update this PR and compare its performance to the original prefetcher.

ehnuje: Please check 21fbc537c66f705af24f5c2b615c378c3c6bebb8

ehnuje:
@aidan-kwon Thanks for your comments. Please check my comments and share your opinion 😄 

> I prefer pre-fetching just one block because of the following reasons.
> 
> 1. Prefetcher goroutines usually work in synchronization mode. When a node has the latest state, the goroutines do nothing in the common environment but the goroutines are maintained. I think it wastes node resources.

Do you know the performance impact or the resource usage of goroutine that does nothing? If it really degrades the performance, maybe we can think about creating and deleting goroutine only if downloader works.

> 2. It is a trade-off between CPU+Memory and DB read. Prefetcher sacrifices CPU and Memory to read DB data faster. However, I cannot sure Klaytn nodes (especially ENs) has enough computation power and memory to process parallel block insert safely. Nodes already have hundreds of goroutines and large size caches.

Regarding EN, I'm not sure how much we should care EN. The block itself is already generated from the high-end instances. Also, as far as I know, Klaytn node does not utilize CPU since block processing is a serial job. If the node serves many requests, using additional CPU might be a problem, but I think using a bit more CPU won't be a problem in most cases.

blockchain/state_prefetcher.go
```diff
@@ -0,0 +1,91 @@
...
+func (p *statePrefetcher) Prefetch(block *types.Block, stateDB *state.StateDB, cfg vm.Config, interrupt *uint64) {
+	var (
+		header = block.Header()
+	)
+	// Iterate over and process the individual transactions
+	for i, tx := range block.Transactions() {
+		// If block precaching was interrupted, abort
+		if interrupt != nil {
+			val := atomic.LoadUint64(interrupt)
+			if val == 1 {
```
jeongkyun-oh, line 64: Isn't it possible to set `interrupt` to 0 in `insertChain` before this method returns?

winnie-byun:
About go routine and resources:
1. The main resource go routine is using is memory usage. (4536bytes are used per routine in go 1.6.2 [link](https://stackoverflow.com/questions/8509152/max-number-of-goroutines#:~:text=On%20a%20machine%20with%204,slightly%20less%20than%201%20million.&text=It%20is%20practical%20to%20create,in%20the%20same%20address%20space.)) 
2. If a go routine is blocked by channel, the routine is only started when there is a signal. (Checkout `Don't interrupt!` in [this article](https://codeburst.io/why-goroutines-are-not-lightweight-threads-7c460c1f155f)) 
3. The uploaded data is stored in state trie cache. If the cache size is in concern, we could resize the cache.
@aidan-kwon @ehnuje

PR #867, opened 2021-01-19
statedb: background loading of FastCache

ehnuje:
## Proposed changes

- Klaytn has codes, which restart the node to avoid continuous round changing.
- To make this work, CN nodes should be started around the same time
- However, as the time used to load saved fastcache can be different by the nodes, this might not work
- With this PR, saved fastache is loaded in the background and will be used after when it is loaded successfully

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/storage/statedb/cache_fastcache.go
+++ b/storage/statedb/cache_fastcache.go
@@ -17,6 +17,9 @@
 package statedb
 
 import (
+	"fmt"
+	"time"
+
 	"github.com/VictoriaMetrics/fastcache"
 	"github.com/rcrowley/go-metrics"
 )
@@ -36,8 +39,15 @@ var (
 	memcacheFastInvalidValueHashErrors = metrics.NewRegisteredGauge("trie/memcache/fast/error/invalid/hash", nil)
 )
 
+const (
+	useNew = iota
+	useBoth
+	useOld
+)
+
 type FastCache struct {
-	fast *fastcache.Cache
+	new *fastcache.Cache // newly created, empty one
+	old *fastcache.Cache // created from saved fastcache, if exists
 }
 
 // NewFastCache creates a FastCache with given cache size.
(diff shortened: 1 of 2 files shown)


storage/statedb/cache_fastcache.go
```diff
@@ -52,30 +62,64 @@ func NewFastCache(config *TrieNodeCacheConfig) TrieNodeCache {
...
 
-	fc := &FastCache{fast: fastcache.LoadFromFileOrNew(config.FastCacheFileDir, config.LocalCacheSizeMB*1024*1024)} // Convert MB to Byte
-	stats := fc.UpdateStats().(fastcache.Stats)
+	fc := &FastCache{new: fastcache.New(config.LocalCacheSizeMB * 1024 * 1024)} // Convert MB to Byte
+	go fc.loadFromBackground(config)
 
-	logger.Info("Initialize local trie node cache (fastCache)",
-		"MaxMB", config.LocalCacheSizeMB, "FilePath", config.FastCacheFileDir,
-		"LoadedBytes", stats.BytesSize, "LoadedEntries", stats.EntriesCount)
+	fmt.Println("Initialize local trie node cache (fastCache)",
```
winnie-byun, line 68: Could you use logger instead?

storage/statedb/cache_fastcache.go
```diff
@@ -36,8 +39,15 @@ var (
 	memcacheFastInvalidValueHashErrors = metrics.NewRegisteredGauge("trie/memcache/fast/error/invalid/hash", nil)
 )
 
+const (
+	useNew = iota
+	useBoth
+	useOld
+)
```
winnie-byun, line 46: There are some unused values.

storage/statedb/cache_fastcache.go
```diff
@@ -52,30 +62,64 @@ func NewFastCache(config *TrieNodeCacheConfig) TrieNodeCache {
...
+	// if there is no saved cache, use only new cache
+	if stats.EntriesCount == 0 {
+		logger.Info("There is no saved cache in the fastcache directory",
+			"FilePath", config.FastCacheFileDir)
+		cache.old = nil
+		return
+	}
+	// if we loaded the saved cache successfully, reset new cache to reclaim the memory
+	cache.new.Reset()
+	fmt.Println("Finished loading fastcache from background", "elapsed", time.Since(start),
```
winnie-byun, line 86: Could you use logger instead?

storage/statedb/cache_fastcache.go
```diff
@@ -52,30 +62,64 @@ func NewFastCache(config *TrieNodeCacheConfig) TrieNodeCache {
...
-		"MaxMB", config.LocalCacheSizeMB, "FilePath", config.FastCacheFileDir,
-		"LoadedBytes", stats.BytesSize, "LoadedEntries", stats.EntriesCount)
+	fmt.Println("Initialize local trie node cache (fastCache)",
+		"MaxMB", config.LocalCacheSizeMB, "FilePath", config.FastCacheFileDir)
 	return fc
 }
 
+func (cache *FastCache) loadFromBackground(config *TrieNodeCacheConfig) {
+	start := time.Now()
+	cache.old = fastcache.LoadFromFileOrNew(config.FastCacheFileDir, config.LocalCacheSizeMB*1024*1024)
```
winnie-byun, line 75: The caches use too much memory and there could be an OOM during upload. What do you think of resizing the cache?

winnie-byun (commented):
1. When there is a saved cache file, `FastCache.old` is the only used. All `Set` and `Get` is done at `old` cache. There is no need for `new` cache.
2. Could you elaborate about the connection between the restart logic and fastcache loading?

ehnuje:
> 1. When there is a saved cache file, `FastCache.old` is the only used. All `Set` and `Get` is done at `old` cache. There is no need for `new` cache.

You're right. But please note that old cache loading is done by background goroutine, not by main goroutine. Therefore, new cache is needed if we don't want to direct access to LevelDB before loading fastcache.

> 2. Could you elaborate about the connection between the restart logic and fastcache loading?

As the node waits until it finishes loading fastcache from file, which takes around 30 seconds, but it may be different across the nodes and this can cause CN nodes to start their first round in different time, which we want to avoid. Because of this, background fastcache loading is needed to decrease the possibility of different node start time.

storage/statedb/cache_fastcache.go
```diff
@@ -52,30 +62,64 @@ func NewFastCache(config *TrieNodeCacheConfig) TrieNodeCache {
...
+		if data := cache.old.Get(nil, k); data != nil {
+			return data
+		}
+	}
+	return cache.new.Get(nil, k)
 }
 
 func (cache *FastCache) Set(k, v []byte) {
-	cache.fast.Set(k, v)
+	if cache.old != nil {
```
ehnuje, line 100: `cache.old` is used to check if background loading has finished or not and can be accessed from different goroutines concurrently.

storage/statedb/cache_fastcache.go
ehnuje: @winnie-byun If we resize the cache, we can't load the stored fastcache since it checks if the current configuration is identical with the previous, stored one.

ehnuje, line 75: And setting `cache.old` is done here, which is also processed in another goroutine. Therefore, read and write of `cache.old` is done in concurrent manner and seems to be a definite data race condition. 

> Unsynchronized, concurrent access to any variable from multiple goroutines where at least one of them is a write is undefined behavior by The Go Memory Model.

https://stackoverflow.com/a/41407827
https://golang.org/ref/mem

KimKyungup:
As @winnie-byun said, this PR has memory problem and race condition. 
I think we need to consider other approach. @ehnuje

ehnuje:
> As @winnie-byun said, this PR has memory problem and race condition.
> I think we need to consider other approach. @ehnuje

Right. Let me close this PR and consider other options.

PR #884, opened 2021-02-01
Added retry max logic after which the chaindata fetcher is stopped

jeongkyun-oh:
## Proposed changes

- Added stop logic when retry is exceeded over max value.
- Used atomic value instead of normal value (it may occur closing channel several times)

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/datasync/chaindatafetcher/chaindata_fetcher.go
+++ b/datasync/chaindatafetcher/chaindata_fetcher.go
@@ -396,7 +396,7 @@ func (f *ChainDataFetcher) resetRequestCh() {
 	}
 }
 
-func (f *ChainDataFetcher) forceStop() {
+func (f *ChainDataFetcher) pause() {
 	f.stopFetching()
 	f.stopRangeFetching()
 	f.resetChainCh()
@@ -424,7 +424,8 @@ func (f *ChainDataFetcher) handleRequest() {
 			}
 
 			if err != nil && err == errMaxRetryExceeded {
-				f.forceStop()
+				logger.Error("the chaindatafetcher reaches the maximum retries. it pauses fetching and clear the channels", "blockNum", ev.Block.NumberU64())
+				f.pause()
 			}
 		case req := <-f.reqCh:
 			numRequestsGauge.Update(int64(len(f.reqCh)))
@@ -436,7 +437,8 @@ func (f *ChainDataFetcher) handleRequest() {
 			}
 			err = f.handleRequestByType(req.ReqType, req.ShouldUpdateCheckpoint, ev)
 			if err != nil && err == errMaxRetryExceeded {
-				f.forceStop()
+				logger.Error("the chaindatafetcher reaches the maximum retries. it pauses fetching and clear the channels", "blockNum", ev.Block.NumberU64())
+				f.pause()
 			}
 		}
 	}
--- a/datasync/chaindatafetcher/chaindata_fetcher.go
+++ b/datasync/chaindatafetcher/chaindata_fetcher.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/klaytn/klaytn/blockchain"
@@ -39,8 +40,14 @@ import (
 	"github.com/rcrowley/go-metrics"
 )
 
+const (
+	stopped = uint32(iota)
+	running
+)
+
 var logger = log.NewModuleLogger(log.ChainDataFetcher)
 var errUnsupportedMode = errors.New("the given chaindatafetcher mode is not supported")
+var errMaxRetryExceeded = errors.New("the number of retries is exceeded over max")
 
 //go:generate mockgen -destination=./mocks/blockchain_mock.go -package=mocks github.com/klaytn/klaytn/datasync/chaindatafetcher BlockChain
 type BlockChain interface {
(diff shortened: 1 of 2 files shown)


datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -370,6 +373,34 @@ func (f *ChainDataFetcher) handleRequestByType(reqType cfTypes.RequestType, shou
...
+
+func (f *ChainDataFetcher) resetRequestCh() {
+	for {
+		select {
+		case <-f.reqCh:
+		default:
+			return
+		}
+	}
+}
```
ehnuje, line 397: Why don't you just assign a new channel to `f.chain` and `f.reqCh`?

datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -398,7 +434,10 @@ func (f *ChainDataFetcher) handleRequest() {
 				logger.Error("making chain event is failed", "err", err)
 				break
 			}
-			f.handleRequestByType(req.ReqType, req.ShouldUpdateCheckpoint, ev)
+			err = f.handleRequestByType(req.ReqType, req.ShouldUpdateCheckpoint, ev)
+			if err != nil && err == errMaxRetryExceeded {
+				f.forceStop()
```
ehnuje, line 439: Shouldn't we return here to terminate `handleRequest`?
```suggestion
				f.forceStop()
				return
```

datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -382,14 +413,19 @@ func (f *ChainDataFetcher) handleRequest() {
...
+				err = f.handleRequestByType(cfTypes.RequestTypeAll, true, ev)
 			case ModeKafka:
-				f.handleRequestByType(cfTypes.RequestTypeGroupAll, true, ev)
+				err = f.handleRequestByType(cfTypes.RequestTypeGroupAll, true, ev)
 			default:
 				logger.Error("the chaindatafetcher mode is not supported", "mode", f.config.Mode, "blockNumber", ev.Block.NumberU64())
 			}
+
+			if err != nil && err == errMaxRetryExceeded {
+				f.forceStop()
```
ehnuje, line 427: Shouldn't we return here to terminate `handleRequest`?
```suggestion
				f.forceStop()
				return
```

datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -370,6 +373,34 @@ func (f *ChainDataFetcher) handleRequestByType(reqType cfTypes.RequestType, shou
...
+	for {
+		select {
+		case <-f.reqCh:
+		default:
+			return
+		}
+	}
+}
+
+func (f *ChainDataFetcher) forceStop() {
```
ehnuje, line 399: What would be the difference between `forceStop` and `Stop`? Why do we need another stop function?

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: Not really. The goroutines are not terminated until the chaindatafetcher is stopped. Here, only fetching/rangeFetching is stopped and it can rerun in the future using API.

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: `Stop` stops chaindatafetcher. I want to make only fetching/rangeFetching jobs to be stopped and clear the channels.

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: Because I want to reuse those channels when the fetching jobs rerun.

datasync/chaindatafetcher/chaindata_fetcher.go
ehnuje: In that purpose, `pause` would be more appropriate. `Stop` terminates ChainDataFetcher, but `forceStop` doesn't. `forceStop` sounds like it will `Stop` anyway though it might be ungraceful.

datasync/chaindatafetcher/chaindata_fetcher.go
ehnuje: > Here, only fetching/rangeFetching is stopped and it can rerun in the future using API.

Thanks for your explanation. I think this should be printed as an error log inside `forceStop`.

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: I renamed `forceStop` to `pause`. Thanks.

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: I think the log messages, but outside of `pause`. PTAL @ehnuje.

ehnuje (approved):

aidan-kwon (approved):
I think `InsertMaxRetry` is too big. How about 10?
Except that, LGTM

jeongkyun-oh:
> I think `InsertMaxRetry` is too big. How about 10?
> Except that, LGTM

@aidan-kwon I think it has 50 blocks buffer at least (it can be increased using flags), so I thought 50 seconds can be fine (for kafka self-recovering). What do you think?

ehnuje:
@jeongkyun-oh Don't we merge this change?

jeongkyun-oh:
@ehnuje It should be. Please take another look when you have a moment @aidan-kwon @KimKyungup.

aidan-kwon:
> @aidan-kwon I think it has 50 blocks buffer at least (it can be increased using flags), so I thought 50 seconds can be fine (for kafka self-recovering). What do you think?

@jeongkyun-oh Okay about it. I was afraid that 50 seconds is too long for KAS SLA, but we can consider it later when KES is really integrated into KAS.

KimKyungup (approved):

PR #888, opened 2021-02-08
Add a condition sending a COMMIT message in handleCommit

aidan-kwon:
## Proposed changes

When CNs try to have a consensus on the hashed-locked block, they may fail to send a `commit` message. Non-hash-locked nodes need to change their state to `prepared` sending a `commit` message when they received more than 2f+1 `prepare` or `commit` messages. But, they don't change their state in the current logic when only `commit` messages are delivered after 2f messages since the state changing logic exists in `prepare` message handling logic. 

With this change, committee nodes can change their state to `prepared` when they received 2f+1 agreement regardless of the message order. 

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

- This bug prolonged consensus time triggering round change more frequently for hash-locked proposals in Cypress network.

### DIFF
--- a/consensus/istanbul/core/commit.go
+++ b/consensus/istanbul/core/commit.go
@@ -119,7 +119,7 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 	// If we already have a proposal, we may have chance to speed up the consensus process
 	// by committing the proposal without PREPARE messages.
 	//logger.Error("### consensus check","len(commits)",c.current.Commits.Size(),"f(2/3)",2*c.valSet.F(),"state",c.state.Cmp(StateCommitted))
-	if c.current.Commits.Size() > 2*c.valSet.F() && c.state.Cmp(StateCommitted) < 0 {
+	if c.state.Cmp(StateCommitted) < 0 && c.current.Commits.Size() > 2*c.valSet.F() {
 		// Still need to call LockHash here since state can skip Prepared state and jump directly to the Committed state.
 		c.current.LockHash()
 		c.commit()
--- a/consensus/istanbul/core/commit.go
+++ b/consensus/istanbul/core/commit.go
@@ -21,9 +21,10 @@
 package core
 
 import (
+	"reflect"
+
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/consensus/istanbul"
-	"reflect"
 )
 
 func (c *core) sendCommit() {
@@ -95,6 +96,16 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 	}
 
 	c.acceptCommit(msg, src)
+
+	// Change to Prepared state if we've received enough PREPARE or COMMIT messages and we are in earlier state
+	// before Prepared state.
+	// Both of PREPARE and COMMIT messages are counted since hashlocked nodes in the previous round skip sending
+	// PREPARE messages.
+	if c.current.GetPrepareOrCommitSize() > 2*c.valSet.F() && c.state.Cmp(StatePrepared) < 0 {
+		c.current.LockHash()
+		c.setState(StatePrepared)
+		c.sendCommit()
+	}
 
 	// Commit the proposal once we have enough COMMIT messages and we are not in the Committed state.
 	//
--- a/consensus/istanbul/core/commit.go
+++ b/consensus/istanbul/core/commit.go
@@ -97,14 +97,21 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 
 	c.acceptCommit(msg, src)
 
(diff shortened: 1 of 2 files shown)


consensus/istanbul/core/commit.go
```diff
@@ -96,6 +97,16 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 
 	c.acceptCommit(msg, src)
 
+	// Change to Prepared state if we've received enough PREPARE or COMMIT messages and we are in earlier state
+	// before Prepared state.
+	// Both of PREPARE and COMMIT messages are counted since hashlocked nodes in the previous round skip sending
+	// PREPARE messages.
+	if c.current.GetPrepareOrCommitSize() > 2*c.valSet.F() && c.state.Cmp(StatePrepared) < 0 {
```
jeongkyun-oh, line 104: Do you have any reason why you skip the following condition `(c.current.IsHashLocked() && prepare.Digest == c.current.GetLockedHash())` ?

aidan-kwon: If a node is hashlocked, it will do the same things when it received a `preprepare` message.

ehnuje: I also wonder why it does not check the digest from the message, like it does in `handlePrepare`, like @jeongkyun-oh asked.

BTW, `LockHash` sounds a bit misleading since it just atomically updates the value of `roundState.lockedHash`. It just avoids concurrent read/write of the variable, not working as a usual lock.

KimKyungup: I think it is very rare for hashlocked node not to receive prepare msg.
So most hashlocked node can change its state to commit state by any prepare msg.
However, I think If there is a hashlocked node doesn't receive any prepare msg, this condition let it change the state to commit state by any commit msg.

consensus/istanbul/core/commit.go
```diff
@@ -96,6 +97,16 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
...
+		c.current.LockHash()
+		c.setState(StatePrepared)
```
yoomee1313, line 106: If 2f+1 commits arrive first, shouldn't the committed also be set? Is it set in other logic?

consensus/istanbul/core/commit.go
aidan-kwon: @jeongkyun-oh @ehnuje @KimKyungup As you mentioned, I added another condition for hash locked nodes. Thank you for letting me know.

consensus/istanbul/core/commit.go
aidan-kwon: @yoomee1313 If 2f+1 commits arrived, the node will commit the proposal in the next lines. It means that the state will be changed to `StateCommitted`. 
https://github.com/klaytn/klaytn/blob/604c761dc94c7d16bb0a2d0867e24435187c292d/consensus/istanbul/core/commit.go#L122-L126
Do I understand your question correctly?

yoomee1313: Oh there it is, thank you!

jeongkyun-oh (approved):

KimKyungup (approved):

KimKyungup:
@alanchchen
@bailantaotao
@markya0616
@tailingchen
@yutelin

ehnuje (approved):

yoomee1313 (approved):

alanchchen:
> @alanchchen @bailantaotao @markya0616 @tailingchen @yutelin
> 
> Hello. I invited you to ask for your opinion on this PR.
> 
> Seeing @yutelin writing https://github.com/ethereum/EIPs/issues/650, I know you developed Quorum IBFT implementation.
> 
> 
> 
> First of all, briefly speaking about our Klaytn blockchain.
> 
> It is based on the Geth and Quorum IBFT. Then, we optimized for 1 second block time and added several feature like account/transaction type, fee delegation, etc. (For more information, visit https://docs.klaytn.com/klaytn, https://scope.klaytn.com/)
> 
> 
> 
> Recently, the Klaytn network has some round change problems. And we investigated and concluded that the IBFT code misses some code like this PR. And also we made a issue/PR(https://github.com/ConsenSys/quorum/issues/1133, https://github.com/ConsenSys/quorum/pull/1134) to Quorum.
> 
> 
> 
> So I think it's the most accurate and honorable to get a review as the original developer.
> 
> If possible, we would appreciate your feedback via this PR or the email below.
> 
> 
> 
> ethan.kim@groundx.xyz
> 
> 
> 
> 

Hi, it's honored to see that Klaytn chooses IBFT. Unfortunately our team haven't continued IBFT development for a long time. I suggest you contact the Quorum team.

BTW, Hyperledger Besu have implemented IBFT 2.0. It might be helpful.

Thank you 😀

PR #891, opened 2021-02-10
Fix invalid validator list in extraData of block header

KimKyungup:
## Proposed changes

This PR fixes the invalid committee list in block header.
Before this PR, the committee is derived with old proposer. So the list was invalid.

This validator list is used for calculating a hash of block.
So it is not easy to save a right committee list of each round.

Therefore, like an original code, I reverted it to save all validator list in the field.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -281,10 +281,7 @@ func (api *APIExtension) getConsensusInfo(block *types.Block) (ConsensusInfo, er
 
 	// get origin proposer at 0 round.
 	originProposer := common.Address{}
-	lastProposer := common.Address{}
-	if blockNumber > 0 {
-		lastProposer = api.istanbul.GetProposer(blockNumber - 1)
-	}
+	lastProposer := api.istanbul.GetProposer(blockNumber - 1)
 
 	newValSet := snap.ValSet.Copy()
 	newValSet.CalcProposer(lastProposer, 0)
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -161,9 +161,19 @@ func (api *APIExtension) GetCouncil(number *rpc.BlockNumber) ([]common.Address, 
 	}
 	// Ensure we have an actually valid block and return the council from its snapshot
 	if header == nil {
-		logger.Trace("Failed to find the requested block", "number", number)
 		return nil, errNoBlockExist // return nil if block is not found.
 	}
+
+	// Since v1.6.1, extra.validators represents a list of council
+	// TODO-Klaytn : replace this to the below calculation logic
+	//istanbulExtra, err := types.ExtractIstanbulExtra(header)
+	//if err == nil {
+	//	return istanbulExtra.Validators, nil
+	//} else {
+	//	return nil, errExtractIstanbulExtra
+	//}
+
+	// Calculate council list from snapshot
 	snap, err := api.istanbul.snapshot(api.chain, header.Number.Uint64(), header.Hash(), nil)
 	if err != nil {
 		logger.Error("Failed to get snapshot.", "hash", header.Hash(), "err", err)
@@ -197,12 +207,34 @@ func (api *APIExtension) GetCommittee(number *rpc.BlockNumber) ([]common.Address
 		return nil, errNoBlockExist
 	}
 
-	istanbulExtra, err := types.ExtractIstanbulExtra(header)
-	if err == nil {
-		return istanbulExtra.Validators, nil
-	} else {
-		return nil, errExtractIstanbulExtra
+	blockNumber := header.Number.Uint64()
+	round := header.Round()
+	view := &istanbul.View{
+		Sequence: new(big.Int).SetUint64(blockNumber),
+		Round:    new(big.Int).SetUint64(uint64(round)),
 	}
+
+	// get the proposer of this block.
+	proposer, err := ecrecover(header)
+	if err != nil {
+		return nil, err
+	}
+
+	// get the snapshot of the previous block.
+	parentHash := header.ParentHash
+	snap, err := api.istanbul.snapshot(api.chain, blockNumber-1, parentHash, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// get the committee list of this block at the view (blockNumber, round)
+	committee := snap.ValSet.SubListWithProposer(parentHash, proposer, view)
+	addresses := make([]common.Address, len(committee))
+	for i, v := range committee {
+		addresses[i] = v.Address()
+	}
+
+	return addresses, nil
 }
 
 func (api *APIExtension) GetCommitteeSize(number *rpc.BlockNumber) (int, error) {
@@ -248,17 +280,12 @@ func (api *APIExtension) getConsensusInfo(block *types.Block) (ConsensusInfo, er
 
 	// get origin proposer at 0 round.
 	originProposer := common.Address{}
-	lastProposer := common.Address{}
-	// TODO-Klaytn add the logic to get the last proposer for other policies. Weighted Random doesn't need it.
-	if istanbul.ProposerPolicy(snap.Policy) == istanbul.WeightedRandom {
-		newValSet := snap.ValSet.Copy()
-		newValSet.CalcProposer(lastProposer, 0)
-		originProposer = newValSet.GetProposer().Address()
-	} else {
-		logger.Warn("getConsensusInfo does not support to get origin proposal on", "proposerPolicy", snap.Policy)
-	}
+	_, lastProposer := api.istanbul.LastProposal()
+	newValSet := snap.ValSet.Copy()
+	newValSet.CalcProposer(lastProposer, 0)
+	originProposer = newValSet.GetProposer().Address()
 
-	// get the committee list of this block.
+	// get the committee list of this block at the view (blockNumber, round)
 	committee := snap.ValSet.SubListWithProposer(parentHash, proposer, view)
 	committeeAddrs := make([]common.Address, len(committee))
 	for i, v := range committee {
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -23,6 +23,9 @@ package backend
 import (
 	"errors"
 	"fmt"
+	"math/big"
+	"reflect"
+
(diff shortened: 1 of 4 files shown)


work/worker.go
```diff
@@ -186,9 +184,6 @@ func newWorker(config *params.ChainConfig, engine consensus.Engine, rewardbase c
 	go worker.update()
 
 	go worker.wait(TxResendUseLegacy)
-	worker.commitNewWork()
```
KimKyungup, line 189: This line is removed not to call commitNewWork before calling`woker.start()`.
In that case, there is nil `chain`. It causes `sb.GetProposer` access nil `chain` to get last block. 

`worker.commitNewWork()` is called by many cases like below. So we can remove this.
image

tests/klay_test_blockchain_test.go
```diff
@@ -120,6 +120,8 @@ func NewBCData(maxAccounts, numValidators int) (*BCData, error) {
 		return nil, err
 	}
 
+	engine.Start(bc, bc.CurrentBlock, bc.HasBadBlock)
```
KimKyungup, line 123: This line is added to set the `chain` of the `engine`.

work/worker.go
ehnuje: I understood that `sb.chain` can be `nil`, however, why it wasn't a problem before?

consensus/istanbul/backend/engine.go
```diff
@@ -324,7 +324,10 @@ func (sb *backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 	header.Vote = sb.governance.GetEncodedVote(sb.address, number)
 
 	// add validators in snapshot to extraData's validators section
-	extra, err := prepareExtra(header, snap.committee(header.ParentHash, sb.currentView.Load().(*istanbul.View)))
+	currentView := sb.currentView.Load().(*istanbul.View)
+	lastProposer := sb.GetProposer(header.Number.Uint64() - 1)
+	snap.ValSet.CalcProposer(lastProposer, currentView.Round.Uint64())
+	extra, err := prepareExtra(header, snap.committee(header.ParentHash, currentView))
```
ehnuje, line 330: Can we make a test for this?

jeongkyun-oh, line 330: @KimKyungup It looks that the snapshot should contain the validators with a proposer on the specific block. How about calculating the proposer in the `apply` method of snapshot?

KimKyungup:
I found this change can save the exact committee list for only round 0.
For other round, every new start round should calculate new committee and save it to header.
However, I noticed that a hash value of a block can be changed by modifying the committee list.
The committee list data is included to the data for hashing.

So I will consider saving all validator list to it.

work/worker.go
KimKyungup: There was not a code (like `Prepare`) to use chain before setting `sb.chain`.
So It was not a problem. But calling `worker.commitNewWork()`  before `worker.start()` is weird.

KimKyungup:
PTAL. :)

consensus/istanbul/backend/api.go
```diff
@@ -248,17 +280,12 @@ func (api *APIExtension) getConsensusInfo(block *types.Block) (ConsensusInfo, er
...
-	lastProposer := common.Address{}
-	// TODO-Klaytn add the logic to get the last proposer for other policies. Weighted Random doesn't need it.
-	if istanbul.ProposerPolicy(snap.Policy) == istanbul.WeightedRandom {
-		newValSet := snap.ValSet.Copy()
-		newValSet.CalcProposer(lastProposer, 0)
-		originProposer = newValSet.GetProposer().Address()
-	} else {
-		logger.Warn("getConsensusInfo does not support to get origin proposal on", "proposerPolicy", snap.Policy)
-	}
+	_, lastProposer := api.istanbul.LastProposal()
```
ehnuje, line 283: Isn't `lastProposer` identical to `proposer`?

consensus/istanbul/backend/engine.go
```diff
@@ -323,8 +323,8 @@ func (sb *backend) Prepare(chain consensus.ChainReader, header *types.Header) er
 	// if there is a vote to attach, attach it to the header
 	header.Vote = sb.governance.GetEncodedVote(sb.address, number)
 
-	// add validators in snapshot to extraData's validators section
-	extra, err := prepareExtra(header, snap.committee(header.ParentHash, sb.currentView.Load().(*istanbul.View)))
+	// add validators (council list) in snapshot to extraData's validators section
+	extra, err := prepareExtra(header, snap.validators())
```
ehnuje, line 327: This would be the core change of this PR, "like an original code, I reverted it to save all validator list in the field." Did I get it right?

KimKyungup: Yes. This is core change of this PR.

consensus/istanbul/backend/api.go
KimKyungup: This function returns the parent block proposer.
I missed to consider the given block number. I will fix it. Thanks.

consensus/istanbul/backend/api.go
```diff
@@ -249,16 +282,15 @@ func (api *APIExtension) getConsensusInfo(block *types.Block) (ConsensusInfo, er
...
 	originProposer := common.Address{}
 	lastProposer := common.Address{}
-	// TODO-Klaytn add the logic to get the last proposer for other policies. Weighted Random doesn't need it.
-	if istanbul.ProposerPolicy(snap.Policy) == istanbul.WeightedRandom {
-		newValSet := snap.ValSet.Copy()
-		newValSet.CalcProposer(lastProposer, 0)
-		originProposer = newValSet.GetProposer().Address()
-	} else {
-		logger.Warn("getConsensusInfo does not support to get origin proposal on", "proposerPolicy", snap.Policy)
+	if blockNumber > 0 {
```
aidan-kwon, line 285: Whether `blockNumber` == 0 is already checked at the beginning of the method. And, the type of `blockNumber` is uint64. Therefore, the checking logic is not necessary.

KimKyungup: Good point. I will reflect.

consensus/istanbul/backend/api.go
```diff
@@ -197,12 +208,34 @@ func (api *APIExtension) GetCommittee(number *rpc.BlockNumber) ([]common.Address
...
+
+	// get the proposer of this block.
+	proposer, err := ecrecover(header)
+	if err != nil {
+		return nil, err
+	}
+
+	// get the snapshot of the previous block.
+	parentHash := header.ParentHash
+	snap, err := api.istanbul.snapshot(api.chain, blockNumber-1, parentHash, nil)
```
aidan-kwon, line 226: Why did you load a snapshot of the previous block instead of the target block?

consensus/istanbul/backend/api.go
```diff
@@ -248,17 +281,13 @@ func (api *APIExtension) getConsensusInfo(block *types.Block) (ConsensusInfo, er
...
-		newValSet.CalcProposer(lastProposer, 0)
-		originProposer = newValSet.GetProposer().Address()
-	} else {
-		logger.Warn("getConsensusInfo does not support to get origin proposal on", "proposerPolicy", snap.Policy)
-	}
+	lastProposer := api.istanbul.GetProposer(blockNumber - 1)
+
+	newValSet := snap.ValSet.Copy()
+	newValSet.CalcProposer(lastProposer, 0)
+	originProposer = newValSet.GetProposer().Address()
```
aidan-kwon, line 288: Why don't you use the snapshot of the target block to calculate originProposer like `targetSnap.CalcProposer(lasstProposer, 0)`
And, we need a comment describing `originProposer` that may confuse other developers. Or, we can rename it to `round0Proposer` or `r0Proposer`.

consensus/istanbul/backend/api.go
KimKyungup: To get the committee list to validate the target block, the parent snapshot is needed.
(I noticed it by seeing the verifySigner/VerifyCommittedSeals/CheckInSubList)

consensus/istanbul/backend/api.go
KimKyungup: The reason why this uses the parent snapshot is same with above comments.

For `originProposer`,  I think it is good suggestion, I will make another PR to change the variable name and RPC output.

consensus/istanbul/backend/api.go
aidan-kwon: Oh.. I see. The sublist is selected based on the previous block's validator set. The method name confuses me. I think `SubListWithProposer` should be changed to `NextBlockCommittee` or something else.

aidan-kwon (approved):

KimKyungup:
@jeongkyun-oh @ehnuje @yoomee1313 PTAL.

yoomee1313 (approved):

consensus/istanbul/backend/api.go
```diff
@@ -161,9 +162,19 @@ func (api *APIExtension) GetCouncil(number *rpc.BlockNumber) ([]common.Address,
 	}
 	// Ensure we have an actually valid block and return the council from its snapshot
 	if header == nil {
-		logger.Trace("Failed to find the requested block", "number", number)
 		return nil, errNoBlockExist // return nil if block is not found.
 	}
+
+	// Since v1.6.1, extra.validators represents a list of council
+	// TODO-Klaytn : replace this to the below calculation logic
```
yoomee1313, line 169: How is the blocks generated before v1.6.1 handled? As far as I understand, those blocks contain invalid committee list.

jeongkyun-oh (approved):

consensus/istanbul/backend/api.go
KimKyungup: As you understood, present blocks` validator list is wrong. 
So this code can return wrong Council member for the past block which is generated by present code.

KimKyungup:
@ehnuje PTAL.

ehnuje (approved):
LGTM, however, is it okay to ignore the linter failure?

KimKyungup:
> LGTM, however, is it okay to ignore the linter failure?

The linter failure is not related with this change. So I didn't fix it. Winnie already merged a PR to fix all linter issue.

PR #903, opened 2021-02-25
API: Introduce init flags for klay_getLogs and klay_getFilterLogs APIs

aidan-kwon:
## Proposed changes

Introduce init flags for `klay_getLogs` and `klay_getFilterLogs` APIs

- `api.filter.getLogs.deadline`: Execution deadline for log collecting filter APIs (Default: 10s)
- `api.filter.getLogs.maxitems`: Maximum allowed number of return items for log collecting filter API (Default: 10000)

**Test result**
```
# curl -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"klay_getLogs","params":[{"fromBlock":"0x3210300","toBlock":"0x3210400"}],"id":1}' http://localhost:8551
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"query returned more than 10000 results"}}
```

```
# curl -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"klay_getLogs","params":[{"fromBlock":"0x3210300","toBlock":"0x3210400"}],"id":1}' http://localhost:8551
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"query timeout exceeded"}}
```

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -51,6 +51,7 @@ import (
 	"github.com/klaytn/klaytn/networks/rpc"
 	"github.com/klaytn/klaytn/node"
 	"github.com/klaytn/klaytn/node/cn"
+	"github.com/klaytn/klaytn/node/cn/filters"
 	"github.com/klaytn/klaytn/node/sc"
 	"github.com/klaytn/klaytn/params"
 	"github.com/klaytn/klaytn/storage/database"
@@ -523,6 +524,16 @@ var (
 	PreloadJSFlag = cli.StringFlag{
 		Name:  "preload",
 		Usage: "Comma separated list of JavaScript files to preload into the console",
+	}
+	APIFilterGetLogsDeadlineFlag = cli.DurationFlag{
+		Name:  "api.filter.getLogs.deadline",
+		Usage: "Execution deadline for log collecting filter APIs (Default: 10s)",
+		Value: filters.GetLogsDeadline,
+	}
+	APIFilterGetLogsMaxItemsFlag = cli.IntFlag{
+		Name:  "api.filter.getLogs.maxitems",
+		Usage: "Maximum allowed number of return items for log collecting filter API (Default: 10000)",
+		Value: filters.GetLogsMaxItems,
 	}
 
 	// Network Settings
@@ -1180,6 +1191,12 @@ func setgRPC(ctx *cli.Context, cfg *node.Config) {
 	}
 }
 
+// setAPIConfig sets configurations for specific APIs.
+func setAPIConfig(ctx *cli.Context) {
+	filters.GetLogsDeadline = ctx.GlobalDuration(APIFilterGetLogsDeadlineFlag.Name)
+	filters.GetLogsMaxItems = ctx.GlobalInt(APIFilterGetLogsMaxItemsFlag.Name)
+}
+
 // MakeAddress converts an account specified directly as a hex encoded string or
 // a key index in the key store to an internal account representation.
 func MakeAddress(ks *keystore.KeyStore, account string) (accounts.Account, error) {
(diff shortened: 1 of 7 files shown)


node/cn/filters/api.go
```diff
@@ -356,6 +365,10 @@ func (api *PublicFilterAPI) UninstallFilter(id rpc.ID) bool {
 // GetFilterLogs returns the logs for the filter with the given id.
 // If the filter could not be found an empty array of logs is returned.
 func (api *PublicFilterAPI) GetFilterLogs(ctx context.Context, id rpc.ID) ([]*types.Log, error) {
+	ctx = context.WithValue(ctx, "maxItem", GetLogsMaxItems)
```
ehnuje, line 368:
```suggestion
	ctx = context.WithValue(ctx, getLogsCxtKeyMaxItems, GetLogsMaxItems)
```
Is it intended?

node/cn/filters/filter.go
```diff
@@ -159,6 +163,14 @@ func (f *Filter) indexedLogs(ctx context.Context, end uint64) ([]*types.Log, err
 	// Iterate over the matches until exhausted or context closed
 	var logs []*types.Log
 
+	maxItems := math.MaxInt32 - 1
+	if val := ctx.Value(getLogsCxtKeyMaxItems); val != nil {
+		val, ok := val.(int)
+		if ok {
+			maxItems = val
+		}
```
ehnuje, line 171: Don't we need to return error for this case?

ehnuje: Also, as the same code is used in line#214~line#220, how about using a function for this?

cmd/utils/flags.go
```diff
@@ -524,6 +525,16 @@ var (
...
 		Usage: "Comma separated list of JavaScript files to preload into the console",
 	}
+	APIFilterGetLogsDeadlineFlag = cli.DurationFlag{
+		Name:  "api.filter.getLogs.deadline",
+		Usage: "Execution deadline for log collecting filter APIs (Default: 10s)",
+		Value: filters.GetLogsDeadline,
+	}
+	APIFilterGetLogsMaxItemsFlag = cli.IntFlag{
+		Name:  "api.filter.getLogs.maxitems",
+		Usage: "Maximum allowed number of return items for log collecting filter API (Default: 10000)",
```
ehnuje, line 535: Isn't "Default" message printed if `Value` is set?

aidan-kwon: I can understand the question. Does "message" mean log message or console message?

node/cn/filters/filter.go
aidan-kwon: I added a function for the duplicated code.

cmd/utils/flags.go
ehnuje: Oh, I mean console message. 
Even default value is not mentioned in `Usage` field, console will automatically print the default value.

Screen Shot 2021-02-25 at 2 47 45 PM
Screen Shot 2021-02-25 at 2 47 51 PM

node/cn/filters/filter.go
aidan-kwon: @ehnuje 
1. `nil` value for `getLogsCxtKeyMaxItems` is allowed since `indexedLogs` can be called without the context key.
2. We can consider returning an error for the type assertion failure even though the code cannot fail without improper code fixes. However, users always return errors if improper code fixes happened. I think using `MaxInt32-1` is better than stopping `klay_getLogs` API serving.

cmd/utils/flags.go
aidan-kwon: Oh, you're right. Thank you for the notice! I removed the default words.

node/cn/filters/filter.go
ehnuje: I considered the second case, to notify the user that he/she entered the wrong parameter. 
BTW, what does "even though the code cannot fail without improper code fixes" mean?

aidan-kwon: The value of `getLogsCxtKeyMaxItems` is set by the init flag `api.filter.getLogs.maxitems`. If the value is given with an invalid type, Klaytn node will fail to start returning an parse error.

ehnuje: Oh, I see. Then we don't need to worry about it!

ehnuje (approved):

KimKyungup (dismissed):

yoomee1313 (dismissed):

KimKyungup (approved):

jeongkyun-oh (approved):

ehnuje (approved):

PR #916, opened 2021-03-08
Enable tx prefetcher for multiple blocks

kjhman21:
## Proposed changes

- Previously, the tx prefetcher was enabled when only one block is passed into BlockChain.insertChain().
- After sync test, we found that the tx prefetcher is useful when multiple blocks are passed to that function.
- Hence, we enable tx prefetcher if multiple blocks are passed.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- https://github.com/klaytn/klaytn/pull/911

PR #911, opened 2021-03-03
Fix the prefech logic for multiple txs of the same sender in a block

kjhman21:
## Proposed changes

**PROBLEM**
- Prefetch worker executes transactions in a block in parallel.
- However, if there are multiple transactions from the same account
  cannot be executed due to nonce check logic.

**SOLUTION**
- For prefetch execution, nonce check logic is disabled.
- Gas was set to the gas value of the transaction. (See
  blockchain/state_transition.go:230)
- To be aware of the prefetch execution, a field `prefetching` is addded
  to vm.Config, state.StateDB
- To check the efficiency of the prefetch,
  memcacheCleanPrefetchMissMeter is added. This value increases when
  a cache miss occurs in the prefetch execution.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #918, opened 2021-03-09
Move prefetcher execution as early as possible

kjhman21:
## Proposed changes

- Prefetching should be executed as early as possible.
- This commit moves the prefetch trigger to the top of the for loop without changing code semantics.
- We found that it shows little bit performance improvement. See below snapshot from grafana. (from block number 51786905)

image
image (5)

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- https://github.com/klaytn/klaytn/pull/911
- https://github.com/klaytn/klaytn/pull/916

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1584,6 +1584,11 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 
 	// Iterate over the blocks and insert when the verifier permits
 	for i, block := range chain {
+		// If the chain is terminating, stop processing blocks
+		if atomic.LoadInt32(&bc.procInterrupt) == 1 {
+			logger.Debug("Premature abort during blocks processing")
+			break
+		}
 		// Create a new trie using the parent block and report an
 		// error if it fails.
 		var parent *types.Block
@@ -1620,11 +1625,6 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 					}
 				}(time.Now())
 			}
-		}
-		// If the chain is terminating, stop processing blocks
-		if atomic.LoadInt32(&bc.procInterrupt) == 1 {
-			logger.Debug("Premature abort during blocks processing")
-			break
 		}
 		// If the header is a banned one, straight out abort
 		if BadHashes[block.Hash()] {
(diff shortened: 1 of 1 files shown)


ehnuje (dismissed):
LGTM, thanks!

KimKyungup (dismissed):

blockchain/blockchain.go
```diff
@@ -1584,6 +1584,43 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
...
+					blockPrefetchExecuteTimer.Update(time.Since(start))
+					if atomic.LoadUint32(&followupInterrupt) == 1 {
+						blockPrefetchInterruptMeter.Mark(1)
+					}
+				}(time.Now())
+			}
+		}
 		// If the chain is terminating, stop processing blocks
 		if atomic.LoadInt32(&bc.procInterrupt) == 1 {
 			logger.Debug("Premature abort during blocks processing")
```
aidan-kwon, line 1626: @kjhman21 Is it better to place the prefetching logic below of this `if` branch? It's minor, but `blockchain` needs to wait to finish prefetching when it is closing.

winnie-byun (approved):

blockchain/blockchain.go
kjhman21: @aidan-kwon Hmm.. Actually prefetch logic just passes data to another goroutine, so I think it does not take up much time. Do you think it would be better even if we are using goroutine?

aidan-kwon: @kjhman21 Oh, yeah. I missed prefetching runs concurrently. Let' me check another thing, concurrency issues...
- Prefetching logic can read Database and TrieNodeCache after Blockchain is stopped. It would be fine. Database and TrieNodeCache don't panic even though read operation is called after the close. 
- `bc.prefechTxCh` can receives items after the channel is closed. It causes panic. It is an existing issue, not caused by this change. However, this change (checking  `bc.procInterrupt` later) may increase the possibility of panic. @kjhman21 Could you review this comment?

kjhman21: @aidan-kwon 
Okay. So, you think it would be better to check `bc.procInterrupt` first. Right?

aidan-kwon: @kjhman21 Yes

kjhman21: @aidan-kwon All right, then. :)

aidan-kwon: @kjhman21 Thanks you!

aidan-kwon (approved):

winnie-byun (approved):

ehnuje (approved):

PR #920, opened 2021-03-12
Updated minimum staking related logic

jeongkyun-oh:
## Proposed changes

- Updated the minimum staking update flag available
- Removed the validators who do not have minimum staking amount of KLAY from committees/proposers
  - Case 1. Some members do not have enough KLAYs
    - Members who do not have enough KLAYs cannot be committee as well as proposers
  - Case 2. No validator has enough KLAYs
    - All members become the committee and a proposer is chosen from them
  - Case 3. Governing node in single mode doesn't have enough KLAYs
    - No one can vote until governing node has enough KLAYs
    - This case will be handled in another PR. In single mode, the governing node will always be in proposers/committee.
- Updated for backward-incompatible change

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/genesis.go
+++ b/blockchain/genesis.go
@@ -238,11 +238,18 @@ func SetupGenesisBlock(db database.DBManager, genesis *Genesis, networkId uint64
 		if storedcfg.Governance == nil {
 			logger.Crit("Failed to read governance. storedcfg.Governance == nil")
 		}
+		if storedcfg.Governance.Reward == nil {
+			logger.Crit("Failed to read governance. storedcfg.Governance.Reward == nil")
+		}
 		if storedcfg.Governance.Reward.StakingUpdateInterval != 0 {
 			params.SetStakingUpdateInterval(storedcfg.Governance.Reward.StakingUpdateInterval)
 		}
 		if storedcfg.Governance.Reward.ProposerUpdateInterval != 0 {
 			params.SetProposerUpdateInterval(storedcfg.Governance.Reward.ProposerUpdateInterval)
+		}
+		if storedcfg.Governance.Reward.MinimumStake != nil &&
+			storedcfg.Governance.Reward.MinimumStake.Cmp(common.Big0) > 0 {
+			params.SetMinimumStakingAmount(storedcfg.Governance.Reward.MinimumStake)
 		}
 	}
 	// Special case: don't change the existing config of a non-mainnet chain if no new
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -634,26 +634,9 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
 	for i := 0; i < len(headers)/2; i++ {
 		headers[i], headers[len(headers)-1-i] = headers[len(headers)-1-i], headers[i]
 	}
-	snap, err := snap.apply(headers, sb.governance, sb.address, sb.governance.Epoch())
+	snap, err := snap.apply(headers, sb.governance, sb.address, sb.governance.ProposerPolicy(), chain)
 	if err != nil {
 		return nil, err
-	}
-	if sb.governance.ProposerPolicy() == uint64(istanbul.WeightedRandom) {
-		// Snapshot of block N (Snapshot_N) should contain proposers for N+1 and following blocks.
-		// And proposers for Block N+1 can be calculated from the nearest previous proposersUpdateInterval block.
-		// Let's refresh proposers in Snapshot_N using previous proposersUpdateInterval block for N+1, if not updated yet.
-		pHeader := chain.GetHeaderByNumber(params.CalcProposerBlockNumber(snap.Number + 1))
-		if pHeader != nil {
-			if err := snap.ValSet.Refresh(pHeader.Hash(), pHeader.Number.Uint64(), sb.chain.Config()); err != nil {
-				// There are three error cases and they just don't refresh proposers
-				// (1) no validator at all
-				// (2) invalid formatted hash
-				// (3) no staking info available
-				logger.Trace("Skip refreshing proposers while creating snapshot", "snap.Number", snap.Number, "pHeader.Number", pHeader.Number.Uint64(), "err", err)
-			}
-		} else {
-			logger.Trace("Can't refreshing proposers while creating snapshot due to lack of required header", "snap.Number", snap.Number)
-		}
 	}
 
 	// If we've generated a new checkpoint snapshot, save to disk
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -644,7 +644,7 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
 		// Let's refresh proposers in Snapshot_N using previous proposersUpdateInterval block for N+1, if not updated yet.
 		pHeader := chain.GetHeaderByNumber(params.CalcProposerBlockNumber(snap.Number + 1))
 		if pHeader != nil {
-			if err := snap.ValSet.Refresh(pHeader.Hash(), pHeader.Number.Uint64()); err != nil {
+			if err := snap.ValSet.Refresh(pHeader.Hash(), pHeader.Number.Uint64(), sb.chain.Config()); err != nil {
 				// There are three error cases and they just don't refresh proposers
 				// (1) no validator at all
 				// (2) invalid formatted hash
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -563,7 +563,7 @@ func (sb *backend) initSnapshot(chain consensus.ChainReader) (*Snapshot, error) 
 	if !ok {
 		committeeSize = params.DefaultSubGroupSize
 	}
-	snap := newSnapshot(sb.governance, 0, genesis.Hash(), validator.NewValidatorSet(istanbulExtra.Validators, istanbul.ProposerPolicy(proposerPolicy), committeeSize, chain), chain.Config())
+	snap := newSnapshot(sb.governance, 0, genesis.Hash(), validator.NewValidatorSet(istanbulExtra.Validators, nil, istanbul.ProposerPolicy(proposerPolicy), committeeSize, chain), chain.Config())
 
 	if err := snap.store(sb.db); err != nil {
 		return nil, err
--- a/consensus/istanbul/backend/snapshot.go
+++ b/consensus/istanbul/backend/snapshot.go
@@ -273,10 +273,11 @@ func (s *Snapshot) toJSONStruct() *snapshotJSON {
 	var proposers []common.Address
 	var proposersBlockNum uint64
 	var validators []common.Address
+	var demotedValidators []common.Address
 
 	// TODO-Klaytn-Issue1166 For weightedCouncil
 	if s.ValSet.Policy() == istanbul.WeightedRandom {
-		validators, rewardAddrs, votingPowers, weights, proposers, proposersBlockNum = validator.GetWeightedCouncilData(s.ValSet)
+		validators, demotedValidators, rewardAddrs, votingPowers, weights, proposers, proposersBlockNum = validator.GetWeightedCouncilData(s.ValSet)
 	} else {
 		validators = s.validators()
 	}
@@ -295,6 +296,7 @@ func (s *Snapshot) toJSONStruct() *snapshotJSON {
 		Weights:           weights,
 		Proposers:         proposers,
 		ProposersBlockNum: proposersBlockNum,
+		DemotedValidators: demotedValidators,
 	}
 }
 
@@ -313,7 +315,7 @@ func (s *Snapshot) UnmarshalJSON(b []byte) error {
 
 	// TODO-Klaytn-Issue1166 For weightedCouncil
 	if j.Policy == istanbul.WeightedRandom {
-		s.ValSet = validator.NewWeightedCouncil(j.Validators, j.RewardAddrs, j.VotingPowers, j.Weights, j.Policy, j.SubGroupSize, j.Number, j.ProposersBlockNum, nil)
+		s.ValSet = validator.NewWeightedCouncil(j.Validators, j.DemotedValidators, j.RewardAddrs, j.VotingPowers, j.Weights, j.Policy, j.SubGroupSize, j.Number, j.ProposersBlockNum, nil)
 		validator.RecoverWeightedCouncilProposer(s.ValSet, j.Proposers)
 	} else {
 		s.ValSet = validator.NewSubSet(j.Validators, j.Policy, j.SubGroupSize)
--- a/consensus/istanbul/backend/snapshot.go
+++ b/consensus/istanbul/backend/snapshot.go
@@ -263,6 +263,7 @@ type snapshotJSON struct {
 	Weights           []uint64         `json:"weight"`
 	Proposers         []common.Address `json:"proposers"`
 	ProposersBlockNum uint64           `json:"proposersBlockNum"`
+	DemotedValidators []common.Address `json:"demotedValidators"`
 }
 
 func (s *Snapshot) toJSONStruct() *snapshotJSON {
--- a/consensus/istanbul/backend/snapshot.go
+++ b/consensus/istanbul/backend/snapshot.go
@@ -24,6 +24,8 @@ import (
 	"bytes"
 	"encoding/json"
 
+	"github.com/klaytn/klaytn/consensus"
+
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/consensus/istanbul"
@@ -142,7 +144,7 @@ func (s *Snapshot) checkVote(address common.Address, authorize bool) bool {
 
 // apply creates a new authorization snapshot by applying the given headers to
 // the original one.
-func (s *Snapshot) apply(headers []*types.Header, gov *governance.Governance, addr common.Address, epoch uint64) (*Snapshot, error) {
+func (s *Snapshot) apply(headers []*types.Header, gov *governance.Governance, addr common.Address, policy uint64, chain consensus.ChainReader) (*Snapshot, error) {
 	// Allow passing in no headers for cleaner code
 	if len(headers) == 0 {
 		return s, nil
@@ -177,6 +179,23 @@ func (s *Snapshot) apply(headers []*types.Header, gov *governance.Governance, ad
 		}
 
 		snap.ValSet, snap.Votes, snap.Tally = gov.HandleGovernanceVote(snap.ValSet, snap.Votes, snap.Tally, header, validator, addr)
+		if policy == uint64(params.WeightedRandom) {
+			// Snapshot of block N (Snapshot_N) should contain proposers for N+1 and following blocks.
+			// And proposers for Block N+1 can be calculated from the nearest previous proposersUpdateInterval block.
+			// Let's refresh proposers in Snapshot_N using previous proposersUpdateInterval block for N+1, if not updated yet.
+			pHeader := chain.GetHeaderByNumber(params.CalcProposerBlockNumber(number + 1))
+			if pHeader != nil {
+				if err := snap.ValSet.Refresh(pHeader.Hash(), pHeader.Number.Uint64(), chain.Config()); err != nil {
+					// There are three error cases and they just don't refresh proposers
+					// (1) no validator at all
+					// (2) invalid formatted hash
+					// (3) no staking info available
+					logger.Trace("Skip refreshing proposers while creating snapshot", "snap.Number", snap.Number, "pHeader.Number", pHeader.Number.Uint64(), "err", err)
+				}
+			} else {
+				logger.Trace("Can't refreshing proposers while creating snapshot due to lack of required header", "snap.Number", snap.Number)
+			}
+		}
 
 		if number%snap.Epoch == 0 {
 			if len(header.Governance) > 0 {
--- a/consensus/istanbul/validator/validator.go
+++ b/consensus/istanbul/validator/validator.go
@@ -39,10 +39,10 @@ func New(addr common.Address) istanbul.Validator {
 	}
 }
 
-func NewValidatorSet(addrs []common.Address, proposerPolicy istanbul.ProposerPolicy, subGroupSize uint64, chain consensus.ChainReader) istanbul.ValidatorSet {
+func NewValidatorSet(addrs, demotedAddrs []common.Address, proposerPolicy istanbul.ProposerPolicy, subGroupSize uint64, chain consensus.ChainReader) istanbul.ValidatorSet {
 	var valSet istanbul.ValidatorSet
 	if proposerPolicy == istanbul.WeightedRandom {
-		valSet = NewWeightedCouncil(addrs, nil, nil, nil, proposerPolicy, subGroupSize, 0, 0, chain)
+		valSet = NewWeightedCouncil(addrs, demotedAddrs, nil, nil, nil, proposerPolicy, subGroupSize, 0, 0, chain)
 	} else {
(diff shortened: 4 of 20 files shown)


consensus/istanbul/backend/engine.go
```diff
@@ -634,10 +634,19 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
 		return nil, err
 	}
 	if sb.governance.ProposerPolicy() == uint64(istanbul.WeightedRandom) {
+		pHeader := chain.GetHeaderByNumber(params.CalcStakingBlockNumber(snap.Number + 1))
```
KimKyungup, line 637: Why is this routine added again?

jeongkyun-oh: The parameter is different. This one is for staking update using `CalcStakingBlockNumber` and the other one is for proposer update using `CalcProposerBlockNumber`.

aidan-kwon:
> Case 3. Governing node in single mode doesn't have enough KLAYs
> * No one can vote until governing node has enough KLAYs

@KimKyungup @jeongkyun-oh The third policy probably becomes a huddle for Service Chain or Private Chain operators. How about letting "Governing node in single mode" propose a block?

consensus/istanbul/validator/weighted.go
```diff
@@ -553,6 +572,36 @@ func (valSet *weightedCouncil) F() int {
 
 func (valSet *weightedCouncil) Policy() istanbul.ProposerPolicy { return valSet.policy }
 
+func (valSet *weightedCouncil) Update(hash common.Hash, blockNum uint64) error {
+	valSet.validatorMu.Lock()
+	defer valSet.validatorMu.Unlock()
+
+	newStakingInfo := reward.GetStakingInfo(blockNum + 1)
```
KimKyungup, line 579: Do you know why this gets the staking info of next block?
```suggestion
	newStakingInfo := reward.GetStakingInfo(blockNum + 1)
```

consensus/istanbul/validator/weighted.go
```diff
@@ -553,6 +572,36 @@ func (valSet *weightedCouncil) F() int {
 
 func (valSet *weightedCouncil) Policy() istanbul.ProposerPolicy { return valSet.policy }
 
+func (valSet *weightedCouncil) Update(hash common.Hash, blockNum uint64) error {
```
KimKyungup, line 575: Please add some description about this function.

consensus/istanbul/validator/weighted.go
```diff
@@ -95,6 +95,7 @@ func newWeightedValidator(addr common.Address, reward common.Address, votingpowe
 
 type weightedCouncil struct {
 	subSize    uint64
+	demoted    istanbul.Validators
```
KimKyungup, line 98:
```suggestion
	demotedValidator    istanbul.Validators
```

KimKyungup: How about add `stakedValidator` instead of `demotedValidator`?
So `stakedValidator` like filtered validators and origin validator list can be managed.

consensus/istanbul/validator/weighted.go
```diff
@@ -592,7 +641,7 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
 		return errors.New("skip refreshing proposers due to no staking info")
 	}
 
-	weightedValidators, stakingAmounts, err := valSet.getStakingAmountsOfValidators(newStakingInfo)
+	weightedValidators, stakingAmounts, err := getStakingAmountsOfValidators(valSet.validators, newStakingInfo)
```
KimKyungup, line 644: Does this `valSet.validators` always return the filtered validators?

blockchain/genesis.go
```diff
@@ -244,6 +244,9 @@ func SetupGenesisBlock(db database.DBManager, genesis *Genesis, networkId uint64
 		if storedcfg.Governance.Reward.ProposerUpdateInterval != 0 {
 			params.SetProposerUpdateInterval(storedcfg.Governance.Reward.ProposerUpdateInterval)
 		}
+		if storedcfg.Governance.Reward.MinimumStake.Cmp(common.Big0) != 0 {
+			params.SetMinimumStakingAmount(storedcfg.Governance.Reward.MinimumStake)
+		}
```
aidan-kwon, line 249: I think `0` is supposed to be allowed as a minimum stake. 
Also, we need to check whether `storedcfg.Governance.Reward` is nil or `storedcfg.Governance.Reward.MinimumStake` is nil.

params/config.go
```diff
@@ -342,6 +342,12 @@ func (c *ChainConfig) SetDefaults() {
 		logger.Warn("Override the default proposer update interval to the chain config", "interval",
 			c.Governance.Reward.ProposerUpdateInterval)
 	}
+
+	if c.Governance.Reward.MinimumStake.Cmp(common.Big0) == 0 {
+		c.Governance.Reward.MinimumStake = MinimumStakingAmount()
+		logger.Warn("Override the default minimum staking amount to the chain config", "amount",
+			c.Governance.Reward.MinimumStake)
+	}
```
aidan-kwon, line 350: I think `0` is supposed to be allowed as a minimum stake.
Also, we need to check whether `storedcfg.Governance.Reward.MinimumStake` is nil.

consensus/istanbul/validator/weighted.go
```diff
@@ -553,6 +572,36 @@ func (valSet *weightedCouncil) F() int {
...
+	valSet.validatorMu.Lock()
+	defer valSet.validatorMu.Unlock()
+
+	newStakingInfo := reward.GetStakingInfo(blockNum + 1)
+
+	valSet.stakingInfo = newStakingInfo
+	if valSet.stakingInfo == nil {
+		// Just return without updating validators
+		return errors.New("skip updating validators due to no staking info")
+	}
```
aidan-kwon, line 585:
```suggestion
	newStakingInfo := reward.GetStakingInfo(blockNum + 1)
	if newStakingInfo == nil {
		// Just return without updating validators
		return errors.New("skip updating validators due to no staking info")
	}
	valSet.stakingInfo = newStakingInfo
```
If you don't mind, could you change it? I think it's more intuitive.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: So, we are going to check the staking information for the next block to calculate staked validators.
Like proposer list which is proposer candidates of the next block, validators in the snapshot will be the ones of the next block, not the current block.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: I will add some comments.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: So, the `validators` and `demoted` are exclusive. The `validators` are the filtered validators.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: Yes, it is filtered validators by minimum staking.

consensus/istanbul/validator/weighted.go
```diff
@@ -607,17 +656,62 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
...
+
+// filterPoorValidators divided the given weightedValidators into two group filtered by the minimum amount of staking.
+func filterPoorValidators(weightedValidators []*weightedValidator, stakingAmounts []float64) ([]*weightedValidator, []float64, []*weightedValidator, []float64) {
+	var (
+		newWeightedValidators []*weightedValidator
+		newWeightedDemoted    []*weightedValidator
+		newValidatorsStaking  []float64
+		newDemotedStaking     []float64
+	)
+	amount := params.MinimumStakingAmount()
```
aidan-kwon, line 686:
```suggestion
	amount := params.MinimumStakingAmount().Uint64()
```
We don't need to convert it to uint64 on every iteration.

consensus/istanbul/validator/weighted.go
```diff
@@ -607,17 +656,62 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
...
+			newWeightedValidators = append(newWeightedValidators, weightedValidators[idx])
+			newValidatorsStaking = append(newValidatorsStaking, val)
+		} else {
+			newWeightedDemoted = append(newWeightedDemoted, weightedValidators[idx])
+			newDemotedStaking = append(newDemotedStaking, val)
+		}
+	}
+
+	if len(newWeightedValidators) <= 0 {
+		return newWeightedDemoted, newDemotedStaking, nil, nil
```
aidan-kwon, line 698: How about returning an empty slice instead of `nil`? If use tried to call a method of the returned value, the node will be panicked.

consensus/istanbul/validator/weighted.go
```diff
@@ -132,7 +133,7 @@ func RecoverWeightedCouncilProposer(valSet istanbul.ValidatorSet, proposerAddrs
 	weightedCouncil.proposers = proposers
 }
 
-func NewWeightedCouncil(addrs []common.Address, rewards []common.Address, votingPowers []uint64, weights []uint64, policy istanbul.ProposerPolicy, committeeSize uint64, blockNum uint64, proposersBlockNum uint64, chain consensus.ChainReader) *weightedCouncil {
+func NewWeightedCouncil(addrs []common.Address, demotedAddrs []common.Address, rewards []common.Address, votingPowers []uint64, weights []uint64, policy istanbul.ProposerPolicy, committeeSize uint64, blockNum uint64, proposersBlockNum uint64, chain consensus.ChainReader) *weightedCouncil {
```
aidan-kwon, line 136: Why don't you call `valiset.Update()` inside of this function?

blockchain/genesis.go
jeongkyun-oh: I have updated it. Thanks.

params/config.go
jeongkyun-oh: I have updated it. Thanks.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: This is from other code fragments and updated it as well.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: Updated! Thanks.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: Updated. Thanks.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: It is not necessary to call it, but it is given as a parameter. The `Update()` function is called while creating snapshot only.

consensus/istanbul/validator/validator.go
```diff
@@ -42,7 +42,7 @@ func New(addr common.Address) istanbul.Validator {
 func NewValidatorSet(addrs []common.Address, proposerPolicy istanbul.ProposerPolicy, subGroupSize uint64, chain consensus.ChainReader) istanbul.ValidatorSet {
 	var valSet istanbul.ValidatorSet
 	if proposerPolicy == istanbul.WeightedRandom {
-		valSet = NewWeightedCouncil(addrs, nil, nil, nil, proposerPolicy, subGroupSize, 0, 0, chain)
+		valSet = NewWeightedCouncil(addrs, nil, nil, nil, nil, proposerPolicy, subGroupSize, 0, 0, chain)
```
aidan-kwon, line 45: Why is `demotedAddrs` always `nil`? Does `backend.getValidators()` needs the value?

KimKyungup:
> The third policy probably becomes a huddle for Service Chain or Private Chain operators. How about letting "Governing node in single mode" propose a block?

@jeongkyun-oh @aidan-kwon 
Does this feature affect to all governing modes?
If the default minimum staking amount is zero, I think it's ok. But the feature of governing mode should be documented and discussed.

consensus/istanbul/validator/validator.go
jeongkyun-oh: `backend.getValidators()` calls `NewValidatorSet` when getting snapshot is failed. `NewValidatorSet` is used with proper parameters when initially creating a snapshot only.

consensus/istanbul/validator/weighted.go
```diff
@@ -193,9 +194,18 @@ func NewWeightedCouncil(addrs []common.Address, rewards []common.Address, voting
...
 	sort.Sort(valSet.validators)
 
+	// init demoted validators
+	valSet.demotedValidators = make([]istanbul.Validator, len(demotedAddrs))
+	for i, addr := range demotedAddrs {
+		valSet.demotedValidators[i] = newWeightedValidator(addr, common.Address{}, 1000, 0)
+	}
+
+	// sort demoted validators
+	sort.Sort(valSet.demotedValidators)
```
ehnuje, line 207: These lines can't be tested since there is no place that calls `NewWeightedCouncil` with non-nil `demotedAddrs`

consensus/istanbul/validator/weighted.go
```diff
@@ -235,6 +245,12 @@ func GetWeightedCouncilData(valSet istanbul.ValidatorSet) (validators []common.A
 			weights[i] = atomic.LoadUint64(&weightedVal.weight)
 		}
 
+		numDemoted := len(weightedCouncil.demotedValidators)
+		demotedValidators = make([]common.Address, numDemoted)
+		for i, val := range weightedCouncil.demotedValidators {
+			demotedValidators[i] = val.Address()
+		}
```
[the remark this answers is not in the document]
jeongkyun-oh: The type of `weightedCouncil.demotedValidators` is not `common.Address`, so it is not possible to copy them as you suggested.

consensus/istanbul/validator/weighted.go
```diff
@@ -607,17 +656,62 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
...
+			newWeightedValidators = append(newWeightedValidators, weightedValidators[idx])
+			newValidatorsStaking = append(newValidatorsStaking, val)
+		} else {
+			newWeightedDemoted = append(newWeightedDemoted, weightedValidators[idx])
+			newDemotedStaking = append(newDemotedStaking, val)
+		}
+	}
+
+	if len(newWeightedValidators) <= 0 {
+		return newWeightedDemoted, newDemotedStaking, []*weightedValidator{}, []float64{}
```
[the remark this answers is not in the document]
jeongkyun-oh: If no validator has enough KLAYs, the demoted validators can still propose blocks to prevent them from not creating blocks.

governance/default.go
```diff
@@ -909,6 +910,9 @@ func (gov *Governance) ReadGovernanceState() {
 	gov.UnmarshalJSON(b)
 	params.SetStakingUpdateInterval(gov.StakingUpdateInterval())
 	params.SetProposerUpdateInterval(gov.ProposerUpdateInterval())
+	if minimumStakingAmount, ok := new(big.Int).SetString(gov.MinimumStake(), 10); ok {
+		params.SetMinimumStakingAmount(minimumStakingAmount)
+	}
```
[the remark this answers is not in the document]
jeongkyun-oh: Right, but `updateMinimumStakingAmount` is a trigger function for voting. There is no propoer key and value for this.

consensus/istanbul/validator/weighted.go
```diff
@@ -607,17 +656,62 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
...
+}
+
 // getStakingAmountsOfValidators calculates stakingAmounts of validators.
 // If validators have multiple staking contracts, stakingAmounts will be a sum of stakingAmounts with the same rewardAddress.
 //  - []*weightedValidator : a list of validators which type is converted to weightedValidator
 //  - []float64 : a list of stakingAmounts.
-func (valSet *weightedCouncil) getStakingAmountsOfValidators(stakingInfo *reward.StakingInfo) ([]*weightedValidator, []float64, error) {
-	numValidators := len(valSet.validators)
+func getStakingAmountsOfValidators(candidates istanbul.Validators, stakingInfo *reward.StakingInfo) ([]*weightedValidator, []float64, error) {
+	numValidators := len(candidates)
```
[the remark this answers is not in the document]
jeongkyun-oh: I think validators is more proper term. Thanks.

consensus/istanbul/validator/weighted.go
```diff
@@ -607,17 +656,62 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
 	return nil
 }
 
+// updateValidators converts weighted validator slice to istanbul.Validators and sets them to the council.
+func (valSet *weightedCouncil) updateValidators(validators []*weightedValidator, demoted []*weightedValidator) {
```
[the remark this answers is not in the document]
jeongkyun-oh: Updated. Thanks!

consensus/istanbul/backend/engine.go
```diff
@@ -634,10 +634,22 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
 		return nil, err
 	}
 	if sb.governance.ProposerPolicy() == uint64(istanbul.WeightedRandom) {
+		// Based on the block in which staking information is updated, the validators will be recalculated.
+		// The validators staking less than minimum staking amount will be demoted into the group and
+		// they will not be in the committee as well as the proposer candidates list.
+		pHeader := chain.GetHeaderByNumber(params.CalcStakingBlockNumber(snap.Number + 1))
+		if pHeader != nil {
+			if err := snap.ValSet.TransitionValidators(pHeader.Hash(), pHeader.Number.Uint64()); err != nil {
```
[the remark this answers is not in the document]
jeongkyun-oh: The logic is updated and moved to into `Refresh` function. Please take another look.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: `NewWeightedCouncil` is only used properly while unmarshalling a snapshot. This is called when read a snapshot from database.

jeongkyun-oh:
I have updated the PR so that the `validators` are selected from `candidates` based on the staking amount of KLAYs. The `candidates` include all the validators no matter how much they stake.

consensus/istanbul/validator/weighted.go
```diff
@@ -681,12 +705,24 @@ func filterPoorValidators(weightedValidators []*weightedValidator, stakingAmount
 		} else {
```
KimKyungup, line 705: @jeongkyun-oh When this loop check the amount, it can check if the node is `govNode`. Then even if the node doesn't have enough staking, it can be skipped when the node is govNode.

This code works like demoting and promoting.
But I think demoting only normal node except govNode is better. What do you think?

KimKyungup (commented):
LGTM except the comment.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: @KimKyungup  The reason is that the governing node should be demoted when no one has enough staking amount of KLAYs. Otherwise, only the governing node will be a proposer and the committee member.

consensus/istanbul/validator/weighted.go
```diff
@@ -585,17 +616,41 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
...
+		var demotedValidators []*weightedValidator
+		// divide the obtained validators into two groups which have enough amount of staking
+		weightedValidators, stakingAmounts, demotedValidators, _ = filterPoorValidators(weightedValidators, stakingAmounts)
+		// update new validators and demoted validators of the council
+		valSet.setValidators(weightedValidators, demotedValidators)
+	} else {
+		candidates := valSet.validators
+		weightedValidators, stakingAmounts, err = getStakingAmountsOfValidators(candidates, newStakingInfo)
+		if err != nil {
+			return err
```
ehnuje, line 650: This code block can be simplified, could you please check?

ehnuje (commented):
LGTM, other than minor comments.

KimKyungup (approved):

consensus/istanbul/validator/weighted.go
```diff
@@ -607,17 +668,82 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
...
+			if !emptyAddress(govNode) && govNode == weightedValidators[idx].Address() {
+				demotedGovNodeIdx = demotedNodeIdx
+			}
+		}
+	}
+
+	if len(newWeightedValidators) <= 0 {
+		return newWeightedDemoted, newDemotedStaking, []*weightedValidator{}, []float64{}
+	} else {
+		// promote the governing node if it has not enough KLAYs
```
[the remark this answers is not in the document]
jeongkyun-oh: This commit is updated. Please take another look.

ehnuje (approved):
LGTM, thanks!

jeongkyun-oh:
@aidan-kwon @winnie-byun @yoomee1313 Please take a look when you have time.

consensus/istanbul/validator/weighted.go
```diff
@@ -585,17 +622,37 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64) error
...
+	blockNumBig := new(big.Int).SetUint64(blockNum)
+	chainRules := config.Rules(blockNumBig)
+
+	candidates := append(valSet.validators, valSet.demotedValidators...)
+	weightedValidators, stakingAmounts, err := getStakingAmountsOfValidators(candidates, newStakingInfo)
 	if err != nil {
 		return err
 	}
+
+	if chainRules.IsIstanbul {
```
yoomee1313, line 640:
```suggestion
	if config.IsIstanbul(new(big.Int).SetUint64(blockNum)) 
```
What about changing like this? 631, 632 lines can be removed.

yoomee1313 (approved):

jeongkyun-oh:
@aidan-kwon As we talked f2f, I will updated the following cases.
1. Comment: The method `NewValidatorSet` shouldn't be called.
2. Comment: The `validatorMu` is used for `validators` and `demotedValidators`.
3. The `Refresh` method should be called for every block, so it would be better to move into `apply` method.
4. The unit-test for `snapshot` is added.

consensus/istanbul/validator/weighted.go
jeongkyun-oh: I will update this on the next PR.

jeongkyun-oh:
@aidan-kwon @ehnuje @KimKyungup @yoomee1313 PTAL

consensus/istanbul/backend/snapshot.go
```diff
@@ -177,6 +179,23 @@ func (s *Snapshot) apply(headers []*types.Header, gov *governance.Governance, ad
...
+					// There are three error cases and they just don't refresh proposers
+					// (1) no validator at all
+					// (2) invalid formatted hash
+					// (3) no staking info available
+					logger.Trace("Skip refreshing proposers while creating snapshot", "snap.Number", snap.Number, "pHeader.Number", pHeader.Number.Uint64(), "err", err)
+				}
+			} else {
+				logger.Trace("Can't refreshing proposers while creating snapshot due to lack of required header", "snap.Number", snap.Number)
+			}
+		}
```
KimKyungup, line 198: Do we need a lock for atomicity of `snap.ValSet`? @jeongkyun-oh

KimKyungup (approved):

consensus/istanbul/backend/snapshot.go
jeongkyun-oh: I think no need to lock `snap.ValSet` which is accessed synchronously.`snap.ValSet` is an interface for `weightedCouncil`. If you worry about the `validators` and `demotedValidators`,  the `validatorsMu` protects them in `weightedCouncil`.

ehnuje (approved):

yoomee1313 (approved):

PR #927, opened 2021-03-18
Updated governance initialization

jeongkyun-oh:
## Proposed changes

Updated initialization of the governance in order to load proper current set. The `currentSet` of the governance is stored in `Governance` and `GovernanceState`. Although the `currentSet` can be loaded from both persistences, but it doesn't consider the order. So, this PR fixes the metioned bug.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/default.go
+++ b/governance/default.go
@@ -443,6 +443,7 @@ func (g *Governance) updateGovernanceParams() {
 	params.SetStakingUpdateInterval(g.StakingUpdateInterval())
 	params.SetProposerUpdateInterval(g.ProposerUpdateInterval())
 
+	// NOTE: HumanReadable related functions are inactivated now
 	if txGasHumanReadable, ok := g.currentSet.GetValue(params.ConstTxGasHumanReadable); ok {
 		params.TxGasHumanReadable = txGasHumanReadable.(uint64)
 	}
@@ -599,6 +600,8 @@ func newGovernanceCache() common.Cache {
 	return cache
 }
 
+// initializeCache reads governance item data from database and updates Governance.itemCache.
+// The latest governance item data of Governance.itemCache is imported to Governance.currentSet.
 func (g *Governance) initializeCache() error {
 	// get last n governance change block number
 	indices, err := g.db.ReadRecentGovernanceIdx(params.GovernanceIdxCacheLimit)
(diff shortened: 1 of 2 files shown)


aidan-kwon (approved):
LGTM

KimKyungup:
@jeongkyun-oh What is difference between Governance and GovernanceState?

jeongkyun-oh:
@KimKyungup 
We have two sorts of persistence. 
`ReadGovernance` / `WriteGovernance` is for storing the governance key/value set applied from a specific block number. For example, the API `governance.itemsAt()` method is the result of `ReadGovernance` method.
```
{
  governance.governancemode: "single",
  governance.governingnode: "0xfa97aa7618c4b84c7f88f2af8c45e48c2bc27bf6",
  governance.unitprice: 25000000000,
  istanbul.committeesize: 22,
  istanbul.epoch: 30,
  istanbul.policy: 2,
  reward.deferredtxfee: true,
  reward.minimumstake: "7000000",
  reward.mintingamount: "9600000000000000000",
  reward.proposerupdateinterval: 30,
  reward.ratio: "34/54/12",
  reward.stakingupdateinterval: 60,
  reward.useginicoeff: true
}
```

On the other hand, `ReadGovernanceState` / `WriteGovernanceState` is for  `Governance` object persistence. This is stored for restoring the `Governance` object quickly instead of replaying all the governance changes. The object also has the current governance key/value set.

The problem is when initializing the `Governance` object which contains also the current governance key/value set as well. The current set is not considered the order of them. So, this PR fixes the problem.

jeongkyun-oh:
@KimKyungup @winnie-byun @yoomee1313 Could you take a look please?

winnie-byun (approved):

KimKyungup (approved):

yoomee1313 (approved):

governance/default.go
```diff
@@ -423,6 +423,7 @@ func NewGovernanceInitialize(chainConfig *params.ChainConfig, dbm database.DBMan
 	ret := NewGovernance(chainConfig, dbm)
 	// nil is for testing or simple function usage
 	if dbm != nil {
+		ret.ReadGovernanceState()
```
yoomee1313, line 426: Can I ask why you move this line to here?

jeongkyun-oh: The `ReadGovernanceState()` first sets the current set, and the `initializeCache()` overwrites the current set if necessary.

PR #943, opened 2021-04-06
Introduce NewRelic web transaction monitoring feature

aidan-kwon:
## Proposed changes

With this PR, Application Performance Monitoring (APM) feature of NewRelic can be enabled. When it is enabled, it also prints JSON RPC error logs reqeusted through fastHTTP server. 

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


node/config.go
```diff
@@ -166,6 +166,9 @@ type Config struct {
 	// ephemeral nodes).
 	GRPCPort int `toml:",omitempty"`
 
+	// APINRAppName string `toml:",omitempty"`
+	// APINRLicense string `toml:",omitempty"`
```
ehnuje, line 170: Will these be used in the upcoming PR, which uses environmental variable to update NewRelic related values?

node/node.go
```diff
@@ -64,6 +64,9 @@ type Node struct {
 	rpcAPIs       []rpc.API
 	inprocHandler *rpc.Server // In-process RPC request handler to process the API requests
 
+	nrAppName string
+	nrLicense string
```
ehnuje, line 68: How about leaving comments about new variables?

networks/rpc/http.go
```diff
@@ -232,6 +235,18 @@ func NewFastHTTPServer(cors []string, vhosts []string, timeouts HTTPTimeouts, sr
 	handler := newCorsHandler(srv, cors)
 	handler = newVHostHandler(vhosts, handler)
 
+	if nrAppName != "" && nrLicense != "" {
```
ehnuje, line 238: Don't we need to notify user if only one variable is set?

networks/rpc/http_newrelic.go
```diff
@@ -0,0 +1,136 @@
...
+	return attrs
+}
+
+// newNewRelicHTTPHandler enables NewRelic web transaction monitor.
+// It also prints error logs when RPC returns contains error messages.
+func newNewRelicHTTPHandler(nrApp *newrelic.Application, handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			if err := recover(); err != nil {
+				logger.Error("NewRelic http handler panic\n"+string(debug.Stack()), "err", err)
```
ehnuje, line 49: How about using `ErrorWithStack`?
```suggestion
				logger.ErrorWithStack("NewRelic http handler panic", "err", err)
```

networks/rpc/http_newrelic.go
```diff
@@ -0,0 +1,136 @@
...
+			if err := json.Unmarshal(dupW.body.Bytes(), &rpcReturns); err == nil {
+				for i, rpcReturn := range rpcReturns {
+					if data, err := json.Marshal(rpcReturn); err == nil {
+						printRPCErrorLog(data, reqs[i].method, r)
+					}
+				}
+			}
+		} else {
+			printRPCErrorLog(dupW.body.Bytes(), reqs[0].method, r)
+		}
```
ehnuje, line 95: Can't we always unmarshal `dupW.body.Bytes()` with `var rpcReturns []interface{}`? (meaning always take `isBatch` branch`) 
```suggestion
		var rpcReturns []interface{}
		if err := json.Unmarshal(dupW.body.Bytes(), &rpcReturns); err == nil {
			for i, rpcReturn := range rpcReturns {
				if data, err := json.Marshal(rpcReturn); err == nil {
					printRPCErrorLog(data, reqs[i].method, r)
				}
			}
		}
```

ehnuje (approved):

networks/rpc/http_newrelic.go
```diff
@@ -0,0 +1,136 @@
...
+		defer func() {
+			if err := recover(); err != nil {
+				logger.Error("NewRelic http handler panic\n"+string(debug.Stack()), "err", err)
+			}
+		}()
+
+		// parse RPC requests
+		reqs, isBatch, err := getRPCRequests(r)
+		if err != nil || len(reqs) < 1 {
+			logger.Error("failed to parse RPC request", "err", err, "len(reqs)", len(reqs))
```
jeongkyun-oh, line 56: Do we need to leave logs for all invalid RPC requests?

jeongkyun-oh (dismissed):

networks/rpc/http_newrelic.go
KimKyungup: `ErrorWithStack` is same with below code from `serverRequest`?
```
			const size = 64 << 10
			buf := make([]byte, size)
			buf = buf[:runtime.Stack(buf, false)]
			logger.Error(string(buf))
```

node/config.go
aidan-kwon: @ehnuje I removed these comments and make the node read os environment variables when NewRelic is configured

node/node.go
aidan-kwon: @ehnuje I removed these variables and make the node read os environment variables when NewRelic is configured

networks/rpc/http.go
aidan-kwon: @ehnuje I changed the logic to print a critical log if only one variable is set. Thanks!

networks/rpc/http_newrelic.go
aidan-kwon: @jeongkyun-oh KAS team wants to print all RPC logs including success cases as well as RPC error logs. Let's consider the proper log level for RPC requests later.

networks/rpc/http_newrelic.go
aidan-kwon: Thank you for the kind suggestion. However, it will fail to parse a single request. If the request is single, `for i, rpcReturn := range rpcReturns {` will iterate with keys of the json return. For example, `rpcReturn` will be `jsonrpc`, `id`, or `result` with your code and the following return. `json` is unmashalled as `[]interface` internally. and `the array of json` is unmarshalled as `[][]interface`.
```
{
    "jsonrpc": "2.0",
    "id": 73,
    "result": "0x0"
}
```

ehnuje (approved):
LGTM, thanks!

networks/rpc/http_newrelic.go
aidan-kwon: @KimKyungup You're right. The same action except for minor configurations. And, this code will handle the panic raised on the handler level code (before `serverRequest`).

jeongkyun-oh (approved):

networks/rpc/http_newrelic.go
KimKyungup: @aidan-kwon This PR prints logs for all success cases?
Which routine does print the success cases?

aidan-kwon: @KimKyungup It doesn't print success cases. I mean we need to consider it later since KAS Team requested it.

yoomee1313 (approved):

aidan-kwon:
@KimKyungup @winnie-byun PTAL

networks/rpc/http_newrelic.go
KimKyungup: I think printing all logs is not a good idea also.

KimKyungup (approved):

PR #957, opened 2021-05-07
Updated istanbul consensus engine to verify headers synchronously

jeongkyun-oh:
## Proposed changes

- **UPDATE**: The addresses cache is updated. Previously, the `ecrecover` method used an ARU cache with block hash as a key, but we need to cache all the addresses recovered from committed seals of a block. So, 65-bytes signatures are used as keys. Furthermore, the size of the cache is increased to 2048 * 30 (max number of downloaded blocks * approximate sub group size of validators).

- After the PR https://github.com/klaytn/klaytn/pull/920, the `Refresh` method must be called all the time to refresh proposer candidates as well as validators.
- The `Refresh` method wasn't always called before because the header verification concurrently occurs while downloading a batch of blocks. (If the header does not exist, it ignores to refresh)
- The `Refresh` method didn't work properly since it was failed to get staking information to verify headers. (It tries to call a contract method on a specific block which is not yet inserted)
- This PR updates the header verification synchronously, while the computation of addresses will be processed concurrently to prevent download performance.

- The simple synchronous verification performs poorly as below panel, so extracting the pub key from seals handled concurrently.
image

- The comparision sync test is attached.
EN0 : master branch (Klaytn v1.6.1+f695312c38)
EN1 : After this PR (Klaytn v1.6.1+d76876935e)
EN2 : Before this PR (Klaytn v1.6.1+37b05fe791)

screencapture-54-180-156-230-3000-d-KnJhY-iWz-klaytn-main-overview-2021-05-17-09_20_39

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1588,7 +1588,15 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		seals[i] = true
 	}
 
-	abort, results := bc.engine.VerifyHeaders(bc, headers, seals)
+	var (
+		abort   chan<- struct{}
+		results <-chan error
+	)
+	if bc.engine.CanVerifyHeadersConcurrently() {
+		abort, results = bc.engine.VerifyHeaders(bc, headers, seals)
+	} else {
+		abort, results = bc.engine.PreprocessHeaderVerification(headers)
+	}
 	defer close(abort)
 
 	// Start a parallel signature recovery (signer will fluke on fork transition, minimal perf loss)
@@ -1647,6 +1655,10 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		bstart := time.Now()
 
 		err := <-results
+		if !bc.engine.CanVerifyHeadersConcurrently() && err == nil {
+			err = bc.engine.VerifyHeader(bc, block.Header(), true)
+		}
+
 		if err == nil {
 			err = bc.validator.ValidateBody(block)
 		}
--- a/consensus/clique/clique.go
+++ b/consensus/clique/clique.go
@@ -245,7 +245,7 @@ func (c *Clique) CanVerifyHeadersConcurrently() bool {
 	return true
 }
 
-// PreprocessHeaderVerification prepare header verification for heavy computation before synchronous header verification such as ecrecover.
+// PreprocessHeaderVerification prepares header verification for heavy computation before synchronous header verification such as ecrecover.
 func (c *Clique) PreprocessHeaderVerification(headers []*types.Header) (chan<- struct{}, <-chan error) {
 	panic("this method is not used for Clique engine")
 }
--- a/consensus/clique/clique.go
+++ b/consensus/clique/clique.go
@@ -240,6 +240,16 @@ func (c *Clique) Author(header *types.Header) (common.Address, error) {
 	return ecrecover(header, c.signatures)
 }
 
+// CanVerifyHeadersConcurrently returns true if concurrent header verification possible, otherwise returns false.
+func (c *Clique) CanVerifyHeadersConcurrently() bool {
+	return true
+}
+
+// PreprocessHeaderVerification prepare header verification for heavy computation before synchronous header verification such as ecrecover.
+func (c *Clique) PreprocessHeaderVerification(headers []*types.Header) (chan<- struct{}, <-chan error) {
+	panic("this method is not used for Clique engine")
+}
+
 // VerifyHeader checks whether a header conforms to the consensus rules.
 func (c *Clique) VerifyHeader(chain consensus.ChainReader, header *types.Header, seal bool) error {
 	return c.verifyHeader(chain, header, nil)
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -63,6 +63,12 @@ type Engine interface {
 	// block.
 	Author(header *types.Header) (common.Address, error)
 
+	// CanVerifyHeadersConcurrently returns true if concurrent header verification possible, otherwise returns false.
+	CanVerifyHeadersConcurrently() bool
+
+	// PreprocessHeaderVerification prepare header verification for heavy computation before synchronous header verification such as ecrecover.
+	PreprocessHeaderVerification(headers []*types.Header) (chan<- struct{}, <-chan error)
+
 	// VerifyHeader checks whether a header conforms to the consensus rules of a
 	// given engine. Verifying the seal may be done optionally here, or explicitly
 	// via the VerifySeal method.
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -66,7 +66,7 @@ type Engine interface {
 	// CanVerifyHeadersConcurrently returns true if concurrent header verification possible, otherwise returns false.
 	CanVerifyHeadersConcurrently() bool
 
-	// PreprocessHeaderVerification prepare header verification for heavy computation before synchronous header verification such as ecrecover.
+	// PreprocessHeaderVerification prepares header verification for heavy computation before synchronous header verification such as ecrecover.
 	PreprocessHeaderVerification(headers []*types.Header) (chan<- struct{}, <-chan error)
 
 	// VerifyHeader checks whether a header conforms to the consensus rules of a
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -96,8 +96,8 @@ var (
 	nonceAuthVote = hexutil.MustDecode("0xffffffffffffffff") // Magic nonce number to vote on adding a new validator
 	nonceDropVote = hexutil.MustDecode("0x0000000000000000") // Magic nonce number to vote on removing a validator.
 
-	inmemoryBlocks             = 100 // Number of blocks to precompute validators' addresses
-	inmemoryValidatorsPerBlock = 30  // Approximate number of validators' addresses from ecrecover
+	inmemoryBlocks             = 2048 // Number of blocks to precompute validators' addresses
+	inmemoryValidatorsPerBlock = 30   // Approximate number of validators' addresses from ecrecover
 	recentAddresses, _         = lru.NewARC(inmemoryBlocks)
 	signatureAddresses, _      = lru.New(inmemoryBlocks * inmemoryValidatorsPerBlock)
 )
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -98,9 +98,22 @@ var (
 
 	inmemoryBlocks             = 2048 // Number of blocks to precompute validators' addresses
 	inmemoryValidatorsPerBlock = 30   // Approximate number of validators' addresses from ecrecover
-	recentAddresses, _         = lru.NewARC(inmemoryBlocks)
-	signatureAddresses, _      = lru.New(inmemoryBlocks * inmemoryValidatorsPerBlock)
+	signatureAddresses, _      = lru.NewARC(inmemoryBlocks * inmemoryValidatorsPerBlock)
 )
+
+// cacheSignatureAddresses extracts the address from the given data and signature and cache them for later usage.
+func cacheSignatureAddresses(data []byte, sig []byte) (common.Address, error) {
+	sigStr := hex.EncodeToString(sig)
+	if addr, ok := signatureAddresses.Get(sigStr); ok {
+		return addr.(common.Address), nil
+	}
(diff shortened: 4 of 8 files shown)


consensus/clique/clique.go
```diff
@@ -240,6 +240,16 @@ func (c *Clique) Author(header *types.Header) (common.Address, error) {
 	return ecrecover(header, c.signatures)
 }
 
+// CanVerifyHeadersConcurrently returns true if concurrent header verification possible, otherwise returns false.
+func (c *Clique) CanVerifyHeadersConcurrently() bool {
+	return true
+}
+
+// PreprocessHeaderVerification prepare header verification for heavy computation before synchronous header verification such as ecrecover.
```
ehnuje, line 248:
```suggestion
// PreprocessHeaderVerification prepares header verification for heavy computation before synchronous header verification such as ecrecover.
```

consensus/consensus.go
```diff
@@ -63,6 +63,12 @@ type Engine interface {
 	// block.
 	Author(header *types.Header) (common.Address, error)
 
+	// CanVerifyHeadersConcurrently returns true if concurrent header verification possible, otherwise returns false.
+	CanVerifyHeadersConcurrently() bool
+
+	// PreprocessHeaderVerification prepare header verification for heavy computation before synchronous header verification such as ecrecover.
```
ehnuje, line 69:
```suggestion
	// PreprocessHeaderVerification prepares header verification for heavy computation before synchronous header verification such as ecrecover.
```

consensus/istanbul/backend/engine.go
```diff
@@ -95,15 +96,68 @@ var (
...
 func (sb *backend) Author(header *types.Header) (common.Address, error) {
 	return ecrecover(header)
 }
 
+// CanVerifyHeadersConcurrently returns true if concurrent header verification possible, otherwise returns false.
+func (sb *backend) CanVerifyHeadersConcurrently() bool {
+	return false
+}
+
+// PreprocessHeaderVerification prepare header verification for heavy computation before synchronous header verification such as ecrecover.
```
ehnuje, line 115:
```suggestion
// PreprocessHeaderVerification prepares header verification for heavy computation before synchronous header verification such as ecrecover.
```

consensus/istanbul/backend/engine.go
```diff
@@ -95,15 +96,68 @@ var (
...
+
+	proposalSeal := istanbulCore.PrepareCommittedSeal(header.Hash())
+	for _, seal := range istanbulExtra.CommittedSeal {
+		sealStr := hex.EncodeToString(seal)
+		if _, ok := signatureAddresses.Get(sealStr); ok {
+			continue
+		}
+		addr, err := istanbul.GetSignatureAddress(proposalSeal, seal)
+		if err != nil {
+			return err
```
ehnuje, line 154: How about returning `errInvalidSignature` in accordance with line 311?

ehnuje (commented):

blockchain/blockchain.go
```diff
@@ -1588,7 +1588,15 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
...
-	abort, results := bc.engine.VerifyHeaders(bc, headers, seals)
+	var (
+		abort   chan<- struct{}
+		results <-chan error
+	)
+	if bc.engine.CanVerifyHeadersConcurrently() {
+		abort, results = bc.engine.VerifyHeaders(bc, headers, seals)
+	} else {
+		abort, results = bc.engine.PreprocessHeaderVerification(headers)
+	}
```
aidan-kwon, line 1599: `PreprocessHeaderVerification` executes some parts of `VerifyHeaders` and the rest parts of the verfication are not executed if `bc.engine.CanVerifyHeadersConcurrently()` is false.

jeongkyun-oh: Right. `VerifyHeaders` is enough to implement the functionality, but the new method is introduced since the `PreprocessHeaderVerification` does not verify headers. Any suggestion?

consensus/clique/clique.go
jeongkyun-oh: Updated the comment. Thanks.

consensus/consensus.go
jeongkyun-oh: Updated the comment. Thanks.

consensus/istanbul/backend/engine.go
jeongkyun-oh: Updated the comment. Thanks.

consensus/istanbul/backend/engine.go
jeongkyun-oh: Reflected. Thanks :)

consensus/istanbul/backend/engine.go
```diff
@@ -95,15 +96,68 @@ var (
...
 
+// CanVerifyHeadersConcurrently returns true if concurrent header verification possible, otherwise returns false.
+func (sb *backend) CanVerifyHeadersConcurrently() bool {
+	return false
+}
+
+// PreprocessHeaderVerification prepares header verification for heavy computation before synchronous header verification such as ecrecover.
+func (sb *backend) PreprocessHeaderVerification(headers []*types.Header) (chan<- struct{}, <-chan error) {
+	abort := make(chan struct{})
+	results := make(chan error, inmemoryBlocks)
```
aidan-kwon, line 118: Why did you use `inmemoryBlocks` instead of `len(headers)`?

aidan-kwon: Is 100 * 30 enough cache size? Did downloader send thousands of blocks at one time?

consensus/istanbul/backend/engine.go
```diff
@@ -95,15 +96,68 @@ var (
 	nonceAuthVote = hexutil.MustDecode("0xffffffffffffffff") // Magic nonce number to vote on adding a new validator
 	nonceDropVote = hexutil.MustDecode("0x0000000000000000") // Magic nonce number to vote on removing a validator.
 
-	inmemoryAddresses  = 20 // Number of recent addresses from ecrecover
-	recentAddresses, _ = lru.NewARC(inmemoryAddresses)
+	inmemoryBlocks             = 100 // Number of blocks to precompute validators' addresses
+	inmemoryValidatorsPerBlock = 30  // Approximate number of validators' addresses from ecrecover
+	recentAddresses, _         = lru.NewARC(inmemoryBlocks)
+	signatureAddresses, _      = lru.New(inmemoryBlocks * inmemoryValidatorsPerBlock)
```
aidan-kwon, line 102: How about combing them into one cache and set/get items insides of `GetSignatureAddress`?

consensus/istanbul/backend/engine.go
jeongkyun-oh: > Why did you use inmemoryBlocks instead of len(headers)?

This is for throttling purpose. If the computing is way faster and it replaces the already computed ones from cache, this precomputing is not useful. It may use more memory, but I think it is not necessary.

> Is 100 * 30 enough cache size? Did downloader send thousands of blocks at one time?

So, the throttling header verification every 100 blocks, and each block may contain 30 signatures at most. So, I put that number.

consensus/istanbul/backend/engine.go
```diff
@@ -95,15 +96,68 @@ var (
...
+	// Retrieve the signature from the header extra-data
+	istanbulExtra, err := types.ExtractIstanbulExtra(header)
+	if err != nil {
+		return err
+	}
+
+	proposalSeal := istanbulCore.PrepareCommittedSeal(header.Hash())
+	for _, seal := range istanbulExtra.CommittedSeal {
+		sealStr := hex.EncodeToString(seal)
+		if _, ok := signatureAddresses.Get(sealStr); ok {
```
KimKyungup, line 149: How about using seal as a key for `signatureAddresses`?
```suggestion
		if _, ok := signatureAddresses.Get(seal); ok {
```

consensus/istanbul/backend/engine.go
```diff
@@ -95,15 +96,68 @@ var (
...
+			continue
+		}
+		addr, err := istanbul.GetSignatureAddress(proposalSeal, seal)
+		if err != nil {
+			return errInvalidSignature
+		}
+		signatureAddresses.Add(sealStr, addr)
```
KimKyungup, line 156:
```suggestion
		addr, err := istanbul.GetSignatureAddress(proposalSeal, seal)
		if err != nil {
			return errInvalidSignature
		}
		signatureAddresses.Add(seal, addr)
```

consensus/istanbul/backend/engine.go
```diff
@@ -95,15 +96,68 @@ var (
 	nonceAuthVote = hexutil.MustDecode("0xffffffffffffffff") // Magic nonce number to vote on adding a new validator
 	nonceDropVote = hexutil.MustDecode("0x0000000000000000") // Magic nonce number to vote on removing a validator.
 
-	inmemoryAddresses  = 20 // Number of recent addresses from ecrecover
-	recentAddresses, _ = lru.NewARC(inmemoryAddresses)
+	inmemoryBlocks             = 100 // Number of blocks to precompute validators' addresses
```
KimKyungup, line 99: This value is enough? 
Could you check the all block header can be preprocessed in normal cases?

```suggestion
	inmemoryBlocks             = 100 // Number of blocks to precompute validators' addresses
```

consensus/istanbul/backend/engine.go
aidan-kwon: I mean throttling 100 blocks doesn't seem to be enough. Downloader could send 2048 blocks at once, but it only preprocesses 100 blocks.

jeongkyun-oh: @aidan-kwon I will increase the number to 2048. I thought the cache is not meaningful after syncing, so the space would be wasted (although only about 4MB is used).

consensus/istanbul/backend/engine.go
jeongkyun-oh: @KimKyungup The byte slice is not hashable, so the raw seal cannot be used. Any other suggestion?

consensus/istanbul/backend/engine.go
jeongkyun-oh: Same reason with the above comment.

consensus/istanbul/backend/engine.go
jeongkyun-oh: I will increase size to 2048 suggested in https://github.com/klaytn/klaytn/pull/957/files#r629000344

consensus/istanbul/backend/engine.go
jeongkyun-oh: I think combining is fine, but the `GetSignatureAddress` is an util function, so putting the cache inside is not good. What do you think?

aidan-kwon: It makes sense. You might create a wrapper function such as `CacheSignatureAddress` or put the caching logic outside as it was.

jeongkyun-oh: I have added `cacheSignatureAddress` for caching. PTAL.

consensus/istanbul/backend/engine.go
KimKyungup: @jeongkyun-oh 
It seems that the key can be byte slice like below? Isn't it?
`hash` is byte array.
`recentAddresses.Add(hash, addr)`

consensus/istanbul/backend/engine.go
```diff
@@ -673,22 +735,16 @@ func sigHash(header *types.Header) (hash common.Hash) {
 
 // ecrecover extracts the Klaytn account address from a signed header.
 func ecrecover(header *types.Header) (common.Address, error) {
-	hash := header.Hash()
```
KimKyungup, line 676: @jeongkyun-oh Why did you remove this code?

consensus/istanbul/backend/engine.go
aidan-kwon: Thanks. I think `len(headers)` is more suitable for the length of `results` channel, though. `inmemoryBlocks` looks fine also.

jeongkyun-oh:
@ehnuje @winnie-byun @yoomee1313 Please take a look.

yoomee1313 (approved):
lgtm

winnie-byun (approved):
It would be nice to put explanation. Nice work:)
EN0 : master branch (Klaytn v1.6.1+f695312c38)
EN1 : After this PR (Klaytn v1.6.1+d76876935e)
EN2 : Before this PR (Klaytn v1.6.1+37b05fe791)

ehnuje (approved):

PR #974, opened 2021-05-20
Updated governance APIs

jeongkyun-oh:
## Proposed changes

The governance-related APIs are updated. The related definitions are as follows.

- Council: the authorized members to be validators and proposers regardless of the staking amount.
  - `getCouncil`: the addresses of council members at the given block.
  - `getCouncilSize`: the number of council members at the given block.
- Validator: the member of council which is a candidate of block proposer and validators. Only the members who have enough KLAYs are qualified.
  - `getValidators`: the validators at the given block number.
  - `getValidatorsAtHash`: the validators at the given block hash.
- DemotedValidator: the permissioned validator, but not participating validation and proposal since the staked KLAYs are not enough.
  - `getDemotedValidators`: the demoted validators at the given block number.
  - `getDemotedValidatorsAtHash`: the demoted validators at the given block hash.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/consensus/istanbul/backend/api.go
+++ b/consensus/istanbul/backend/api.go
@@ -66,37 +66,40 @@ func (api *API) GetSnapshotAtHash(hash common.Hash) (*Snapshot, error) {
 	return api.istanbul.snapshot(api.chain, header.Number.Uint64(), header.Hash(), nil)
 }
 
-// GetValidators retrieves the list of authorized validators at the specified block.
+// GetValidators retrieves the list of authorized validators with the given block number.
 func (api *API) GetValidators(number *rpc.BlockNumber) ([]common.Address, error) {
-	// Retrieve the requested block number (or current if none requested)
-	var header *types.Header
-	if number == nil || *number == rpc.LatestBlockNumber {
-		header = api.chain.CurrentHeader()
-	} else {
-		header = api.chain.GetHeaderByNumber(uint64(number.Int64()))
-	}
-	// Ensure we have an actually valid block and return the validators from its snapshot
-	if header == nil {
-		return nil, errUnknownBlock
-	}
-	snap, err := api.istanbul.snapshot(api.chain, header.Number.Uint64(), header.Hash(), nil)
+	snap, err := api.GetSnapshot(number)
 	if err != nil {
 		return nil, err
 	}
 	return snap.validators(), nil
 }
 
-// GetValidatorsAtHash retrieves the state snapshot at a given block.
+// GetValidatorsAtHash retrieves the list of authorized validators with the given block hash.
 func (api *API) GetValidatorsAtHash(hash common.Hash) ([]common.Address, error) {
-	header := api.chain.GetHeaderByHash(hash)
-	if header == nil {
-		return nil, errUnknownBlock
-	}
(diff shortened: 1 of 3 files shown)


consensus/istanbul/backend/api.go
```diff
@@ -66,39 +66,42 @@ func (api *API) GetSnapshotAtHash(hash common.Hash) (*Snapshot, error) {
...
+	snap, err := api.GetSnapshotAtHash(hash)
 	if err != nil {
 		return nil, err
 	}
 	return snap.validators(), nil
 }
 
+// GetDemotedValidators retrieves the list of authorized, but demoted validators with the given block number.
+func (api *API) GetDemotedValidators(number *rpc.BlockNumber) ([]common.Address, error) {
+	snap, err := api.GetSnapshot(number)
```
hyochan-brown, line 89: Is it okay to call api in api?
I'm curious about other people's opinions.

hyochan-brown (commented):
LGTM!
I have a one question.

consensus/istanbul/backend/snapshot.go
```diff
@@ -243,6 +243,15 @@ func (s *Snapshot) validators() []common.Address {
 	return sortValidatorArray(validators)
 }
 
+// demotedValidators retrieves the list of authorized but demoted validators in ascending order.
```
yoomee1313, line 246: What about adding comma like `GetDemotedValidatorsAtHash` comment?
```suggestion
// demotedValidators retrieves the list of authorized, but demoted validators in ascending order.
```

consensus/istanbul/backend/api.go
yoomee1313: I think it's okay.

aidan-kwon (dismissed):

consensus/istanbul/backend/api.go
jeongkyun-oh: I think it makes it simpler and less complex and it is fine logically as well. If you have any other suggestion, I appreciate for your opinion.

yoomee1313 (approved):

consensus/istanbul/backend/api.go
hyochan-brown: This is just my opinion that it is awkward for API to call API.
I don't want it to sound rude.
1. API is code written for external communication. It feels awkward to be called by code from inside.
2. The GetDemotedValidators API will be affected if modifications or deletions are made to the GetSnapshot API. I think this will be a less flexible structure.
3. So If there is a duplicate code between multiple APIs, I think it is appropriate to separate that part into other method.

jeongkyun-oh: @hyochan-brown Thanks for your opinion. According to the other cases, I think this would be fine. There are already many cases for the API structure used its method. I think it makes more consistent with other APIs. What do you think about this?

`s.DoEstimateGas`
`s.ChainId()`
`s.SendTransaction`
`s.SendTransactionAsFeePayer`
`s.SignAndSendTransaction`
...

hyochan-brown: Oh..
Consistency is more important than anything else, so I think it's good. 
Thank you for your kind reply.

hyochan-brown (approved):

aidan-kwon (approved):

PR #977, opened 2021-05-26
Prevent nil receipts and added retry logic for making chain events

jeongkyun-oh:
## Proposed changes

Making a chain event is done locally after getting items from database. It was hard to get an error, but the error case is not handled properly. This change checks if receipt list is nil or not, and added retrying logic for making chain events.

Unlike retrying to produce Kafka messages, it should not fail to get data from the database. It retries inifinately, but after 100 retry it leaves error logs.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/datasync/chaindatafetcher/chaindata_fetcher.go
+++ b/datasync/chaindatafetcher/chaindata_fetcher.go
@@ -273,9 +273,13 @@ func (f *ChainDataFetcher) makeChainEventWithRetry(blockNumber uint64) (blockcha
 		case <-f.stopCh:
 			return ev, err
 		default:
-			logger.Warn("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
-			if retryCount >= retryMax {
-				logger.Error("the max retry exceeded")
+			if retryCount < retryMax {
+				logger.Warn("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
+			}
+
+			// leaves an error log periodically after max retry
+			if retryCount%retryMax == 0 {
+				logger.Error("max retry exceeded. retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
 			}
 			time.Sleep(retryInterval)
 			ev, err = f.makeChainEvent(blockNumber)
--- a/datasync/chaindatafetcher/chaindata_fetcher.go
+++ b/datasync/chaindatafetcher/chaindata_fetcher.go
@@ -268,13 +268,13 @@ func (f *ChainDataFetcher) retryMakeChainEvent(blockNumber uint64) (blockchain.C
 		retryInterval = 1 * time.Second
 	)
 	ev, err := f.makeChainEvent(blockNumber)
-	for i := 1; err != nil; i++ {
+	for retryCount := 1; err != nil; retryCount++ {
 		select {
 		case <-f.stopCh:
 			return ev, err
 		default:
-			logger.Warn("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", i, "retryMax", retryMax, "err", err)
-			if i >= retryMax {
+			logger.Warn("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
+			if retryCount >= retryMax {
 				logger.Error("the max retry exceeded")
 			}
 			time.Sleep(retryInterval)
--- a/datasync/chaindatafetcher/chaindata_fetcher.go
+++ b/datasync/chaindatafetcher/chaindata_fetcher.go
@@ -265,15 +265,18 @@ func (f *ChainDataFetcher) stopRangeFetching() error {
 func (f *ChainDataFetcher) retryMakeChainEvent(blockNumber uint64) (blockchain.ChainEvent, error) {
 	var (
 		retryMax      = 100
-		retryInterval = 300 * time.Millisecond
+		retryInterval = 1 * time.Second
 	)
 	ev, err := f.makeChainEvent(blockNumber)
-	for i := 0; i < retryMax && err != nil; i++ {
+	for i := 1; err != nil; i++ {
 		select {
 		case <-f.stopCh:
 			return ev, err
 		default:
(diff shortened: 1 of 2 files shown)


datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -262,13 +262,38 @@ func (f *ChainDataFetcher) stopRangeFetching() error {
...
+		retryMax      = 100
+		retryInterval = 1 * time.Second
+	)
+	ev, err := f.makeChainEvent(blockNumber)
+	for i := 1; err != nil; i++ {
+		select {
+		case <-f.stopCh:
+			return ev, err
+		default:
+			logger.Warn("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", i, "retryMax", retryMax, "err", err)
```
hyochan-brown, line 276: I wonder why you named it "i", not "retryCount".

hyochan-brown (approved):
LGTM!!
I have only one question.

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: I think `retryCount` is better to read. Reflected, thanks :)

aidan-kwon (dismissed):

ehnuje (approved):

datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -262,13 +262,38 @@ func (f *ChainDataFetcher) stopRangeFetching() error {
 	return nil
 }
 
+func (f *ChainDataFetcher) retryMakeChainEvent(blockNumber uint64) (blockchain.ChainEvent, error) {
```
winnie-byun, line 265:
```suggestion
func (f *ChainDataFetcher) makeChainEventWithRetry(blockNumber uint64) (blockchain.ChainEvent, error) {
```
`retryMakeChainEvent` sounds like a chain event already occured.

datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -262,13 +262,38 @@ func (f *ChainDataFetcher) stopRangeFetching() error {
...
+	)
+	ev, err := f.makeChainEvent(blockNumber)
+	for retryCount := 1; err != nil; retryCount++ {
+		select {
+		case <-f.stopCh:
+			return ev, err
+		default:
+			logger.Warn("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
+			if retryCount >= retryMax {
+				logger.Error("the max retry exceeded")
```
winnie-byun, line 278: * Shouldn't this case be paused?
* When this happens, it would be hard to check the log file. How about making this log every 100 retries, for example?

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: I agreed. Reflected :)

datasync/chaindatafetcher/chaindata_fetcher.go
jeongkyun-oh: - As I mentioned in the description, I wanted to retry the `makeChainEvent` infinitely. So, the code is not returned, but it leaves error log after the threshold.
- Do you think that this is too verbose? I think it is fine since the interval is 1 second. How about increasing log level instead of the `Warn` after `retryMax` instead?

aidan-kwon (approved):

ehnuje (approved):

datasync/chaindatafetcher/chaindata_fetcher.go
winnie-byun: I don't think the changing the logging level doesn't affect much.
I think it is verbose to print every second. It would be the only log we could find when this happens. Could you please change it to log every 10 seconds at least.

jeongkyun-oh: @winnie-byun I just updated the log frequencies. It leaves warning logs for the number of max retry. After that it leaves an error log every 100 retries.

datasync/chaindatafetcher/chaindata_fetcher.go
```diff
@@ -262,13 +262,42 @@ func (f *ChainDataFetcher) stopRangeFetching() error {
...
+		case <-f.stopCh:
+			return ev, err
+		default:
+			if retryCount < retryMax {
+				logger.Warn("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
+			}
+
+			// leaves an error log periodically after max retry
+			if retryCount%retryMax == 0 {
+				logger.Error("max retry exceeded. retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
```
winnie-byun, line 282:
```suggestion
				logger.Warn("max retry exceeded. retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
			}

			// leaves an error log periodically after max retry
			if retryCount%retryMax == 1 {
				logger.Error("retrying to make a chain event...", "blockNumber", blockNumber, "retryCount", retryCount, "retryMax", retryMax, "err", err)
```
* I think the log is switched
* Changing 0 to 1 could make the log at the first try.

jeongkyun-oh: @winnie-byun Umm. I think the first retries warn that it is retrying. After that, the max retries exceeded and we consider it error situation, so it leaves error log.

winnie-byun: @jeongkyun-oh Sorry My bad

winnie-byun (approved):

KimKyungup:
@jeongkyun-oh What is the root cause of not having a receipt?

jeongkyun-oh:
@KimKyungup Unexpected errors occur in DynamoDB unlike LevelDB. The retry logic is added for unexpected temporary errors.

jeongkyun-oh:
Dear reviewers, sorry for confusion.

For now, https://github.com/klaytn/klaytn/pull/982 PR fixes the current issue. 
For retry, I will make another PR in the near future.
I made another PR for checking `nil` receipts in https://github.com/klaytn/klaytn/pull/983.

PR #1008, opened 2021-07-09
Implemented error callback and expiration of message

jeongkyun-oh:
## Proposed changes

- Make `Setup` / `Cleanup` method for sarama library configurable
- Add `ExpirationTime` for a messsage to be expired if it is in buffer too long
- Add `ErrCallback` in order to handle error situation

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/datasync/chaindatafetcher/kafka/consumer.go
+++ b/datasync/chaindatafetcher/kafka/consumer.go
@@ -49,6 +49,7 @@ var (
 	noHandlerErrorMsg          = "the handler does not exist for the given topic"
 	emptySegmentErrorMsg       = "there is no segment in the segment slice"
 	bufferOverflowErrorMsg     = "the number of items in buffer exceeded the maximum"
+	msgExpiredErrorMsg         = "the message is expired"
 )
 
 // TopicHandler is a handler function in order to consume published messages.
@@ -189,7 +190,6 @@ func (c *Consumer) Subscribe(ctx context.Context) error {
 			Logger.Printf("[ERROR] the consumption is failed [err: %s]\n", err.Error())
 			return err
 		}
-		// TODO-Chaindatafetcher add retry logic and error callback here
 	}
 }
 
@@ -307,6 +307,22 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
 	return nil
 }
 
+// renewExpireMsg updates expiration timer and message if the expiring message is handled.
+func (c *Consumer) renewExpireMsg(buffer [][]*Segment, timer *time.Timer, oldestMsg *sarama.ConsumerMessage) *sarama.ConsumerMessage {
+	if c.config.ExpirationTime > time.Duration(0) {
+		if len(buffer) == 0 {
+			timer.Stop()
+			return nil
+		} else if oldestMsg != buffer[0][0].orig {
+			timer.Reset(c.config.ExpirationTime)
+			return buffer[0][0].orig
+		} else {
+			return oldestMsg
+		}
+	}
+	return nil
+}
+
 // ConsumeClaim must start a consumer loop of ConsumerGroupClaim's Messages().
 // Once the Messages() channel is closed, the Handler must finish its processing
 // loop and exit.
@@ -332,7 +348,7 @@ func (c *Consumer) consumeClaim(cgs sarama.ConsumerGroupSession, cgc sarama.Cons
 	for {
 		select {
 		case <-expireTimer.C:
-			return buffer, errors.New("the message is expired")
+			return buffer, errors.New(msgExpiredErrorMsg)
 		case msg, ok := <-cgc.Messages():
 			if !ok {
 				return buffer, nil
@@ -359,15 +375,8 @@ func (c *Consumer) consumeClaim(cgs sarama.ConsumerGroupSession, cgc sarama.Cons
 				return buffer, err
 			}
 
-			if c.config.ExpirationTime > time.Duration(0) {
-				if len(buffer) == 0 {
-					expireTimer.Stop()
-					expireMsg = nil
-				} else if expireMsg != buffer[0][0].orig {
-					expireTimer.Reset(c.config.ExpirationTime)
-					expireMsg = buffer[0][0].orig
-				}
-			}
+			// update expire message and timer if it is handled
+			expireMsg = c.renewExpireMsg(buffer, expireTimer, expireMsg)
 
 			// mark offset of the oldest message to be read
 			if err := c.updateOffset(buffer, msg, cgs); err != nil {
@@ -377,7 +386,7 @@ func (c *Consumer) consumeClaim(cgs sarama.ConsumerGroupSession, cgc sarama.Cons
 	}
 }
 
-func (c *Consumer) handleError(buffer [][]*Segment, cgs sarama.ConsumerGroupSession, parentErr error) error {
+func (c *Consumer) handleError(buffer [][]*Segment, cgs ConsumerGroupSession, parentErr error) error {
 	if len(buffer) > 0 {
 		oldestMsg := buffer[0][0].orig
 		key := string(oldestMsg.Key)
--- a/datasync/chaindatafetcher/kafka/consumer.go
+++ b/datasync/chaindatafetcher/kafka/consumer.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"time"
 
 	"github.com/Shopify/sarama"
 )
@@ -197,13 +198,13 @@ func (c *Consumer) Subscribe(ctx context.Context) error {
 
 // Setup is called at the beginning of a new session, before ConsumeClaim.
 func (c *Consumer) Setup(s sarama.ConsumerGroupSession) error {
-	return nil
+	return c.config.Setup(s)
 }
 
 // Cleanup is called at the end of a session, once all ConsumeClaim goroutines have exited
 // but before the offsets are committed for the very last time.
 func (c *Consumer) Cleanup(s sarama.ConsumerGroupSession) error {
-	return nil
+	return c.config.Cleanup(s)
 }
 
 // insertSegment inserts the given segment to the given buffer.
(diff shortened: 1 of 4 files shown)


datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
...
-		// handle the buffered messages if any message can be reassembled
-		buffer, err = c.handleBufferedMessages(buffer)
-		if err != nil {
-			return err
+	for {
+		select {
+		case <-expireTimer.C:
+			return buffer, errors.New(msgExpiredErrorMsg)
+		case msg, ok := <-cgc.Messages():
+			if !ok {
```
yoomee1313, line 353: @jeongkyun-oh It returns `buffer, nil` when ok is false.. is it okay to return like that?

datasync/chaindatafetcher/kafka/consumer_test.go
```diff
@@ -556,3 +558,157 @@ func Test_V1Segment_MultiProducers(t *testing.T) {
...
+		{
+			"callback_success_an_item_in_buffer",
+			func() { session.EXPECT().MarkMessage(gomock.Eq(testSegment.orig), gomock.Eq("")).Times(1) },
+			func(s string) error { return nil },
+			[][]*Segment{{testSegment}},
+			parentErr,
+			nil,
+		},
+		{
+			"callback_success_two_items_in_buffer",
```
yoomee1313, line 618: Can I ask why you add this tc? because i think `callback_success_two_items_in_buffer` tc and `callback_failure` tc have same purpose..

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: @yoomee1313 here, okay is false when the channel is closed. So, if message channel is closed, then it is returned as a normal case.
https://tour.golang.org/concurrency/4

datasync/chaindatafetcher/kafka/consumer_test.go
jeongkyun-oh: @yoomee1313 Oh, `callback_success_two_items_in_buffer` is a successful case by calling error callback. On the other hand, the `callback_failure` is a failure case although the error callback is called.

datasync/chaindatafetcher/kafka/consumer.go
yoomee1313: Oh, i didn't know that channel closing is normal case. Thanks!

datasync/chaindatafetcher/kafka/consumer_test.go
yoomee1313: Oh, setup is different.. thanks for the explanation!

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
 	return nil
 }
 
+// renewExpireMsg updates expiration timer and message if the expiring message is handled.
+func (c *Consumer) renewExpireMsg(buffer [][]*Segment, timer *time.Timer, oldestMsg *sarama.ConsumerMessage) *sarama.ConsumerMessage {
+	if c.config.ExpirationTime > time.Duration(0) {
+		if len(buffer) == 0 {
+			timer.Stop()
+			return nil
+		} else if oldestMsg != buffer[0][0].orig {
```
yoomee1313, line 316: Is `buffer` is affected when timer.Reset is executed? If not, what about changing the code to return only `buffer[0][0].orig`?
```
if oldestMsg != buffer[0][0].orig {
    timer.Reset(c.config.ExpirationTime)
}
return buffer[0][0].orig
```

yoomee1313 (commented):
it looks good to me except a minor comment

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
...
+		oldestMsg := buffer[0][0].orig
+		key := string(oldestMsg.Key)
+
+		if c.config.ErrCallback != nil {
+			if err := c.config.ErrCallback(key); err != nil {
+				return err
+			}
+
+			buffer = buffer[1:]
+			return c.updateOffset(buffer, oldestMsg, cgs)
```
aidan-kwon, line 400: The error can occur while consuming a message not placed in buffer[0], but this method always removes buffer[0]. Is that you intended?

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
...
+			return nil
+		} else if oldestMsg != buffer[0][0].orig {
+			timer.Reset(c.config.ExpirationTime)
+			return buffer[0][0].orig
+		} else {
+			return oldestMsg
+		}
+	}
+	return nil
+}
```
hyochan-brown, line 324: I'm sorry I couldn't give you a good review.
How do you think of the style below?
I learned that more else makes it less legible in Clean code.
```
func (c *Consumer) renewExpireMsg(buffer [][]*Segment, timer *time.Timer, oldestMsg *sarama.ConsumerMessage) *sarama.ConsumerMessage {
	if c.config.ExpirationTime <= time.Duration(0) {
		return nil
	}
	if len(buffer) == 0 {
		timer.Stop()
		return nil
	}
	if oldestMsg != buffer[0][0].orig {
		timer.Reset(c.config.ExpirationTime)
		return buffer[0][0].orig
	}
	return oldestMsg
}
```

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
...
+func (c *Consumer) renewExpireMsg(buffer [][]*Segment, timer *time.Timer, oldestMsg *sarama.ConsumerMessage) *sarama.ConsumerMessage {
+	if c.config.ExpirationTime > time.Duration(0) {
+		if len(buffer) == 0 {
+			timer.Stop()
+			return nil
+		} else if oldestMsg != buffer[0][0].orig {
+			timer.Reset(c.config.ExpirationTime)
+			return buffer[0][0].orig
+		} else {
+			return oldestMsg
```
aidan-kwon, line 320: Why time is not reset in this case?

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Applied. Thanks :)

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: @yoomee1313 The code is updated, could you take a look once more?

yoomee1313: @jeongkyun-oh How about returning buffer[0][0].orig instead of returning oldMsg? It would be helpful to understand the logic easily. As long as I understood, buffer[0][0].orig is always the oldest msg, right?

jeongkyun-oh: @yoomee1313 Oh, they may not be the same. So, if they are the same, it returns `buffer[0][0].orig` without resetting the timer. Otherwise, it returns `oldestMsg` with restting the timer.

aidan-kwon (approved):
I think removing the oldest message when an error occurs on the latest message is a little bit confusing. But, this feature enables auto-recovery from consumer errors and can be improved later with other users' feedback.

yoomee1313 (dismissed):

aidan-kwon (approved):

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
 	return nil
 }
 
+// renewExpireMsg updates expiration timer and message if the expiring message is handled.
+func (c *Consumer) renewExpireMsg(buffer [][]*Segment, timer *time.Timer, oldestMsg *sarama.ConsumerMessage) *sarama.ConsumerMessage {
```
ehnuje, line 311: Does `expire` mean `oldest` here? It looks confusing since you used variable `oldestMsg` here, but what is passed is `expireMsg`. 

Also, this function. does not `renew` a message, instead it resets the timer if the given oldestMsg is handled before this function is called (did I understand correctly?)

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
 	return nil
 }
 
+// renewExpireMsg updates expiration timer and message if the expiring message is handled.
```
ehnuje, line 310: It does not update the message, message is updated by the returned value of this function at caller-side.

datasync/chaindatafetcher/kafka/consumer.go
```diff
@@ -306,37 +307,99 @@ func (c *Consumer) updateOffset(buffer [][]*Segment, lastMsg *sarama.ConsumerMes
...
 
-		segment, err := newSegment(msg)
-		if err != nil {
-			return err
-		}
+func (c *Consumer) consumeClaim(cgs sarama.ConsumerGroupSession, cgc sarama.ConsumerGroupClaim) ([][]*Segment, error) {
+	var (
+		buffer      [][]*Segment // TODO-Chaindatafetcher better to introduce segment buffer structure with useful methods
+		expireMsg   *sarama.ConsumerMessage
+		expireTimer = time.NewTimer(c.config.ExpirationTime)
```
ehnuje, line 342: Doesn't it always go to line 353 with default value of 0?

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: 1. Right. So, actually the oldest message is the candidate to be expired.
2. Yes, it resets or stops the timer if necessary.

I updated the method names and variable. Please take a look.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Right. The timer is reset only, so I renamed it to `resetTimer`. PTAL.

datasync/chaindatafetcher/kafka/consumer.go
jeongkyun-oh: Initially, it consumes the first signal on lines 346-347.

ehnuje (approved):

yoomee1313 (approved):

aidan-kwon (approved):

PR #1009, opened 2021-07-13
blockchain: make chainConfig updatable by using hard-coded chainConfigs

yoomee1313:
## Proposed changes

- When `init` or `start`, SetupGenesisBlock function is used to generate/store genesis block, and return the chainConfig to use it in the node. However, chainConfig is not updatable with hard-coded chainConfigs unlike ethereum does. However, chainConfig must be updatable if we want to update hard fork block numbers.
- There's several important variables. 
  - `genesis` is the structure which contains information of genesis block and config. When `start`, it is always nil (when `init`, it is always not nil`)
  - `storedCfg` is the chainConfig derived from stored chainConfig.
  - `newCfg` is the chainConfig newly generated in SetupGenesisBlock. If `genesis` is not nil, it is derived from `genesis`.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/genesis.go
+++ b/blockchain/genesis.go
@@ -255,7 +255,7 @@ func SetupGenesisBlock(db database.DBManager, genesis *Genesis, networkId uint64
 	// Special case: don't change the existing config of a non-mainnet chain if no new
 	// config is supplied. These chains would get AllProtocolChanges (and a compat error)
 	// if we just continued here.
-	if genesis == nil && isPrivate {
+	if genesis == nil && params.CypressGenesisHash != stored && params.BaobabGenesisHash != stored {
 		return storedcfg, stored, nil
 	}
 
--- a/blockchain/genesis.go
+++ b/blockchain/genesis.go
@@ -255,7 +255,7 @@ func SetupGenesisBlock(db database.DBManager, genesis *Genesis, networkId uint64
 	// Special case: don't change the existing config of a non-mainnet chain if no new
 	// config is supplied. These chains would get AllProtocolChanges (and a compat error)
 	// if we just continued here.
-	if genesis == nil && stored != params.CypressGenesisHash {
+	if genesis == nil && isPrivate {
 		return storedcfg, stored, nil
 	}
 
@@ -277,6 +277,10 @@ func (g *Genesis) configOrDefault(ghash common.Hash) *params.ChainConfig {
 	switch {
 	case g != nil:
 		return g.Config
+	case ghash == params.CypressGenesisHash:
+		return params.CypressChainConfig
+	case ghash == params.BaobabGenesisHash:
+		return params.BaobabChainConfig
 	default:
 		return params.AllGxhashProtocolChanges
 	}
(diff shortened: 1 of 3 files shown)


yoomee1313:
I'll reopen this PR after decision is made.
option 1. make a new command to update only hard fork block number
option 2. use this PR

aidan-kwon (approved):

jeongkyun-oh (approved):

winnie-byun (approved):

blockchain/genesis.go
```diff
@@ -255,7 +255,7 @@ func SetupGenesisBlock(db database.DBManager, genesis *Genesis, networkId uint64
 	// Special case: don't change the existing config of a non-mainnet chain if no new
 	// config is supplied. These chains would get AllProtocolChanges (and a compat error)
 	// if we just continued here.
-	if genesis == nil && stored != params.CypressGenesisHash {
+	if genesis == nil && params.CypressGenesisHash != stored && params.BaobabGenesisHash != stored {
```
winnie-byun, line 258: Could you know why go-eth decided to check only the mainnet hash?
go-eth code : 
```
	// Special case: don't change the existing config of a non-mainnet chain if no new
	// config is supplied. These chains would get AllProtocolChanges (and a compat error)
	// if we just continued here.
	if genesis == nil && stored != params.MainnetGenesisHash {
		return storedcfg, stored, nil
	}
```

yoomee1313: @winnie-byun Oh, it's really good question. If you look at SetEthKlay, you can see that mainnet flag support is recent. https://github.com/ethereum/go-ethereum/pull/21932/commits/ae00010b95b3c6c92eb34d55bdee139b54ef7f30. However, above code is written at 2017. I think that checking the mainnetGenesisHash was necessary in the past, but it seems to be unnecessary now as the  https://github.com/ethereum/go-ethereum/pull/21932/commits/ae00010b95b3c6c92eb34d55bdee139b54ef7f30 enters. (when mainnet, genesis != not nil) What do you think?
Also, I did some tests about changing hard fork block number in the mainnet and private network without `sored != params.MainnetGenesisHash`, the same result came out :)

winnie-byun:
I don't see why this PR make chainConfig updatable.
Doesn't this line make that happen? :
```
	// Check whether the genesis block is already written.
	if genesis != nil {
		// If overwriteGenesis is true, overwrite existing genesis block with the new one.
		// This is to run a test with pre-existing data.
		if overwriteGenesis {
			headBlock := findBlockWithState(db)
			logger.Warn("Trying to overwrite original genesis block with the new one",
				"headBlockHash", headBlock.Hash().String(), "headBlockNum", headBlock.NumberU64())
			newGenesisBlock, err := genesis.Commit(headBlock.Root(), db)
			return genesis.Config, newGenesisBlock.Hash(), err
		}
```

yoomee1313:
@hyochan-brown please take a look

blockchain/genesis.go
winnie-byun:
```suggestion
	if genesis == nil && (params.CypressGenesisHash != stored || params.BaobabGenesisHash != stored) {
```

ehnuje (commented):
@yoomee1313 How about making tests to check if chainConfig becomes updatable with this change?

yoomee1313:
@ehnuje Oh, thanks for good idea. I'll add the unit test in later PR!!

PR #1018, opened 2021-07-29
Fixed governance votes application

jeongkyun-oh:
## Proposed changes

The multiple votes on consecutive epochs are not applied properly. The votes on the previous epoch are ignored. The order of governance items is not right and this PR updates it to handle votes properly.

This PR updated the following.
1. The governance items are applied on the `currentSet`, but it should be the one read from database.
2. The votes are not removed if they are not casted yet.
3. The governance items are applied first before the votes are handled on epoch block.

Briefly, the governance votes moves in the following order.
1. The user cast a vote by API call.
2. The vote is added to the governance structure `voteMap`.
3. When the node creates a block, the vote is attached to the block header `voteData`.
4. The block is propagated and inserted.
5. The snapshot is created and the vote is decoded and added to the `changeSet` of governance structure.
6. When a block is created on epoch period, the `changeSet` is encoded and added to the block header `governanceData`.
7. The block is propagated and inserted.
8. The snapshot is created and the governance data is decoded
9. The governance data is overwritten on the `currentSet` which is applied governance items at the moment.
10. The `currentSet` is updated to the set saved on the previous epoch.
11. It clears all temporary data including `voteMap` and `changeSet`.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/consensus/istanbul/backend/engine_test.go
+++ b/consensus/istanbul/backend/engine_test.go
@@ -840,22 +840,24 @@ func TestGovernance_Votes(t *testing.T) {
 	testcases := []testcase{
 		{
 			votes: []vote{
-				{"governance.governancemode", "none"},     // voted on 1 block
-				{"istanbul.committeesize", uint64(4)},     // voted on 2 block
-				{"governance.unitprice", uint64(2000000)}, // voted on 3 block
-				{"reward.mintingamount", "96000000000"},   // voted on 4 block
-				{"reward.ratio", "34/33/33"},              // voted on 5 block
-				{"reward.useginicoeff", true},             // voted on 6 block
-				{"reward.minimumstake", "5000000"},        // voted on 7 block
+				{"governance.governancemode", "none"},     // voted on block 1
+				{"istanbul.committeesize", uint64(4)},     // voted on block 2
+				{"governance.unitprice", uint64(2000000)}, // voted on block 3
+				{},                                      // voted on block 4
+				{"reward.mintingamount", "96000000000"}, // voted on block 5
+				{"reward.ratio", "34/33/33"},            // voted on block 6
+				{},                                      // voted on block 7
+				{"reward.useginicoeff", true},           // voted on block 8
+				{"reward.minimumstake", "5000000"},      // voted on block 9
 			},
 			expected: []governanceItem{
-				{vote{"governance.governancemode", "none"}, 4},
+				{vote{"governance.governancemode", "none"}, 6},
 				{vote{"istanbul.committeesize", uint64(4)}, 6},
-				{vote{"governance.unitprice", uint64(2000000)}, 6},
-				{vote{"reward.mintingamount", "96000000000"}, 8},
-				{vote{"reward.ratio", "34/33/33"}, 8},
-				{vote{"reward.useginicoeff", true}, 10},
-				{vote{"reward.minimumstake", "5000000"}, 10},
+				{vote{"governance.unitprice", uint64(2000000)}, 9},
+				{vote{"reward.mintingamount", "96000000000"}, 9},
+				{vote{"reward.ratio", "34/33/33"}, 12},
+				{vote{"reward.useginicoeff", true}, 12},
+				{vote{"reward.minimumstake", "5000000"}, 15},
 				// check governance items on current block
 				{vote{"governance.governancemode", "none"}, 0},
 				{vote{"istanbul.committeesize", uint64(4)}, 0},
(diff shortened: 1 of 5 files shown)


aidan-kwon (commented):
This change looks fine, but `gov.UpdateGovernance` still has a dependency for the order of `gov.UpdateCurrentGovernance`. 
If `gov.WriteGovernance` uses a previous epoch's governance got from `gov.ReadGovernance` instead of `gov.currentSet`, `gov.UpdateGovernance` can be called regardless of the order of `gov.UpdateCurrentGovernance`.  This way could prevent possible future mistakes.

aidan-kwon (commented):
It would be better if you add an unit test for this fix.

jeongkyun-oh:
> This change looks fine, but `gov.UpdateGovernance` still has a dependency for the order of `gov.UpdateCurrentGovernance`.
> If `gov.WriteGovernance` uses a previous epoch's governance got from `gov.ReadGovernance` instead of `gov.currentSet`, `gov.UpdateGovernance` can be called regardless of the order of `gov.UpdateCurrentGovernance`. This way could prevent possible future mistakes.

I applied your comment. Please take another look.

consensus/istanbul/backend/snapshot.go
```diff
@@ -178,6 +178,19 @@ func (s *Snapshot) apply(headers []*types.Header, gov *governance.Governance, ad
...
+			if len(header.Governance) > 0 {
+				gov.UpdateGovernance(number, header.Governance)
+			}
+			gov.UpdateCurrentGovernance(number)
+			gov.ClearVotes(number)
+
+			// Reload governance values because epoch changed
+			snap.Epoch, snap.Policy, snap.CommitteeSize = getGovernanceValue(gov, number)
+			snap.Votes = make([]governance.GovernanceVote, 0)
+			snap.Tally = make([]governance.GovernanceTallyItem, 0)
```
yoomee1313, line 191: `snap.Votes` and `snap.Tally` are used in `gov.HandleGovernanceVote`. Is it okay to initialize them in here?

consensus/istanbul/backend/engine_test.go
```diff
@@ -812,3 +821,184 @@ func TestSnapshot(t *testing.T) {
 		engine.Stop()
 	}
 }
+
+func TestGovernance_Votes(t *testing.T) {
+	type vote struct {
```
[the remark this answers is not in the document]
jeongkyun-oh: @yoomee1313 Isn't `item` too general? I think `vote` is clearer. We can think the expected `vote` field as applied vote.

consensus/istanbul/backend/engine_test.go
```diff
@@ -812,3 +821,184 @@ func TestSnapshot(t *testing.T) {
...
+	var configItems []interface{}
+	configItems = append(configItems, proposerPolicy(params.WeightedRandom))
+	configItems = append(configItems, epoch(3))
+	configItems = append(configItems, governanceMode("single"))
+	configItems = append(configItems, blockPeriod(0)) // set block period to 0 to prevent creating future block
+	for _, tc := range testcases {
+		chain, engine := newBlockChain(1, configItems...)
+
+		// test initial governance items
+		assert.Equal(t, uint64(3), engine.governance.Epoch())
```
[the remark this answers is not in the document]
jeongkyun-oh: @yoomee1313 Right. There is the same problem for the epoch since it uses the current epoch only. I think it is not possible to add the test at this moment. I will make an issue for this.

consensus/istanbul/backend/engine_test.go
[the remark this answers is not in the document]
yoomee1313: @jeongkyun-oh Understood.. thanks!

governance/default_test.go
```diff
@@ -240,6 +240,8 @@ func TestGovernance_ClearVotes(t *testing.T) {
 		if ret != val.e {
 			t.Errorf("Want %v, got %v for %v and %v", val.e, ret, val.k, val.v)
 		}
+		avt := gov.adjustValueType(val.k, val.v)
+		gov.RemoveVote(val.k, avt, 0)
```
ehnuje, line 244: Why is `RemoveVote` added to test?

hqjang-pepper: I'm curious too

hqjang-pepper (approved):
I ran additional tests on TestGovernance_votes, and they all passed.
As long as the need for `removeVote` becomes clear, I think it's a good PR to approve.

governance/default_test.go
jeongkyun-oh: @ehnuje @hqjang-pepper The original test include `voteMap.Clear()` method which clears all votes whether or not the votes are casted. This PR updates the method where only casted votes are cleared. So, `RemoveVote` is the method that sets the vote casted. That is why I added it.

hqjang-pepper: oh I see, it makes sense.

hqjang-pepper (approved):

consensus/istanbul/backend/engine_test.go
```diff
@@ -812,3 +821,184 @@ func TestSnapshot(t *testing.T) {
...
+			currentBlock = makeBlockWithSeal(chain, engine, previousBlock)
+			_, err = chain.InsertChain(types.Blocks{currentBlock})
+			assert.NoError(t, err)
+		}
+
+		for _, item := range tc.expected {
+			blockNumber := item.appliedBlockNumber
+			if blockNumber == 0 {
+				blockNumber = chain.CurrentBlock().NumberU64()
+			}
```
aidan-kwon, line 996: It's not a general way to indicate the latest block number. You better use `LatestBlockNumber` or `-1`. 
Pleaes, fix this later

consensus/istanbul/backend/engine_test.go
```diff
@@ -812,3 +821,184 @@ func TestSnapshot(t *testing.T) {
...
+				{"governance.governancemode", "single"}, // voted on block 6
+				{"governance.governancemode", "none"},   // voted on block 7
+				{"governance.governancemode", "single"}, // voted on block 8
+				{"governance.governancemode", "none"},   // voted on block 9
+			},
+			expected: []governanceItem{
+				{vote{"governance.governancemode", "single"}, 6},
+				{vote{"governance.governancemode", "none"}, 9},
+				{vote{"governance.governancemode", "single"}, 12},
+				{vote{"governance.governancemode", "none"}, 15},
```
aidan-kwon, line 884: I can't understand this.
The last vote "none" was in block 9 and the epoch was 3. I think the vote is applied from block 12.

aidan-kwon: My mistake. The vote is effective after 2 * epoch later, right? Then, your code makes sense.

aidan-kwon (approved):

consensus/istanbul/backend/engine_test.go
jeongkyun-oh: I will update this later. Thanks.

hyochan-brown (approved):
LGTM!!

consensus/istanbul/backend/engine_test.go
jeongkyun-oh: Yes, it is applied after 2 epochs.

PR #1052, opened 2021-10-29
Revise SetHead method before applying the snapshot package

hqjang-pepper:
## Proposed changes

- This is the first revision of `SetHead` method, which should be applied before snapshot package is merged.
- `SetHead` method rewinds the chain to the given number, including currentBlock, currentFastBlock and currentHeader.
- If a target block doesn't have state information, the currentBlock is further rewinded to the block in which state is stored.
- You can rewind node by calling `debug_setHead` api.
- Added a flag called `tmpStop` to prevent synchronization temporarily while setHead is in progress.
- When the snapshot package is merged, it is necessary to apply it to setHead.

For PR reviewers, these are the two main changes of setHead from this PR
- For loop repeats until it finds 1. a block with a state 2. that is less than the target number.
- While rewinding a chain, downloader starts and get a chain segment from a peer. After block rewind successfully ends, node tries to add the chain segments to `rewinded chain`, which causes sync error. That's the reason why `tmpStop` is needed. 

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


jeongkyun-oh:
@hqjang-pepper I guess this is not referenced from the recent Ethereum codes. Could you add a PR link which you referenced?

blockchain/blockchain.go
```diff
@@ -450,54 +450,66 @@ func (bc *BlockChain) loadLastState() error {
...
-// above the new head will be deleted and the new one set. In the case of blocks
-// though, the head may be further rewound if block bodies are missing (non-archive
-// nodes after a fast sync).
 func (bc *BlockChain) SetHead(head uint64) error {
-	logger.Info("Rewinding blockchain", "target", head)
+	logger.Warn("Rewinding blockchain", "target", head)
 
-	bc.mu.Lock()
-	defer bc.mu.Unlock()
+	bc.chainmu.Lock()
```
jeongkyun-oh, line 456: mutex is changed, is this okay?

storage/database/db_manager.go
```diff
@@ -823,6 +824,22 @@ func (dbm *databaseManager) DeleteCanonicalHash(number uint64) {
...
 }
 
+// ReadAllHashes retrieves all the hashes assigned to blocks at a certain heights,
+// both canonical and reorged forks included.
+func (dbm *databaseManager) ReadAllHashes(number uint64) []common.Hash {
+	db := dbm.getDatabase(headerDB)
+	prefix := headerKeyPrefix(number)
+
+	hashes := make([]common.Hash, 0, 1)
+	it := db.NewIterator(prefix, nil)
```
jeongkyun-oh, line 834: Don't we need `it.Release` after using it?

storage/database/db_manager.go
```diff
@@ -823,6 +824,22 @@ func (dbm *databaseManager) DeleteCanonicalHash(number uint64) {
 	dbm.cm.writeCanonicalHashCache(number, common.Hash{})
 }
 
+// ReadAllHashes retrieves all the hashes assigned to blocks at a certain heights,
+// both canonical and reorged forks included.
+func (dbm *databaseManager) ReadAllHashes(number uint64) []common.Hash {
```
jeongkyun-oh, line 829: Where is this method used?

hqjang-pepper:
> @hqjang-pepper I guess this is not referenced from the recent Ethereum codes. Could you add a PR link which you referenced?

- SetHead of blockchain.go - https://github.com/ethereum/go-ethereum/commit/542df8898ef6d718647058c129069804bc463ea5
- SetHead of headerchain.go - https://github.com/ethereum/go-ethereum/commit/b02958b9c57f35d0df085ed8e7057a52131373a3
- https://github.com/ethereum/go-ethereum/commit/8cbdc8638fd28693f84d7bdbbdd587e8c57f6383

blockchain/blockchain.go
hqjang-pepper: Ethereum also changed mutex at https://github.com/ethereum/go-ethereum/commit/f25f776c9f9d6a0538cd9028d9d82fc6965d9123. Take a look.

storage/database/db_manager.go
hqjang-pepper: I reflected. Thanks for reviewing.

storage/database/db_manager.go
hqjang-pepper: @jeongkyun-oh 
Oh, this method will be used in `headerchain.go` **after** snapshot package is ported.

blockchain/blockchain.go
aidan-kwon: @hqjang-pepper Your reference combined two mutexes into `bc.chainmu`. But your suggestion changes only one usage of the `bc.mu` mutex. It means that `SetHead` is not a critical section of `bc.mu` mutex and other codes still using `bc.mu` can be executed concurrently with `SetHead`. 
I think this mutex change is not proper for the intend of this PR.

hqjang-pepper: @aidan-kwon I reflected. Thanks for your comment.

node/cn/handler.go
```diff
@@ -122,6 +122,8 @@ type ProtocolManager struct {
 
 	nodetype          common.ConnType
 	txResendUseLegacy bool
+
+	tmpstop bool
```
yoomee1313, line 126: The `tmpstop` is a little bit confusing to me to figure out what it does. What about changing to the more specific name?

node/cn/gen_config.go
```diff
@@ -25,14 +25,19 @@ func (c Config) MarshalTOML() (interface{}, error) {
 		NetworkId               uint64
 		SyncMode                downloader.SyncMode
 		NoPruning               bool
+		WorkerDisable           bool
```
yoomee1313, line 28: Can I ask why these flags are added?

node/cn/api_backend.go
```diff
@@ -75,8 +75,10 @@ func (b *CNAPIBackend) CurrentBlock() *types.Block {
 }
 
 func (b *CNAPIBackend) SetHead(number uint64) {
-	//b.cn.protocolManager.downloader.Cancel()
+	b.cn.protocolManager.Downloader().Cancel()
+	b.cn.protocolManager.SetTmpStop(true)
 	b.cn.blockchain.SetHead(number)
+	b.cn.protocolManager.SetTmpStop(false)
```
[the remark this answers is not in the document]
hqjang-pepper: 1. There is no way to call CN's protocolmanager in methods of `blockchain` package.
2. blockchain.SetHead is called only in two functions except API call. 
-  `NewBlockChain` is an initialize function so we don't need to care about peer sync.
- `ResetWithGenesisBlock`is called only in `loadLastState()` when Head block is missing or got some problem with db. And `loadLastState()` is used when making a new blockchain or reset the canonical blocks. Kind of initializing process. 

so I think in case of non-api function calls of `setHead`, we don't need to care about sync stop process.

node/cn/handler.go
hqjang-pepper: Thx. I'll reflect.

blockchain/blockchain.go
```diff
@@ -450,54 +450,90 @@ func (bc *BlockChain) loadLastState() error {
 	return nil
 }
 
-// SetHead rewinds the local chain to a new head. In the case of headers, everything
-// above the new head will be deleted and the new one set. In the case of blocks
-// though, the head may be further rewound if block bodies are missing (non-archive
-// nodes after a fast sync).
 func (bc *BlockChain) SetHead(head uint64) error {
```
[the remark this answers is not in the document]
hqjang-pepper: There is no **specific** reference version of this function.
Because necessary features are spread over several versions, and function format is different with Ethereum.
But mostly from these following PR --> https://github.com/ethereum/go-ethereum/commit/8cbdc8638fd28693f84d7bdbbdd587e8c57f6383

+ I'll add some comments about setHead. Don't remember why I just deleted. Thanks for reminding.

yoomee1313:
@hqjang-pepper If you haven't, please test the setHead with istanbulCompatibleBlock set to larger than 0.

ehnuje:
@hqjang-pepper This change got merge conflicts. Could you please resolve them and update the PR?

hqjang-pepper:
> @hqjang-pepper This change got merge conflicts. Could you please resolve them and update the PR?

resolved.

api/mocks/backend_mock.go
```diff
@@ -162,6 +162,7 @@ func (mr *MockBackendMockRecorder) Engine() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Engine", reflect.TypeOf((*MockBackend)(nil).Engine))
 }
 
+
```
kjhman21, line 165: unnecessary change.

node/cn/handler.go
```diff
@@ -367,6 +370,11 @@ func (pm *ProtocolManager) Stop() {
 	logger.Info("Klaytn protocol stopped")
 }
 
+// SetSyncStop sets value of syncStop flag. If it's true, peer sync process does not proceed.
+func (pm *ProtocolManager) SetSyncStop(flag bool) {
+	pm.syncStop = flag
```
kjhman21, line 375: It seems that writing this variable and reading it can be done by different thread. In this case we need to use an atomic operation.

node/cn/sync.go
```diff
@@ -182,8 +182,8 @@ func (pm *ProtocolManager) getSyncMode(currentBlock *types.Block) downloader.Syn
 
 // synchronise tries to sync up our local block chain with a remote peer.
 func (pm *ProtocolManager) synchronise(peer Peer) {
-	// Short circuit if no peers are available
-	if peer == nil {
+	// Short circuit if no peers are available or syncStop flag is set to true
+	if peer == nil || pm.syncStop {
```
kjhman21, line 186: same here. Don't we need to use atomic operation?

storage/database/db_manager.go
```diff
@@ -823,6 +824,24 @@ func (dbm *databaseManager) DeleteCanonicalHash(number uint64) {
...
+	db := dbm.getDatabase(headerDB)
+	prefix := headerKeyPrefix(number)
+
+	hashes := make([]common.Hash, 0, 1)
```
kjhman21, line 833: Why do you make an array of hashes with capacity 1?

storage/database/schema.go
```diff
@@ -105,6 +105,18 @@ type TxLookupEntry struct {
...
+// encodeBlockNumber encodes a block number as big endian uint64
+func encodeBlockNumber(number uint64) []byte {
+	enc := make([]byte, 8)
+	binary.BigEndian.PutUint64(enc, number)
+	return enc
+}
+
+// headerKeyPrefix = headerPrefix + num (uint64 big endian)
+func headerKeyPrefix(number uint64) []byte {
+	return append(headerPrefix, encodeBlockNumber(number)...)
```
kjhman21, line 117: Why don't you use `common.Int64ToByteBigEndian(number)` as shown below?

blockchain/blockchain_sethead_test.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	// Set the head of the chain back to the requested number
+	chain.SetHead(tt.setheadBlock)
+
+	// Iterate over all the remaining blocks and ensure there are no gaps
+	//verifyNoGaps(t, chain, true, canonblocks)
+	//verifyNoGaps(t, chain, false, sideblocks)
+	//verifyCutoff(t, chain, true, canonblocks, tt.expCanonicalBlocks)
+	//verifyCutoff(t, chain, false, sideblocks, tt.expSidechainBlocks)
```
aidan-kwon, line 160: Why did you comment out these?

node/cn/api_backend.go
```diff
@@ -76,8 +76,10 @@ func (b *CNAPIBackend) CurrentBlock() *types.Block {
 }
 
 func (b *CNAPIBackend) SetHead(number uint64) {
-	//b.cn.protocolManager.downloader.Cancel()
+	b.cn.protocolManager.Downloader().Cancel()
```
aidan-kwon, line 79: How downloader start again after `SetHead`?

aidan-kwon:
`SetHead` removes old blocks but doesn't governance data. I think it should remove the governance data also because it can be stored in `misc` DB and affect the node later. @hqjang-pepper  Did you consider governance data also?

kjhman21:
related to #1144

hqjang-pepper:
> `SetHead` removes old blocks but doesn't governance data. I think it should remove the governance data also because it can be stored in `misc` DB and affect the node later. @hqjang-pepper Did you consider governance data also?

The fundamental question. Is clearing the governance DB data needed?
Even if `setHead()` method does not delete the governance DB data, I think it's okay.
While catching up the block, if there is governanceData that has already been entered in the governance DB, it will be fine because there is a logic to return without writing itself at all.

In function `WriteGovernance()`
```
if err := dbm.WriteGovernanceIdx(num); err != nil {
   if err == errGovIdxAlreadyExist {
      // Overwriting existing data is not allowed, but the attempt is not considered as a failure.
      return nil
   }
   return err
}
```

How do you think? @aidan-kwon @jeongkyun-oh

node/cn/handler.go
hqjang-pepper: changed SetSyncStop method atomically.

node/cn/sync.go
hqjang-pepper: Changed atomically, too.

jeongkyun-oh:
> > `SetHead` removes old blocks but doesn't governance data. I think it should remove the governance data also because it can be stored in `misc` DB and affect the node later. @hqjang-pepper Did you consider governance data also?
> 
> The fundamental question. Is clearing the governance DB data needed? Even if `setHead()` method does not delete the governance DB data, I think it's okay. While catching up the block, if there is governanceData that has already been entered in the governance DB, it will be fine because there is a logic to return without writing itself at all.
> 
> In function `WriteGovernance()`
> 
> ```
> if err := dbm.WriteGovernanceIdx(num); err != nil {
>    if err == errGovIdxAlreadyExist {
>       // Overwriting existing data is not allowed, but the attempt is not considered as a failure.
>       return nil
>    }
>    return err
> }
> ```
> 
> How do you think? @aidan-kwon @jeongkyun-oh

As we talked, `setHead` may be called under the unexpected circumstance. If corrupted data may be (somehow) inserted and want to `setHead`, I think we should be able to insert new data. However, if wrong data may be inserted and we want to retry to recover governance data, it will not be since the logic you mentioned.

aidan-kwon (approved):

blockchain/blockchain.go
```diff
@@ -450,53 +450,93 @@ func (bc *BlockChain) loadLastState() error {
...
+// that the rewind must pass the specified state root. The method will try to
+// delete minimal data from disk whilst retaining chain consistency.
 func (bc *BlockChain) SetHead(head uint64) error {
-	logger.Info("Rewinding blockchain", "target", head)
+	logger.Warn("Rewinding blockchain", "target", head)
 
 	bc.mu.Lock()
 	defer bc.mu.Unlock()
+	bc.chainmu.Lock()
+	defer bc.chainmu.Unlock()
```
jeongkyun-oh, line 462: Should we update this after https://github.com/klaytn/klaytn/pull/1099 ?

kjhman21 (approved):

blockchain/blockchain.go
hqjang-pepper: I applied this change by rebasing.

node/cn/api_backend.go
hqjang-pepper: After setHead logic is done, Td gets smaller than peer's. 
If local Td is smaller than peer's Td, then new goroutine starts downloader automatically.
```
if trueTD.Cmp(pm.blockchain.GetTd(currentBlock.Hash(), currentBlock.NumberU64())) > 0 {
	go pm.synchronise(p)
}
```

blockchain/blockchain_sethead_test.go
hqjang-pepper: I commented this because of golint complexity.
PR https://github.com/klaytn/klaytn/pull/1137 is merged, so I uncommented.

storage/database/schema.go
hqjang-pepper: I'll reflect this change. Thanks.

aidan-kwon (approved):

jeongkyun-oh (approved):

kjhman21 (approved):

PR #1076, opened 2021-12-03
A draft structure of eth namespace API for compatibility

jimni1222:
## Proposed changes

This PR introduces a draft of the structure to support `eth` namespace APIs.

`EthereumAPI` is newly implemented to support this, and this will be implemented using Klaytn API functions.
For using Klaytn API functions, the current draft implementation has all the `klay` namespace APIs as variables. (except for `PublicKlayAPI` because eth API does not support `Rewardbase`)
However, after the implementation is completed, unused variables will be cleaned up.
Please let me know if there is better idea for this. :) 

After this PR, another PR that defines API functions to be supported will be uploaded.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"github.com/klaytn/klaytn/datasync/downloader"
 	"github.com/klaytn/klaytn/governance"
 	"github.com/klaytn/klaytn/node/cn/filters"
 )
@@ -8,8 +9,9 @@ import (
 // EthereumAPI provides an API to access the Klaytn through the `eth` namespace.
 // TODO-Klaytn: Removed unused variable
 type EthereumAPI struct {
-	publicFilterAPI   *filters.PublicFilterAPI
-	governanceKlayAPI *governance.GovernanceKlayAPI
+	publicFilterAPI     *filters.PublicFilterAPI
+	governanceKlayAPI   *governance.GovernanceKlayAPI
+	publicDownloaderAPI *downloader.PublicDownloaderAPI
 
 	publicKlayAPI            *PublicKlayAPI
 	publicBlockChainAPI      *PublicBlockChainAPI
@@ -22,7 +24,7 @@ type EthereumAPI struct {
 // Therefore, it is necessary to use APIs defined in two different packages(cn and api),
 // so those apis will be defined through a setter.
 func NewEthereumAPI() *EthereumAPI {
-	return &EthereumAPI{nil, nil, nil, nil, nil, nil}
+	return &EthereumAPI{nil, nil, nil, nil, nil, nil, nil}
 }
 
 // SetPublicFilterAPI sets publicFilterAPI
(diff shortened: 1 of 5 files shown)


jimni1222:
I will upload PR again after resolving CI errors.

aeharvlee (dismissed):

api/api_ethereum.go
```diff
@@ -0,0 +1,63 @@
...
+type EthereumAPI struct {
+	publicFilterAPI     *filters.PublicFilterAPI
+	governanceKlayAPI   *governance.GovernanceKlayAPI
+	publicDownloaderAPI *downloader.PublicDownloaderAPI
+
+	publicKlayAPI            *PublicKlayAPI
+	publicBlockChainAPI      *PublicBlockChainAPI
+	publicTransactionPoolAPI *PublicTransactionPoolAPI
+	publicAccountAPI         *PublicAccountAPI
+}
```
aeharvlee, line 20: @jimni1222 
Can we handle the `Syncing` APIs using the current design?
There are two `Syncing` APIs, one in `PublicDownloaderAPI` and the other in `PublicKlayAPI`. (Same name but different function signatures).

Because Golang does not support method overloading(it means using same name and different arguments), 
we should separate APIs.

jimni1222: thank you for comment. i modified this logic. can you check?

node/cn/backend.go
```diff
@@ -480,12 +488,12 @@ func (s *CN) APIs() []rpc.API {
...
 			Version:   "1.0",
-			Service:   downloader.NewPublicDownloaderAPI(s.protocolManager.Downloader(), s.eventMux),
+			Service:   publicDownloaderAPI,
 			Public:    true,
 		}, {
-			Namespace: "klay",
+			Namespace: "eth",
 			Version:   "1.0",
-			Service:   filters.NewPublicFilterAPI(s.APIBackend, false),
+			Service:   publicFilterAPI,
```
aeharvlee, line 496:
```suggestion
			Service:   publicDownloaderAPI,
```
@jimni1222 
Currently, only the method `Syncing` is duplicated(one in PublicDownloaderAPI which is subscription method and the other in `PublicKlayAPI` which is callback method), but we need to think about whether a similar case may occur in the future.

If so, we should separate all the structures as separately like below.
```go
type EthereumFilterAPI struct {
	publicFilterAPI   *filters.PublicFilterAPI
}
type EthereumPublicBlockChainAPI struct {
	publicBlockChainAPI      *PublicBlockChainAPI
}
// ...
```

I'm not sure which one is the more maintainable design.
cc. @aidan-kwon

jimni1222: for now, i modified wrong service name.
And in my opinion, we can just do at one struct for now (except for downloader api), and if duplication is happened in the future, and then we can separate the structure.

api/backend.go
```diff
@@ -89,23 +89,34 @@ type Backend interface {
 	GetTxLookupInfoAndReceiptInCache(Hash common.Hash) (*types.Transaction, common.Hash, uint64, uint64, *types.Receipt)
 }
 
-func GetAPIs(apiBackend Backend) []rpc.API {
+func GetAPIs(apiBackend Backend, ethAPI *EthereumAPI) []rpc.API {
```
aidan-kwon, line 92: This function name starts with `Get`, but it sets something insider for `ethAPI`. How about declaring a new EthereumAPI inside and returning it as a second return value?
`func GetAPIs(apiBackend Backend) ([]rpc.API,  *EthereumAPI)`

jimni1222: That sounds great. I modified the source code in https://github.com/klaytn/klaytn/pull/1076/commits/6b9302cb2ab04f7c18c797e7c62edf0a7fc6941c

jeongkyun-oh (approved):

yoomee1313 (approved):

aeharvlee (approved):

aidan-kwon (approved):

hqjang-pepper (approved):

sirano11 (approved):

PR #1079, opened 2021-12-08
Added CumulativeGasUsed in Receipt struct

jimni1222:
## Proposed changes

This PR introduces adding `CumulativeGasUsed` field in Receipt struct for compatibility with Ethereum rpc call return type.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -2443,7 +2443,7 @@ func (bc *BlockChain) ApplyTransaction(chainConfig *params.ChainConfig, author *
 	statedb.Finalise(true, false)
 	*usedGas += gas
 
-	receipt := types.NewReceipt(kerr.Status, tx.Hash(), gas)
+	receipt := types.NewReceipt(kerr.Status, tx.Hash(), gas, *usedGas)
 	// if the transaction created a contract, store the creation address in the receipt.
 	msg.FillContractAddress(vmenv.Context.Origin, receipt)
 	// Set the receipt logs and create a bloom for filtering
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -2443,7 +2443,8 @@ func (bc *BlockChain) ApplyTransaction(chainConfig *params.ChainConfig, author *
 	statedb.Finalise(true, false)
 	*usedGas += gas
 
-	receipt := types.NewReceipt(kerr.Status, tx.Hash(), gas, *usedGas)
+	receipt := types.NewReceipt(kerr.Status, tx.Hash(), gas)
+	receipt.SetCumulativeGasUsed(*usedGas)
 	// if the transaction created a contract, store the creation address in the receipt.
 	msg.FillContractAddress(vmenv.Context.Origin, receipt)
 	// Set the receipt logs and create a bloom for filtering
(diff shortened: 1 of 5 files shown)


blockchain/blockchain.go
```diff
@@ -2444,6 +2444,7 @@ func (bc *BlockChain) ApplyTransaction(chainConfig *params.ChainConfig, author *
 	*usedGas += gas
 
 	receipt := types.NewReceipt(kerr.Status, tx.Hash(), gas)
+	receipt.SetCumulativeGasUsed(*usedGas)
```
jeongkyun-oh, line 2447: It's very minor, but it seems to appear our of thin air. 
How about adding `NewReceipt` parameter or just assign like `receipt.Logs = statedb.GetLogs(tx.Hash())` for consistency?

jimni1222: Can you explain a little more? I could not understand. ;-;

jeongkyun-oh: Oh, I meant that it seems to appear `SetCumulativeGasUsed` only, but not other fields. So, I suggested to add `*usedGas` while creating new receipt or just assign it `receipt.cumulativeGasUsed = *usedGas` for consistency. It is very minor, though.

jimni1222: Do you meaning that passing *usedGas as the last parameter when doing `NewReceipt` and assigning it to `receipt.CumulativeGas = *usedGas` internally?
If so, early implementations did just that.
However, the test generates an error because CumulativeGasUsed is not used in other parts that use receipt, for example, the part that processes ReceiptsMsg.
In that test, the receipt test data is generated by `NewReceipt`, and because cumulativeGasUsed is added as a last parameter of the `NewReceipt`, that should use the cumulativeGasUsed also.
But when handle ReceiptsMsg, cumulativeGasUsed is not used so test will be fail because what expected (generated from NewReceipt) and got (receipt object used during handling ReceiptsMsg) is different.
This was discovered because some tests were failing, so I changed the logic to exclude CumulativeGasUsed from NewReceipt and assign it separately in this [commit](https://github.com/klaytn/klaytn/pull/1079/commits/14fd0e7011c56e179e261aa071c358ff3abf4c34).

jimni1222: @jeongkyun-oh I think i misunderstood what you said, ;;; i think now i understood what you saying, and i applied in [here](https://github.com/klaytn/klaytn/pull/1079/commits/ec78c6dcf44cf30daa96f22b6fe8659d4076b0ce). Sorry for miscommunication

blockchain/types/receipt.go
```diff
@@ -101,6 +101,8 @@ type Receipt struct {
 	TxHash          common.Hash    `json:"transactionHash" gencodec:"required"`
 	ContractAddress common.Address `json:"contractAddress"`
 	GasUsed         uint64         `json:"gasUsed" gencodec:"required"`
+
+	CumulativeGasUsed uint64 `json:"cumulativeGasUsed"`
```
jeongkyun-oh, line 105: Is this okay to add a field to `Receipt`? Shouldn't it be hard-forked to add a field of receipt?
If we are considering only API returns, how about adding the hard-coded field to API response format?

jimni1222: I've been discussing this.
For now, We've decided to go the safer way, and I'll close this PR and modify the implementation to be counted every time the API is called.
Thank you for your review :)

blockchain/blockchain.go
jeongkyun-oh: Sorry for vague explanation

blockchain/types/receipt.go
jeongkyun-oh: Ah, I agree with the way you talked. Thanks.

PR #1087, opened 2021-12-15
Implemented eth namespace tx read api

jimni1222:
## Proposed changes

This PR introduces eth namespace read transaction api.
Below apis are implemented newly to return in ethereum format (different with Klaytn return format)
- GetTransactionByBlockNumberAndIndex
- GetTransactionByBlockHashAndIndex
- GetTransactionByHash
- GetTransactionReceipt
- PendingTransactions

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -572,14 +572,16 @@ func newEthRPCPendingTransaction(tx *types.Transaction) (*EthRPCTransaction, err
 // GetTransactionByBlockNumberAndIndex returns the transaction for the given block number and index.
 func (api *EthereumAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) *EthRPCTransaction {
 	block, err := api.publicTransactionPoolAPI.b.BlockByNumber(ctx, blockNr)
-	if block != nil && err == nil {
-		ethTx, err := newEthRPCTransactionFromBlockAndIndex(block, uint64(index))
-		if ethTx == nil || err != nil {
-			return nil
-		}
-		return ethTx
+	if block == nil || err != nil {
+		return nil
 	}
-	return nil
+
+	ethTx, err := newEthRPCTransactionFromBlockAndIndex(block, uint64(index))
+	if ethTx == nil || err != nil {
+		return nil
+	}
+
+	return ethTx
 }
 
 // GetTransactionByBlockHashAndIndex returns the transaction for the given block hash and index.
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -598,7 +598,7 @@ func (api *EthereumAPI) GetTransactionByBlockHashAndIndex(ctx context.Context, b
 // GetRawTransactionByBlockNumberAndIndex returns the bytes of the transaction for the given block number and index.
 func (api *EthereumAPI) GetRawTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) hexutil.Bytes {
 	rawTx, err := api.publicTransactionPoolAPI.GetRawTransactionByBlockNumberAndIndex(ctx, blockNr, index)
-	if rawTx == nil || err != nil {
+	if err != nil {
 		return nil
 	}
 
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -18,6 +18,7 @@ package api
 
 import (
 	"context"
+	"errors"
 	"math/big"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


api/api_ethereum.go
```diff
@@ -502,52 +503,218 @@ type EthRPCTransaction struct {
...
+		return nil, nil
+	}
+
+	from := getFrom(tx)
+
+	// If to is nil, it is fills with from.
+	to := tx.To()
+	if to == nil {
+		to = &from
+	}
```
aidan-kwon, line 529: Contract deploying transaction has `nil` for `to` address. This logic should be applied only for few Klaytn tx types.

api/api_ethereum.go
```diff
@@ -502,52 +503,218 @@ type EthRPCTransaction struct {
...
 func (api *EthereumAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) *EthRPCTransaction {
-	// TODO-Klaytn: Not implemented yet.
+	block, err := api.publicTransactionPoolAPI.b.BlockByNumber(ctx, blockNr)
+	if block != nil && err == nil {
+		ethTx, err := newEthRPCTransactionFromBlockAndIndex(block, uint64(index))
+		if ethTx == nil || err != nil {
+			return nil
+		}
+		return ethTx
+	}
```
aidan-kwon, line 581: How about changing these if branches to more simple structures?
```suggestion
	block, err := api.publicTransactionPoolAPI.b.BlockByNumber(ctx, blockNr)
	if block == nil || err != nil {
	        return nil 
	}
	ethTx, err := newEthRPCTransactionFromBlockAndIndex(block, uint64(index))
	if ethTx == nil || err != nil {
	        return nil 
	}
	return ethTx
	}
```

api/api_ethereum.go
```diff
@@ -502,52 +503,218 @@ type EthRPCTransaction struct {
...
 	return nil
 }
 
 // GetRawTransactionByBlockNumberAndIndex returns the bytes of the transaction for the given block number and index.
 func (api *EthereumAPI) GetRawTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) hexutil.Bytes {
-	// TODO-Klaytn: Not implemented yet.
-	return nil
+	rawTx, err := api.publicTransactionPoolAPI.GetRawTransactionByBlockNumberAndIndex(ctx, blockNr, index)
+	if rawTx == nil || err != nil {
```
aidan-kwon, line 601: I think one comparition is enough. `return rawTx` will return `nil` if `rawTx == nil`
```suggestion
	if err != nil {
```

api/api_ethereum.go
```diff
@@ -502,52 +503,218 @@ type EthRPCTransaction struct {
...
+		return nil, nil
+	}
+
+	from := getFrom(tx)
+
+	// If to is nil, it is fills with from.
+	to := tx.To()
+	if to == nil {
+		to = &from
+	}
```
aidan-kwon, line 679: Same here

api/api_ethereum.go
```diff
@@ -502,52 +503,218 @@ type EthRPCTransaction struct {
...
+	fields := map[string]interface{}{
+		"blockHash":         blockHash,
+		"blockNumber":       hexutil.Uint64(blockNumber),
+		"transactionHash":   tx.Hash(),
+		"transactionIndex":  hexutil.Uint64(index),
+		"from":              getFrom(tx),
+		"to":                to,
+		"gasUsed":           hexutil.Uint64(receipt.GasUsed),
+		"cumulativeGasUsed": hexutil.Uint64(cumulativeGasUsed),
+		"contractAddress":   nil,
```
aidan-kwon, line 698: Why this field is always `nil`? I think contract deploying tx should return an address.

api/api_ethereum.go
jimni1222: I thought when result of the `tx.To()` is `nil`, then fill the to field to from can cover only few Klaytn tx types. Is there any worry about this?

api/api_ethereum.go
jimni1222: In `L713`, contractAddress filed will be filled when receipt.ContractAddress is existed.

api/api_ethereum.go
jimni1222: That sounds good :) I applied in [this commit](https://github.com/klaytn/klaytn/pull/1087/commits/bf237b7ede566b340b6bbe2a5a85afc9ee009a66).

api/api_ethereum.go
aeharvlee: I think when the transaction type is `SmartContractDeploy`, we have to leave `nil` as it is.
So additional tx type checking logic should be appended.
(**should consider FeeDelegation and PartialFeeDelegation also**)

* When the type of Tx is `SmartContractDeploy` transaction. | => Use `nil` value as it is.
* When the type of Tx is `AccountUpdate` transaction . | => Use `from` value instead of `nil`.
* When the type of Tx is `Cancel` transaction. | => Use `from` value instead of `nil`.
* When the type of Tx is `ChainDataAnchoring` transaction. | => Use `from` value instead of `nil`.

api/api_ethereum.go
aeharvlee: When the transaction type is SmartContractDeploy, we have to leave nil as it is.
So additional tx type checking logic should be appended.
(should consider FeeDelegation and PartialFeeDelegation also)

When the type of Tx is SmartContractDeploy transaction. | => Use nil value as it is.
When the type of Tx is AccountUpdate transaction . | => Use from value instead of nil.
When the type of Tx is Cancel transaction. | => Use from value instead of nil.
When the type of Tx is ChainDataAnchoring transaction. | => Use from value instead of nil.

aeharvlee:
Hi reviewers. 
I'll re-open and continue this PR based on @jimni1222 branch which is currently being discussed.
I talked with @jimni1222 already about this :) vacation issue 😉

PR #1088, opened 2021-12-15
Added eth namespace Block Header apis

aeharvlee:
## Proposed changes

Added `Engine()` to `backend` interface and `CNAPIBackend` also.
* We need it when calculating proposer of the block. This proposer will be used by `(&EthereumAPI{}).Miner` method.
* Regenerated `backend_mock.go` also because new method `Engine` is added.

Added `rpcMarshalHeader` function.
* This is for supporting Etheruem compatible format.

Added Methods for `BlockNonce`
* Added implementations from Ethereum. 
* You can check the Ethereum implementations [here](https://github.com/ethereum/go-ethereum/blob/eae3b1946a276ac099e0018fc792d9e8c3bfda6d/core/types/block.go#L39-L64).

Made `BaseFee` as protocol params for eth namespace.
* Value of `BaseFee` and `Extradata` are considered by klaytn developers.
* While defining `BaseFee` as protocol params, the parts that previously defined baseFee were also modified.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -19,10 +19,11 @@ package api
 import (
 	"context"
 	"encoding/binary"
-	"github.com/klaytn/klaytn/params"
 	"math/big"
 	"sync/atomic"
 	"time"
+
+	"github.com/klaytn/klaytn/params"
 
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -369,10 +369,9 @@ func (api *EthereumAPI) GetProof(ctx context.Context, address common.Address, st
 // * When blockNr is -1 the chain head is returned.
 // * When blockNr is -2 the pending chain head is returned.
 func (api *EthereumAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockNumber) (map[string]interface{}, error) {
-	klaytnHeader, err := api.publicBlockChainAPI.GetHeaderByNumber(ctx, number)
 	// In Ethereum, err is always nil because the backend of Ethereum always return nil.
-	err = nil
-	if klaytnHeader != nil && err == nil {
+	klaytnHeader, _ := api.publicBlockChainAPI.GetHeaderByNumber(ctx, number)
+	if klaytnHeader != nil {
 		response := api.rpcMarshalHeader(klaytnHeader)
 		if number == rpc.PendingBlockNumber {
 			// Pending header need to nil out a few fields
@@ -382,7 +381,7 @@ func (api *EthereumAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockN
 		}
 		return response, nil
 	}
-	return nil, err
+	return nil, nil
 }
 
 // GetHeaderByHash returns the requested header by hash.
@@ -734,7 +733,7 @@ func (api *EthereumAPI) Accounts() []common.Address {
 func (api *EthereumAPI) rpcMarshalHeader(head *types.Header) map[string]interface{} {
 	proposer, err := api.publicKlayAPI.b.Engine().Author(head)
 	if err != nil {
-		logger.Error("Failed to fetch author during marshaling header: %s", err.Error(), nil)
+		logger.Error("Failed to fetch author during marshaling header", "err", err.Error())
 	}
 	result := map[string]interface{}{
 		"number":          (*hexutil.Big)(head.Number),
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -372,7 +372,10 @@ func (api *EthereumAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockN
 	// In Ethereum, err is always nil because the backend of Ethereum always return nil.
 	klaytnHeader, _ := api.publicBlockChainAPI.GetHeaderByNumber(ctx, number)
 	if klaytnHeader != nil {
-		response := api.rpcMarshalHeader(klaytnHeader)
+		response, err := api.rpcMarshalHeader(klaytnHeader)
+		if err != nil {
+			return nil, err
+		}
 		if number == rpc.PendingBlockNumber {
 			// Pending header need to nil out a few fields
 			for _, field := range []string{"hash", "nonce", "miner"} {
@@ -386,9 +389,14 @@ func (api *EthereumAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockN
 
 // GetHeaderByHash returns the requested header by hash.
 func (api *EthereumAPI) GetHeaderByHash(ctx context.Context, hash common.Hash) map[string]interface{} {
+	// In Ethereum, err is always nil because the backend of Ethereum always return nil.
 	klaytnHeader, _ := api.publicBlockChainAPI.GetHeaderByHash(ctx, hash)
 	if klaytnHeader != nil {
-		return api.rpcMarshalHeader(klaytnHeader)
+		response, err := api.rpcMarshalHeader(klaytnHeader)
+		if err != nil {
+			return nil
+		}
+		return response
 	}
 	return nil
 }
@@ -729,11 +737,14 @@ func (api *EthereumAPI) Accounts() []common.Address {
 	return nil
 }
 
-// rpcMarshalHeader marshal block header as Ethereum compatible format
-func (api *EthereumAPI) rpcMarshalHeader(head *types.Header) map[string]interface{} {
+// rpcMarshalHeader marshal block header as Ethereum compatible format.
+// It returns error when fetching Author which is block proposer is failed.
+func (api *EthereumAPI) rpcMarshalHeader(head *types.Header) (map[string]interface{}, error) {
 	proposer, err := api.publicKlayAPI.b.Engine().Author(head)
 	if err != nil {
+		// miner is the field Klaytn should provide the correct value. It's not the field dummy value is allowed.
 		logger.Error("Failed to fetch author during marshaling header", "err", err.Error())
+		return nil, err
 	}
 	result := map[string]interface{}{
 		"number":          (*hexutil.Big)(head.Number),
@@ -759,5 +770,5 @@ func (api *EthereumAPI) rpcMarshalHeader(head *types.Header) map[string]interfac
 		"baseFeePerGas":    (*hexutil.Big)(new(big.Int).SetUint64(params.BaseFee)),
 	}
 
-	return result
+	return result, nil
 }
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -738,17 +738,19 @@ func (api *EthereumAPI) rpcMarshalHeader(head *types.Header) map[string]interfac
 		logger.Error("Failed to fetch author during marshaling header: %s", err.Error(), nil)
 	}
 	result := map[string]interface{}{
-		"number":           (*hexutil.Big)(head.Number),
-		"hash":             head.Hash(),
-		"parentHash":       head.ParentHash,
-		"nonce":            BlockNonce{},  // There is no block nonce concept in Klaytn, so it must be empty.
-		"mixHash":          common.Hash{}, // Klaytn does not use mixHash, so it must be empty.
-		"sha3Uncles":       common.HexToHash(EmptySha3Uncles),
-		"logsBloom":        head.Bloom,
-		"stateRoot":        head.Root,
-		"miner":            proposer,
-		"difficulty":       (*hexutil.Big)(head.BlockScore),
-		"totalDifficulty":  (*hexutil.Big)(api.publicKlayAPI.b.GetTd(head.Hash())),
+		"number":          (*hexutil.Big)(head.Number),
+		"hash":            head.Hash(),
+		"parentHash":      head.ParentHash,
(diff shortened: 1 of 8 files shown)


api/api_ethereum.go
```diff
@@ -18,10 +18,13 @@ package api
 
 import (
 	"context"
+	"encoding/binary"
 	"math/big"
 	"sync/atomic"
 	"time"
 
+	"github.com/klaytn/klaytn/params"
+
```
yoomee1313, line 27: This is minor, but what about combining them together?

api/api_ethereum.go
```diff
@@ -679,3 +728,34 @@ func (api *EthereumAPI) Accounts() []common.Address {
 	// TODO-Klaytn: Not implemented yet.
 	return nil
 }
+
+// rpcMarshalHeader marshal block header as Ethereum compatible format
+func (api *EthereumAPI) rpcMarshalHeader(head *types.Header) map[string]interface{} {
```
jimni1222, line 733: How about changing the function name so that it knows that it is marshaling to the Ethereum header format?
EthereumAPI is inside the api package, so I think it can prevent confusion in the future if it is differentiated.

aeharvlee: @jimni1222 
I thought it is clear because the `rpcMarshalHeader` is a method of the struct `EthereumAPI`, so I believe we can guess this method will be do formatting header to Ethereum header.

api/api_ethereum.go
```diff
@@ -334,13 +370,28 @@ func (api *EthereumAPI) GetProof(ctx context.Context, address common.Address, st
 // * When blockNr is -1 the chain head is returned.
 // * When blockNr is -2 the pending chain head is returned.
 func (api *EthereumAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockNumber) (map[string]interface{}, error) {
-	// TODO-Klaytn: Not implemented yet.
-	return nil, nil
+	klaytnHeader, err := api.publicBlockChainAPI.GetHeaderByNumber(ctx, number)
+	// In Ethereum, err is always nil because the backend of Ethereum always return nil.
+	err = nil
```
aeharvlee, line 375: It could be looked weird, but currently Ethereum backend always return nil at error position.

[Ethereum source code link](https://github.com/ethereum/go-ethereum/blob/356bbe343a30789e77bb38f25983c8f2f2bfbb47/eth/api_backend.go#L65-L76).

blockchain/vm/runtime/runtime.go
```diff
@@ -89,7 +89,7 @@ func setDefaults(cfg *Config) {
 		}
 	}
 	if cfg.BaseFee == nil {
-		cfg.BaseFee = big.NewInt(0)
+		cfg.BaseFee = new(big.Int).SetUint64(params.BaseFee)
```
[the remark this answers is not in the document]
aeharvlee: @yoomee1313 
Your welcome 😉

api/api_ethereum.go
```diff
@@ -334,13 +370,28 @@ func (api *EthereumAPI) GetProof(ctx context.Context, address common.Address, st
 // * When blockNr is -1 the chain head is returned.
 // * When blockNr is -2 the pending chain head is returned.
 func (api *EthereumAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockNumber) (map[string]interface{}, error) {
-	// TODO-Klaytn: Not implemented yet.
-	return nil, nil
+	klaytnHeader, err := api.publicBlockChainAPI.GetHeaderByNumber(ctx, number)
```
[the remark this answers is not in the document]
aeharvlee: @yoomee1313 
About Q1.
The reason why I use "klaytn" at the return value of `api.publicBlockChainAPI.GetHeaderByNumber(ctx, number)` is to clarify the meaning like 
"The data type of returned value is Header (which is different data structure from Ethereum), so we should convert it to Ethereum Header format.".

About Q2.
Nope. There is no variable or data structure named `etheruemHeader` in the PR :)

api/api_ethereum.go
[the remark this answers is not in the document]
aeharvlee: No problem.
I updated it here :)
https://github.com/klaytn/klaytn/pull/1088/commits/bc824a7aa1c6f4a7b6e70f5e69e212a689bfb24f

yoomee1313 (dismissed):

api/api_ethereum_test.go
```diff
@@ -0,0 +1,158 @@
...
+		blockParam = header.Hash()
+		mockBackend.EXPECT().HeaderByHash(gomock.Any(), gomock.Any()).Return(header, nil)
+	}
+
+	results := reflect.ValueOf(&api).MethodByName(testAPIName).Call(
+		[]reflect.Value{
+			reflect.ValueOf(context.Background()),
+			reflect.ValueOf(blockParam),
+		},
+	)
```
hqjang-pepper, line 69: Reflect package is used due to call multiple api tests in a time

hqjang-pepper (dismissed):

jimni1222 (dismissed):

api/api_ethereum.go
```diff
@@ -679,3 +729,36 @@ func (api *EthereumAPI) Accounts() []common.Address {
 	// TODO-Klaytn: Not implemented yet.
 	return nil
 }
+
+// rpcMarshalHeader marshal block header as Ethereum compatible format
+func (api *EthereumAPI) rpcMarshalHeader(head *types.Header) map[string]interface{} {
+	proposer, err := api.publicKlayAPI.b.Engine().Author(head)
+	if err != nil {
+		logger.Error("Failed to fetch author during marshaling header: %s", err.Error(), nil)
```
aidan-kwon, line 737: Parameters are not proper for the logger

params/protocol_params.go
```diff
@@ -163,6 +163,9 @@ const (
 	TxGasContractExecution uint64 = 21000
 
 	TxDataGas uint64 = 100
+
+	// BaseFee exists for Supporting Ethereum compatible data structure.
```
aidan-kwon, line 167:
```suggestion
	// BaseFee exists for supporting Ethereum compatible data structure.
```

api/api_ethereum.go
aeharvlee: Thanks for your feedback.

Applied your feedback. Now log is printed well.
```
ERROR[01/10,10:32:26 +09] [3] Failed to fetch author during marshaling header  err="invalid istanbul header extra-data"
```

aidan-kwon: It returns an empty proposer when err is not nil. Is it okay?

jimni1222 (dismissed):

api/api_ethereum.go
aeharvlee: @aidan-kwon 
As you said, an empty proposer will be served if error happens at `Author()`.
Although it is not perfect information(miner field would be empty address `0x0000000000000000000000000000000000000000` in error case), I thought we should provide information about other fields if possible.

aeharvlee: @aidan-kwon 
So, there could be 2 options about this issue.

**Option 1**: Provide header even if a miner field is empty address
In this case, user will be received header like below.
```json
{
  baseFeePerGas: "0x0",
  difficulty: "0x1",
  extraData: "0x",
  gasLimit: "0x3b9ac9ff",
  gasUsed: "0x0",
  hash: "0xc8036293065bacdfce87debec0094a71dbbe40345b078d21dcc47adb4513f348",
  logsBloom: "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  miner: "0x0000000000000000000000000000000000000000",
  mixHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
  nonce: "0x0000000000000000",
  number: "0x3",
  parentHash: "0x8c73a925cf0ec8019162b32bb75b6f1df706213df4468de131dfde9612e33331",
  receiptsRoot: "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470",
  sha3Uncles: "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
  size: "0x2cc",
  stateRoot: "0xae2f08edc892858210ce258f3bd0d70fde29eefe85e16e3535802a36e15a974c",
  timestamp: "0x61d53853",
  totalDifficulty: "0x4",
  transactionsRoot: "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"
}
```

**Option2**: Do not provide header
In this case, user got a `null` about the header because fetching apis in Ethereum namespace does not return error, so user will always be got `null` as result.

aidan-kwon: I think Option2 is better. All block signatures should be recoverable, and `miner` is the field Klaytn should provide the correct value. It's not the field dummy value is allowed. 
Even though Ethereum doesn't return errors for this API, I think returning errors will not make formatting errors because SDK usually defines a formatter for an entire API instead of each API. We can build rules like the following. 
- For the error cases Ethereum returns nil, Klaytn returns nil also. 
- For the error cases Ethereum doesn't expect, Klaytn returns errors.

aeharvlee: @aidan-kwon 

Great. I had a few questions about your comment.
There are two apis returning Block Header in this PR.
```go
func (api *EthereumAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockNumber) (map[string]interface{}, error) {}

func (api *EthereumAPI) GetHeaderByHash(ctx context.Context, hash common.Hash) map[string]interface{} {}
```

I understood your comment like below.
* `GetHeaderByNumber` should return error with `nil` at block header when error happened during fetching miner address.
* `GetHeaderByHash` should return `nil` because it only returns header, not err by it's function signature.

Am I understanding correctly?

aeharvlee: @aidan-kwon 
Applied your feedback [here](https://github.com/klaytn/klaytn/commit/84e603725f56d282ee29a8a48538a2cdb8dce47b).

Now if error happened during fetching miner address,
user would will got response like below.

```shell
$ curl http://localhost:8551 \
-X POST \
-H "Content-Type: application/json" \
-d '{"jsonrpc":"2.0","method":"eth_getHeaderByNumber","params":["0x4"],"id":1}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"recovery failed"}}
```
> It returns error because the function signature of `eth_getHeaderByNumber` returns both header and error.


```shell
$ curl http://localhost:8551 \
-X POST \
-H "Content-Type: application/json" \
-d '{"jsonrpc":"2.0","method":"eth_getHeaderByHash","params":["0xcb2a163d2658d718e031c513088ef2064a2deea8320cff4a3ed1d481cfd1cbe9"],"id":1}'
{"jsonrpc":"2.0","id":1,"result":null}
```
> It returns null as result because the function signature of `eth_getHeaderByHash` returns only header.

api/api_ethereum_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Creates a MockEngine.
+	mockEngine := mocks.NewMockEngine(mockCtrl)
+	// GetHeader APIs calls internally below methods.
+	mockBackend.EXPECT().Engine().Return(mockEngine)
+	// Author is called when calculates miner field of Header.
+	mockEngine.EXPECT().Author(gomock.Any()).Return(common.Address{}, nil)
+	mockBackend.EXPECT().GetTd(gomock.Any()).Return(new(big.Int))
+
+	header := types.CopyHeader(&types.Header{}) // Creates an empty header
```
aidan-kwon, line 53: Why don't you put values in this header? Without this, the following tests only check field existence and dummy values after calling `rpcMarshalHeader`. If this header contains proper values corresponding to the json example below, this TC can test whether `rpcMarshalHeader` assignes all field values properly.

aeharvlee: @aidan-kwon 
Thanks for your feedback.
As you said, I should have been written test cases like you describe.
Applied your feedback [here](https://github.com/klaytn/klaytn/pull/1088/commits/9db424abe98f41656def6bc7693e93068ef6b4be).

Now TCs can checks whether `rpcMarshalHeader` assigns all field values properly or not.
In the [line 157 - 177](https://github.com/klaytn/klaytn/pull/1088/commits/9db424abe98f41656def6bc7693e93068ef6b4be#diff-5929e238aaa795539ea5f0eb411cd553e712ae2bc234cdf6cde9f28257c87149R157-R177) we can make sure that `rpcMarshalHeader` didn't modify it's original values of header.

aeharvlee: @aidan-kwon 
Refactored test case with just checking json marshaled data [here](https://github.com/klaytn/klaytn/pull/1088/commits/521fb3c022a7f020bbff4cc370c8685acfd7a2ad).

Thanks.

aidan-kwon (approved):
Thank you!

jimni1222 (approved):

yoomee1313 (approved):

sirano11 (approved):

PR #1092, opened 2021-12-21
Introduce a txpool spam throttler prototype

aidan-kwon:
## Proposed changes

- Implement a prototype of txpool spam throttler. 
- admin namespace APIs: startSpamThrottler, stopSpamThrottler, spamThrottlerConfig, setSpamThrottlerWhiteList, getSpamThrottlerWhiteList, getSpamThrottlerThrottleList, getSpamThrottlerCandidateList
- Add a unit test for spam throttle 

### Test Results 
**Environment** 
- 1 CN / 2 PN / 4 EN (m5.2xlarge)
  - EN1, EN2, and EN3 receives massive revert txs (revertTxTC)
  - EN4, receives normal txs (newValueTransferTC)
- Request: 100 RPS normal txs + 3000 RPS revert txs  with 1000 to addresses
- Spam throttler configuration: DefaultSpamThrottlerConfig

**Applying spam throttler result**
- Average TPS: 100 TPS normal txs + 100 TPS revert txs 
- TPS of revert txs peak every 300 seconds (Default throttling duration)
- Some EN's txpool that keep receiving massive revert txs are continuously full of txs 
- The throttler threshold decreased to MinimumThreshold because  the revert ratio of blocks keep exceeding TargetFailRatio
image
image

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/spam_throttler.go
+++ b/blockchain/spam_throttler.go
@@ -45,10 +45,10 @@ var (
 type throttler struct {
 	config *ThrottlerConfig
 
-	candidates map[*common.Address]int  // throttle candidates with spam weight. Not for concurrent use
-	throttled  map[*common.Address]int  // throttled addresses. It requires mu.lock for concurrent use
-	allowed    map[*common.Address]bool // white listed addresses. It requires mu.lock for concurrent use
-	mu         *sync.RWMutex            // mutex for throttled and allowed
+	candidates map[common.Address]int  // throttle candidates with spam weight. Not for concurrent use
+	throttled  map[common.Address]int  // throttled addresses. It requires mu.lock for concurrent use
+	allowed    map[common.Address]bool // white listed addresses. It requires mu.lock for concurrent use
+	mu         *sync.RWMutex           // mutex for throttled and allowed
 
 	threshold  int
 	throttleCh chan *types.Transaction
@@ -130,10 +130,10 @@ func (t *throttler) adjustThreshold(ratio uint) {
 }
 
 // newAllowed generates a new allowed list of throttler.
-func (t *throttler) newAllowed(allowed []*common.Address) {
+func (t *throttler) newAllowed(allowed []common.Address) {
 	t.mu.Lock()
 
-	a := make(map[*common.Address]bool, len(allowed))
+	a := make(map[common.Address]bool, len(allowed))
 	for _, addr := range allowed {
 		a[addr] = true
 	}
@@ -147,8 +147,8 @@ func (t *throttler) newAllowed(allowed []*common.Address) {
 }
 
 // updateThrottled removes outdated addresses from the throttle list and adds new addresses to the list.
-func (t *throttler) updateThrottled(newThrottled []*common.Address) {
-	var removeThrottled []*common.Address
+func (t *throttler) updateThrottled(newThrottled []common.Address) {
+	var removeThrottled []common.Address
 	t.mu.Lock()
 
 	// Decrease spam weight for all throttled addresses.
@@ -177,8 +177,8 @@ func (t *throttler) updateThrottled(newThrottled []*common.Address) {
 
 // updateThrottlerState updates the throttle list by calculating spam weight of candidates.
 func (t *throttler) updateThrottlerState(txs types.Transactions, receipts types.Receipts) {
-	var removeCandidate []*common.Address
-	var newThrottled []*common.Address
+	var removeCandidate []common.Address
+	var newThrottled []common.Address
 
 	startTime := time.Now()
 	numFailed := 0
@@ -195,14 +195,15 @@ func (t *throttler) updateThrottlerState(txs types.Transactions, receipts types.
 				continue
 			}
 
-			weight := t.candidates[toAddr]
+			weight := t.candidates[*toAddr]
 			if weight == 0 {
 				if mapSize >= t.config.WeightMapSize {
 					continue
 				}
 				mapSize++
 			}
-			t.candidates[toAddr] = weight + t.config.IncreaseWeight
+
+			t.candidates[*toAddr] = weight + t.config.IncreaseWeight
 		}
 	}
 
@@ -211,7 +212,7 @@ func (t *throttler) updateThrottlerState(txs types.Transactions, receipts types.
 		newWeight := weight - t.config.DecreaseWeight
 
 		switch {
-		case newWeight < 0:
+		case newWeight <= 0:
 			removeCandidate = append(removeCandidate, addr)
 
 		case newWeight > t.threshold:
@@ -249,7 +250,7 @@ func (t *throttler) classifyTxs(txs types.Transactions) (types.Transactions, typ
 
 	t.mu.RLock()
 	for _, tx := range txs {
-		if tx.To() != nil && t.throttled[tx.To()] > 0 && t.allowed[tx.To()] == false {
+		if tx.To() != nil && t.throttled[*tx.To()] > 0 && t.allowed[*tx.To()] == false {
 			throttleTxs = append(throttleTxs, tx)
 		} else {
 			allowTxs = append(allowTxs, tx)
@@ -261,20 +262,17 @@ func (t *throttler) classifyTxs(txs types.Transactions) (types.Transactions, typ
 }
 
 // SetAllowed resets the allowed list of throttler. The previous list will be abandoned.
-func (t *throttler) SetAllowed(addrs []*common.Address) {
+func (t *throttler) SetAllowed(addrs []common.Address) {
 	t.mu.Lock()
-	t.allowed = make(map[*common.Address]bool)
+	t.allowed = make(map[common.Address]bool)
 	for _, addr := range addrs {
-		if addr == nil {
-			continue
-		}
 		t.allowed[addr] = true
 	}
 	t.mu.Unlock()
 }
 
-func (t *throttler) GetAllowed() []*common.Address {
-	var addrs []*common.Address
+func (t *throttler) GetAllowed() []common.Address {
+	var addrs []common.Address
 
 	t.mu.RLock()
 	for addr := range t.allowed {
@@ -285,8 +283,8 @@ func (t *throttler) GetAllowed() []*common.Address {
 	return addrs
 }
 
-func (t *throttler) GetThrottled() []*common.Address {
-	var addrs []*common.Address
+func (t *throttler) GetThrottled() []common.Address {
+	var addrs []common.Address
 
 	t.mu.RLock()
 	for addr := range t.throttled {
@@ -297,7 +295,7 @@ func (t *throttler) GetThrottled() []*common.Address {
 	return addrs
 }
 
-func (t *throttler) GetCandidates() map[*common.Address]int {
+func (t *throttler) GetCandidates() map[common.Address]int {
 	t.mu.RLock()
 	defer t.mu.RUnlock()
 
--- a/blockchain/spam_throttler.go
+++ b/blockchain/spam_throttler.go
@@ -30,9 +30,8 @@ import (
 // TODO-Klaytn: move these variables into TxPool when BlockChain struct contains a TxPool interface
 // spamThrottler need to be accessed by both of TxPool and BlockChain.
 var (
-	DisableSpamThrottlerAtRuntime            = false
-	spamThrottler                 *throttler = nil
-	spamThrottlerMu                          = new(sync.RWMutex)
+	spamThrottler   *throttler = nil
+	spamThrottlerMu            = new(sync.RWMutex)
 )
 
 var (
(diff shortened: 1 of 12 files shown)


cmd/utils/flags.go
```diff
@@ -1442,6 +1448,11 @@ func setTxPool(ctx *cli.Context, cfg *blockchain.TxPoolConfig) {
 	if ctx.GlobalIsSet(TxPoolLifetimeFlag.Name) {
 		cfg.Lifetime = ctx.GlobalDuration(TxPoolLifetimeFlag.Name)
 	}
+
+	// PN specific txpool setting
+	if ctx.GlobalIsSet(TxPoolSpamThrottlerDisalbleFlag.Name) {
+		blockchain.EnableSpamThrottlerAtRuntime = !ctx.GlobalBool(TxPoolSpamThrottlerDisalbleFlag.Name)
```
jeongkyun-oh, line 1454: Doesn't this always set `false`?

aidan-kwon: EN/CN will return `false` for `ctx.GlobalIsSet(TxPoolSpamThrottlerDisalbleFlag.Name)`. So, this logic will not set `blockchain.EnableSpamThrottlerAtRuntime` for EN/CN.
PN will set `blockchain.EnableSpamThrottlerAtRuntime` by default.

jeongkyun-oh (dismissed):

blockchain/spam_throttler.go
```diff
@@ -0,0 +1,258 @@
...
+	throttled  map[*common.Address]int  // throttled addresses. It requires mu.lock for concurrent use
+	allowed    map[*common.Address]bool // white listed addresses. It requires mu.lock for concurrent use
+	mu         *sync.RWMutex            // mutex for throttled and allowed
+
+	threshold  int
+	throttleCh chan *types.Transaction
+	quitCh     chan struct{}
+}
+
+type throttlerConfig struct {
```
jeongkyun-oh, line 57: @aidan-kwon CI said the fields of this config should be public with json tag.

blockchain/spam_throttler.go
```diff
@@ -0,0 +1,303 @@
...
+
+	return nil
+}
+
+// adjustThreshold adjusts the spam weight threshold of throttler in an adaptive way.
+func (t *throttler) adjustThreshold(ratio uint) {
+	var newThreshold int
+	// Decrease threshold if a fail ratio is bigger than target value to put more addresses in throttled map
+	if ratio > t.config.TargetFailRatio {
+		newThreshold = t.threshold - t.config.IncreaseWeight
```
KimKyungup, line 113: How about this code?
```suggestion
	if t.threshold > t.config.MinimumThreshold {	
		newThreshold = t.threshold - t.config.IncreaseWeight
	} else {	
	        newThreshold = t.config.MinimumThreshold
	}        
```

blockchain/spam_throttler.go
```diff
@@ -0,0 +1,303 @@
...
+
+		// Set minimum threshold
+		if newThreshold < t.config.MinimumThreshold {
+			newThreshold = t.config.MinimumThreshold
+		}
+
+		// Increase threshold if a fail ratio is smaller than target ratio until it exceeds InitialThreshold
+	} else {
+		newThreshold = t.threshold + t.config.IncreaseWeight
```
KimKyungup, line 122: I think modifying this like the above comment is better.

blockchain/tx_pool.go
```diff
@@ -924,10 +933,114 @@ func (pool *TxPool) HandleTxMsg(txs types.Transactions) {
...
+
+		// Activate spam throttler when pool has enough txs
+		if poolSize > uint64(spamThrottler.config.ActivateTxPoolSize) {
+			allowTxs, throttleTxs := spamThrottler.classifyTxs(txs)
+
+			for _, tx := range throttleTxs {
+				select {
+				case spamThrottler.throttleCh <- tx:
+				default:
+					logger.Trace("drop a tx when throttleTxs channel is full", "txHash", tx.Hash())
```
KimKyungup, line 952: If we add a counter metric for dropped tx, It can be helpful.

blockchain/spam_throttler.go
aidan-kwon: Reflected. Please check this agains

blockchain/spam_throttler.go
aidan-kwon: Reflected. Please check this agains

blockchain/tx_pool.go
```diff
@@ -924,10 +933,114 @@ func (pool *TxPool) HandleTxMsg(txs types.Transactions) {
...
+				}
+			}
+
+			txs = allowTxs
+		}
+	}
+
+	// TODO-Klaytn: Consider removing the next line and move the above logic to `addTx` or `AddRemotes`
```
KimKyungup, line 960: I think moving this line is not good. Because it is better to recover as soon as possible

blockchain/tx_pool.go
aidan-kwon: added

blockchain/tx_pool.go
KimKyungup: I think you want to consider the throttled transactions. For that reason, we can consider it later.

blockchain/tx_pool.go
```diff
@@ -924,10 +933,116 @@ func (pool *TxPool) HandleTxMsg(txs types.Transactions) {
...
+
+	go pool.throttleLoop(t)
+
+	spamThrottler = t
+	logger.Info("Start spam throttler", "config", *conf)
+	return nil
+}
+
+func (pool *TxPool) StopSpamThrottler() {
+	spamThrottlerMu.Lock()
```
KimKyungup, line 1031: Don't we need any wait group?

console/web3ext/web3ext.go
```diff
@@ -318,6 +318,33 @@ web3._extend({
 			call: 'admin_setMaxSubscriptionPerWSConn',
 			params: 1
 		}),
+		new web3._extend.Method({
+			name: 'startSpamThrottler',
+			call: 'admin_startSpamThrottler',
+			params: 1,
+			inputFormatter: [null]
```
KimKyungup, line 325: Can we call like this without any parameter? How is `config *blockchain.ThrottlerConfig` included?
``` admin.startSpamThrottler()```

File: blockchain/tx_pool.go (hunk 6 above)
Comment (aidan-kwon) [reply blockchain/tx_pool.go:961]: I added this comment because I think calling `sendCacher.recover` here doesn't make processing faster but enqueues senderCacher task channels twice.
`sendCacher.recover` is called again after checking txpool capacity in `pool.addTxs`. And, `pool.txMsgCh <- txs` is hard to block something because the channel pops txs as soon as the channel is filled without any blocker.

File: console/web3ext/web3ext.go (hunk 8 above)
Comment (aidan-kwon) [reply console/web3ext/web3ext.go:325]: With this inputFormatter, we can call in both ways of `admin.startSpamThrottler()` and `admin.startSpamThrottler({"activate_tx_pool_size": 1000, ...})`

File: blockchain/tx_pool.go (hunk 7 above)
Comment (aidan-kwon) [reply blockchain/tx_pool.go:1031]: I thought using methods of the spamThrottler instance is okay after `spamThrottler =nil`. `StopSpamThrottler` doesn't free a throttler, but GC does.

File: blockchain/spam_throttler.go [hunk 9]
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	return nil
+}
+
+// adjustThreshold adjusts the spam weight threshold of throttler in an adaptive way.
+func (t *throttler) adjustThreshold(ratio uint) {
+	var newThreshold int
+	// Decrease threshold if a fail ratio is bigger than target value to put more addresses in throttled map
+	if ratio > t.config.TargetFailRatio {
```
KimKyungup, line 113: How about change variable name like below?
- IncreaseWeight -> failRatioWeight.   (This will be used for both increasing and decreasing thresholds)
- DecreaseWeight -> timedecreaseWeight

KimKyungup (approved):
LGTM

aidan-kwon:
Reviewers, PTAL

blockchain/spam_throttler.go
```diff
@@ -0,0 +1,304 @@
...
+	// Decrease spam weight for all candidates and update throttle lists in throttled.
+	for addr, weight := range t.candidates {
+		newWeight := weight - t.config.DecreaseWeight
+
+		switch {
+		case newWeight <= 0:
+			removeCandidate = append(removeCandidate, addr)
+
+		case newWeight > t.threshold:
+			removeCandidate = append(removeCandidate, addr)
```
aeharvlee, line 217: @aidan-kwon 
Hi aidan. Sorry for late review, but have a question here.

Does `newWeight > t.threshold` means "it should be limited because it exceeds threshold"? 
If so, I can understand Line 218, but hard to grasp meaning of Line 217 because I think `removeCandidate` means "the address list which should be removed because now it is under threshold."

Can you explain why Line 217 is needed? 
Thanks in advance.

blockchain/spam_throttler.go
```diff
@@ -107,18 +110,19 @@ func (t *throttler) adjustThreshold(ratio uint) {
...
-		newThreshold = t.threshold - t.config.IncreaseWeight
-
-		// Set minimum threshold
-		if newThreshold < t.config.IncreaseWeight {
-			newThreshold = t.config.IncreaseWeight
+		if t.threshold-t.config.ThresholdAdjustment > t.config.MinimumThreshold {
+			newThreshold = t.threshold - t.config.ThresholdAdjustment
+		} else {
+			// Set minimum threshold
+			newThreshold = t.config.MinimumThreshold
```
KimKyungup, line 117: This value is changed from 5 to 100 . Is it ok?

aidan-kwon: Yes. it's intended. the minimum threshold is changed from IncreaseWeight (5) to MinimumThreshold (300)

blockchain/spam_throttler.go
aidan-kwon: 1) For the ease of managing `candidates` and `throttled`
Both threshold and weight are variable. Even though a weight keeps decreasing, it can be re-registered in the throttled list if the threshold decrease more sharply. It could be false-positive detecting of throttled address. And, decreasing weight to 0 takes many times (100 ~ 500 blocks). The throttle takes lots of time to manage weights of already throttled addresses

2) For continuous throttling 
Even though an address is in the `throttled`, it may keep generating revert txs exceeding the threshold. In that case, current logic allows resetting `ThrottleSeconds` of the throttled address by accumulating the weight again

blockchain/spam_throttler.go
```diff
@@ -0,0 +1,304 @@
...
+	mu         *sync.RWMutex           // mutex for throttled and allowed
+
+	threshold  int
+	throttleCh chan *types.Transaction
+	quitCh     chan struct{}
+}
+
+type ThrottlerConfig struct {
+	ActivateTxPoolSize uint `json:"activate_tx_pool_size"`
+	TargetFailRatio    uint `json:"target_fail_ratio"`
```
aeharvlee, line 61: @aidan-kwon 
Very minor thing, but how about changing the word Ratio to Percentile? Maybe Percentile is more adequate word I think.
You can ignore this comment though :) 

If there is a percentage `50%`, it can be represented as
Ratio representation: `0.5`
Percentile representation: `50`

aidan-kwon: Can I change this in another PR? Since this PR is included in v1.7.3 already. I want to change the name in the next version. The late change may not result in compatibility issues because it is a prototype feature only for CCOs. 
Thank you for the comment!

aeharvlee: @aidan-kwon 
Sure, why not?
Thanks aidan.

blockchain/spam_throttler.go
```diff
@@ -0,0 +1,304 @@
...
+	t := spamThrottler
+	spamThrottlerMu.RUnlock()
+	return t
+}
+
+func validateConfig(conf *ThrottlerConfig) error {
+	if conf == nil {
+		return errors.New("nil ThrottlerConfig")
+	}
+	if conf.TargetFailRatio > 100 {
```
aeharvlee, line 98:
```suggestion
	if conf.TargetFailRatio > 100 || conf.TargetFailRatio < 0 {
```

Based on error message, I think is should be handle when the given `conf.TargetFailRatio` is under zero.

aidan-kwon: I didn't put the condition because the type of TargetFailRatio is `uint`, but I will consider adding the condition for make sure later

aeharvlee: @aidan-kwon 
My miss..!! I didn't consider that variable is `uint`. Thanks for your reply

KimKyungup (approved):

jeongkyun-oh (approved):

aeharvlee (approved):

jeongkyun-oh:
It would be better if you attach the configuration setting of smapfilter with the test result.

PR #1099, opened 2022-01-03
Get rid of dual mutex in blockchain.go

hqjang-pepper:
## Proposed changes

- This PR removes `chainmu` in blockchain.go.
- The mu supposed to protect internal state from getting corrupted whilst chaimu from multiple modifications to the chain. 
- I found that functions such as Rollback or SetHead may cause race condition problems with chain insertion, since they use different mutex but can be called at the same time.

Related Ethereum PR : https://github.com/ethereum/go-ethereum/pull/18436/files

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


jeongkyun-oh:
@hqjang-pepper Could you please check why the test cases are failed.

hqjang-pepper:
> 



> @hqjang-pepper Could you please check why the test cases are failed.

Following check-fails are not directly from this PR changes.
Since I had no problems in local tests, so it's hard to reason with..
Gotta need some time to figure it out to deal with.

blockchain/blockchain.go
```diff
@@ -159,8 +159,7 @@ type BlockChain struct {
 	scope         event.SubscriptionScope
 	genesisBlock  *types.Block
 
-	mu      sync.RWMutex // global mutex for locking chain operations
-	chainmu sync.RWMutex // blockchain insertion lock
+	mu sync.RWMutex // global mutex for locking chain operations
```
jeongkyun-oh, line 162: @hqjang-pepper In Ethereum, `mu` is removed, but this PR removes `chainmu`. Do you have any reason for this?
Furthermore, in the given Ethereum PR, with removing `mu`, some logic is modified such as `WriteBlockWithState`. Is this okay?

hqjang-pepper: Fixed some part of `WriteBlockWithState`. Thanks for your advice.
`mu` of Klaytn has exactly same role with `chainmu` in Ethereum. Just naming issue. 
Since mutex in blockchain.go is protecting globally-used fields , not just chain insertion. 
So I left `mu` instead of `chainmu`.

blockchain/blockchain.go
```diff
@@ -1405,9 +1401,6 @@ func (bc *BlockChain) writeBlockWithStateParallel(block *types.Block, receipts [
 			"hash", block.Hash(), "parentHash", block.ParentHash())
 		return WriteResult{Status: NonStatTy}, consensus.ErrUnknownAncestor
 	}
-	// Make sure no inconsistent state is leaked during insertion
-	bc.mu.Lock()
-	defer bc.mu.Unlock()
```
jeongkyun-oh, line 1410: @hqjang-pepper Removing this is okay?

hqjang-pepper: There are only 2 cases where `WriteBlockWithState` is used.
1. `wait()` in `worker.go` -> protects worker's state with `stateMu`.
2. `InsertChain()` in `blockchain.go` -> protects state with `mu`.

I think `stateMu` cannot protect state in chain.
Additional version of `writeBlockwithState` (which contains `mu` lock/unlock, for sake of `worker.go`)will be needed.

thanks for your advice!

blockchain/blockchain.go
yoomee1313, line 162: What about bringing https://github.com/ethereum/go-ethereum/pull/22853? This changes the mutex type to ClosableMutexType

blockchain/blockchain.go
```diff
@@ -2279,26 +2279,6 @@ func (bc *BlockChain) InsertHeaderChain(chain []*types.Header, checkFreq int) (i
...
-// writeHeader writes a header into the local chain, given that its parent is
-// already known. If the total blockscore of the newly inserted header becomes
-// greater than the current known TD, the canonical chain is re-routed.
-//
-// Note: This method is not concurrent-safe with inserting blocks simultaneously
-// into the chain, as side effects caused by reorganisations cannot be emulated
-// without the real blocks. Hence, writing headers directly should only be done
-// in two scenarios: pure-header mode of operation (light clients), or properly
-// separated header/block phases (non-archive clients).
-func (bc *BlockChain) writeHeader(header *types.Header) error {
```
yoomee1313, line 2291: Can I ask why you remove this function?

hqjang-pepper: Because this function is in no-usage.
Ethereum also removed this function.

blockchain/blockchain.go
hqjang-pepper: If it's needed, I'll bring it in another PR.
This PR is only about getting rid of `chainmu`.

ehnuje (approved):
LGTM, however, have you found the root cause of timeout issue on CircleCI test?

jeongkyun-oh (approved):

hqjang-pepper:
> LGTM, however, have you found the root cause of timeout issue on CircleCI test?

I missed removing mutex at `WriteBlockWithState`. 
Without removing, it causes race condition when `InsertChain` is called.
Changes were applied at this following PR. https://github.com/klaytn/klaytn/pull/1099/commits/e93aae4d58d66121dd16ba0b687ee7fe0fed3ff4

ehnuje:
> > LGTM, however, have you found the root cause of timeout issue on CircleCI test?
> 
> I missed removing mutex at `WriteBlockWithState`. Without removing, it causes race condition when `InsertChain` is called. Changes were applied at this following PR. [e93aae4](https://github.com/klaytn/klaytn/commit/e93aae4d58d66121dd16ba0b687ee7fe0fed3ff4)

@hqjang-pepper If so, why didn't it make any problem on the local test? Due to timing issue?

blockchain/blockchain.go
aidan-kwon: If it is not an important reason, please consider using the same name with Ethreeum for compatibility the next time. It would be better to cherry-pick their improvements.

aidan-kwon (approved):

hqjang-pepper:
> > > LGTM, however, have you found the root cause of timeout issue on CircleCI test?
> > 
> > 
> > I missed removing mutex at `WriteBlockWithState`. Without removing, it causes race condition when `InsertChain` is called. Changes were applied at this following PR. [e93aae4](https://github.com/klaytn/klaytn/commit/e93aae4d58d66121dd16ba0b687ee7fe0fed3ff4)
> 
> @hqjang-pepper If so, why didn't it make any problem on the local test? Due to timing issue?

@ehnuje Sorry for late response. 
There was no timeout issue when I tested through `make test`, which was different with results from CircleCi. I'm not sure that the reason no timeout failure was timing issue. But after I checked again, timeout issue also occurred in my local environment, and I found the cause by debugging.

blockchain/blockchain.go
```diff
@@ -1721,9 +1723,9 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 				winner[j], winner[len(winner)-1-j] = winner[len(winner)-1-j], winner[j]
 			}
 			// Import all the pruned blocks to make the state available
-			bc.chainmu.Unlock()
+			bc.mu.Unlock()
```
kjhman21, line 1726: I don't know it is a good approach. What about applying the same approach as WriteBlockWithState adn writeBlockWithState? 
Calling locking and unlocking in InsertChain?

kjhman21:
@hqjang-pepper It requires heavy testing especially changing calling locations of locks. How did you conduct testing?

blockchain/blockchain.go
```diff
@@ -2262,15 +2264,13 @@ func (bc *BlockChain) InsertHeaderChain(chain []*types.Header, checkFreq int) (i
...
-	bc.chainmu.Lock()
-	defer bc.chainmu.Unlock()
+	bc.mu.Lock()
+	defer bc.mu.Unlock()
 
 	bc.wg.Add(1)
 	defer bc.wg.Done()
 
 	whFunc := func(header *types.Header) error {
-		bc.mu.Lock()
```
kjhman21, line 2272: @hqjang-pepper I'm not sure we can remove this. This function is called in InsertrHeaderChain below.

If we remove here, InsertHeaderChain will be executed without holding mutex.


Could you please run this modification with klaytn-deploy with heavy load?

hqjang-pepper: Got it. I'll conduct load test with this code.

kjhman21 (approved):

yoomee1313 (approved):

hqjang-pepper (commented):
I conducted load test on two version, before/after mutex integration.
As a result of sending enough requests(more than 1m), the TPS difference between the two versions was 1~10. 
It's not that big difference, so we decided to merge this PR.

Issue #1105, opened 2022-01-07
addVote api improvement - addValidator/removeValidator vote value accepts multiple nodes


yoomee1313:
**Describe the solution you'd like**
addValidator/removeValidator vote value accepts multiple nodes

**Additional context**
AS-IS
```
> governance.vote("governance.addValidator", "0xe93a890fb7ec5f973b1a7fd77b0d13a0763eff3d,0x88fa5858c542f351086f3de88de8789e4c599000")
Error: Your vote couldn't be placed. Please check your vote's key and value
    at web3.js:3276:20
    at web3.js:6450:15
    at web3.js:5214:36
    at <anonymous>:1:1
>
```

TO-BE
```
> governance.vote("governance.addValidator", "0xe93a890fb7ec5f973b1a7fd77b0d13a0763eff3d,0x88fa5858c542f351086f3de88de8789e4c599000")
"Your vote was successfully placed."
> 
```

PR #1110, opened 2022-01-13
governance: accept multiple validator addresses when voting

yoomee1313:
## Proposed changes

- This PR enables accepting multiple validator addresses when voting `governance.addvalidator` and `governance.removevalidator`
- The parsing logic of vote api (when key is `governance.addvalidator`, `governance.removevalidator`) has changed
  - dev: it transforms the value to address.
  - pr: it counts the number of node addresses. If number of node addresses is 1, it transforms the value to address. If not, it transforms the value to the address array.

closes https://github.com/klaytn/klaytn/issues/1105.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/governance/api.go
+++ b/governance/api.go
@@ -98,7 +98,7 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 		if _, ok := api.governance.ValidateVote(&GovernanceVote{Key: key, Value: val}); !ok {
 			return "", errInvalidKeyValue
 		}
-		if api.isRemovingSelf(val) {
+		if api.isRemovingSelf(val.(string)) {
 			return "", errRemoveSelf
 		}
 	}
@@ -108,8 +108,8 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 	return "", errInvalidKeyValue
 }
 
-func (api *PublicGovernanceAPI) isRemovingSelf(val interface{}) bool {
-	for _, str := range strings.Split(val.(string), ",") {
+func (api *PublicGovernanceAPI) isRemovingSelf(val string) bool {
+	for _, str := range strings.Split(val, ",") {
 		str = strings.Trim(str, " ")
 		if common.HexToAddress(str) == api.governance.nodeAddress.Load().(common.Address) {
 			return true
--- a/governance/api.go
+++ b/governance/api.go
@@ -110,6 +110,7 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 
 func (api *PublicGovernanceAPI) isRemovingSelf(val interface{}) bool {
 	for _, str := range strings.Split(val.(string), ",") {
+		str = strings.Trim(str, " ")
 		if common.HexToAddress(str) == api.governance.nodeAddress.Load().(common.Address) {
 			return true
 		}
--- a/governance/api.go
+++ b/governance/api.go
@@ -19,7 +19,6 @@ package governance
 import (
 	"errors"
 	"math/big"
-	"reflect"
 	"strings"
 	"sync/atomic"
 
@@ -96,11 +95,7 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 		return "", errPermissionDenied
 	}
 	if strings.ToLower(key) == "governance.removevalidator" {
-		if reflect.TypeOf(val).String() != "string" {
-			return "", errInvalidKeyValue
-		}
-		target := val.(string)
-		if !common.IsHexAddress(target) {
+		if _, ok := api.governance.ValidateVote(&GovernanceVote{Key: key, Value: val}); !ok {
 			return "", errInvalidKeyValue
 		}
 		if api.isRemovingSelf(val) {
@@ -114,13 +109,12 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 }
 
 func (api *PublicGovernanceAPI) isRemovingSelf(val interface{}) bool {
-	target := val.(string)
-
-	if common.HexToAddress(target) == api.governance.nodeAddress.Load().(common.Address) {
-		return true
-	} else {
-		return false
+	for _, str := range strings.Split(val.(string), ",") {
+		if common.HexToAddress(str) == api.governance.nodeAddress.Load().(common.Address) {
+			return true
+		}
 	}
+	return false
 }
 
 func (api *PublicGovernanceAPI) ShowTally() []*returnTally {
--- a/governance/default.go
+++ b/governance/default.go
@@ -499,7 +499,6 @@ func (g *Governance) getKey(k string) string {
 func (g *Governance) RemoveVote(key string, value interface{}, number uint64) {
 	k := GovernanceKeyMap[key]
 	if isEqualValue(k, g.voteMap.GetValue(key).Value, value) {
-		//if GovernanceItems[k].isEqual(g.voteMap.GetValue(key).Value, value) {
 		g.voteMap.SetValue(key, VoteStatus{
 			Value:  value,
 			Casted: true,
@@ -519,7 +518,7 @@ func (g *Governance) ClearVotes(num uint64) {
 	logger.Info("Governance votes are cleared", "num", num)
 }
 
-// parseVoteValue parse vote.Value from []uint8 to appropriate type
+// parseVoteValue parse vote.Value from []uint8, [][]uint8 to appropriate type
 func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, error) {
 	var val interface{}
 	k, ok := GovernanceKeyMap[gVote.Key]
--- a/governance/default.go
+++ b/governance/default.go
@@ -530,12 +530,23 @@ func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, err
 	// checking type of gVote.Value
 	switch k {
 	case params.AddValidator, params.RemoveValidator:
-		// gVote.Value type should be [][]interface{}(multiple node removal/addition) or []uint8(single node removal/addition)
 		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([]interface{}{}) && reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{}) {
 			return nil, ErrValueTypeMismatch
 		}
+
+		// if value contains multiple addresses, gVote.Value type should be [][]uint8{}
+		if reflect.TypeOf(gVote.Value) == reflect.TypeOf([]interface{}{}) {
+			for _, item := range gVote.Value.([]interface{}) {
+				if reflect.TypeOf(item) != reflect.TypeOf([]uint8{}) {
+					return nil, ErrValueTypeMismatch
+				}
+			}
+			break
+		}
+		// if value contains single addresses, gVote.Value type should be []uint8{}
+		fallthrough
 	default:
-		// gVote.Value type should be []uint8
+		// Neither governance.addvalidator key nor governance.removevalidator key, gVote.Value type should be []uint8{}
 		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{}) {
 			return nil, ErrValueTypeMismatch
 		}
@@ -547,9 +558,11 @@ func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, err
 	case params.GoverningNode:
 		val = common.BytesToAddress(gVote.Value.([]uint8))
 	case params.AddValidator, params.RemoveValidator:
-		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([]interface{}{}) {
+		if reflect.TypeOf(gVote.Value) == reflect.TypeOf([]uint8{}) {
+			// value contains single node address, so convert the value to common.address type
 			val = common.BytesToAddress(gVote.Value.([]uint8))
 		} else {
+			// value contains multiple node addresses, so convert the value to array of common.address
 			var nodeAddresses []common.Address
 			for _, item := range gVote.Value.([]interface{}) {
 				nodeAddresses = append(nodeAddresses, common.BytesToAddress(item.([]uint8)))
--- a/governance/default.go
+++ b/governance/default.go
@@ -527,48 +527,30 @@ func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, err
 		return nil, ErrUnknownKey
 	}
 
-	// checking type of gVote.Value
-	switch k {
-	case params.AddValidator, params.RemoveValidator:
-		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([]interface{}{}) && reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{}) {
-			return nil, ErrValueTypeMismatch
-		}
-
-		// if value contains multiple addresses, gVote.Value type should be [][]uint8{}
-		if reflect.TypeOf(gVote.Value) == reflect.TypeOf([]interface{}{}) {
-			for _, item := range gVote.Value.([]interface{}) {
-				if reflect.TypeOf(item) != reflect.TypeOf([]uint8{}) {
-					return nil, ErrValueTypeMismatch
-				}
-				if len(item.([]uint8)) != common.AddressLength {
-					return nil, ErrValueTypeMismatch
-				}
+	// if value contains multiple addresses, gVote.Value type should be [][]uint8{}
+	if v, ok := gVote.Value.([]interface{}); ok && len(v) != 0 {
+		for _, item := range v {
+			if in, ok := item.([]uint8); !ok || len(in) != common.AddressLength {
+				return nil, ErrValueTypeMismatch
 			}
-			break
 		}
-		// if value contains single addresses, gVote.Value type should be []uint8{}
-		fallthrough
-	default:
-		// Neither governance.addvalidator key nor governance.removevalidator key, gVote.Value type should be []uint8{}
-		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{}) {
+		// in other correct cases, gVote.Value type should be []uint8{}
+	} else if v, ok := gVote.Value.([]uint8); ok {
+		if GovernanceItems[k].checkValueType(common.Address{}) && len(v) != common.AddressLength {
 			return nil, ErrValueTypeMismatch
 		}
-		if GovernanceItems[k].checkValueType(common.Address{}) && len(gVote.Value.([]uint8)) != common.AddressLength {
-			return nil, ErrValueTypeMismatch
-		}
+		// otherwise, return ErrValueTypeMismatch (e.g. gVote.Value = nil)
+	} else {
+		return nil, ErrValueTypeMismatch
 	}
 
 	switch k {
 	case params.GovernanceMode, params.MintingAmount, params.MinimumStake, params.Ratio:
 		val = string(gVote.Value.([]uint8))
-	case params.GoverningNode:
-		val = common.BytesToAddress(gVote.Value.([]uint8))
-	case params.AddValidator, params.RemoveValidator:
-		if reflect.TypeOf(gVote.Value) == reflect.TypeOf([]uint8{}) {
-			// value contains single node address, so convert the value to common.address type
-			val = common.BytesToAddress(gVote.Value.([]uint8))
+	case params.GoverningNode, params.AddValidator, params.RemoveValidator:
+		if v, ok := gVote.Value.([]uint8); ok {
+			val = common.BytesToAddress(v)
 		} else {
-			// value contains multiple node addresses, so convert the value to array of common.address
 			var nodeAddresses []common.Address
 			for _, item := range gVote.Value.([]interface{}) {
 				nodeAddresses = append(nodeAddresses, common.BytesToAddress(item.([]uint8)))
--- a/governance/default.go
+++ b/governance/default.go
@@ -534,7 +534,13 @@ func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, err
 			return nil, ErrValueTypeMismatch
 		}
 		val = string(v)
-	case params.GoverningNode, params.AddValidator, params.RemoveValidator:
+	case params.GoverningNode:
+		v, ok := gVote.Value.([]uint8)
+		if !ok || len(v) != common.AddressLength {
+			return nil, ErrValueTypeMismatch
+		}
+		val = common.BytesToAddress(v)
+	case params.AddValidator, params.RemoveValidator:
 		if v, ok := gVote.Value.([]uint8); ok {
 			// if value contains single address, gVote.Value type should be []uint8{}
 			if len(v) != common.AddressLength {
@@ -557,7 +563,6 @@ func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, err
 		} else {
 			return nil, ErrValueTypeMismatch
 		}
-
 	case params.Epoch, params.CommitteeSize, params.UnitPrice, params.StakeUpdateInterval, params.ProposerRefreshInterval, params.ConstTxGasHumanReadable, params.Policy, params.Timeout:
 		v, ok := gVote.Value.([]uint8)
 		if !ok {
--- a/governance/default.go
+++ b/governance/default.go
@@ -527,42 +527,50 @@ func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, err
 		return nil, ErrUnknownKey
 	}
 
-	// if value contains multiple addresses, gVote.Value type should be [][]uint8{}
-	if v, ok := gVote.Value.([]interface{}); ok && len(v) != 0 {
-		for _, item := range v {
-			if in, ok := item.([]uint8); !ok || len(in) != common.AddressLength {
-				return nil, ErrValueTypeMismatch
-			}
-		}
-		// in other correct cases, gVote.Value type should be []uint8{}
-	} else if v, ok := gVote.Value.([]uint8); ok {
-		if GovernanceItems[k].checkValueType(common.Address{}) && len(v) != common.AddressLength {
-			return nil, ErrValueTypeMismatch
-		}
-		// otherwise, return ErrValueTypeMismatch (e.g. gVote.Value = nil)
-	} else {
-		return nil, ErrValueTypeMismatch
-	}
-
 	switch k {
 	case params.GovernanceMode, params.MintingAmount, params.MinimumStake, params.Ratio:
-		val = string(gVote.Value.([]uint8))
+		v, ok := gVote.Value.([]uint8)
+		if !ok {
+			return nil, ErrValueTypeMismatch
+		}
+		val = string(v)
 	case params.GoverningNode, params.AddValidator, params.RemoveValidator:
 		if v, ok := gVote.Value.([]uint8); ok {
+			if GovernanceItems[k].checkValueType(common.Address{}) && len(v) != common.AddressLength {
+				return nil, ErrValueTypeMismatch
+			}
 			val = common.BytesToAddress(v)
-		} else {
+		} else if addresses, ok := gVote.Value.([]interface{}); ok {
+			// if value contains multiple addresses, gVote.Value type should be [][]uint8{}
+			if len(addresses) == 0 {
+				return nil, ErrValueTypeMismatch
+			}
 			var nodeAddresses []common.Address
-			for _, item := range gVote.Value.([]interface{}) {
+			for _, item := range addresses {
+				if in, ok := item.([]uint8); !ok || len(in) != common.AddressLength {
+					return nil, ErrValueTypeMismatch
+				}
 				nodeAddresses = append(nodeAddresses, common.BytesToAddress(item.([]uint8)))
 			}
 			val = nodeAddresses
+		} else {
+			return nil, ErrValueTypeMismatch
 		}
+
 	case params.Epoch, params.CommitteeSize, params.UnitPrice, params.StakeUpdateInterval, params.ProposerRefreshInterval, params.ConstTxGasHumanReadable, params.Policy, params.Timeout:
-		gVote.Value = append(make([]byte, 8-len(gVote.Value.([]uint8))), gVote.Value.([]uint8)...)
-		val = binary.BigEndian.Uint64(gVote.Value.([]uint8))
+		v, ok := gVote.Value.([]uint8)
+		if !ok {
+			return nil, ErrValueTypeMismatch
+		}
+		v = append(make([]byte, 8-len(v)), v...)
+		val = binary.BigEndian.Uint64(v)
 	case params.UseGiniCoeff, params.DeferredTxFee:
-		gVote.Value = append(make([]byte, 8-len(gVote.Value.([]uint8))), gVote.Value.([]uint8)...)
-		if binary.BigEndian.Uint64(gVote.Value.([]uint8)) != uint64(0) {
+		v, ok := gVote.Value.([]uint8)
+		if !ok {
+			return nil, ErrValueTypeMismatch
+		}
+		v = append(make([]byte, 8-len(v)), v...)
+		if binary.BigEndian.Uint64(v) != uint64(0) {
 			val = true
 		} else {
 			val = false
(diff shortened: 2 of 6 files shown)


governance/handler.go
```diff
@@ -127,20 +173,43 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
 
 func (g *Governance) adjustValueType(key string, val interface{}) interface{} {
 	k := GovernanceKeyMap[key]
-	reqType := GovernanceItems[k].t
 	var x interface{}
 
 	// When an int value comes from JS console, it comes as a float64
-	if reqType == uint64T && reflect.TypeOf(val) == float64T {
+	if GovernanceItems[k].checkValueType(uint64(0)) && reflect.TypeOf(val) == float64T {
```
aeharvlee, line 179: @yoomee1313 
Is it ok to passing `uint64(0)` to `checkValueType`, not `val` from caller?

governance/handler.go
```diff
@@ -127,20 +173,43 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
+			return val
+		}
+
+		if len(addresses) == 1 {
+			x = common.HexToAddress(val.(string))
+			return x
+		}
+
+		var nodeAddresses []common.Address
+		for _, str := range strings.Split(val.(string), ",") {
```
aeharvlee, line 200:
```suggestion
		for _, str := range addresses {
```

I think we can use pre-defined local variable `addresses` in here.

governance/handler.go
```diff
@@ -127,20 +173,43 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
+				return val
+			}
+		}
+		x = nodeAddresses
+		return x
+	}
+
 	// address comes as a form of string from JS console
-	if reqType == addressT && reflect.TypeOf(val) == stringT {
+	if GovernanceItems[k].checkValueType(common.Address{}) && reflect.TypeOf(val) == stringT {
```
aeharvlee, line 212: @yoomee1313 
Is it ok to pass `common.Address{}` to checkValueType, not val from caller?

governance/default.go
```diff
@@ -521,18 +521,35 @@ func (g *Governance) ClearVotes(num uint64) {
 // parseVoteValue parse vote.Value from []uint8 to appropriate type
 func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, error) {
 	var val interface{}
-	k := GovernanceKeyMap[gVote.Key]
+	k, ok := GovernanceKeyMap[gVote.Key]
+	if ok == false {
```
jeongkyun-oh, line 525:
```suggestion
	if !ok {
```
Very minor comment

governance/default.go
```diff
@@ -543,8 +560,6 @@ func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, err
 		} else {
 			val = false
 		}
-	default:
-		logger.Warn("Unknown key was given", "key", k)
```
jeongkyun-oh, line 547: Why this line is removed? I think leaving log is better.

governance/default.go
```diff
@@ -521,18 +521,35 @@ func (g *Governance) ClearVotes(num uint64) {
...
 func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, error) {
 	var val interface{}
-	k := GovernanceKeyMap[gVote.Key]
+	k, ok := GovernanceKeyMap[gVote.Key]
+	if ok == false {
+		logger.Warn("Unknown key was given", "key", k)
+		return nil, ErrUnknownKey
+	}
+
+	if (k == params.AddValidator || k == params.RemoveValidator) && (reflect.TypeOf(gVote.Value) != reflect.TypeOf([]interface{}{}) && reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{})) {
```
kjhman21, line 530: Wouldn't it be better to make a switch statement for this and the next conditional statements?

governance/default.go
yoomee1313: sorry for the late. Thanks, reflected at https://github.com/klaytn/klaytn/pull/1110/commits/680be2ebb4e2fc84f7ecab4319657906761f9377.

governance/default.go
yoomee1313: Thanks for the nice advice! reflected at https://github.com/klaytn/klaytn/pull/1110/commits/680be2ebb4e2fc84f7ecab4319657906761f9377

governance/default.go
yoomee1313: This log is also logged at the front of the `ParseVoteValue` function, so I removed it.
```
func (g *Governance) ParseVoteValue(gVote *GovernanceVote) (*GovernanceVote, error) {
	var val interface{}
	k, ok := GovernanceKeyMap[gVote.Key]
	if !ok {
		logger.Warn("Unknown key was given", "key", k)
		return nil, ErrUnknownKey
	}
```

governance/handler.go
yoomee1313: @aeharvlee sorry for the late! I want to check the `reqType` of the key `k` is uint64T. `val` type is checked at the condition at right.

governance/handler.go
yoomee1313: Thanks!! I reflected at https://github.com/klaytn/klaytn/pull/1110/commits/680be2ebb4e2fc84f7ecab4319657906761f9377.

governance/handler.go
yoomee1313: As I wrote at above comment, the condition checks the `reqType` of key `k`.

aeharvlee: @yoomee1313 

Understood. Thanks :)

governance/default.go
```diff
@@ -521,18 +521,41 @@ func (g *Governance) ClearVotes(num uint64) {
...
+	}
 
-	// filter out if vote value is an interface list
-	if reflect.TypeOf(gVote.Value) == reflect.TypeOf([]interface{}{}) {
-		return nil, ErrValueTypeMismatch
+	// checking type of gVote.Value
+	switch k {
+	case params.AddValidator, params.RemoveValidator:
+		// gVote.Value type should be [][]interface{}(multiple node removal/addition) or []uint8(single node removal/addition)
+		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([]interface{}{}) && reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{}) {
```
kjhman21, line 534:
```suggestion
		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([][]interface{}{}) && reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{}) {
```

Need double `[]`?

kjhman21: Or you need to check once more after casting []interface{}.

kjhman21: Actually we only allow `[][]uint8`, don't we?

governance/default_test.go
```diff
@@ -549,6 +545,30 @@ func TestVoteValueNilInterface(t *testing.T) {
 		_, err := gov.ParseVoteValue(gVote)
 		assert.Equal(t, nil, err)
 	}
+
+	gVote.Key = "governance.addvalidator"
+	{
+		gVote.Value = [][]uint8{{0x3, 0x4}, {0x5, 0x6}}
```
kjhman21, line 551: Don't we need to validate those values as if they were addresses?
Do we also allow "0x0304"?

governance/default.go
yoomee1313: Yes, we only allow [][]uint8. Current code doesn't check type of the inner array.. Thanks for the advice! I fixed the code to check the inner array type, too! reflected at https://github.com/klaytn/klaytn/pull/1110/commits/b1e90f018639923869c1fd9e7521b53032ecd69e

governance/default_test.go
yoomee1313: Yes, we allow 0x0304 at vote api. It is converted to `0x0000000000000000000000000000000000000304` while processing vote api(`adjustValueType`).
e.g.
```
> governance.vote("governance.addvalidator", "0x0304")
"Your vote was successfully placed."
> istanbul.getValidators()
["0x0000000000000000000000000000000000000304", "0x00000000000003233494f27b1cae261197d18f54", ...]
> 
```

So, as you suggested, ParseVoteValue should not accept 0x0304. If you agree with it, I'll add check logic and change the test case's results. @kjhman21

kjhman21: @yoomee1313 In that case, it is okay, but we also have test cases for valid addresses. (I'm not sure we have it, but it does not show in this PR review.)

governance/api.go
```diff
@@ -114,13 +109,12 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 }
 
 func (api *PublicGovernanceAPI) isRemovingSelf(val interface{}) bool {
-	target := val.(string)
-
-	if common.HexToAddress(target) == api.governance.nodeAddress.Load().(common.Address) {
-		return true
-	} else {
-		return false
+	for _, str := range strings.Split(val.(string), ",") {
```
aidan-kwon, line 112: 1. Can we ensure that `val` cannot be `nil`? If it is, this line will cause a panic.
2. I think we need to trim `str`. If not, `common.HexToAddress(str)` can return invalid values.

governance/default_test.go
yoomee1313: @kjhman21 I'm sorry, the `adjustValueType` checking logic was wrong. Vote api should not accepts non-address length, too. I added address length checking logic to vote api and parseVoteValue. It is reflected at https://github.com/klaytn/klaytn/pull/1110/commits/4b2028fb3a43a13c90bb2c448aba3ad9f91fae21.

governance/api.go
yoomee1313: > Can we ensure that val cannot be nil? If it is, this line will cause a panic.

`val` cannot be `nil` because those votes are filtered by `api.governance.ValidateVote(&GovernanceVote{Key: key, Value: val})`.

> I think we need to trim str. If not, common.HexToAddress(str) can return invalid values.

Currently,, the vote api does not accept the `str` which contains whitespaces.. thanks for the advice. Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/abdc7e4170e95f7550593e1e957cd0f762258529.

governance/default.go
```diff
@@ -521,18 +521,60 @@ func (g *Governance) ClearVotes(num uint64) {
...
+		if GovernanceItems[k].checkValueType(common.Address{}) && len(gVote.Value.([]uint8)) != common.AddressLength {
+			return nil, ErrValueTypeMismatch
+		}
 	}
 
 	switch k {
 	case params.GovernanceMode, params.MintingAmount, params.MinimumStake, params.Ratio:
 		val = string(gVote.Value.([]uint8))
-	case params.GoverningNode, params.AddValidator, params.RemoveValidator:
+	case params.GoverningNode:
```
kjhman21, line 564: Hmm.. after updating like this, I'm curious that do we really have two switch statements? we can check the casting failure here to detect type mismatch like:

```go
val, ok = string(gVote.Value.([]uint8))
```

governance/default_test.go
```diff
@@ -132,6 +132,13 @@ var tstData = []voteValue{
 	{k: "istanbul.timeout", v: true, e: false},
 	{k: "istanbul.timeout", v: "10", e: false},
 	{k: "istanbul.timeout", v: 5.3, e: false},
+	{k: "governance.addvalidator", v: "0x639e5ebfc483716fbac9810b230ff6ad487f366c", e: true},
+	{k: "governance.addvalidator", v: " 0x639e5ebfc483716fbac9810b230ff6ad487f366c ", e: true},
+	{k: "governance.addvalidator", v: "0x639e5ebfc483716fbac9810b230ff6ad487f366c,0x828880c5f09cc1cc6a58715e3fe2b4c4cf3c5869", e: true},
+	{k: "governance.addvalidator", v: "0x639e5ebfc483716fbac9810b230ff6ad487f366c,0x828880c5f09cc1cc6a58715e3fe2b4c4cf3c58", e: false},
+	{k: "governance.addvalidator", v: "0x639e5ebfc483716fbac9810b230ff6ad487f366c,0x639e5ebfc483716fbac9810b230ff6ad487f366c", e: false},
+	{k: "governance.addvalidator", v: "0x639e5ebfc483716fbac9810b230ff6ad487f366c, 0x828880c5f09cc1cc6a58715e3fe2b4c4cf3c5869, ", e: false},
```
kjhman21, line 140: Can we add a case of double `,`?

```go
{k: "governance.addvalidator", v: "0x639e5ebfc483716fbac9810b230ff6ad487f366c,, 0x828880c5f09cc1cc6a58715e3fe2b4c4cf3c5869, ", e: false},
```

governance/default_test.go
```diff
@@ -277,39 +285,33 @@ func TestGovernance_ClearVotes(t *testing.T) {
...
-			assert.Equal(t, nil, err)
-		}
-
-		if v.Value != gov.voteMap.GetValue(v.Key).Value {
-			t.Errorf("Encoded vote and Decoded vote are different! Encoded: %v, Decoded: %v\n", gov.voteMap.GetValue(v.Key).Value, v.Value)
-		}
-		gov.RemoveVote(v.Key, v.Value, 1000)
+		v, err := gov.ParseVoteValue(v)
+		assert.Equal(t, nil, err)
+		assert.Equal(t, gov.voteMap.GetValue(v.Key).Value, v.Value,
```
kjhman21, line 302: Equal is enough? what about DeepEqual?

governance/default_test.go
```diff
@@ -318,13 +320,10 @@ func TestGovernance_ParseVoteValue(t *testing.T) {
...
-		if d, err = gov.ParseVoteValue(d); err != nil {
-			assert.Equal(t, nil, err)
-		}
-
-		if !reflect.DeepEqual(v, d) {
-			t.Errorf("Parse was not successful! %v %v \n", v, d)
-		}
+		d, err := gov.ParseVoteValue(d)
+		assert.Equal(t, nil, err)
+		assert.Equal(t, reflect.DeepEqual(v, d), true,
```
kjhman21, line 325: need to try `assert.DeepEqual` instead.

governance/handler.go
```diff
@@ -127,22 +156,45 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
 		}
 		return val
 	}
 
-	// address comes as a form of string from JS console
-	if reqType == addressT && reflect.TypeOf(val) == stringT {
-		if common.IsHexAddress(val.(string)) {
-			x = common.HexToAddress(val.(string))
+	if GovernanceItems[k].checkValueType(common.Address{}) && reflect.TypeOf(val) == stringT {
+		addresses := strings.Split(val.(string), ",")
```
kjhman21, line 171: This is not a good practice. If val.(string) fails, the process can panic. This cannot be true because there is a condition statement of type checking, but it would be better to use like below:

```go
addressesStr, ok := val.(string)
if !ok {
  // do something for a failure case
}
// do something for a successful case
```

governance/handler.go
```diff
@@ -127,22 +156,45 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
+
+		if len(addresses) == 0 {
```
kjhman21, line 173: Also here we can make a switch statement here.

governance/handler.go
```diff
@@ -253,6 +304,23 @@ func checkBigInt(k string, v interface{}) bool {
 }
 
 func checkAddress(k string, v interface{}) bool {
+	// if value contains single address, return true
+	if reflect.TypeOf(v) == reflect.TypeOf(common.Address{}) {
```
kjhman21, line 308: You can use reflection, but I think this is more readable approach:

```go
if addresses, ok := v.(common.Address); ok {
  return true
}
addresses, ok := v.([]common.Address)
if !ok {
  return false
}
for _, address := range addresses {
  // check the duplication.
}
```

aidan-kwon (dismissed):

governance/default.go
```diff
@@ -521,18 +521,60 @@ func (g *Governance) ClearVotes(num uint64) {
...
+	k, ok := GovernanceKeyMap[gVote.Key]
+	if !ok {
+		logger.Warn("Unknown key was given", "key", k)
+		return nil, ErrUnknownKey
+	}
+
+	// checking type of gVote.Value
+	switch k {
+	case params.AddValidator, params.RemoveValidator:
+		if reflect.TypeOf(gVote.Value) != reflect.TypeOf([]interface{}{}) && reflect.TypeOf(gVote.Value) != reflect.TypeOf([]uint8{}) {
```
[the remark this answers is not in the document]
yoomee1313: @kjhman21 The inner type is `interface{}` when `[][]uint8` came in, and `interface{}` cannot be recognized as `[]uint8` directly. So, It's not possible to check `[][]uint8` directly.

governance/default.go
yoomee1313: Thanks, reflected at https://github.com/klaytn/klaytn/pull/1110/commits/70ad97142c82c7962c5150839e6fb074ae5091f3.

governance/default_test.go
yoomee1313: Of course. Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/70ad97142c82c7962c5150839e6fb074ae5091f3.

governance/default_test.go
yoomee1313: Thanks! It's really helping comment.. Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/70ad97142c82c7962c5150839e6fb074ae5091f3.

governance/default_test.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/70ad97142c82c7962c5150839e6fb074ae5091f3.

governance/handler.go
yoomee1313: I refactored whole `adjustValueType` at https://github.com/klaytn/klaytn/pull/1110/commits/70ad97142c82c7962c5150839e6fb074ae5091f3. Please take another look.

governance/handler.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/70ad97142c82c7962c5150839e6fb074ae5091f3.

governance/handler.go
yoomee1313: Thanks, reflected at https://github.com/klaytn/klaytn/pull/1110/commits/70ad97142c82c7962c5150839e6fb074ae5091f3.

aidan-kwon (dismissed):

governance/default.go
yoomee1313: @kjhman21 I changed the code to integrate type mismatch checking logic with parsing logic. Please take another look. https://github.com/klaytn/klaytn/pull/1110/commits/e28a478c428d7627141a5c9583d47b3dde921028

governance/handler.go
```diff
@@ -127,37 +156,62 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
+	if !ok {
+		return val
+	}
+	if GovernanceItems[k].checkValueType(common.Address{}) {
+		addresses := strings.Split(v, ",")
+		switch len(addresses) {
+		case 0:
+			return val
+		case 1:
+			str := strings.Trim(val.(string), " ")
```
kjhman21, line 185:
```suggestion
			str := strings.Trim(v, " ")
```

governance/default_test.go
kjhman21: @yoomee1313 Shouldn't this be failed due to the wrong length of addresses?

governance/handler.go
```diff
@@ -253,6 +306,22 @@ func checkBigInt(k string, v interface{}) bool {
...
 
 func checkAddress(k string, v interface{}) bool {
+	if _, ok := v.(common.Address); ok {
+		return true
+	}
+	if _, ok := v.([]common.Address); !ok {
+		return false
+	}
+
+	// there should not be identical addresses, if value contains multiple addresses
```
kjhman21, line 316:
```suggestion
	// there should not be duplicated addresses, if value contains multiple addresses
```

governance/handler.go
```diff
@@ -253,6 +306,22 @@ func checkBigInt(k string, v interface{}) bool {
 }
 
 func checkAddress(k string, v interface{}) bool {
```
kjhman21, line 308: Function name is very ambiguous.
something like checkAddressOrListOfUniqueAddresses ?

governance/handler.go
```diff
@@ -286,12 +355,28 @@ func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes
 				return valset, votes, tally
 			}
 		case params.AddValidator:
-			if !gov.checkVote(gVote.Value.(common.Address), true, valset) {
-				return valset, votes, tally
+			if reflect.TypeOf(gVote.Value) == addressT {
+				if !gov.checkVote(gVote.Value.(common.Address), true, valset) {
+					return valset, votes, tally
+				}
```
kjhman21, line 361:
```suggestion
			if addr, ok := gVote.Value.(common.Address); ok {
				if !gov.checkVote(addr, true, valset) {
					return valset, votes, tally
				}
```

governance/default_test.go
yoomee1313: @kjhman21 oh, sorry for the confuse.. i mean `vote api` didn't accept non-address string. And this PR changed it to accept non-address length . Than, I reverted it, and also changed parseVoteValue not to accept non-address value.

governance/handler.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/25b86fa38cffb37df13e62505554b020e7daf6a6. I have refactored some code, too. Please take another look.

governance/handler.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/25b86fa38cffb37df13e62505554b020e7daf6a6.

governance/handler.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/25b86fa38cffb37df13e62505554b020e7daf6a6.

governance/api.go
```diff
@@ -114,13 +109,13 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 }
 
 func (api *PublicGovernanceAPI) isRemovingSelf(val interface{}) bool {
```
kjhman21, line 111: If you always receive string here, what about changing of the parameter type to string instead of interface?

kjhman21: Of you need to check once again to avoid panic here.
```go
s, ok := val.(string)
```

governance/handler.go
```diff
@@ -32,37 +32,66 @@ import (
...
+
+func isBoolType(v interface{}) bool { return boolT == reflect.TypeOf(v) }
+
+func isAddressType(v interface{}) bool {
+	return addressSliceT == reflect.TypeOf(v) || addressT == reflect.TypeOf(v)
+}
+
+func compareEqual(v1 interface{}, v2 interface{}) bool {
+	return v1 == v2
+}
```
kjhman21, line 82: Hmm.. It seeks ok for now, but these kinds of functions can be used in various functions. What about moving theses functions into common package or similar?

kjhman21: and also, it can be done like this also:

```go
func isStringType(v interface{}) bool { _, ok := v.(string); return ok }
```

governance/handler.go
```diff
@@ -127,37 +156,62 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
-	if reqType == uint64T && reflect.TypeOf(val) == float64T {
-		x = uint64(val.(float64))
-		if float64(x.(uint64)) == val.(float64) {
+	if GovernanceItems[k].checkValueType(uint64(0)) {
+		v, ok := val.(float64)
+		if !ok {
+			return val
+		}
+		x = uint64(v)
+		if float64(x.(uint64)) == v {
```
kjhman21, line 168:
```suggestion
		if float64(x) == v {
```

This would be also okay because x is already the type of uint64.

governance/handler.go
```diff
@@ -127,37 +156,62 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
+		case 0:
+			return val
+		case 1:
+			str := strings.Trim(v, " ")
+			if common.IsHexAddress(str) {
+				x = common.HexToAddress(str)
+				return x
+			} else {
+				return val
+			}
```
kjhman21, line 191: Seems like it is a common pattern shown various area. What about making a function for this?

```go
func ConvertAddress(str string) (common.Address, error) {
  if common.IsHexAddress(str) {
    x = common.HexToAddress(str)
    return x
  } else {
    return val
  }
}

```

governance/api.go
yoomee1313: Yes, it always receives `string` because non-string is handled by the above `ValidateVote`. 
Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/4caf232a66b2d7eaaa5b94a02ef1fd127745dc6e.

governance/handler.go
yoomee1313: Yepp.. There was no need to implement like this. The `checkValueType` functions are moved to a `checkValue` function.
Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/4caf232a66b2d7eaaa5b94a02ef1fd127745dc6e.

governance/handler.go
yoomee1313: `x` is declared as interface{} type. So, It needs to be converted to `uint64`.

governance/handler.go
yoomee1313: Hmm.. I think the benefits of functionalization of this logic are unlikely to be significant. Looking at `case 1` and `case 2`, it is likely that many overlapping parts will occur again due to different logic.

governance/handler.go
kjhman21: okay. I will suggest this in another way.

governance/handler.go
```diff
@@ -127,40 +127,80 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
+		v, ok := val.(float64)
+		if !ok {
+			return val
+		}
+		x = uint64(v)
+		if float64(x.(uint64)) == v {
 			return x
 		}
 		return val
 	}
```
kjhman21, line 143:
```suggestion
		if float64(v.(uint64)) == v {
			return v
		}
		return val
	}
```

governance/handler.go
```diff
@@ -127,40 +127,80 @@ func (g *Governance) AddVote(key string, val interface{}) bool {
...
+			for _, str := range addresses {
+				str = strings.Trim(str, " ")
+				if common.IsHexAddress(str) {
+					nodeAddresses = append(nodeAddresses, common.HexToAddress(str))
+				} else {
+					return val
+				}
+			}
+			x = nodeAddresses
 			return x
```
kjhman21, line 174:
```suggestion
			return nodeAddresses
```

governance/handler.go
kjhman21: Ok then. Leave it as it is.

governance/handler.go
```diff
@@ -414,11 +493,22 @@ func (gov *Governance) addNewVote(valset istanbul.ValidatorSet, votes []Governan
 			switch GovernanceKeyMap[gVote.Key] {
 			case params.AddValidator:
 				//reward.GetStakingInfo()
-				valset.AddValidator(gVote.Value.(common.Address))
+				if reflect.TypeOf(gVote.Value) == addressT {
+					valset.AddValidator(gVote.Value.(common.Address))
```
kjhman21, line 497:
```suggestion
				if addr, ok := gVote.Value.(common.Address); ok {
					valset.AddValidator(addr)
```

governance/handler.go
```diff
@@ -414,11 +493,22 @@ func (gov *Governance) addNewVote(valset istanbul.ValidatorSet, votes []Governan
...
+					for _, address := range gVote.Value.([]common.Address) {
+						valset.AddValidator(address)
+					}
+				}
 			case params.RemoveValidator:
-				target := gVote.Value.(common.Address)
-				valset.RemoveValidator(target)
-				votes = gov.removeVotesFromRemovedNode(votes, target)
+				if reflect.TypeOf(gVote.Value) == addressT {
+					valset.RemoveValidator(gVote.Value.(common.Address))
```
kjhman21, line 505:
```suggestion
				if addr, ok := gVote.Value.(common.Address); ok {
					valset.AddValidator(addr)
```

governance/handler.go
yoomee1313: The suggested code is not working so, I refactored the code like below.
```
		if float64(uint64(v)) == v {
			return uint64(v)
		}
```

Also, after refactoring like this, `x` is no longer needed, so `x` is removed.

Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/fc1a4dd9144c6b78bef9ff326d804c47d7ea0b95. Please take another look.

governance/handler.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/fc1a4dd9144c6b78bef9ff326d804c47d7ea0b95.

governance/handler.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/fc1a4dd9144c6b78bef9ff326d804c47d7ea0b95.

governance/handler.go
yoomee1313: Reflected at https://github.com/klaytn/klaytn/pull/1110/commits/fc1a4dd9144c6b78bef9ff326d804c47d7ea0b95.

governance/handler.go
```diff
@@ -252,10 +286,38 @@ func checkBigInt(k string, v interface{}) bool {
...
+func isEqualValue(k int, v1 interface{}, v2 interface{}) bool {
+	if reflect.TypeOf(v1) != reflect.TypeOf(v2) {
+		return false
+	}
+
+	if GovernanceItems[k].t != addressT || reflect.TypeOf(v1) == addressT {
+		return v1 == v2
+	}
+
+	return reflect.DeepEqual(v1.([]common.Address), v2.([]common.Address))
```
aidan-kwon, line 318: I think the following code is more intuitive.

```suggestion
	if GovernanceItems[k].t == addressT && reflect.TypeOf(v1) != addressT {
		return reflect.DeepEqual(v1.([]common.Address), v2.([]common.Address))
	}

	return v1 == v2
```

governance/default.go
```diff
@@ -518,33 +518,64 @@ func (g *Governance) ClearVotes(num uint64) {
...
 	case params.GovernanceMode, params.MintingAmount, params.MinimumStake, params.Ratio:
-		val = string(gVote.Value.([]uint8))
+		v, ok := gVote.Value.([]uint8)
+		if !ok {
+			return nil, ErrValueTypeMismatch
+		}
+		val = string(v)
 	case params.GoverningNode, params.AddValidator, params.RemoveValidator:
-		val = common.BytesToAddress(gVote.Value.([]uint8))
+		if v, ok := gVote.Value.([]uint8); ok {
```
aidan-kwon, line 538: Does `params.GoverningNode` also allows arrays?

kjhman21 (approved):
LGTM. Thanks for changes.

governance/default.go
yoomee1313: @aidan-kwon `params.GoverningNode` does not allow arrays. Thanks for the advice. It is reflected at https://github.com/klaytn/klaytn/pull/1110/commits/81771918b0d94ab90c7f28fc8df92912dbe6280d

governance/handler.go
yoomee1313: Thanks, it is reflected at https://github.com/klaytn/klaytn/pull/1110/commits/81771918b0d94ab90c7f28fc8df92912dbe6280d

governance/handler.go
```diff
@@ -47,10 +47,10 @@ var (
 
 var GovernanceItems = map[int]check{
 	params.GovernanceMode:          {stringT, checkGovernanceMode, nil},
-	params.GoverningNode:           {addressT, checkAddress, nil},
+	params.GoverningNode:           {addressT, checkAddressOrListOfUniqueAddresses, nil},
```
aidan-kwon, line 50: `checkAddressOrListOfUniqueAddresses` should be `checkAddress` , right?

yoomee1313: Yes, you're right. I restored checkAddress. https://github.com/klaytn/klaytn/pull/1110/commits/7bfb3aedc3fea9ad8c0ff61c74ba98bf338b333f.

aidan-kwon (approved):

kjhman21 (approved):

jeongkyun-oh (approved):

hqjang-pepper (approved):

PR #1141, opened 2022-02-03
cmd: move unlock logic to the end of startNode to fix temporary bad block when node starts

### LINKED ISSUE #1103: A bad block occurs when downloader starts before the node's init job is done
**Describe the bug**
When a node starts, a bad block appears if the downloader starts before the whole node's init job is done.
Then, it doesn't reproduce a bad block any more after it downloads the identical block from the other peer.

**How to reproduce**
step1. make a node's init job heavy (ex. unlock 20 or more local accounts)
step2. start up a cypress or baobab en.

**Expected behavior**
No bad block appears.

**Environment (please complete the following information)**
 - Klaytn version: v1.7.3
 - go version: v1.17.5
 - ubuntu20.04

yoomee1313:
## Proposed changes
- It moves unlock logic to the end of `startNode`.
- It fixes #1103.
- It seems the node used uninitilialized istanbul chainConfig because sync starts before istanbul chainConfig is set.

closes https://github.com/klaytn/klaytn/issues/1103

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/cmd/utils/nodecmd/defaultcmd.go
+++ b/cmd/utils/nodecmd/defaultcmd.go
@@ -63,16 +63,6 @@ func startNode(ctx *cli.Context, stack *node.Node) {
 	// Start up the node itself
 	utils.StartNode(stack)
 
-	// Unlock any account specifically requested
-	ks := stack.AccountManager().Backends(keystore.KeyStoreType)[0].(*keystore.KeyStore)
-
-	passwords := utils.MakePasswordList(ctx)
-	unlocks := strings.Split(ctx.GlobalString(utils.UnlockedAccountFlag.Name), ",")
-	for i, account := range unlocks {
-		if trimmed := strings.TrimSpace(account); trimmed != "" {
-			UnlockAccount(ctx, ks, trimmed, i, passwords)
-		}
-	}
 	// Register wallet event handlers to open and auto-derive wallets
 	events := make(chan accounts.WalletEvent, 16)
 	stack.AccountManager().Subscribe(events)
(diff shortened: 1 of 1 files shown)


cmd/utils/nodecmd/defaultcmd.go
```diff
@@ -120,6 +110,17 @@ func startNode(ctx *cli.Context, stack *node.Node) {
 	} else {
 		startKlaytnAuxiliaryService(ctx, stack)
 	}
+
```
aidan-kwon, line 113: You mean that the above code should be finished earlier than other code being executed in another goroutine. So, you move the following code at the bottom of the function, right? Can you explain details about the disorder of which goroutines makes the bad block issue? 
Your change seems that it can mitigate the issue somewhat, but cannot resolve it completely.

aidan-kwon: I think this change is okay if we need lots of engineering to resolve the issue, but we need to figure out whether other race issues can occur because of the same reason.

yoomee1313: @aidan-kwon It seems that the node used uninitilialized istanbul chainConfig because sync starts before istanbul chainConfig is set.

The following functions are called to set istanbul chainConfig. startKlaytnAuxiliaryService -> cn.StartMining -> go s.miner.Start() -> self.worker.start() -> istanbul.Start -> sb.SetChain(chain)

At sb.SetChain(chain), `chainConfig` is set by setting `chain` to `sb.chain`. So, I changed the code to call `startKlaytnAuxiliaryService` earlier than the unlocking process.

aidan-kwon: I cannot get the explanation. As far as I know, the moved code is not affected by chainConfig. What is the code using uninitilialized istanbul chainConfig?

aidan-kwon (approved):

jeongkyun-oh (approved):

Issue #1133, opened 2022-01-27
Solidity version of test contracts need to be updated 


aidan-kwon:
Most contracts in this repository have solidity version 0.4.x or 0.5.x.
Those are too old and solc 0.4.24 is even outdated now, so we need to update them. 

This issue only cares contracts used in test cases.
Other contracts used in Cypress will be handled in another issue because it requires 3rd party audits and reviews.

**Sub tasks**
- [x] Make test cases independent on solidity compiler installation.
     https://github.com/klaytn/klaytn/pull/1149
- [x] Update solidity version of contracts in test cases to the recent version.
     https://github.com/klaytn/klaytn/pull/1189
  - [x] contracts/computationcost/opcodeBench.sol - keep it to 0.4.24
  - [x] contracts/feepayer/fee_payer_indirect_test.sol
  - [x] contracts/feepayer/fee_payer_test.sol
  - [x] contracts/internal_tx_contract/internal_tx_contract.sol - (included in PR#1149)
  - [x] contracts/storagetrie/StorageTrieStoreTest.sol
  - [x] node/cn/tracers/testdata/contracts/*.sol - solc not invoked in tests
- [x] Update Klaytn build_base image to use recent solidity compiler version.
      https://github.com/klaytn/klaytn/pull/1320

PR #1149, opened 2022-02-08
Add compiler.CompileSolidityOrLoad for contract testing

blukat29:
## Proposed changes

Add `compiler.CompileSolidityOrLoad()` function that falls back to a json file in case compiler is not available.

1. Parse solidity files for `pragma solidity`
2. Detect installed `solc` version
3. If `solc` exists and can compile the files, then compile.
4. Otherwise try to load `*.sol.json` file

- Some tests can now be modified to run in some environments that lacks solidity compiler.
- `CompilerSolidityOrLoad` is primarily for test purposes. Non-test codes like `abigen` will use the original `CompileSolidity`.
- `Contract.Info.CompilerOptions` is empty when compilation is loaded from json file. But it should be no problem, as `CompilerOptions` field is not used anywhere in klaytn codebase.
- Modify `blockchain_test.go` to depend on `CompileSolidityOrLoad` instead of hardcoded binary and abi.
  Other contract-related tests are to be modified in follow-up PRs (but can be changed if needed in this PR)

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- This PR is first part of https://github.com/klaytn/klaytn/issues/1133


blukat29:
Rebased to latest dev.

aidan-kwon (approved):

aidan-kwon:
Reviewers, PTAL

kjhman21 (approved):
LGTM. But in this case, we should have a solc in the local machine.

We can improve this in the two way:
1. Add a documentation how to install solc.
2. Use [solc docker image](https://hub.docker.com/r/ethereum/solc) so that any solidity version we want can be used without solc installation.

aidan-kwon:
> LGTM. But in this case, we should have a solc in the local machine.

@kjhman21  This change can remove dependcy of solc for some contracts by loading hard-coded bytecode. 
I will add solc installation guide in README [here](https://github.com/klaytn/klaytn/issues/1307).
I think using docker in Klaytn test code is too heavy and requires dependency.

common/compiler/helpers.go
```diff
@@ -21,9 +21,12 @@ import (
 	"bytes"
 	"io/ioutil"
 	"regexp"
+
+	semver "github.com/Masterminds/semver/v3"
 )
 
 var versionRegexp = regexp.MustCompile(`([0-9]+)\.([0-9]+)\.([0-9]+)`)
+var pragmaVersionRegexp = regexp.MustCompile(`(?m)^pragma solidity (.+)\;$`)
```
hyunsooda, line 29: This regex didn't consider whitespaces(ws) and failed if the input string contains multiple ws between words.
Consider change with this one:
```
regexp.MustCompile(`(?m)^(\s*)pragma(\s*) solidity(\s*) (.+)\;(\s*)$`)
```
Now this regex have interesting matched target is in the index 4. (i.e., line78 should be fixed too) (0 = input string, 1 = \s+, 2 = \s+, 3 = \s+ 4 = matched version). 

FYI, the regex is not the correct implementation if this regex is solely used since any string would be matched by `(.+)`. However, this incorrectly matched string would be filtered at `solcCanCompile()` function by semver package if the result of the regex is passed to this function before being used.

common/compiler/solidity_test.go
```diff
@@ -39,6 +41,24 @@ func skipWithoutSolc(t *testing.T) {
...
+	tests := []struct {
+		source   string
+		expected []string
+	}{
+		{"pragma solidity ^0.4.24;\n", []string{"^0.4.24"}},
+		{"pragma solidity 0.4.24;\n", []string{"0.4.24"}},
+		{"pragma solidity ^0.4.24;\n", []string{"^0.4.24"}},
+		{"pragma solidity 0.4.24;\npragma solidity 0.4.25;\n", []string{"0.4.24", "0.4.25"}},
+		{"pragma solidity 0.4.24;\n//pragma solidity 0.4.25;\n", []string{"0.4.24"}},
+		{"//pragma solidity 0.4.24;\npragma solidity 0.4.25;\n", []string{"0.4.25"}},
```
hyunsooda, line 54: Missing test cases may be added such as `pragma solidity >=0.6.0 <0.9.0;`

common/compiler/solidity_test.go
```diff
@@ -70,9 +90,51 @@ func TestSolidityCompiler(t *testing.T) {
...
+func TestSolcCanCompile(t *testing.T) {
+	solcVersion := "0.8.11"
+	tests := []struct {
+		version  string
+		expected bool
+	}{
+		{"^0.8.11", true},
+		{"^0.4.24", false},
+		{"^0.5.6", false},
+		{"0.5.6", false},
```
hyunsooda, line 110: Missing test cases may be added such as `pragma solidity >=0.6.0 <0.9.0;`

aidan-kwon:
Reviewers, please check this code again. 
I removed the solidity version parsing logic from the previous request. 
This change just compiles the target contract with the available `solc` and loads bytecode if it failed.

common/compiler/solidity_test.go
aidan-kwon: I removed the solidity version parsing logic from the previous request.
This latst change just compiles the target contract with the available solc and loads bytecode if it failed.

common/compiler/solidity_test.go
aidan-kwon: I removed the solidity version parsing logic from the previous request.
This latst change just compiles the target contract with the available solc and loads bytecode if it failed.

hyunsooda:
Looks more stable

hyunsooda (approved):

kjhman21 (approved):
LGTM

aidan-kwon (approved):

PR #1154, opened 2022-02-09
Updated legacy account type to Klaytn account type

jeongkyun-oh:
## Proposed changes

- Ethereum account type is removed and substitute them with KIaytn account type
- Ethereum account try to save storage, so it implemented `SlimAccount` type which substitute `emptyHash` and `emptyCodeHash` to `nil` value. In this PR, the `SlimAccount` concept is deleted since we have two different types of Account.
- Simply put, we do encoding/decoding `account.Account` type.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


snapshot/generate.go
```diff
@@ -615,31 +616,27 @@ func (dl *diskLayer) generate(stats *generatorStats) {
...
-		if err := rlp.DecodeBytes(val, &acc); err != nil {
+		serializer := account.NewAccountSerializer()
+		if err := rlp.DecodeBytes(val, serializer); err != nil {
 			logger.Crit("Invalid account encountered during snapshot creation", "err", err)
 		}
+		acc := serializer.GetAccount()
 		// If the account is not yet in-progress, write it out
 		if accMarker == nil || !bytes.Equal(accountHash[:], accMarker) {
 			dataLen := len(val) // Approximate size, saves us a round of RLP-encoding
+			// TODO-Klaytn-Snapshot consider slim account in order to save space for emptyRootHash, emptyCodeHash
```
kjhman21, line 627: @jeongkyun-oh Do we need something more to implement here?

snapshot/generate_test.go
```diff
@@ -118,6 +149,10 @@ func hashData(input []byte) common.Hash {
 
 // Tests that snapshot generation with existent flat state.
 func TestGenerateExistentState(t *testing.T) {
+	if true {
+		enableLog()
+	}
```
kjhman21, line 154: why do we need this condition?

snapshot/generate.go
jeongkyun-oh: @kjhman21 The original implementation modified their account model to reduce data size. So, we could do that also but it is optional so I leave a comment. It is fine as it is, though.

snapshot/generate_test.go
jeongkyun-oh: This is not necessary, I will remove this.

snapshot/generate.go
kjhman21: @jeongkyun-oh Can we do more? I think we do not have emptyRootHash or codeHash for EOAs.

snapshot/generate_test.go
```diff
@@ -118,6 +149,10 @@ func hashData(input []byte) common.Hash {
 
 // Tests that snapshot generation with existent flat state.
 func TestGenerateExistentState(t *testing.T) {
+	if false {
```
kjhman21, line 152: hmm??? false??
@jeongkyun-oh What about adding a comment here something like "replace it with true if you want to see logs"

snapshot/generate.go
jeongkyun-oh: @kjhman21 How about SmartContractAccount?

snapshot/generate.go
```diff
@@ -657,7 +654,14 @@ func (dl *diskLayer) generate(stats *generatorStats) {
 		}
 		// If the iterated account is the contract, create a further loop to
 		// verify or regenerate the contract storage.
-		if acc.Root == emptyRoot {
+		contractAcc, ok := acc.(*account.SmartContractAccount)
+		if !ok {
+			accMarker = nil
+			return nil
+		}
```
aidan-kwon, line 661: Don't we need to call `wipeKeyRange` for EOA?

kjhman21 (approved):

snapshot/generate.go
jeongkyun-oh: @aidan-kwon Can you check this again?

aidan-kwon (approved):

aidan-kwon:
@KimKyungup @ehnuje PTAL

snapshot/generate.go
```diff
@@ -657,7 +645,8 @@ func (dl *diskLayer) generate(stats *generatorStats) {
 		}
 		// If the iterated account is the contract, create a further loop to
 		// verify or regenerate the contract storage.
-		if acc.Root == emptyRoot {
+		contractAcc, ok := acc.(*account.SmartContractAccount)
+		if !ok {
 			// If the root is empty, we still need to ensure that any previous snapshot
```
yoomee1313, line 650: Is the `root empty` comment's location here? Shouldn't it be moved down?

snapshot/generate.go
```diff
@@ -615,31 +616,18 @@ func (dl *diskLayer) generate(stats *generatorStats) {
...
 		// If the account is not yet in-progress, write it out
 		if accMarker == nil || !bytes.Equal(accountHash[:], accMarker) {
 			dataLen := len(val) // Approximate size, saves us a round of RLP-encoding
 			if !write {
-				if bytes.Equal(acc.CodeHash, emptyCode[:]) {
-					dataLen -= 32
-				}
-				if acc.Root == emptyRoot {
-					dataLen -= 32
-				}
```
hqjang-pepper, line 637: Why did you remove this logic? Is it unnecessary in Klaytn?

yoomee1313 (approved):

snapshot/generate.go
jeongkyun-oh: @hqjang-pepper This is for approximate data size. So, Etheruem uses slim account in order to save the space. It reduces `dataLen` since `nil` is stored instead of `emptyRootHash` or `emptyCodeHash`.

PR #1155, opened 2022-02-10
Updated statedb for snapshot functionality

jeongkyun-oh:
## Proposed changes

- Updated statedb logic for snapshot functionality
  - added logic for direct access from snapshot for accounts and storages
- Added a edge test case for a reverted  account creation on the self-destructed address

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/state/state_object.go
+++ b/blockchain/state/state_object.go
@@ -232,13 +232,13 @@ func (self *stateObject) GetCommittedState(db Database, key common.Hash) common.
 	}
 	// If the snapshot is unavailable or reading from it fails, load from the database.
 	if self.db.snap == nil || err != nil {
-		if meter != nil {
-			// If we already spent time checking the snapshot, account for it
-			// and reset the readStart
-			*meter += time.Since(readStart)
-			readStart = time.Now()
-		}
 		if EnabledExpensive {
+			if meter != nil {
+				// If we already spent time checking the snapshot, account for it
+				// and reset the readStart
+				*meter += time.Since(readStart)
+				readStart = time.Now()
+			}
 			meter = &self.db.StorageReads
 		}
 		// Load from DB in case it is missing.
--- a/blockchain/state/state_object.go
+++ b/blockchain/state/state_object.go
@@ -247,7 +247,6 @@ func (self *stateObject) GetCommittedState(db Database, key common.Hash) common.
 			return common.Hash{}
 		}
 	}
-
 	if len(enc) > 0 {
 		_, content, _, err := rlp.Split(enc)
 		if err != nil {
--- a/blockchain/state/state_object.go
+++ b/blockchain/state/state_object.go
@@ -76,7 +76,7 @@ func (self Storage) Copy() Storage {
 // Finally, call CommitStorageTrie to write the modified storage trie into a database.
 type stateObject struct {
 	address  common.Address
-	addrHash common.Hash // hash of ethereum address of the account
+	addrHash common.Hash // hash of the address of the account
 	account  account.Account
 	db       *StateDB
 
--- a/blockchain/state/state_object.go
+++ b/blockchain/state/state_object.go
@@ -75,9 +75,10 @@ func (self Storage) Copy() Storage {
 // Account values can be accessed and modified through the object.
 // Finally, call CommitStorageTrie to write the modified storage trie into a database.
 type stateObject struct {
-	address common.Address
-	account account.Account
-	db      *StateDB
+	address  common.Address
+	addrHash common.Hash // hash of ethereum address of the account
+	account  account.Account
+	db       *StateDB
 
 	// DB error.
 	// State objects are used by the consensus core and VM which are
@@ -120,6 +121,7 @@ func newObject(db *StateDB, address common.Address, data account.Account) *state
 	return &stateObject{
 		db:            db,
 		address:       address,
+		addrHash:      crypto.Keccak256Hash(address[:]),
 		account:       data,
 		originStorage: make(Storage),
 		dirtyStorage:  make(Storage),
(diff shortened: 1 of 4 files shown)


blockchain/state/statedb.go
```diff
@@ -109,16 +120,25 @@ func New(root common.Hash, db Database) (*StateDB, error) {
...
+		snaps:                    nil,
 		stateObjects:             make(map[common.Address]*stateObject),
 		stateObjectsDirtyStorage: make(map[common.Address]struct{}),
 		stateObjectsDirty:        make(map[common.Address]struct{}),
 		logs:                     make(map[common.Hash][]*types.Log),
 		preimages:                make(map[common.Hash][]byte),
 		journal:                  newJournal(),
-	}, nil
+	}
+	if sdb.snaps != nil {
```
kjhman21, line 134: This will not be executed because snaps is always nil. Is there any more implementation done for this? Then, please add a TODO comment.

jeongkyun-oh: Right. `snaps` will be inserted with a parameter. I will add a comment.

blockchain/state/state_object.go
```diff
@@ -197,15 +199,55 @@ func (self *stateObject) GetCommittedState(db Database, key common.Hash) common.
...
+		//      have been handles via pendingStorage above.
+		//   2) we don't have new values, and can deliver empty response back
+		if _, destructed := self.db.snapDestructs[self.addrHash]; destructed {
+			return common.Hash{}
+		}
+		enc, err = self.db.snap.Storage(self.addrHash, crypto.Keccak256Hash(key.Bytes()))
+	}
+	// If the snapshot is unavailable or reading from it fails, load from the database.
+	if self.db.snap == nil || err != nil {
+		if meter != nil {
```
kjhman21, line 235: It seems that `meter` is only set when EnabledExpensive is true, right?
Then it would be better to include this in the next conditional statement.

blockchain/state/state_object.go
```diff
@@ -197,15 +199,55 @@ func (self *stateObject) GetCommittedState(db Database, key common.Hash) common.
...
+		if EnabledExpensive {
+			meter = &self.db.StorageReads
+		}
+		// Load from DB in case it is missing.
+		if enc, err = self.getStorageTrie(db).TryGet(key[:]); err != nil {
+			self.setError(err)
+			return common.Hash{}
+		}
 	}
+
```
kjhman21, line 250: Is this needed?

blockchain/state/statedb.go
```diff
@@ -541,34 +572,62 @@ func (self *StateDB) deleteStateObject(stateObject *stateObject) {
...
+		acc account.Account
+		err error
+	)
+	if self.snap != nil {
+		if EnabledExpensive {
+			defer func(start time.Time) { self.SnapshotAccountReads += time.Since(start) }(time.Now())
+		}
+		if acc, err = self.snap.Account(crypto.Keccak256Hash(addr.Bytes())); err == nil {
+			if acc == nil {
+				return nil
```
kjhman21, line 605: It seems that snap.Account() returns nil if the account is not found.
Then don't we need to check the state trie?

blockchain/state/statedb.go
```diff
@@ -603,7 +662,15 @@ func (self *StateDB) GetOrNewSmartContract(addr common.Address) *stateObject {
 // createObject creates a new state object. If there is an existing account with
 // the given address, it is overwritten and returned as the second return value.
 func (self *StateDB) createObject(addr common.Address) (newobj, prev *stateObject) {
-	prev = self.getStateObject(addr)
+	prev = self.getDeletedStateObject(addr) // Note, prev might have been deleted, we need that!
```
aidan-kwon, line 665: It seems that this code changes the semantics of the previous logic. Is it okay?
In the previous code, `prev` becomes `nil` when it was deleted. And, the journal will append `createObjectChange` in the following logic.
With your change, `prev` becomes the old `obj` when it was deleted before. And, the journal will append `resetObjectChange` in the follwoing logic.

blockchain/state/state_object.go
jeongkyun-oh: I agreed. Updated.

blockchain/state/state_object.go
jeongkyun-oh: Removed. :)

blockchain/state/statedb.go
jeongkyun-oh: Should we? A snapshot contains all leaf nodes only based on the specific root hash. 
But, if an account is not found with an error, then it will try to get from state trie. Can this be the answer your question?

kjhman21: okay. got it.

kjhman21 (dismissed):
LGTM

blockchain/state/state_object.go
```diff
@@ -197,14 +199,53 @@ func (self *stateObject) GetCommittedState(db Database, key common.Hash) common.
...
+	if self.db.snap == nil || err != nil {
+		if EnabledExpensive {
+			if meter != nil {
+				// If we already spent time checking the snapshot, account for it
+				// and reset the readStart
+				*meter += time.Since(readStart)
+				readStart = time.Now()
+			}
+			meter = &self.db.StorageReads
+		}
```
yoomee1313, line 243: Ethereum code is like below.
```
		if meter != nil {
			// If we already spent time checking the snapshot, account for it
			// and reset the readStart
			*meter += time.Since(readStart)
			readStart = time.Now()
		}
		if metrics.EnabledExpensive {
			meter = &s.db.StorageReads
		}
```

Can I ask why you change like this?

blockchain/state/state_object.go
```diff
@@ -287,13 +331,25 @@ func (self *stateObject) updateStorageTrie(db Database) Trie {
...
 			self.setError(tr.TryDelete(key[:]))
-			continue
+		} else {
+			// Encoding []byte cannot fail, ok to ignore the error.
+			v, _ = rlp.EncodeToBytes(bytes.TrimLeft(value[:], "\x00"))
+			self.setError(tr.TryUpdate(key[:], v))
+		}
+		// If state snapshotting is active, cache the data til commit
+		if self.db.snap != nil {
+			if storage == nil {
```
hqjang-pepper, line 344: `storage` is always nil because it's local vaiable. Isn't it?

blockchain/state/statedb.go
```diff
@@ -541,34 +572,62 @@ func (self *StateDB) deleteStateObject(stateObject *stateObject) {
...
-// Retrieve a state object given by the address. Returns nil if not found.
+// getStateObject retrieves a state object given by the address, returning nil if
+// the object is not found or was deleted in this execution context. If you need
+// to differentiate between non-existent/just-deleted, use getDeletedStateObject.
 func (self *StateDB) getStateObject(addr common.Address) *stateObject {
+	if obj := self.getDeletedStateObject(addr); obj != nil && !obj.deleted {
+		return obj
+	}
+	return nil
+}
```
hqjang-pepper, line 583: Is `obj.deleted` atomic value or if statement is thread-safe? 
Then we can surely believe about object state in execution context.

blockchain/state/state_object.go
```diff
@@ -75,9 +75,10 @@ func (self Storage) Copy() Storage {
 // Account values can be accessed and modified through the object.
 // Finally, call CommitStorageTrie to write the modified storage trie into a database.
 type stateObject struct {
-	address common.Address
-	account account.Account
-	db      *StateDB
+	address  common.Address
+	addrHash common.Hash // hash of ethereum address of the account
```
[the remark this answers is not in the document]
jeongkyun-oh: Updated. Thanks.

blockchain/state/state_object.go
jeongkyun-oh: I have updated since I agreed with this comment https://github.com/klaytn/klaytn/pull/1155#discussion_r804370332.

blockchain/state/state_object.go
jeongkyun-oh: Not always `nil`. Please take a look line 346 and 347.

blockchain/state/statedb.go
jeongkyun-oh: Hmm.. it is not, but is it necessary? Do you have any concern about this?

kjhman21 (approved):

aidan-kwon (approved):

yoomee1313 (approved):

hqjang-pepper (approved):

blockchain/state/statedb.go
hqjang-pepper: I think it's ok because the DB guarantees concurrency.

Issue #1226, opened 2022-03-08
Should we add KIP7, KIP17 contracts in klaytn/contracts repository?


hqjang-pepper:
Currently, FT and NFT interface and token contracts(KIP7,KIP17) are only in [klaytn-contracts](https://github.com/klaytn/klaytn-contracts) repository.
ServiceChain cell is supporting the service chain code to support the kip token standard.
It would be convenient in many ways if the kip contract code is in the Klaytn repository.

PR #249, opened 2019-09-25
SC: remove uri in requetValueTransferEvent and add the routine to get uri #201

KimKyungup:
## Proposed changes

This PR is derived by merged [PR](https://github.com/klaytn/klaytn/pull/201) at master branch.

```
This PR removed uri in requetValueTransferEvent and add the routine to get uri before handling value transfer.

This PR can reduce the gas used of request value transfer transactions.
```

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1239, opened 2022-03-14
[SC] Added URI property in the requestValueTransfer event

hyunsooda:
## Proposed changes

URI property is retrieved when the ERC721 token is transferred via `requestERC721Transfer()`. However, [the URI property is always empty unless the bridge contract counterpart does not own the corresponding token ID](https://github.com/klaytn/klaytn/blob/dev/node/sc/bridge_manager.go#L327). If the bridge contract on the other side already owns the corresponding token ID, the code works as expected, while if not, the transfer is successfully done with an empty URI. In this PR, we simply introduce the URI property and it is directly used to retrieve URI value.

I simply added `uri` property, but I have no idea which option is better. The first option is just adding `uri` property in the `requestValuTransfer` event. Another option is splitting two structures for token transfer (ERC20 and KIP7) and NFT transfer (ERC721 and KIP17). For example:
```solidity
struct TokenTransfer {
    uint256 amount
    unit256 fee
}
struct NFTTransfer {
    uint256 tokenId
    string uri
}
```

TODO:
Currently, [no KIP7 and KIP17 contract code is placed in the official Klaytn repository](https://github.com/klaytn/klaytn/issues/1226). Still, the value transfer with two-step from KIP7 and KIP17 contract binaries can be executed via ERC20 and ERC721 interface contracts. The value transfer with one-step (`requestValueTransfer()`) is not executed with the compatible interface since there's no corresponding service chain contract. [Implementation is already done](https://github.com/hyunsooda/klaytn/commits/SC-support-KIP7-KIP17), but nothing is decided whether supports or not. 

cc. @2dvorak 

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


kjhman21:
> However, [the URI property is always empty unless the bridge contract counterpart does not own the corresponding token ID](https://github.com/klaytn/klaytn/blob/dev/node/sc/bridge_manager.go#L327).

Let's make this clear first. Is the ownership is important? I think the important thing is that the NFT is burnt before getting its URI.

hyunsooda:
> > However, [the URI property is always empty unless the bridge contract counterpart does not own the corresponding token ID](https://github.com/klaytn/klaytn/blob/dev/node/sc/bridge_manager.go#L327).
> 
> Let's make this clear first. Is the ownership is important? I think the important thing is that the NFT is burnt before getting its URI.

Hi @kjhman21,

Token burning is triggered in the counterpart chain as far as I know. Suppose that chain A requested value transfer (sends a NFT token) to chain B. The bridge contract on chain A burns that token and the corresponding event is emitted. Chain B handles the corresponding event. If chain B already has the token ID with the corresponding URI, that's not a problem. However, chain A does not know whether chain B owns the same URI with the corresponding token ID. Therefore, this PR prevents this wrong expectation and directly insert URI value from the chain that sent the request.

contracts/bridge/BridgeTransferERC721.sol
```diff
@@ -84,6 +84,11 @@ contract BridgeTransferERC721 is BridgeTokens, IERC721BridgeReceiver, BridgeTran
     {
         require(isRunning, "stopped bridge");
 
+        (bool success, bytes memory uri) = _tokenAddress.call(abi.encodePacked(ERC721Metadata(_tokenAddress).tokenURI.selector, abi.encode(_tokenId)));
+        if (success == false) {
+            uri = "";
+        }
+
```
KimKyungup, line 91: This code is newly added unlike original code https://github.com/klaytn/klaytn/pull/249/files?diff=split&w=0.
Also this false exception case can be tested in `TestBridgeManagerERC721_notSupportURI`.

It's good change. Thanks.

KimKyungup (approved):

kjhman21 (approved):
LGTM.

hyunsooda:
A merge for these changes should be delayed until #1248 is merged due to the compatibility issue.

contracts/kip13/InterfaceIdentifier.go
```diff
@@ -4,34 +4,37 @@
...
-func DeployInterfaceIdentifier(auth *bind.TransactOpts, backend bind.ContractBackend) (common.Address, *types.Transaction, *InterfaceIdentifier, error) {
-	parsed, err := abi.JSON(strings.NewReader(InterfaceIdentifierABI))
-	if err != nil {
-		return common.Address{}, nil, nil, err
-	}
-	address, tx, contract, err := bind.DeployContract(auth, parsed, common.FromHex(InterfaceIdentifierBin), backend)
-	if err != nil {
-		return common.Address{}, nil, nil, err
-	}
-	return address, tx, &InterfaceIdentifier{InterfaceIdentifierCaller: InterfaceIdentifierCaller{contract: contract}, InterfaceIdentifierTransactor: InterfaceIdentifierTransactor{contract: contract}, InterfaceIdentifierFilterer: InterfaceIdentifierFilterer{contract: contract}}, nil
```
hqjang-pepper, line 34: Can you explain why did you removed this function?

hyunsooda: This file is generated by `abigen`. There's no contract implementation in `contracts/kip13`. The previous commit seemed to run `abigen` with another contract file that has identifier contract implementation. Currently, there's only an interface in the `contracts/kip13` dir.

PR #1189, opened 2022-02-18
Upgrade test solidity files to v0.8

blukat29:
## Proposed changes

- This is follow-up of https://github.com/klaytn/klaytn/pull/1149.
- Upgrade solidity files that are exclusively used for testing to version 0.8.11.
	- fee_payer_test.sol
	- fee_payer_indirect_test.sol
	- StorageTrieStoreTest.sol
- Include precompiled binary and abi of contract that cannot be upgraded
    - opcodeBench.sol: some opcodes (dup1, 
- Fix test/benchmark cases related to above files, so that they work even when compiler is not available.

## Types of changes

Please put an x in the boxes related to your change.

- [x] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- This PR is second part of https://github.com/klaytn/klaytn/issues/1133.

PR #1248, opened 2022-03-18
[SC] Version comparison added 

hyunsooda:
## Proposed changes

The `kscn` node's behavior depends on the bridge binding handler generated by several bridge contracts. For example, the [bug](https://github.com/klaytn/klaytn/pull/1239) in the logic of value transfer request is not easy without a change of event structure, in turn, the outdated `kscn` node's value transfer request would not be correctly handled by the updated `kscn` nodes. 

This PR adds three types of version comparison.
1. Protocol version of SC
2. Bridge contract version

Both are compared at the time of handshake and at the time of registration of the deployed contract, respectively.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


node/sc/api_bridge.go
```diff
@@ -253,6 +254,21 @@ func (sb *SubBridgeAPI) GetAnchoring() bool {
 	return sb.subBridge.GetAnchoringTx()
 }
 
+func checkBridgeContractVersion(cBridge, pBridge *bridge.Bridge) bool {
+	errStr := "Bridge contract version mistmatched"
```
kjhman21, line 258: Can we reuse the global variable `ErrBridgeContractVersionMismatch` here instead of redefining the same string?

node/sc/api_bridge.go
```diff
@@ -253,6 +254,21 @@ func (sb *SubBridgeAPI) GetAnchoring() bool {
...
+	cv, cvErr := cBridge.VERSION(nil)
```
kjhman21, line 259: @hyunsooda What is the return value if VERSION is not defined in the contract?

node/sc/bridgepeer.go
```diff
@@ -48,15 +48,16 @@ const (
 // BridgePeerInfo represents a short summary of the Klaytn Bridge sub-protocol metadata known
 // about a connected peer.
 type BridgePeerInfo struct {
-	Version int    `json:"version"` // Klaytn Bridge protocol version negotiated
-	Head    string `json:"head"`    // SHA3 hash of the peer's best owned block
+	ProtocolVersion string `json:"protocolversion"` // Klaytn Bridge protocol version negotiated
```
kjhman21, line 51: I am not sure it is okay to change the "version" to "protocolversion". Did you check this is working well with previous version of klaytn??

hyunsooda: Why do you think this renaming causes a problem? For example, version v1.7.3 uses `Version` variable as version representation, while the node v1.8.0 uses `ProtocolVersion` variable. Could you give more specific context if I misunderstood your concerns?

node/sc/api_bridge.go
hyunsooda: Thank you. Good catch. 

I think we need to establish a semi-standard for the contract's version query. As developers would customize bridge contracts in the future, a standard function that returns the bridge contract's version seems needed so that the variable's name can be renamed anything as long as they implement that function. Please add how do you feel.

node/sc/api_bridge.go
hyunsooda: Refactored. PTAL.

node/sc/bridgepeer.go
kjhman21: What if a klayn-v1.7.3 node is trying to connect a klaytn-v1.8.0 node? Is this okay? For Baobab and Cypress node, it should not be the problem.

node/sc/api_bridge.go
kjhman21: Yeah, I agree with you to some extent. We have [KIP-13](https://kips.klaytn.com/KIPs/kip-13) to query interfaces in a standard way. It does not exactly fit to this, though.

By the way, what I wanted to say is that whether this works well with previously deployed bridge contracts without this function.

contracts/bridge/BridgeTestAnotherVersion.sol
```diff
@@ -0,0 +1,24 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity 0.5.6;
+
+contract BridgeAnotherVersion {
```
kjhman21, line 19: We need to have a better name than this. ;)

node/sc/api_bridge.go
hyunsooda: Bridge registration is failed in this PR If the variable `VERSION` does not exist in the bridge contract.

node/sc/bridgepeer.go
hyunsooda: Suppose that `kscn` node uses v1.7.3 in the SC network and `ken` with this PR version in the baobab or cypress network. Each node in its own network cannot establish a P2P connection since their node versions are different from each other. This PR was designed to guarantee to have the same internal bridge logic to avoid surprise malfunctioning due to the different implementation. 

If the compatibility between old and new bridge implementations is necessary, I think two options can be considered.
1. The node must be updated with the latest version. (Current PR's one)
2. The node must be updated with a minimum compatible node version.

If your concern is neither first nor second, both protocol version and node version check should be reverted again.

contracts/bridge/BridgeTestAnotherVersion.sol
hyunsooda: Okay, then how about `TestBridgeContractVersionDiff`? I would appreciate it if I can get your naming consideration. This is to avoid committing multiple files again due to its renaming.

kjhman21:
@ElliotHan @aidan-kwon Please also take a look.

node/sc/mainbridge.go
```diff
@@ -347,17 +354,23 @@ func (mb *MainBridge) Start(srvr p2p.Server) error {
 	return nil
 }
 
-func (mb *MainBridge) newPeer(pv int, p *p2p.Peer, rw p2p.MsgReadWriter) BridgePeer {
-	return newBridgePeer(pv, p, newMeteredMsgWriter(rw))
+func (mb *MainBridge) newPeer(pv string, p *p2p.Peer, rw p2p.MsgReadWriter) BridgePeer {
+	return newBridgePeer(pv, params.Version, p, newMeteredMsgWriter(rw))
```
kjhman21, line 358: Why do we add params.Version here?

contracts/bridge/BridgeTestAnotherVersion.sol
kjhman21: What about `BridgeVersionController`? We can later use this also in the real bridge contract.

node/sc/api_bridge.go
kjhman21: I thought that we don't have VERSION variable for the current bridge contract implementation. It seems that I was wrong, and this version check is good to go. Thanks.

contracts/bridge/BridgeTestAnotherVersion.sol
kjhman21: BridgeVersionController is a class that has a variable VERSION and a getter function of this.

The bridge contract inherits this class, and this variable is initialized in the Bridge contract constructor.

kjhman21:
@hyunsooda By the way, what is the reason we need to have "node version"?

hyunsooda:
@kjhman21 The variable `NodeVersion` represents the version of binary such as `cn` `pn`, `en`, `scn`, `spn`, `sen`.  I thought this was required to avoid incompatible behavior between two nodes (parent/child). For example, a node with v1.7.3 has A behavior on the SC functionalities, while a node B with v1.8.0 has incompatible functionalities with v1.7.3. Let me give you a concrete example. If a new change is created in the bridge contract, then the corresponding bridge handler should be updated too. However, some nodes with previous versions don't know this change and can not be aware of newly changed request handling. This PR's node accepts the only same version of a newbie, but minimum boundary for compatible versions seems necessary as [I mentioned earlier](https://github.com/klaytn/klaytn/pull/1248#discussion_r830700637) .

contracts/bridge/BridgeTestAnotherVersion.sol
hyunsooda: Okay, then do you want me to update origin contracts too so that they inherit this new change? Or, just only do rename for this function for now?

kjhman21: For this PR, please do not update the original contract. Just build preparation. Then we can make another PR that update the original contract. Sounds good to you?

hyunsooda: Revised.

contracts/bridge/BridgeTestAnotherVersion.sol
```diff
@@ -0,0 +1,24 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity 0.5.6;
+
+import "./BridgeVersionController.sol";
+
+contract BridgeAnotherVersion is BridgeVersionController {
+    constructor(uint64 version) BridgeVersionController(version)  public {
```
kjhman21, line 22:
```suggestion
    constructor() BridgeVersionController(2)  public {
```

I think this value is set by the contract it self.

hyunsooda: Thank you. Changed.

hyunsooda:
The `NodeVersion`, which stood for node's version comparison, was removed since the `ProtocolVersion` is the same thing. According to this change, the draft of PR was revised too.

node/sc/bridgepeer.go
kjhman21: @hyunsooda Do we need to change the name of this?

hyunsooda: Restored.

contracts/bridge/BridgeTestAnotherVersion.sol
```diff
@@ -0,0 +1,26 @@
...
+//
+// The klaytn library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity 0.5.6;
```
kjhman21, line 17: now, can we use later version of solidity?

kjhman21: oh, it might require more changes to other contracts. Let's do this later.

aidan-kwon: We are planning to update the solidity version in the build base docker image that is used by Circl ci. 
Can you update this version to 0.8.x  or make json file with compiled data?
You better check https://github.com/klaytn/klaytn/pull/1189/files first to understand my request.

hyunsooda: I see. I was working on implementing the bridge interface checker and KIP contracts for value transfer on SC. Let me go this up first. If you have another consideration for the contract, please let me know early.

node/sc/bridgepeer.go
```diff
@@ -48,15 +48,15 @@ const (
 // BridgePeerInfo represents a short summary of the Klaytn Bridge sub-protocol metadata known
 // about a connected peer.
 type BridgePeerInfo struct {
-	Version int    `json:"version"` // Klaytn Bridge protocol version negotiated
```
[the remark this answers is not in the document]
kjhman21: @hyunsooda What makes you require '1.x'? Isn't '2' enough?

hqjang-pepper:
LGTM if it's clear to use protocol version with string ,as @kjhman21 said.

node/sc/bridgepeer.go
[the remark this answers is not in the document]
hyunsooda: I just followed the convention of the Klaytn node's version representation. Why do you think the version on SC is okay with int type? If we keep it that way, the version semantic may cause confusion to developers.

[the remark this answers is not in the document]
kjhman21: I just wanted not to change if it is really necessary.

contracts/bridge/BridgeTestAnotherVersion.sol
hyunsooda: Once the reconstruction of the official contract library for Klaytn is done, I'll import them.

node/sc/bridgepeer.go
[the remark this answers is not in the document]
hyunsooda: Changed again. PTAL.

FYI, the current bridge contract's version number comparison is forced to have the same bridge contract type. The version comparison is not available for customizable bridge contracts, even if their contract version is the same. Currently, the bridge contract type is hardcoded with the provided bridge contract in the Klaytn team. The user's custom bridge contract type is unknown in our SC implementation. In order to port this compatibility, we need an interface structure that wraps a standard bridge contract interface. I'll add it todo-list.

hyunsooda:
Closed.

1. The comparison of protocol version is not necessary because the p2p module checks protocol versions and discard a peer that has an unmatched protocol version number.
2. The contract version comparison with the corresponding features would be newly developed on the contracts compatible ^v0.8.0

PR #1277, opened 2022-03-25
[SC] Alias-dedicated APIs in external bridge APIs

hyunsooda:
## Proposed changes

There has been a lack of bridge management in ServiceChain (SC). Some of the registered bridges have been internally managed with raw addresses, which may cause inconvenient usage to end-developers. This PR introduces a new bridge pairs management by alias access. In this alias access manner, the user now does not have to keep individually the raw bridge addresses they created; just keep the alias instead of the raw address. Note that these changes are compatible with old versions since they add new alias-dedicated APIs and keep the origin still.

The newly added external APIs have `ByAlias` suffix:
- subscribeBridgeByAlias
- unsubscribeBridgeByAlias
- registerBridgeByAlias
- deregisterBridgeByAlias
- registerTokenByAlias
- deregisterTokenByAlias
- getBridgePairByAlias
- changeBridgeAlias

The behavior of these new APIs is exactly the same as existing APIs but takes bridge an alias parameter instead of raw addresses. 

**----------------------------------------------------------------------- Editted -----------------------------------------------------------------------** 
- Removed the exported APIs that have `ByAlias` suffcix
- The existing APIs is now aware of both formats of raw address and alias.
- A few example:
```
subbridge_registerBridge(0XCHILD_BRIDGE, 0XPARENT_BRIDGE, ALIAS) 
or 
subbridge_registerBridge(0XCHILD_BRIDGE, 0XPARENT_BRIDGE)

subbridge_subscribeBridge(ALIAS)
or
subbridge_subscribeBridge(0XCHILD_BRIDGE, 0XPARENT_BRIDGE)

subbridge_registerToken(ALIAS, 0XCHILD_TOKEN, 0XPARENT_TOKEN)
or
subbridge_registerToken(0XCHILD_BRIDGE, 0XPARENT_BRIDGE, 0XCHILD_TOKEN, 0XPARENT_TOKEN)
``` 

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


node/sc/bridge_manager.go
```diff
@@ -432,20 +433,22 @@ func (bi *BridgeInfo) GetCurrentBlockNumber() (uint64, error) {
 // DecodeRLP decodes the Klaytn
 func (b *BridgeJournal) DecodeRLP(s *rlp.Stream) error {
 	var elem struct {
+		Alias         string
 		LocalAddress  common.Address
 		RemoteAddress common.Address
 		Paired        bool
 	}
 	if err := s.Decode(&elem); err != nil {
```
KimKyungup, line 441: Can this logic decode the legacy journal (without Alias data)?

KimKyungup, line 441: Can this logic decode the legacy journal (without Alias data)?

hyunsooda: Closed duplicated thread.

hyunsooda: Thanks for pointing out what I couldn't consider. I uploaded a new commit that resolves the compatibility issue. The current one can load both legacy journal and aliased-journal. 

Say this PR is a v0.0.3. The node with v0.0.3 can read legacy and aliased journals. However, some nodes with v0.0.1 and v0.0.2 can't load all the journals created from v0.0.3 nodes. Of course, we can make legacy node load aliased-journal, but it leads to the nodes with higher version enforce generating two journal files at every insertion for legacy structure's journal and aliased-journal. So, I didn't do that because I'm not sure it's essential or not. I hope to hear your opinion.

node/sc/bridge_manager.go
```diff
@@ -57,13 +57,16 @@ const (
...
+	ErrInvalidTokenPair        = errors.New("invalid token pair")
+	ErrNoBridgeInfo            = errors.New("bridge information does not exist")
+	ErrDuplicatedBridgeInfo    = errors.New("bridge information is duplicated")
+	ErrDuplicatedToken         = errors.New("token is duplicated")
+	ErrNoRecovery              = errors.New("recovery does not exist")
+	ErrAlreadySubscribed       = errors.New("already subscribed")
+	ErrBridgeRestore           = errors.New("restoring bridges is failed")
+	ErrBridgeAliasFormatDecode = errors.New("restoring bridges is failed")
+
+	//IsLegacyBridgeJournal = false
```
2dvorak, line 69:
```suggestion
```
Is this line unintended?

hyunsooda: I forgot to remove it. Thank you.

node/sc/bridge_manager.go
```diff
@@ -431,21 +435,40 @@ func (bi *BridgeInfo) GetCurrentBlockNumber() (uint64, error) {
...
-	if err := s.Decode(&elem); err != nil {
-		return err
+	var BridgeAddrInfo struct {
+		BridgeAlias   string
+		LocalAddress  common.Address
+		RemoteAddress common.Address
+		Paired        bool
+	}
+	if !b.isLegacyBridgeJournal {
+		if err := s.Decode(&BridgeAddrInfo); err != nil {
```
kjhman21, line 450: Don't we need to decode once again with legacy one?

hyunsooda: Do you consider a circumstance for lower compatibility so that a node with a lower version can read the aliased journal? If so, please check [here](https://github.com/klaytn/klaytn/pull/1277#discussion_r840261142) and add an opinion again here.

kjhman21: Hmm.. Okay, but I think it is quite complicated. Why don't we try to decode once again here? instead of using isLegacyBridgeJournal? like this?
https://github.com/klaytn/klaytn/blob/dev/blockchain/types/tx_internal_data_serializer.go#L67-L77

hyunsooda: The code you give seems to decode type first and then data. Suppose we are going to have this way. In that case, we need to reconstruct the bridge journal struct by decomposing the data structure (`LocalAddress`, `RemoteAddress`, and `Paired`) with a type that indicates bridge or legacy. It seems not bad in my view. Please add how do you think.

kjhman21: No, it's not. It just decode once again in the DecodeRLP() if the decoding fails.

hyunsooda: Let me ask you how to do that without reconstructing? I understood as the decode function reads and moves the internal pointer. Suppose that we first read something as a unit of struct and failed due to an incompatible legacy struct. At this point, if we try to read with another structure again, the next struct is decoded and the first one was already gone away. That's why I did reset and read again with another struct. The pattern that decodes the type of struct and decodes the actual interesting data again looks good to me, as I said above.

hyunsooda: I retook a look at the decode failure logic. The code you provided works well even if the decoding failure exists. The reason is that the target failed type is scalar type (`uint16` = `TxType`). In our case, however, the failed type is a struct, not a scalar. The [struct decoder](https://github.com/klaytn/klaytn/blob/dev/rlp/decode.go#L432) decodes one by one with an iteration of all the fields. The problem is that the struct decoder successfully decodes some fields but not some. For example, there are two structures `struct A { a,b int}, struct B { a,b,c int} `. Say a target encoded structure is A and we try to decode with struct B. The properties `a` and `b` are successfully decoded but, not c and get an error. Thus, the next read always fails since the read pointer does not point to the proper starting position of the expected encoded structure.

kjhman21: @hyunsooda hmm but it seems like the first field of `BridgeAddrInfo` is a string while the first field of `LegacyBridgeAddrInfo` is a common.address (byte[20]). Then I think when it tries to decode the first field, it should fail. What do you think?

hyunsooda: The first failure output is correct with your concern. From all the following failure outputs, the error `expected input list for struct { ... }` would continuously pop up. The struct decoder calls the` List()` function that does actually read, meaning that the next read is not a failed type again. 

Let me show you two concrete examples. The first example is a primitive type passing to the `Decode()` function. The second one is when a different type of struct is passed.

1) Failed to decode struct value with a primitive type, but never do read anything actually. Next time we call `Decode()` again with a proper struct type and it succeeds. 

2) Failed to decode struct value with a primitive type and have read that size of the struct. Next time we call `Decode()` again with a proper struct type and fail due to the incorrect read pointer (i.e., malformed read pointer).   

If you have a doubt still, I recommend you to test and debug these two cases above.

kjhman21: Got it. Understood. I think we can do it here like what you did [here](https://github.com/klaytn/klaytn/pull/1277/files#diff-5354626cdfa967dd6b4dbe6fd55e43f5bc198018bb25bb38fc6d21b18fab9f2eR95-R104).

Is there any differences?

kjhman21: @hyunsooda Okay. I got it. Let's go then what it is.

aidan-kwon (commented):
Another way of implementing bridge alias feature in APIs is extending input parameter types of existing APIs to handle both `common.Address` and `string`. Then, users can call bridge with either address or alias with one restriction, alias cannot start with `0x`.

hyunsooda:
@aidan-kwon, Thanks for the suggestion. 

How do you guys (@2dvorak , @kjhman21, @hqjang-pepper, @henry-will, @nohkwak) think about the suggested scheme? In my view, Aidan's one can reduce an additionally exported function list with an implicit rule of parameters. The current one does not have implicit rules and has apparent API usage, but not a small amount of additionally exported function lists.

2dvorak:
> @aidan-kwon, Thanks for the suggestion.
> 
> How do you guys (@2dvorak , @kjhman21, @hqjang-pepper, @henry-will, @nohkwak) think about the suggested scheme? In my view, Aidan's one can reduce an additionally exported function list with an implicit rule of parameters. The current one does not have implicit rules and has apparent API usage, but not a small amount of additionally exported function lists.

I've seen other APIs accept multiple type of parameters.. I think that's a good idea.

kjhman21:
> Another way of implementing bridge alias feature in APIs is extending input parameter types of existing APIs to handle both `common.Address` and `string`. Then, users can call bridge with either address or alias with one restriction, alias cannot start with `0x`.

You can check out how getBlock works with hash and block number.

hyunsooda:
@aidan-kwon, @2dvorak, @kjhman21, The alias-dedicated APIs take bridge alias and the other existing methods take two alias raw addresses. The different number of parameters is an obstacle to re-implement without changing the parameter structure. We should not change the existing methods' number of parameters due to incompatibility. The advantage of the bridge alias consists of (1) human-readable format (alias string) and usage with only one alias parameter than two raw bridge address parameters.

aidan-kwon:
> The different number of parameters is an obstacle to re-implement without changing the parameter structure. We should not change the existing methods' number of parameters due to incompatibility. 

@hyunsooda I think we can implement a multi-type parameter without the parameter number increasing. An example is [`rpc.BlockNumberOrHash` parameter used in GetBalance API](https://github.com/klaytn/klaytn/blob/c46e2fc7a1fb02456b7d82ab4b5083c476090b20/api/api_public_blockchain.go#L123).
Do I miss something?

hyunsooda:
@aidan-kwon, Thanks for the link. The parameter `blockNrOrHash` holds either block number or hash without no additional parameters. Let me explain my case again. The bridge alias tasks one parameter (bridge alias), and the existing method takes two raw addresses, which means the two parameters should be able to represent the bridge alias, or the bridge alias parameter should represent two raw addresses. How can you design it as a multi-type scheme? Of course, there are many workarounds, such as concatenating two raw addresses to a single address format, but it seems not a good idea. Please add if I'm misunderstanding.

aidan-kwon:
@hyunsooda I missed that alias is for a pair, but not for a single address. In that case, we can choose another solution.  API can get [optional parameters as you can see in the GetCommitee API.](https://github.com/klaytn/klaytn/blob/c46e2fc7a1fb02456b7d82ab4b5083c476090b20/consensus/istanbul/backend/api.go#L199). If the first parameter is a multi-type and the second parameter is optional, we can implement multi-type parameter APIs

hyunsooda:
@aidan-kwon, I'm sorry, I didn't find what you explained from the link you provided. That function has a single parameter.

hyunsooda:
@aidan-kwon, Please take a look at the updated PR description and added commits.

console/web3ext/web3ext.go
```diff
@@ -1394,6 +1400,14 @@ web3._extend({
...
 		}),
 		new web3._extend.Method({
+			name: 'getBridgePairByAlias',
+			call: 'subbridge_getBridgePairByAlias',
+			params: 1
+		}),
+		new web3._extend.Method({
+			name: 'changeBridgeAlias',
+			call: 'subbridge_changeBridgeAlias',
+			params: 2
```
aidan-kwon, line 1410:
```suggestion
			params: 2
		new web3._extend.Method({
```

hyunsooda: :( I should have checked the final commits out. Fixed. Thank you.

console/jsre/deps/web3.js
```diff
@@ -3850,6 +3850,11 @@ var inputBlockNumberFormatter = function (blockNumber) {
     return utils.toHex(blockNumber);
 };
 
+var inputEmptyFormatter = function (a) {
+  if (a === undefined) return "";
+  else                 return a;
+}
```
aidan-kwon, line 3856: After updating `web3.js`, you need to generate `bindata.go` again. Refer to `console/jsre/deps/deps.go`

node/sc/api_bridge.go
```diff
@@ -361,6 +442,21 @@ func (sb *SubBridgeAPI) RegisterToken(cBridgeAddr, pBridgeAddr, cTokenAddr, pTok
 	return nil
 }
 
+func (sb *SubBridgeAPI) RegisterToken(cBridgeAddrStrP, pBridgeAddrStrP, cTokenAddrStrP, pTokenAddrStrP *string) error {
```
aidan-kwon, line 445: It seems that `cBridgeAddrStrP` can be either a child bridge address or an alias. But, the name of the parameter confused me. How about changing the name to `cBridgeAddrOrAlias, `firstParam`, or a more intuitive name. 

`parseBridgeAddrWithAlias` also looks similar.

hyunsooda: Considering this API call:
```
subbridge_registerToken(ALIAS, 0XCHILD_TOKEN, 0XPARENT_TOKEN)
or
subbridge_registerToken(0XCHILD_BRIDGE, 0XPARENT_BRIDGE, 0XCHILD_TOKEN, 0XPARENT_TOKEN)
```
The first and second params are perceived as the following two sets:
1. alias and child token address
2. child bridge address and parent bridge address

Your suggestion hits for the first param, but not the second param in the case above. `FirstParam`, `SecondParam` does not have any identifiable info at all too. Considering the current namings, developers can figure out all the convertible cases by referring to the number of params from `console/web3ext/web3ext.go`.

We may consider changing all the names verbosely, like those representing possibly convertible cases (e.g., cBridgeAddrOrAlias, pBridgeOrChildToken). Please feel free to add further comments.

aidan-kwon: I prefer the verbose style names than inaccurate name.

hyunsooda: Renamed.

node/sc/api_bridge.go
```diff
@@ -179,8 +202,22 @@ func (sb *SubBridgeAPI) SubscribeBridge(cBridgeAddr, pBridgeAddr common.Address)
...
+	cBridgeAddrOrAliasStr, pBridgeAddrOrEmptyStr := stringDeref(cBridgeAddrOrAlias), stringDeref(pBridgeAddrOrEmpty)
+
+	var err error
+	bridgeAlias, cBridgeAddr, pBridgeAddr, _ := parseBridgeAddrWithAlias(cBridgeAddrOrAliasStr, pBridgeAddrOrEmptyStr)
+	if bridgeAlias != "" {
+		cBridgeAddr, pBridgeAddr, err = sb.subBridge.bridgeManager.getAddrByAlias(bridgeAlias)
+		if err != nil {
+			return err
+		}
+	}
```
aidan-kwon, line 215: This logic is repeated when `parseBridgeAddrWithAlias` is called. 
How about processing `getAddrByAlias` inside of `parseBridgeAddrWithAlias` ?
Then, `parseBridgeAddrWithAlias` is gonna return `(cBridgeAddr, pBridgeAddr, args, err)`

hyunsooda: Good suggestion. Refactored.

node/sc/bridge_addr_journal.go
```diff
@@ -24,31 +24,38 @@ import (
...
 // bridgeAddrJournal is a rotating log of addresses with the aim of storing locally
 // created addresses to allow deployed bridge contracts to survive node restarts.
 type bridgeAddrJournal struct {
-	path   string         // Filesystem path to store the addresses at
-	writer io.WriteCloser // Output stream to write new addresses into
-	cache  map[common.Address]*BridgeJournal
+	path       string         // Filesystem path to store the addresses at
+	writer     io.WriteCloser // Output stream to write new addresses into
+	cache      map[common.Address]*BridgeJournal
+	aliasCache map[string]common.Address
```
aidan-kwon, line 50: It seems that the concurrent access error happens if more than two APIs are concurrently requested. What do you think about adding mutex for these maps?

aidan-kwon (approved):
I think it's good to go, but please consider my last comment also.

hqjang-pepper (dismissed):

node/sc/bridge_addr_journal.go
hyunsooda: The relevant functions that access these two maps are not expected to be called concurrently, but they are potential bugs without considering critical sections. I think concurrent access guard must be considered too. However, this change must be cautiously implemented, and enough test cases too. How about moving this consideration to another PR?

aidan-kwon: @hyunsooda I agree with you.

aidan-kwon: The topic in this conversation is handled in https://github.com/klaytn/klaytn/pull/1438

aidan-kwon:
Close this because all contents are handled in https://github.com/klaytn/klaytn/pull/1438

PR #1298, opened 2022-04-01
Add MixedEngine as pass-through to governance.Governance

blukat29:
## Proposed changes

- This is follow-up of https://github.com/klaytn/klaytn/pull/1297
- Introduce `MixedEngine` which is a a new implementation of `governance.Engine`
- For now, MixedEngine is simply a wrapper to `governance.Governance`. All function calls pass-through to underlying Governance instance.
- In the future, MixedEngine will contain multiple engines in it and redirect function calls to them depending on configuration and block number.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/governance/mixed.go
+++ b/governance/mixed.go
@@ -1,0 +1,198 @@
+package governance
+
+import (
+	"github.com/klaytn/klaytn/blockchain/types"
+	"github.com/klaytn/klaytn/common"
+	"github.com/klaytn/klaytn/consensus/istanbul"
+	"github.com/klaytn/klaytn/params"
+	"github.com/klaytn/klaytn/storage/database"
+)
+
+type MixedEngine struct {
+	initialConfig *params.ChainConfig
+
+	// Data sources
+	db database.DBManager
+
+	// Subordinate engines
+	defaultGov *Governance
+}
+
+func newMixedEngine(config *params.ChainConfig, db database.DBManager, doInit bool) *MixedEngine {
+	e := &MixedEngine{
+		initialConfig: config,
+		db:            db,
+	}
+
+	if doInit {
+		e.defaultGov = NewGovernanceInitialize(config, db)
+	} else {
+		e.defaultGov = NewGovernance(config, db)
+	}
+	return e
+}
+
+// Most users are encouraged to call this constructor
+func NewMixedEngine(config *params.ChainConfig, db database.DBManager) *MixedEngine {
+	return newMixedEngine(config, db, true)
+}
+
+// For test purposes
+func NewMixedEngineNoInit(config *params.ChainConfig, db database.DBManager) *MixedEngine {
+	return newMixedEngine(config, db, false)
+}
+
+// Pass-through to HeaderEngine
+func (e *MixedEngine) AddVote(key string, val interface{}) bool {
+	return e.defaultGov.AddVote(key, val)
+}
+func (e *MixedEngine) ValidateVote(vote *GovernanceVote) (*GovernanceVote, bool) {
+	return e.defaultGov.ValidateVote(vote)
+}
+func (e *MixedEngine) CanWriteGovernanceState(num uint64) bool {
+	return e.defaultGov.CanWriteGovernanceState(num)
+}
+func (e *MixedEngine) WriteGovernanceState(num uint64, isCheckpoint bool) error {
+	return e.defaultGov.WriteGovernanceState(num, isCheckpoint)
+}
+func (e *MixedEngine) ReadGovernance(num uint64) (uint64, map[string]interface{}, error) {
+	return e.defaultGov.ReadGovernance(num)
+}
+func (e *MixedEngine) WriteGovernance(num uint64, data GovernanceSet, delta GovernanceSet) error {
+	return e.defaultGov.WriteGovernance(num, data, delta)
+}
+func (e *MixedEngine) GetEncodedVote(addr common.Address, number uint64) []byte {
+	return e.defaultGov.GetEncodedVote(addr, number)
+}
+func (e *MixedEngine) GetGovernanceChange() map[string]interface{} {
+	return e.defaultGov.GetGovernanceChange()
+}
+func (e *MixedEngine) VerifyGovernance(received []byte) error {
+	return e.defaultGov.VerifyGovernance(received)
+}
+func (e *MixedEngine) ClearVotes(num uint64) {
+	e.defaultGov.ClearVotes(num)
+}
+func (e *MixedEngine) WriteGovernanceForNextEpoch(number uint64, governance []byte) {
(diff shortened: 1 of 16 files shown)


blukat29:
Rebased to latest dev.

governance/mixed.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/klaytn/klaytn/blockchain/types"
+	"github.com/klaytn/klaytn/common"
+	"github.com/klaytn/klaytn/consensus/istanbul"
+	"github.com/klaytn/klaytn/params"
+	"github.com/klaytn/klaytn/storage/database"
+)
+
+type MixedEngine struct {
+	// Subordinate engines
+	defaultGov *Governance
```
hyunsooda, line 13: Not the `Engine` interface?

governance/mixed.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func newMixedEngine(config *params.ChainConfig, db database.DBManager, doInit bool) *MixedEngine {
```
hyunsooda, line 16: Description omitted.

governance/mixed.go
```diff
@@ -0,0 +1,188 @@
...
+	// Setup subordinate engines
+	if doInit {
+		e.defaultGov = NewGovernanceInitialize(config, db)
+	} else {
+		e.defaultGov = NewGovernance(config, db)
+	}
+	return e
+}
+
+// Most users are encouraged to call this constructor
```
hyunsooda, line 28:
```suggestion
// NewMixedEngine instantiate a new MixedEnginestruct. Developers are encouraged to call this constructor in most cases.
```

hyunsooda (commented):
1. I have a doubt for `MixedEngine` wrapper is really necessary. I think the Engine interface may be positioned as a wrapper of governance implementation to be implemented in the future. 

2. If the `MixedEngine` is really necessary for some reasons behind the description of this PR, consideration for renaming seems required. Of course, I can feel your naming, but the keyword `mixed` seems awkward.

governance/mixed.go
blukat29: Fixed in https://github.com/klaytn/klaytn/pull/1298/commits/4cc46318076f6ec2fb03ebf1ea4f13108ee51a6d.

governance/mixed.go
blukat29: Fixed in https://github.com/klaytn/klaytn/pull/1298/commits/4cc46318076f6ec2fb03ebf1ea4f13108ee51a6d.

blukat29:
`interface Engine` and `struct MixedEngine` have different roles.

- `interface Engine`
  - This is what other modules perceive.
  - Being an interface (as opposed to struct) allows easily adding a mock, testing new implementations.
  - Currently it is equal to `HeaderEngine`, but more methods will be added later and some unused methods will be removed.
- `struct MixedEngine`
  - This is an implementation of `Engine`.
  - It switches between subordinate engines.
  - For example, HeaderEngine shall be used only until a certain hardfork block, and ContractEngine should be used after. (refer to https://github.com/blukat29/klaytn/blob/gov/try3/governance/mixed.go)
  - Only a struct can have such code, but interface cannot.

@hyunsooda Given that, let me answer your questions.

1. Engine cannot replace MixedEngine because an interface cannot have concrete code in itself, and we have to put the logic somewhere.
2. MixedEngine switches between implementations, hence the name Mixed. The word 'mixed' could be used to describe two different things coexisting as in Mixed reality or Mixed signal.
3. One could ask, then why don't you remove `interface Engine` and  rename the `struct MixedEngine` to `struct Engine`.
  a. I want to separate interface and struct because having an interface adds flexibility. Governance code appears many places in our codebase. Changing the type name or call sites are quite a hassle. I think we still have a lot to improve about governance, therefore the flexibility should be helpful for future development.

governance/mixed.go
blukat29: Maybe I should use `HeaderEngine` instead.

ian0371:
Do you think we need to copy `default_test.go` into `mixed_test.go` and update related variables so we can test MixedEngine's defaultGov behaviour in a similar way?

blukat29:
> Do you think we need to copy `default_test.go` into `mixed_test.go` and update related variables so we can test MixedEngine's defaultGov behaviour in a similar way?

I don't think so. MixedEngine's defaultGov behavior won't change. What I think `mixed_test.go` should have is, the switching (or handover) between engines. Since we only have one engine, there seems nothing to test just yet.

ian0371 (approved):

hyunsooda (approved):
@blukat29, I had felt overlapped stuff and unnecessary stuff when I first saw this change. I don't know your blueprint for now. I will figure it out at the later PRs then.

blukat29:
@hyunsooda Thanks!

Should have wrote enough backgrounds from the first. I'm sorry to confuse you.

I have built a prototype implementation for the new contract-based governance in my personal repository (https://github.com/blukat29/klaytn/tree/gov/try3). Now I am composing a revised version in 'feat/contractgov' branch (https://github.com/klaytn/klaytn/commits/feat/contractgov).

This contract-based governance feature is rather huge, so it won't fit in a single PR.

During recent weeks I am gradually merging those into klaytn/klaytn (this PR included). I want PRs to be as small as possible so that reviewers be confident about their reviews. I want each changes to be throughly validated.

Some PRs intentionally doesn't add any feature. Some layers may seem unnecessary but they are there for a reason. They are merely refactoring. I'm making a separate refactoring PR because the change is not small.

If you are curious about how these changes are to be utilized, please refer to above GitHub repository and our Confluence.

hyunsooda:
Okay, now I see your strategy. Let me just put one comment here. I don't have the governance background and I'm learning what it is from your PRs; thus, you don't need to consider my comments seriously. Unfortounelty, I guess my comments would be questions generally, not the helpful improvement for your changes.

blukat29:
@hyunsooda I am always grateful about your interest; Any kind of inputs are welcome - either it's about clarity, style, harmony with other code, etc. Those questions make my intentions more clear and verbose, which is very important for all dev team and future developers who wish to work with our codebase. :)

yoomee1313 (approved):

blukat29:
@aeharvlee I need your approval to change a line in `api/api_ethereum_test.go`. Can you have a quick look?

blukat29:
@2dvorak I need you approval to change a couple lines in `cmd/utils/nodecmd/chaincmd.go`. It is purely syntactic change. Please take a look.

2dvorak (approved):

blukat29:
@2dvorak Thanks!

blukat29:
@aeharvlee, @jimni1222, @sirano11, I need your approval to change one line in api/api_ethereum_test.go. Can you have a quick look?

aeharvlee (approved):

Issue #1300, opened 2022-04-04
Would it be better to print the location of an imported account?


hyunsooda:
The account import takes the user's private key with the data directory to be located imported keystore directory. It seems better to explicitly print the location of the imported account. If the data directory is not explicitly given in the cmd argument, then it would be imported to the default location (i.e.,`~/.kcn`, `~/.ken`, and so forth). It would save time for the reasoning why the account is not actually imported when they forget to give a data directory argument or there's a typo of the data directory argument. If you guys think it seems helpful, I'll upload PR.

kjhman21:
@hyunsooda Yeah, why not? You can also upload PR as you want and then we can discuss the usefulness and necessity in the PR. :)

henry-will:
Thanks.

PR #1301, opened 2022-04-04
[Keystore] Added output of imported account directory

hyunsooda:
## Proposed changes

The account import takes the user's private key with the data directory to be located imported keystore directory. It seems better to explicitly print the location of the imported account. If the data directory is not explicitly given in the cmd argument, then it would be imported to the default location (i.e.,~/.kcn, ~/.ken, and so forth). It would save time for the reasoning why the account is not actually imported when they forget to give a data directory argument or there's a typo of the data directory argument. (the description is grabbed from #1300.)

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/accounts/keystore/keystore.go
+++ b/accounts/keystore/keystore.go
@@ -539,10 +539,12 @@ func (ks *KeyStore) ImportECDSA(priv *ecdsa.PrivateKey, passphrase string) (acco
 }
 
 func (ks *KeyStore) importKey(key Key, passphrase string) (accounts.Account, error) {
-	a := accounts.Account{Address: key.GetAddress(), URL: accounts.URL{Scheme: KeyStoreScheme, Path: ks.storage.JoinPath(keyFileName(key.GetAddress()))}}
+	accDir := ks.storage.JoinPath(keyFileName(key.GetAddress()))
+	a := accounts.Account{Address: key.GetAddress(), URL: accounts.URL{Scheme: KeyStoreScheme, Path: accDir}}
 	if err := ks.storage.StoreKey(a.URL.Path, key, passphrase); err != nil {
 		return accounts.Account{}, err
 	}
+	fmt.Println("Your account is imported at", accDir)
 	ks.cache.add(a)
 	ks.refreshWallets()
 	return a, nil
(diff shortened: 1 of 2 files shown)


accounts/keystore/keystore.go
```diff
@@ -539,10 +539,12 @@ func (ks *KeyStore) ImportECDSA(priv *ecdsa.PrivateKey, passphrase string) (acco
 }
 
 func (ks *KeyStore) importKey(key Key, passphrase string) (accounts.Account, error) {
-	a := accounts.Account{Address: key.GetAddress(), URL: accounts.URL{Scheme: KeyStoreScheme, Path: ks.storage.JoinPath(keyFileName(key.GetAddress()))}}
+	accDir := ks.storage.JoinPath(keyFileName(key.GetAddress()))
+	a := accounts.Account{Address: key.GetAddress(), URL: accounts.URL{Scheme: KeyStoreScheme, Path: accDir}}
 	if err := ks.storage.StoreKey(a.URL.Path, key, passphrase); err != nil {
 		return accounts.Account{}, err
 	}
+	fmt.Println("Your account is imported at", accDir)
```
hqjang-pepper, line 547:
```suggestion
	logger.Info("Your account is imported at", "path",accDir)
```
We rarely use `fmt.println` for giving information. How about using `logger.Info`?

hyunsooda: Thanks for the suggestion. When the `account import` command is used, all prints are output to `stdout` using `fmt.Println`. That's why I just followed their usage. Of course, we can push it to logs, but it may be harder or worse to find that.

aidan-kwon: `personal` API also calls this method, and this stdout output can make the node log massy. If you are intended to print this message with homi or node commands, you better print this in `homi` and `nodecmd` packages.

hyunsooda: Hi @aidan-kwon, Thanks for the advice. I found the function `accountImport()` is only called when the import command is triggered. I think placing this print log at the end of that function seems to be the correct location.

kjhman21 (approved):
LGTM

aidan-kwon (approved):

PR #1303, opened 2022-04-05
Add new GovParamSet type as governance parameter model

blukat29:
## Proposed changes

- Introduce new struct `params.GovParamSet`
- It is an immutable set of governance parameters.
- It unifies tedious and error-prone governance data processing
  - Adjust json.Unmarshal'd data (formerly at `adjustDecodedSet()`)
  - Parse []byte encoded data (to read contract output)
  - Nominal getters (like `Governance.Epoch()`)
  - General purpose getters (like `Governance.GetGovernanceValue(key int)`)
- A crucial role of new governance engine will be serving GovParamSet at different block numbers.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/params/governance_paramset.go
+++ b/params/governance_paramset.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 
 	"github.com/klaytn/klaytn/common"
-	"github.com/klaytn/klaytn/log"
 )
 
 type govParamType struct {
@@ -221,22 +220,6 @@ var govParamNames = map[string]int{
 	"reward.proposerupdateinterval": ProposerRefreshInterval,
 }
 
-var govParamDefaults = map[int]interface{}{
-	GovernanceMode:          DefaultGovernanceMode,
-	GoverningNode:           DefaultGoverningNode,
-	Epoch:                   DefaultEpoch,
-	Policy:                  DefaultProposerPolicy,
-	CommitteeSize:           DefaultSubGroupSize,
-	UnitPrice:               DefaultUnitPrice,
-	MintingAmount:           DefaultMintingAmount,
-	Ratio:                   DefaultRatio,
-	UseGiniCoeff:            DefaultUseGiniCoeff,
-	DeferredTxFee:           DefaultDefferedTxFee,
-	MinimumStake:            DefaultMinimumStake,
-	StakeUpdateInterval:     DefaultStakeUpdateInterval,
-	ProposerRefreshInterval: DefaultProposerRefreshInterval,
-}
-
 var govParamNamesReverse = map[int]string{}
 
 func init() {
@@ -348,91 +331,7 @@ func (p *GovParamSet) IntMap() map[int]interface{} {
 }
 
 // Returns a parameter value and a boolean indicating success.
-// Use Get() to be careful about existence of a parameter.
 func (p *GovParamSet) Get(key int) (interface{}, bool) {
 	v, ok := p.items[key]
 	return v, ok
-}
-
-// Returns a parameter value or default value for the key if the value is not set.
-// Use MustGet() if you are confident that the value is set.
-// To simply quote a parameter in an expression, MustGet() can come in handy.
-func (p *GovParamSet) MustGet(key int) interface{} {
-	if v, ok := p.items[key]; ok {
-		return v
-	}
-	if d, ok := govParamDefaults[key]; ok {
-		// If key is a known parameter, fallback to default
-		return d
-	}
-	// This path is very unlikely.
-	logger := log.NewModuleLogger(log.Governance)
-	logger.Crit("Unknown governance param key")
-	return nil
-}
-
-func (p *GovParamSet) GovernanceMode() string {
-	return p.MustGet(GovernanceMode).(string)
-}
-
-func (p *GovParamSet) GoverningNode() common.Address {
-	return p.MustGet(GoverningNode).(common.Address)
-}
-
-func (p *GovParamSet) Epoch() uint64 {
-	return p.MustGet(Epoch).(uint64)
-}
-
-func (p *GovParamSet) Policy() uint64 {
-	return p.MustGet(Policy).(uint64)
-}
-
-func (p *GovParamSet) CommitteeSize() uint64 {
-	return p.MustGet(CommitteeSize).(uint64)
-}
-
-func (p *GovParamSet) UnitPrice() uint64 {
-	return p.MustGet(UnitPrice).(uint64)
-}
-
-func (p *GovParamSet) MintingAmountStr() string {
-	return p.MustGet(MintingAmount).(string)
-}
-
-func (p *GovParamSet) MintingAmountBig() *big.Int {
-	n, _ := new(big.Int).SetString(p.MintingAmountStr(), 10)
-	return n
-}
-
-func (p *GovParamSet) Ratio() string {
-	return p.MustGet(Ratio).(string)
-}
-
-func (p *GovParamSet) UseGiniCoeff() bool {
-	return p.MustGet(UseGiniCoeff).(bool)
-}
-
-func (p *GovParamSet) DeferredTxFee() bool {
-	return p.MustGet(DeferredTxFee).(bool)
-}
-
-func (p *GovParamSet) MinimumStakeStr() string {
-	return p.MustGet(MinimumStake).(string)
-}
-
-func (p *GovParamSet) MinimumStakeBig() *big.Int {
-	n, _ := new(big.Int).SetString(p.MinimumStakeStr(), 10)
-	return n
-}
-
-func (p *GovParamSet) StakeUpdateInterval() uint64 {
-	return p.MustGet(StakeUpdateInterval).(uint64)
-}
-
-func (p *GovParamSet) ProposerRefreshInterval() uint64 {
-	return p.MustGet(ProposerRefreshInterval).(uint64)
-}
-
-func (p *GovParamSet) Timeout() uint64 {
-	return p.MustGet(Timeout).(uint64)
 }
--- a/params/governance_paramset.go
+++ b/params/governance_paramset.go
@@ -1,0 +1,438 @@
+package params
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"math/big"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/klaytn/klaytn/common"
+	"github.com/klaytn/klaytn/log"
+)
+
+type govParamType struct {
+	// The canonical type
+	canonicalType reflect.Type
+
+	// Parse arbitrary typed value into canonical type
+	// Return false if not possible
+	parseValue func(v interface{}) (interface{}, bool)
+
+	// Parse byte array into canonical type
+	// Return false if not possible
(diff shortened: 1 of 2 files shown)


cmd/homi/setup/flags.go
```diff
@@ -321,7 +321,7 @@ var (
 	cliquePeriodFlag = cli.IntFlag{
 		Name:  "clique-period",
 		Usage: "clique period",
-		Value: params.DefaultPeriod,
+		Value: int(params.DefaultPeriod),
```
kjhman21, line 324:
```suggestion
		Value: params.DefaultPeriod,
```

params/governance_params.go
```diff
@@ -111,19 +111,22 @@ const (
...
+	DefaultGoverningNode           = "0x0000000000000000000000000000000000000000"
+	DefaultEpoch                   = uint64(604800)
+	DefaultProposerPolicy          = uint64(0)
+	DefaultSubGroupSize            = uint64(21)
+	DefaultMintingAmount           = big.NewInt(0)
+	DefaultRatio                   = "100/0/0"
+	DefaultUseGiniCoeff            = false
+	DefaultDefferedTxFee           = false
+	DefaultUnitPrice               = uint64(250000000000)
+	DefaultPeriod                  = uint64(1)
```
kjhman21, line 126:
```suggestion
	DefaultPeriod                  = int64(1)
```

cmd/homi/setup/flags.go
blukat29: @kjhman21 I will use `cli.Uint64Flag` instead.

blukat29:
Squashed and rebased to latest dev.

params/governance_paramset.go
```diff
@@ -0,0 +1,419 @@
...
+
+// If the value is not set, logs a warning and returns the default value.
+func (p *GovParamSet) MustGet(key int) interface{} {
+	if v, ok := p.items[key]; ok {
+		return v
+	}
+	if d, ok := govParamDefaults[key]; ok {
+		// If key is a known parameter, fallback to default
+		return d
+	}
```
yoomee1313, line 363: The comment says `if the value is not set, logs a warning`. According to the comments, shouldn't a warning log be added? 

By the way, if there is no value corresponding to the `key`, it seems it reads the default value. If the default value changes, it gets a different value compared to the past. Is it okay to read different value?

params/governance_paramset.go
```diff
@@ -0,0 +1,419 @@
...
+	m := map[int]interface{}{}
+	for key, value := range p.items {
+		m[key] = value
+	}
+	return m
+}
+
+func (p *GovParamSet) Get(key int) (interface{}, bool) {
+	v, ok := p.items[key]
+	return v, ok
```
yoomee1313, line 352: What about this?
```suggestion
	return p.items[key]
```

params/governance_paramset.go
```diff
@@ -0,0 +1,419 @@
...
+
+func (p *GovParamSet) IntMap() map[int]interface{} {
+	m := map[int]interface{}{}
+	for key, value := range p.items {
+		m[key] = value
+	}
+	return m
+}
+
+func (p *GovParamSet) Get(key int) (interface{}, bool) {
```
yoomee1313, line 350: By the way, why two types of `get` are provided? If you don't mind, can you write the comment about the difference?

params/governance_paramset.go
blukat29: That is not possible at least in go 1.18.

```
./governance_paramset.go:353:9: not enough return values
        have (interface{})
        want (interface{}, bool)
```

params/governance_paramset.go
blukat29: I added the comment about them in https://github.com/klaytn/klaytn/pull/1303/commits/e16a3626158d89e1ed87712e71f0c34edfc5f3cf.

`Get(key)` returns a boolean along with value. https://github.com/klaytn/klaytn/blob/v1.8.3/consensus/istanbul/backend/engine.go#L619 is a use case.

`MustGet(key)` always returns some value.
Main role of `MustGet(key)` is to implement nominal(noun-like) getters like `Epoch()`. 
Nominal getters should be the most popular way to read parameters.  https://github.com/klaytn/klaytn/blob/v1.8.3/consensus/istanbul/backend/engine.go#L230 is a use case.

As of now, I cannot determine if any one of them is unnecessary. I decided have both to serve both existing use cases.
When I replace all governance getters with GovParamSet (like in https://github.com/blukat29/klaytn/commit/147007c0), I will re-think about these two.

Feel free to comment. Thanks!

blukat29:
Rebased to latest dev.

params/governance_paramset.go
yoomee1313: I mean, if the codebase's default value is changes, it reads the difference value compared to the past. Are you meaning that the constants will not be changed?

params/governance_paramset.go
yoomee1313: Thanks for the comments

params/governance_paramset.go
blukat29: Understood. The fallback value is there to make sure calls like `goverance.Epoch()` does not panic. These nominal getters are widely used throughout our codebase. The problem is when there is no value.

We have some choices here.

1. Do not provide MustGet and nominal getters at all
    - I'd say it's the safest way. All users are aware that some governance parameter can be undefined and do their own error handling.
    - Proper error handling is required. 
    ```go
    var epoch uint64
    if val, ok := gov.Params().Get(params.Epoch); ok {
        epoch = val.(uint64)
    } else {
        logger.Crit("Cannot get istanbul.epoch")
    }
    ```
2. MustGet returns an empty value like `0` or `""`
    - Empty value may be silently used. Potentially hard-to-catch bug.
    ```go
    epoch := gov.Params().Epoch() // could be 0 on error
    ```
3. MustGet returns designated default value (i.e. params.DefaultXX)
    - It is hard to distinguish between fallback value and actual value.
    - Default value may be silently used.
    - The default may change later. If some code might be relying on the default value (even if the developer didn't know). Potentially hard-to-catch bug.
    ```go
    epoch := gov.Params().Epoch() // could be 86400 on error
    ```
4. Each GovParamSet instance holds an explicit default value
    - For instance, ChainConfig can be used as default value.
    - This behavior is similar to current default Governance, because currentSet is first populated with chainConfig and then overwritten by DB content.
    ```go
    // governance/engine.go
    func NewMyEngine(config *params.ChainConfig, ...) *MyEngine {
        return &MyEngine{
            ChainConfig: config,
            DefaultSet: GovParamSetChainConfig(config),
        }
    }
    func (gov *Engine) Params() *params.GovParamSet {
        data := gov.ReadGovernance()
        return GovParamSetStrMap(data, gov.DefaultSet)
    }
    // params/governance_paramset.go
    func (p *GovParamSet) MustGet(key int) interface{} {
        if v, ok := p.Get(key); ok {
            return v
        } else if v, ok := p.Defaults.Get(key); ok {
            return v
        } else {
            logger.Crit("no default value for", key)
        }
    }            
    ```
    ```go
    epoch := gov.Params().Epoch() // could be config.Istanbul.Epoch on error
    ```

Current PR is choice3. But I think moving to choice1 or choice4 would be safer.
@yoomee1313 What do you think?

yoomee1313: Thanks for organizing it so well. I agree with you that choice 1 or choice 4 is safer. Will you fix the code as choice 1 or choice4?

blukat29: Yes, preferrably in this PR.

blukat29: @yoomee1313 How about going for choice1 in this PR (leave only `Get()`), then add MustGet and nominal getters later? If you are okay, I will add commits.

yoomee1313: @blukat29 Sure, I'm okay with option1.

blukat29:
@yoomee1313 Removed MustGet and related features. Please look again.

yoomee1313 (approved):
it looks good to me.

params/governance_paramset_test.go
```diff
@@ -0,0 +1,197 @@
...
+	}{
+		{govParamTypeGovMode, []byte("single"), "single", true},
+		{govParamTypeGovMode, []byte(""), nil, false},
+
+		{govParamTypeAddress, zeroAddr.Bytes(), zeroAddr, true},
+		{govParamTypeAddress, []byte(zeroAddr.Hex()), nil, false},
+		{govParamTypeAddress, []byte(""), nil, false},
+
+		{govParamTypeUint64, []byte{0x12, 0x34}, uint64(0x1234), true},
+		{govParamTypeUint64, []byte{}, uint64(0), true},
```
ian0371, line 87: What makes an empty slice ok when another case (line 101) is not?

blukat29: - I assumed that there could be some case where `0` is stored as an empty array in solidity.
That could happen because `big.NewInt(0).Bytes() == []byte{}`.

- For boolean, I wanted to be more strict so that only `0x00` and `0x01` are valid values.
This prevents accidently plugging in non-boolean values to boolean parameter.

blukat29: However, we can choose to interpret all truthy-values to true, falsey-values to value. i.e.

```go
govParamTypeBool = &govParamType{
    parseBytes = func(b []byte) (interface{}, bool) {
        return new(big.Int).SetBytes(b).Cmp(common.Big0) != 0, true
    }
}
```

@ian0371 What do you think about this change?

ian0371: I see. The strict policy would be better.

params/governance_paramset.go
```diff
@@ -0,0 +1,337 @@
...
+		parseBytes:    parseBytesString,
+		validate: func(v interface{}) bool {
+			strs := strings.Split(v.(string), "/")
+			if len(strs) != 3 {
+				return false
+			}
+			for _, s := range strs {
+				n, err := strconv.Atoi(s)
+				if err != nil || n < 0 {
+					return false
```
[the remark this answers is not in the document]
blukat29: Fixed in https://github.com/klaytn/klaytn/pull/1303/commits/03db1879b84d97a477d4e1ecfa7b69d371e6a294.

params/governance_paramset.go
```diff
@@ -0,0 +1,337 @@
...
+				return v.(uint64), true
+			case float64:
+				return uint64(v.(float64)), v.(float64) >= 0
+			default:
+				return nil, false
+			}
+		},
+		parseBytes: func(b []byte) (interface{}, bool) {
+			// Must not exceed uint64 range
+			return new(big.Int).SetBytes(b).Uint64(), len(b) <= 8
```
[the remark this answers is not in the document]
blukat29: Fixed in https://github.com/klaytn/klaytn/pull/1303/commits/03db1879b84d97a477d4e1ecfa7b69d371e6a294.

ian0371 (dismissed):

blukat29:
Rebased to latest dev.

ian0371 (approved):

yoomee1313 (approved):

PR #1311, opened 2022-04-08
Added panic guards for prefetchers

aidan-kwon:
## Proposed changes

Klaytn increased sync performance starting prefetching before block body validation in https://github.com/klaytn/klaytn/pull/918.
However, this change can trigger issues because the prefetcher executes blocks having unvalidated headers and bodies. 
For example, `nil` parent because of future block insertion or validation errors causes panic returning `invalid memory address or nil pointer dereference` error. 

So, I adopted two logic to be safe in unexpected cases
1. Add the parent is not `nil` condition for next block prefetcher execution.
2. Recover panic from prefetcher. The invalid execution of the prefetcher doesn't affect to the integrity of Klaytn system, so we can recover the error and keep operating the node. 

I think another option is just reverting https://github.com/klaytn/klaytn/pull/918 to get confidence on the blocks that will be executed in the prefetcher. If you can think like this, please let me know. I can change this PR in that way. 

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1780,11 +1780,12 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		// transactions and probabilistically some of the account/storage trie nodes.
 		var followupInterrupt uint32
 
-		if bc.cacheConfig.TrieNodeCacheConfig.NumFetcherPrefetchWorker > 0 {
+		if bc.cacheConfig.TrieNodeCacheConfig.NumFetcherPrefetchWorker > 0 && parent != nil {
 			var snaps *snapshot.Tree
 			if bc.cacheConfig.TrieNodeCacheConfig.UseSnapshotForPrefetch {
 				snaps = bc.snaps
 			}
+
 			// Tx prefetcher is enabled for all cases (both single and multiple block insertion).
 			for ti := range block.Transactions() {
 				select {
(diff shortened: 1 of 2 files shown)


blockchain/blockchain_test.go
```diff
@@ -1753,3 +1753,52 @@ func TestDeleteCreateRevert(t *testing.T) {
 		t.Fatalf("block %d: failed to insert into chain: %v", n, err)
 	}
 }
+
+// TestBlockChain_InsertChain_InsertFuntureBlocks inserts future blocks that have missing ancestor.
```
kjhman21, line 1757:
```suggestion
// TestBlockChain_InsertChain_InsertFutureBlocks inserts future blocks that have a missing ancestor.
```

kjhman21 (approved):
LGTM

blockchain/blockchain.go
```diff
@@ -1796,6 +1797,12 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 				// current block is not the last one, so prefetch the right next block
 				followup := chain[i+1]
 				go func(start time.Time) {
+					defer func() {
+						if err := recover(); err != nil {
+							logger.Error("Got panic and recovered from prefetcher", "err", err)
+						}
+					}()
+
```
jeongkyun-oh, line 1805: Does this recovering logic necessary to `txPrefetcher` as well? on blockchain.go#350

aidan-kwon: It might, but I am not sure yet. tx prefechers process the inserting block and it should not have processing errors. 
I will consider it later in another PR if needed

jeongkyun-oh (approved):

blockchain/blockchain.go
jeongkyun-oh: Alright. Thanks :)

PR #1314, opened 2022-04-08
Fix intermittent failure of TestWalletNotifications

aidan-kwon:
## Proposed changes

TestWalletNotifications intermittently fails if the event collector goroutine is closed before receiving all events. The fix checks and receives unhandled events. 
The change is imported from https://github.com/ethereum/go-ethereum/pull/21703

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/accounts/keystore/keystore_test.go
+++ b/accounts/keystore/keystore_test.go
@@ -344,7 +344,9 @@ func TestWalletNotifications(t *testing.T) {
 
 	// Shut down the event collector and check events.
 	sub.Unsubscribe()
-	<-updates
+	for ev := range updates {
+		events = append(events, walletEvent{ev, ev.Wallet.Accounts()[0]})
+	}
 	checkAccounts(t, live, ks.Wallets())
 	checkEvents(t, wantEvents, events)
 }


accounts/keystore/keystore_test.go
```diff
@@ -344,7 +344,9 @@ func TestWalletNotifications(t *testing.T) {
 
 	// Shut down the event collector and check events.
 	sub.Unsubscribe()
-	<-updates
+	for ev := range updates {
+		events = append(events, walletEvent{ev, ev.Wallet.Accounts()[0]})
```
kjhman21, line 348: Hmm.. in that case I think we have a concurrency problem to add into 'events' object.
What about using wait group and waiting until [this goroutine](https://github.com/klaytn/klaytn/pull/1314/files#diff-ef18fafacf01f667fab3f8b8f44c3117410e61f79fe4d452059eca75a425bc2bR304-R314) ends?

aidan-kwon: It is tricky. I couldn't generate the race issue manipulating the execution timing of the event subscription loop close and main for loop calling `ks.NewAccount` and `ks.Delete`. Perhaps, there is a race between event feed subscription and send. In that case, it is hard to add a wait group in this test. 
So, I just apply the same mitigation geth applied, and I think it is acceptable even thous it doesn't prevent generating race issues. Because it is a test code.

KimKyungup (approved):

accounts/keystore/keystore_test.go
kjhman21: okay. got it.

kjhman21 (approved):

PR #1316, opened 2022-04-11
Fix txpool loop reset logic to support setHead

hqjang-pepper:
## Proposed changes
When setHead function is executed, panic has occured intermittently when syncing chain segments.
This error occured when txpool tries to reinject discarded transactions of blocks, but can't get block because blockchain is rewinded.
This PR is containing this error handling.

I tested enough with private network (4cn) environment.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


ehnuje:
Have you tried implementing a unit test for this case?

aidan-kwon (commented):
I am not sure this suggestion is proper mitigation for your issue. Can you explain the detail of your issue or share the panic log? 
It seems that all behaviors of txpool as well as `txpool.reset` should be stopped during the `SetHead`. In that case, we can consider locking txpool mutex or other way.

hqjang-pepper:
Sorry for the late response.
First, I will explain in detail the situation when the error occurs after executed the setHead method.
When performing TxPool.reset(oldHead,newHead)
`oldHead` is the head of the chain before setHead, and `newHead` is the head which is during syncing after completing setHead (arbitrary).
This newHead may have a number larger or smaller than the oldHead, which depends on how the downloader receives the chunk.
It is not a problem when oldHead is smaller than newHead, but when it is large, the block corresponding to oldHead does not exist in `pool.chain`, due to the block is deleted by `setHead` function.
There was no logic to handle this in the existing txpool.go code. This is because the situation which **the header exists but the block does not exist** does not occur when the chain operates normally.

@ehnuje I tried to write test code, but it was difficult to reproduce the situation where only the block header exists and block doesn't.

@aidan-kwon What I modified is handling of a special error that occurs only when the block corresponding to oldHead is nil when an event occurs in chainHeadCh.  The reason that event does not occur in chainHeadCh during setHead is the downloader function is turned off while SetHead. 
I don't think there is any logic to stop txpool with this code change.

blockchain/tx_pool.go
```diff
@@ -378,33 +378,49 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
...
-				discarded = append(discarded, rem.Transactions()...)
-				if rem = pool.chain.GetBlock(rem.ParentHash(), rem.NumberU64()-1); rem == nil {
-					logger.Error("Unrooted old chain seen by tx pool", "block", oldHead.Number, "hash", oldHead.Hash())
+			if rem == nil {
+				// This can happen if a setHead function is performed.
+				// In this case we can simply discard the old head.
+
+				// If newNum >= oldNum, then it's not a case of setHead.
+				if newNum >= oldNum {
+					logger.Warn("Transaction pool reset with missing oldhead",
```
aidan-kwon, line 387: If an unexpected situation happens and stops something to do, the log should be ERROR.

aidan-kwon (commented):
I understand that this PR is not harmful in normal situations. However, I don't think it is the correct mitigation for the setHead panic issue. I think we are supposed to find a way that prevents calling `reset` with no existing old blockhead. This change could be a partial solution to the issue, but it could conceal the real problem our code might have. 
Can you explain more detail about the situation the empty oldHead is used?

aidan-kwon (dismissed):

aidan-kwon (approved):
@hqjang-pepper Please, make an issue for the following task

jeongkyun-oh (approved):

PR #1321, opened 2022-04-13
[SC] A new value transfer function and value transfer event

hyunsooda:
## Proposed changes

The event `requestERC721Transfer()` emits a new event `RequestValueTransferEncoded`. The new event is a universal structure containing packed bytes that can be any type of data. No events do not need to be created for this packed data if a need for a new event comes in. The required work is only to add the corresponding decoder into `event_parse.go`.

Refer #1239 to track the previous history.

Compatibility issue:
```
          [Subbridge]                      [Mainbridge]
(1) New node + New contract          New node + New contract
(2) New node + New contract          Old node + New contract 
(3) Old node + New contract          New node + New contract
```
Except for the third case, all the other cases work regardless of who calls VT. Especially if the main bridge calls VT in the third case, the old version's node is not aware of the new event. The unhandled events would be correctly handled when the new version's node is booted with the value transfer recovery option.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


contracts/bridge/BridgeTransferERC721.sol
```diff
@@ -76,28 +76,49 @@ contract BridgeTransferERC721 is BridgeTokens, IERC721BridgeReceiver, BridgeTran
...
+                0,
+                _extraData
+            );
+        } else {
+            (bool success, bytes memory uri) = _tokenAddress.call(abi.encodePacked(ERC721Metadata(_tokenAddress).tokenURI.selector, abi.encode(_tokenId)));
+            if (success == false) {
+                uri = "";
+            }
 
+            emit RequestValueTransferEncoded(
```
kjhman21, line 106: You should move this after the burn. even if the burn fails, the event anyway occurs if it is here.
You can just declare a variable to store `abi.encode(string(uri))` here and then pass the variable to the event.

contracts/bridge/BridgeTransferERC721.sol
```diff
@@ -110,7 +131,21 @@ contract BridgeTransferERC721 is BridgeTokens, IERC721BridgeReceiver, BridgeTran
     )
         public
     {
-        _requestERC721Transfer(msg.sender, _from, _to, _tokenId, _extraData);
+        uint V1 = 1;
```
kjhman21, line 134: I think this is to be V2?

hyunsooda: This function is originally an existing function that I didn't touch except for an additional version parameter. The variable `V2` is located in `onERC721ReceivedV2()` function, which is called through `requestValueTransferV2()`.

contracts/bridge/BridgeTransferERC721.sol
hyunsooda: The `burn()` function reverts if the function fails. Could you elaborate If your concern is not this?

kjhman21: hmm.. yeah you're right. I thought that it can be open to reentrance attack, but it seems not.

contracts/bridge/BridgeTransferERC721.sol
kjhman21: `onERC721Received` 
This is a standard interface of ERC-721. So if we leave this as it is, this function is anyway called and V1 will be executed even on the updated contract.

https://eips.ethereum.org/EIPS/eip-721

hyunsooda: That's my expected usage. An end-developer should call `RequestValueTransferV2()` which calls `onERC721Received2()` if he/she want to emit newly added event which fixes URI-loss problem. I think your expected usage seems the other way against my expected usage. Do you want me to change that `onERC721Received()` would call `_requestERC721Transfer()` with version 2?

contracts/bridge/BridgeTransfer.sol
```diff
@@ -75,6 +75,31 @@ contract BridgeTransfer is BridgeHandledRequests, BridgeFee, BridgeOperator {
         bytes extraData
     );
 
+    /**
+     * Event to log the request value transfer from the Bridge.
+     * @param tokenType is the type of tokens (KLAY/ERC20/ERC721).
```
aidan-kwon, line 80: No description about `ver`

node/sc/bridge_manager_test.go
```diff
@@ -284,38 +319,59 @@ func TestBridgeManager(t *testing.T) {
...
-	// 11. Check KLAY balance
+	// 12. Check KLAY balance
 	{
 		balance, err = sim.BalanceAt(context.Background(), bob.From, nil)
 		assert.Equal(t, nil, err)
 		assert.Equal(t, testKLAY.String(), balance.String())
 	}
 
-	// 12. Check NFT owner
+	// 13. Check NFT owner
```
hqjang-pepper, line 365:
```suggestion
	// 13. Check NFT owner sent with RequestValueTransfer()
```

node/sc/bridge_manager_test.go
```diff
@@ -284,38 +319,59 @@ func TestBridgeManager(t *testing.T) {
...
 	{
-		owner, err := nft.OwnerOf(nil, big.NewInt(int64(nftTokenID)))
+		owner, err := nft.OwnerOf(nil, big.NewInt(int64(nftTokenID1)))
+		assert.Equal(t, nil, err)
+		assert.Equal(t, bob.From, owner)
+	}
+
+	// 14. Check NFT owner
```
hqjang-pepper, line 372:
```suggestion
	// 14. Check NFT owner sent with RequestValueTransferV2()
```

contracts/bridge/BridgeTransfer.sol
hyunsooda: Added.

contracts/bridge/BridgeTransferERC721.sol
hyunsooda: I think the `onERC721Received()` function should not call `requestValueTransferV2()` for a compatibility issue. Consider the following example.

- Mainbridge and subbridge nodes use the new version's binary and the old version's binary, respectively. (Or you are okay to think vice versa)
The new event will be emitted if the mainbridge node calls `requestValueTransfer()` on the ERC-721 contract. The subbridge node is not listening to the new event and nothing happens. Thus, I think we should keep the way of the current PR.

contracts/bridge/BridgeTransfer.sol
```diff
@@ -75,6 +75,32 @@ contract BridgeTransfer is BridgeHandledRequests, BridgeFee, BridgeOperator {
...
+        uint ver,
+        TokenType tokenType,
+        address indexed from,
+        address indexed to,
+        address indexed tokenAddress,
+        uint256 valueOrTokenId,
+        uint64 requestNonce,
+        uint256 fee,
+        bytes encodedEvent,
+        bytes extraData
```
kjhman21, line 101:
```suggestion
    event RequestValueTransferEncoded(
        TokenType tokenType,
        address indexed from,
        address indexed to,
        address indexed tokenAddress,
        uint256 valueOrTokenId,
        uint64 requestNonce,
        uint256 fee,
        bytes extraData,
        uint encodedVersion,
        bytes encodedData
```

hyunsooda:
@aidan-kwon, @kjhman21, @hqjang-pepper, @henry-will, @2dvorak, PTAL. Please read again the revised description too.

contracts/sc_erc721/ERC721ServiceChain.sol
```diff
@@ -42,7 +42,6 @@ contract ERC721ServiceChain is ERC721, Ownable {
 
     function requestValueTransfer(uint256 _uid, address _to, bytes calldata _extraData) external {
         transferFrom(msg.sender, bridge, _uid);
-
```
kjhman21, line 45: Do we really need this change?

node/sc/bridge_manager.go
```diff
@@ -12,7 +12,6 @@
 // GNU Lesser General Public License for more details.
 //
 // You should have received a copy of the GNU Lesser General Public License
-// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
```
kjhman21, line 15: What is the reason we remove this?

node/sc/bridge_manager.go
```diff
@@ -203,41 +211,49 @@ func (bi *BridgeInfo) GetCounterPartToken(token common.Address) common.Address {
...
 
 	logger.Trace("Get ready request value transfer event", "len(readyEvent)", len(ReadyEvent), "len(pendingEvent)", bi.pendingRequestEvent.Len())
 
-	for idx, ev := range ReadyEvent {
-		if ev.RequestNonce < bi.lowerHandleNonce || bi.handledEvent.Exist(ev.RequestNonce) {
-			logger.Trace("handled requests can be ignored", "RequestNonce", ev.RequestNonce, "lowerHandleNonce", bi.lowerHandleNonce)
-			continue
+	for idx, _ev := range ReadyEvent {
+		switch ev := _ev.(type) {
+		case *RequestValueTransferEvent:
```
kjhman21, line 236: It seems that we can make an interface to make this simple...

contracts/sc_erc721/ERC721ServiceChain.sol
hyunsooda: Is not it an unnecessary newline?

node/sc/bridge_manager.go
hyunsooda: What interface do you mean? The event struct is already interfaced.

node/sc/bridge_manager.go
hyunsooda: I've no idea why it was removed because I didn't touch it if I correctly remember it. Recovered. Thank you.

contracts/sc_erc721/ERC721ServiceChain.sol
kjhman21: The only change in this file is removing the newline.
In this case, I suggest you to make another PR to avoid any confusion of readers.

node/sc/bridge_manager.go
kjhman21: Interface of RequestValueTransferEvent and RequestValueTransferEncodedEvent.
Then, I think we don't need this switch statement.

hyunsooda: The concrete type conversion (type assertion) must proceed before do member access.

kjhman21: Yeah you can do that using the interface.

hyunsooda: 1. The event is interfaced.
2. The concrete type conversion is with switch stmt.

If my comments didn't answer your question, please describe what you are seeing in detail.

node/sc/vt_recovery.go
```diff
@@ -46,8 +46,8 @@ type valueTransferRecovery struct {
 
 	child2parentHint *valueTransferHint
 	parent2childHint *valueTransferHint
-	childEvents      []*RequestValueTransferEvent
-	parentEvents     []*RequestValueTransferEvent
+	childEvents      []interface{}
+	parentEvents     []interface{}
```
hqjang-pepper, line 50: Why did you change this VTRecovery struct?

node/sc/bridge_manager_test.go
hyunsooda: Thanks. Changed.

node/sc/vt_recovery.go
hyunsooda: Now that the number of the listening events is not the only one, I changed it to an interfaced structure.

contracts/sc_erc721/ERC721ServiceChain.sol
hyunsooda: Reverted

node/sc/subbridge.go
```diff
@@ -191,13 +193,14 @@ func NewSubBridge(ctx *node.ServiceContext, config *SCConfig) (*SubBridge, error
...
 		// txCh:            make(chan blockchain.NewTxsEvent, transactionChanSize),
-		requestEventCh: make(chan *RequestValueTransferEvent, requestEventChanSize),
-		handleEventCh:  make(chan *HandleValueTransferEvent, handleEventChanSize),
-		quitSync:       make(chan struct{}),
-		maxPeers:       config.MaxPeer,
-		onAnchoringTx:  config.Anchoring,
-		bootFail:       false,
-		rpcSendCh:      make(chan []byte),
+		chanReqVTev:        make(chan *RequestValueTransferEvent, chanReqVTevanSize),
+		chanReqVTencodedEv: make(chan *RequestValueTransferEncodedEvent, chanReqVTevanSize),
```
hqjang-pepper, line 197: Since now we support 2 kinds of requestVTEvent, if you create two channels like this, the total capacity will increase from 10,000 to 20,000. Is this intended?

node/sc/bridge_manager.go
```diff
@@ -271,25 +288,46 @@ func (bi *BridgeInfo) UpdateInfo() error {
...
+	case *RequestValueTransferEncodedEvent:
+		tokenType, tokenAddr, txHash, from, to = ev.TokenType, ev.TokenAddress, ev.Raw.TxHash, ev.From, ev.To
+		valueOrTokenId, requestNonce, blkNumber, extraData, contractAddr = ev.ValueOrTokenId, ev.RequestNonce, ev.Raw.BlockNumber, ev.ExtraData, ev.Raw.Address
+		verEv = uint(ev.EncodingVer.Uint64())
+	default:
+		return ErrUnknownEvent
+	}
+
+	ctpartTokenAddr := bi.GetCounterPartToken(tokenAddr)
 	// TODO-Klaytn-Servicechain Add counterpart token address in requestValueTransferEvent
```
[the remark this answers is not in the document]
hyunsooda: No. The counterpart token address is not in the event struct. If the counterpart token gets into the event, we need another work for encoding and decoding. The workaround is to add a counterpart token in that `if` block. Furthermore, `ctpartTokenAddr := bi.GetCounterPartToken(tokenAddr)` is wrong semantic. It always fails to get the counterpart token address.

node/sc/subbridge.go
hyunsooda: Two channels have different capacities. They have rooms for size 10000 independently. Could you explain what `total capacity is? The first channel is more used than the newly added channel as it is used for ERC-20 token transfer.

hqjang-pepper: In subbridge loop, it can process by receiving events from `chanReqVTev` and `chanReqVTencodedev` respectively. I understood that up to 20,000 requestVTevents that can be processed can be stored in runtime memory.

hyunsooda: The value 10000 came from the origin code. I didn't increase or decrease it.

TL;DR,
The incoming events are immediately processed in the loop you've figured out in `subbridge.go`. A channel blocking delay would be created if the channel capacity is lower. By the way, the current request event handler does not handle each event with a separate go-routine. Actually, the channel capacity seems not meaningless in the current implementation. I just came up with a new improvement from your comment. We may need a benchmark test for value transfer requests per second (VTRS) in a multi-thread manner. I left details about this new improvement at #1332.

hyunsooda:
@hqjang-pepper, @aidan-kwon, @kjhman21, Refactored. PTAL.

node/sc/bridge_manager.go
```diff
@@ -304,48 +296,42 @@ func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEv
...
-		}
-
-		uri, err = erc721.TokenURI(nil, ev.ValueOrTokenId)
-		if err != nil {
-			if err.Error() == vm.ErrExecutionReverted.Error() {
-				logger.Debug("Unable to get an ERC721 URI", "erc721", ev.TokenAddress.String(), "onParent", bi.onChildChain, "tokenId", ev.ValueOrTokenId.String())
-			} else {
-				return err
+		switch evType := ev.(type) {
+		case RequestValueTransferEvent:
```
kjhman21, line 312: @hyunsooda Can we also make this refactored using interface?

hyunsooda: I was going to do that, but others held me back to do that. That's why I reverted my initial things and changed to this PR. Could you guys @henry-will @hqjang-pepper explain?

henry-will: @hyunsooda @hqjang-pepper 
May I suggest my opinion with my code. 

https://github.com/henry-will/klaytn/commit/79d70c80a8df2b7ad297caaef606ccf2835fc139

hyunsooda: Please check to compile first. It's not compiled, even failed to run the test. Some of the suggestion for variable naming seems good to me. 

Please check below:
1. Check compilation and test pass.
   - Moreover, you need to test the compatibility issue I mentioned in the PR description. This is the top-most important thing to us. It's behind the test of this PR.
3. Understand how Ethereum implemented event watcher and channel pipeline for solidity binding. 
FYI, the reason why I created two event channels is because of type enforcement [here](https://github.com/klaytn/klaytn/blob/dev/event/feed.go#L146). 
4. Why did you negelect `encodingVer`? Without this, we have no idea how we should decode correctly.
5. Refrain from making a `errors.new(...)` dynamically.
6. Two interface functions `GetVersion()` and `GetUri()` are not implemented for `RequestValueTransferEvent` struct. [It's wrong](https://github.com/henry-will/klaytn/commit/79d70c80a8df2b7ad297caaef606ccf2835fc139#diff-187d76a03ba63a723bebb7fc7901a8d74ac492124c0a8ebb73ae3404c6feae67R312). It's better to make an empty function to reduce verbose type assertion.

henry-will: I am sorry for compile error. It is just for share my opinion to you, because it is hard to explain about difference IRequestValueTransferEvent interface from RequestValueTransferEncodedEvent struct.

In "case ERC721:", If RequestValueTransferEvent transfer is empty uri "", I think that it isn't good response. And such case does not happen. Is It for netmarble case?

```go
		case RequestValueTransferEvent:
			handleTx, err = bi.bridge.HandleERC721Transfer(auth, txHash, from, to, ctpartTokenAddr, valueOrTokenId, requestNonce, blkNumber, "", extraData)
```
EncodingVer is only 2 currently. When other version add later, we must change some code. It is possible to decode now. I think that ERC721 should have uri, but your code skip or pass uri with "".

I don't know when `RequestValueTransferEvent` need version and uri. I think that RequestValueTransferEncodedEvent need version and uri only.

hyunsooda: I added detailed comments in your commit. Please check it out. Before you refactor the interface, let me emphasize the important stuff again.

1. The code must be compilable and pass the test.
2. For the compatibility test, clone [this repo](https://github.com/klaytn/servicechain-value-transfer-examples) first and run ERC-20, ERC-721, and KLAY examples. The `build` directory contains the old contract's abi and bin. You must replace them with the uploaded ones in this PR because the old contract does not have a new event emit. The repository does not have a bidirectional value transfer example. There's only one direction from parent chain to child chain for the ERC-20 and ERC-721. You have to make another value transfer test from the child chain to the parent chain. Please leave it if you have a question for testing.

henry-will: It committed at [79284e5](https://github.com/klaytn/klaytn/pull/1321/commits/79284e5cd168a71f4eb4550b3b598fb3630d2d55)

node/sc/event_parse.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil
+	}
+}
+
+func GetURI(ev IRequestValueTransferEvent) string {
+	uri := ""
+	switch evType := ev.(type) {
+	case RequestValueTransferEncodedEvent:
+		decoded := UnpackEncodedData(evType.EncodingVer, evType.EncodedData)
+		uri = decoded["uri"].(string)
```
henry-will, line 52: What about check if decoded value is nil?

henry-will (changes requested):

node/sc/event_parse.go
hyunsooda: Nice find. Applied.

node/sc/event_interface.go
```diff
@@ -0,0 +1,96 @@
...
+
+//////////////////// type RequestValueTransferEvent struct ////////////////////
+// RequestValueTransferEvent from Bridge contract
+type RequestValueTransferEvent struct {
+	*bridgecontract.BridgeRequestValueTransfer
+}
+
+func (rEv RequestValueTransferEvent) Nonce() uint64 {
+	return rEv.RequestNonce
+}
```
aidan-kwon, line 33: It has both `Nonce` and `GetRequestNonce` methods, and the two methods return the same value.

hyunsooda: This originated from the origin codebase. The origin codebase used jumbly two methods. I didn't change it because I experienced some of the guys holding my changes back about wrong semantics in origin even though the change is logically right.

node/sc/bridge_manager_test.go
```diff
@@ -76,6 +76,44 @@ func CheckReceipt(b bind.DeployBackend, tx *types.Transaction, duration time.Dur
...
+		tokenType      = ev.GetTokenType()
+		valueOrTokenId = ev.GetValueOrTokenId()
+		from           = ev.GetFrom()
+		to             = ev.GetTo()
+		contractAddr   = ev.GetRaw().Address
+		tokenAddr      = ev.GetTokenAddress()
+		requestNonce   = ev.GetRequestNonce()
+		txHash         = ev.GetRaw().TxHash
+	)
+	fmt.Println("Request Event",
```
aidan-kwon, line 90: Can you use t.Log instead of fmt.Println? CI cannot control this out

node/sc/bridge_manager.go
```diff
@@ -203,41 +192,37 @@ func (bi *BridgeInfo) GetCounterPartToken(token common.Address) common.Address {
...
+		if ev.GetRequestNonce() < bi.lowerHandleNonce || bi.handledEvent.Exist(ev.GetRequestNonce()) {
+			logger.Trace("handled requests can be ignored", "RequestNonce", ev.GetRequestNonce(), "lowerHandleNonce", bi.lowerHandleNonce)
 			continue
 		}
 
 		if err := bi.handleRequestValueTransferEvent(ev); err != nil {
 			bi.AddRequestValueTransferEvents(ReadyEvent[idx:])
 			logger.Debug("Failed handle request value transfer event", "err", err, "len(RePutEvent)", len(ReadyEvent[idx:]))
-			return err
+			return
```
aidan-kwon, line 223: Does an error occur frequently or normally? If not, we better increase the log level from debug to warn because it is the last point the error is handled

hyunsooda: Rarely happen. How about changing it to `Error`?

node/sc/bridge_manager_test.go
hyunsooda: Before changing those outputs, could you elaborate on the relationship between the test output and CI? It just prints the info to ensure the correct values are filled in, not related to the expected output of the test. This question is to understand the need for change.

If the standard output matter to CI, we can simply remove all the print logs.

node/sc/subbridge.go
```diff
@@ -152,10 +152,12 @@ type SubBridge struct {
...
-	requestEventCh  chan *RequestValueTransferEvent
-	requestEventSub event.Subscription
-	handleEventCh   chan *HandleValueTransferEvent
-	handleEventSub  event.Subscription
+	chanReqVTev        chan RequestValueTransferEvent
+	chanReqVTencodedEv chan RequestValueTransferEncodedEvent
+	reqVTevSub         event.Subscription
+	reqVTencodedEvSub  event.Subscription
+	chanHandleVTev     chan *HandleValueTransferEvent
+	handleVTevSub      event.Subscription
```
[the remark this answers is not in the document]
hyunsooda: This is related to the camel case formatting and the semantics of the variable's name. If you think we need to change the variable's name, please suggest the names and the reasons together.

aidan-kwon (approved):
It looks good to me except for my small comments.

node/sc/bridge_manager.go
aidan-kwon: @hyunsooda Error is a possible option. If it makes some failures of value transfer requests permanently, I think Error is better. If we have another retrier of other back-up plans, Warn is okay for that.

node/sc/event_interface.go
aidan-kwon: Okay. Let's consider this later when we need to refine the code

node/sc/bridge_manager_test.go
aidan-kwon: Test output and CI doesn't different, but our CI show limited lines of logs during the test on the initial web page. If we print normal outputs too much, it is annoying when we find the real error logs.

aidan-kwon: So, it is not urgent to remove all stdout during the test. But, we better reduce them when we make a new test case.

hyunsooda: Okay. Replaced from `fmt.Println` to `t.Log` and `fmt.Printf` to `t.Logf`

node/sc/bridge_manager.go
hyunsooda: By call this `bi.AddRequestValueTransferEvents(ReadyEvent[idx:]), the event failed gets have a chance to be handled again. However, it is not always guaranteed to be successfully handled because of some reasons such as invalid nonce. It results in stacks of the pending transactions forever. Please refer to the fix in #1295. Logically, this is a failed behavior. So, I think `Error` seems reasonable.

henry-will (approved):

hqjang-pepper (dismissed):

aidan-kwon:
@kjhman21 PTAL

aidan-kwon (dismissed):

hqjang-pepper (approved):

aidan-kwon (approved):

henry-will (approved):

sirano11 (approved):

PR #1326, opened 2022-04-14
[SC] Fixed misleading value transfer log

hyunsooda:
## Proposed changes

The message had misleading log information in the previous value transfer logs (i.e., removed logs in this PR). The message `Bridge succeeded to ...` is incorrect because the right above function calls `bi.bridge.HandleERC20Transfer`, `bi.bridge.HandleKLAYTransfer`, and `bi.bridge.HandleERC721Transfer` just makes a transaction, not the transaction invokes. This PR considers receipt status received from the counterpart chain too. If the receipt is invalid, a status and transaction hash are pushed to logs.

This PR makes an observable value transfer path. The end-developer can find where the value transfer failed with the added and fixed log. Let me describe two examples.
1. Failed value transfer request from _child_ to _parent_
- The end-developer can figure out where the transaction is failed by tracking this log `logger.Error("Transaction's receipt sent from child chain is invalid", "status", receipt.Status, "txHash", txHash.String())`. This log says the failure happened from the parent chain and the transaction hash now can be searched on the parent chain.
2. Failed value transfer request from _parent_ to _child_
- The developer should search this log `logger.Trace("Bridge contract transaction is created", "VTDirection", "parent--->child", "contractCall", funcName, "nonce", reqNonce, "tx", txHash)` and it says the value transfer handle was failed from the child chain. Now what the developer should do is search the transaction hash on the child chain.
## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/sc/bridge_manager.go
+++ b/node/sc/bridge_manager.go
@@ -156,6 +156,18 @@ func NewBridgeInfo(sb *SubBridge, addr common.Address, bridge *bridgecontract.Br
 	return bi, nil
 }
 
+// handleValueTrnsferLog pushes value transfer transaction's log
+func handleValueTrnsferLog(isChild bool, funcName, txHash string, reqNonce uint64) {
+	// Note the `isChild` should be interpreted as reverse. Refer `ProcessRequestEvent()` in sub_event_handler.go
+	if isChild {
+		logger.Trace("Bridge contract transaction is created", "VTDirection", "parent--->child",
+			"contractCall", funcName, "nonce", reqNonce, "tx", txHash)
+	} else {
+		logger.Trace("Bridge contract transaction is created", "VTDirection", "child--->parent",
+			"contractCall", funcName, "nonce", reqNonce, "tx", txHash)
+	}
+}
+
 func (bi *BridgeInfo) loop() {
 	ticker := time.NewTicker(time.Second)
 	defer ticker.Stop()
@@ -308,14 +320,13 @@ func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEv
 		if err != nil {
 			return err
 		}
-		logger.Trace("Bridge succeeded to HandleKLAYTransfer", "nonce", ev.RequestNonce, "tx", handleTx.Hash().String())
-
+		handleValueTrnsferLog(bi.onChildChain, "handleKLAYTransfer", handleTx.Hash().String(), ev.RequestNonce)
 	case ERC20:
 		handleTx, err = bi.bridge.HandleERC20Transfer(auth, ev.Raw.TxHash, ev.From, ev.To, tokenAddr, ev.ValueOrTokenId, ev.RequestNonce, ev.Raw.BlockNumber, ev.ExtraData)
 		if err != nil {
 			return err
 		}
-		logger.Trace("Bridge succeeded to HandleERC20Transfer", "nonce", ev.RequestNonce, "tx", handleTx.Hash().String())
+		handleValueTrnsferLog(bi.onChildChain, "handleERC20Transfer", handleTx.Hash().String(), ev.RequestNonce)
 	case ERC721:
 		// get URI of the ERC721
 		var uri string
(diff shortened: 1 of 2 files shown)


node/sc/bridge_manager.go
```diff
@@ -156,6 +156,19 @@ func NewBridgeInfo(sb *SubBridge, addr common.Address, bridge *bridgecontract.Br
 	return bi, nil
 }
 
+// handleValueTrnsferLog pushes value transfer transaction's log
```
kjhman21, line 159:
```suggestion
// handleValueTransferLog pushes value transfer transaction's log
```

hyunsooda: Thanks. Fixed.

node/sc/bridge_manager.go
```diff
@@ -156,6 +156,19 @@ func NewBridgeInfo(sb *SubBridge, addr common.Address, bridge *bridgecontract.Br
 	return bi, nil
 }
 
+// handleValueTransferLog pushes value transfer transaction's log
+func handleValueTransferLog(isChild bool, funcName, txHash string, reqNonce uint64) {
```
kjhman21, line 160:
```suggestion
func handleValueTransferLog(toChild bool, funcName, txHash string, reqNonce uint64) {
```

What about changing 'isChild' to 'toChild'?

kjhman21: In the implementation of this function shows that this variable actually indicates the direction. What is the reason that you think that 'is' is more appropriate?

KimKyungup (dismissed):

node/sc/bridge_manager.go
kjhman21:
```suggestion
func handleValueTransferLog(onChild bool, funcName, txHash string, reqNonce uint64) {
```

What about this? It seems that call site uses the name of 'onChildChain'.

hyunsooda: Applied.

node/sc/sub_bridge_handler.go
```diff
@@ -344,6 +344,10 @@ func (sbh *SubBridgeHandler) writeServiceChainTxReceipts(bc *blockchain.BlockCha
 	for _, receipt := range receipts {
 		txHash := receipt.TxHash
 		if tx := sbh.subbridge.GetBridgeTxPool().Get(txHash); tx != nil {
+			if receipt.Status != types.ReceiptStatusSuccessful {
+				logger.Error("Transaction's receipt sent from child chain is invalid",
```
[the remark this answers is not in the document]
hyunsooda: Okay. Do you have other suggestions or concerns?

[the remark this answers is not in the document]
kjhman21: What do you mean by "removed"?

I understood that the failed txs will be resent and reexecuted until it succeeds, right?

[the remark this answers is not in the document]
hyunsooda: No. I was confused with the other stuff. The failed txs are not resent. The only txs in the bridge tx pool (a dedicated tx pool for value transfers) are resent. Otherwise, no tx is recent (e.g., removed tx from the pool). The current implementation unconditionally remove the tx as long the corresponding receipt arrives. We may change to resend failed tx, but those resent tx are not valid due to the low nonce. It leads to stack pending txs. Please refer to the fix at #1295.

FYI, there are two resending logics, but they seems meaningless. The first one was described above. The second one is triggered when making a tx for value transfer handle, not executing the tx. If the failed txs get to be handled again in the counterpart chain, the value transfer recovery is only one correct answer, which finds failed event and makes a new valid tx (i.e., no low nonce issue happened).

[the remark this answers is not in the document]
kjhman21: So, I see that the recovery logic should be revised.
This suggestion is just for the message, what do you think my suggestion? "invalid" is not appropriate based on your description?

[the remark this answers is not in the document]
hyunsooda: What do you mean? 
> So, I see that the recovery logic should be revised.

The suggestion is applied.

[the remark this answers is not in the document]
kjhman21: > What do you mean?
> 
> > So, I see that the recovery logic should be revised.

As you described, "there are two resending logics, but they seems meaningless.", so isn't it revised? of course this should be done in another PR.

I saw the the suggestion is applied. I resolve this comment.

node/sc/sub_bridge_handler.go
```diff
@@ -344,6 +344,10 @@ func (sbh *SubBridgeHandler) writeServiceChainTxReceipts(bc *blockchain.BlockCha
 	for _, receipt := range receipts {
 		txHash := receipt.TxHash
 		if tx := sbh.subbridge.GetBridgeTxPool().Get(txHash); tx != nil {
+			if receipt.Status != types.ReceiptStatusSuccessful {
+				logger.Error("A transaction sent from the child chain is failed",
```
ian0371, line 348:
```suggestion
				logger.Error("A transaction sent from the child chain failed",
```

hqjang-pepper (approved):

Issue #1334, opened 2022-04-18
Mining waiting time improvement


jiseongnoh:
**Is your request related to a problem? Please describe.**
Current version of Klaytn waits for one second if a block has been mined within a second interval, resulting in potential block generation delay.
For example, if a block was mined at 19.9 seconds, then the worker waits for one second, and starts the mining work at at 20.9 seconds, which is supposed to start at each (20.0) second. 

**Describe the solution you'd like**
I suggest the worker can wait for a fraction of seconds, so the next mining can be started at the scheduled timestamp.

**Describe alternatives you've considered**

**Additional context**

aidan-kwon:
Close this via https://github.com/klaytn/klaytn/pull/1340

PR #1340, opened 2022-04-20
Added dynamic waiting time between block generation

jiseongnoh:
## Proposed changes
Klaytn generates a block by one second interval.
When the network is not congested, the block generation time can be less than one second. 
In that case, block generator can pause mining a block for a moment in order to keep the one-second interval.
Currently, the waiting time is fixed to one second, resulting in potential block generation delay. 
For example, when a block was mined in 0.5 seconds, the worker waits for one second and starts the mining at 1.5 second.
The waiting period can be optimized while still keeping the block generation time.
  
I propose the waiting period can be [nextTimestampInteger - currentTimestamp], enabling dynamic waiting time as necessary to keep the one-second block generation interval.

I implemented a prototype of dynamic waiting time and tested block generation performance on a congested network environment (4000 transactions per second), and the result shows the block generation performance can increase up to 15 percent by applying the dynamic waiting time.

<meta charset="utf-8"><b style="font-weight:normal;">
Block generation performance result (50 CN, 4000tps)
  | original binary | dynamic waiting binary | enhancement
-- | -- | -- | -- 
50 committee | 0.825/sec | 0.927/sec | 12%
40 committee | 0.839/sec | 0.941/sec | 12% 
25 committee | 0.831/sec | 0.957/sec | 15%

</b>
 
Original Klaytn 1.8.2 log shows one second wait time like below:

```
INFO[04/13,11:18:07 +09] [24] Committed                                 number=33821 hash=552042…c489f7                                                      address=0xFeb4C00cA4Fe77e2D1382b8eE16B3e328aDa441A
INFO[04/13,11:18:07 +09] [5] Inserted a new block                      number=33821 hash=552042…c489f7                                                      txs=3469 gas=72849000  elapsed=419.480ms   processTxs=274.332ms finalize=27.005ms  validateState=8.182ms   totalWrite=62.070212ms  trieWrite=42.055712ms
INFO[04/13,11:18:07 +09] [50] Mining too far in the future             wait=1s
INFO[04/13,11:18:09 +09] [50] Commit new mining work                    number=33822 hash=e0bc23…f4f955                                                      txs=4126 elapsed=226.251ms   commitTime=191.521ms finalizeTime=34.725ms
```
The log applying the proposed feature shows dynamic waiting time like below:

```
INFO[04/11,13:52:48 +09] [24] Committed                                 number=27427 hash=7d3eca…b18ebb                                                      address=0xbc0014bA28F7dF2235bc1806bA8C702F7194983d
INFO[04/11,13:52:48 +09] [5] Inserted a new block                      number=27427 hash=7d3eca…b18ebb                                                      txs=4136 gas=86856000  elapsed=274.472ms    processTxs=133.115ms finalize=29.601ms  validateState=10.305ms  totalWrite=46.203502ms  trieWrite=24.746015ms
INFO[04/11,13:52:48 +09] [50] Mining too far in the future              wait=79.218ms
INFO[04/11,13:52:49 +09] [50] Commit new mining work                    number=27428 hash=a9ad61…8ea696                                                      txs=5942 elapsed=305.218ms    commitTime=257.074ms finalizeTime=48.135ms

```
This PR implements https://github.com/klaytn/klaytn/issues/1334
## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

- [] Lint and unit tests pass locally with my changes (`$ make test`)
- [] I have added necessary documentation (if appropriate)


work/worker.go
```diff
@@ -500,11 +500,11 @@ func (self *worker) commitNewWork() {
 	tstart := time.Now()
 	tstamp := tstart.Unix()
 	if self.nodetype == common.CONSENSUSNODE {
-		ideal := parent.Time().Int64() + params.BlockGenerationInterval
+		ideal := time.Unix(parent.Time().Int64()+params.BlockGenerationInterval, 0)
```
ehnuje, line 503: If you also use `TimeFoS`, calculation would be more precise.
https://github.com/klaytn/klaytn/blob/58581aa1c43324a5d8a2fa67477bdd03ab921ffb/consensus/istanbul/backend/engine.go#L398-L402

aidan-kwon (approved):
LGTM

work/worker.go
ehnuje: `ideal` is always underestimated unless we take `TimeFoS` into account, since `TimeFoS` is always a positive value. And we compare `tstart` and `ideal` to determine whether we take a sleep here or not, there is higher probability of taking a sleep when not reflecting `TimeFoS`.

- ideal w/o TimeFoS = 1000
- ideal w/   TimeFoS = 1100
- tstart                      = 1050

In this case, if we use TimeFoS, we can avoid sleeping but we need to take 50ms of sleep if we don't use `TimeFoS`. Please correct me if I misunderstood.

aidan-kwon: @ehnuje  In your case, I think it doesn't sleep w/ TimeFoS because `start` is before `ideal w/o TimeFoS` 
I think underestimating `ideal` is better to keep maintaining 1 second block time. 
Do I miss something?

jiseongnoh: > 

I also agree with Aidan's comment. I expect the block miner should wait as little as possible as well as the two different blocks have different timestamp. I believe removing the fraction of seconds can help reduce the waiting time. Please let me know if my idea is incorrect.

yoomee1313 (approved):

PR #1357, opened 2022-05-03
[SC] Try to retrieve handle token address from child chain

hyunsooda:
## Proposed changes

- Moved runtime token register to the time when an initial token register is called via bridge setup phase.
- Removed invalid bridge pair comparison. This comparison becomes invalid when a node is rebooted.
- The current handle retrieve logic asks to parent chain to retrieve the corresponding handle token address. This PR changes to try to retrieve the handle token address from its own subbridge chain.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/sc/api_bridge.go
+++ b/node/sc/api_bridge.go
@@ -331,8 +331,16 @@ func (sb *SubBridgeAPI) RegisterToken(cBridgeAddr, pBridgeAddr, cTokenAddr, pTok
 	if err != nil {
 		return err
 	}
+	err = cBi.RegisterToken(pTokenAddr, cTokenAddr)
+	if err != nil {
+		return err
+	}
 
 	err = pBi.RegisterToken(pTokenAddr, cTokenAddr)
+	if err != nil {
+		return err
+	}
+	err = pBi.RegisterToken(cTokenAddr, pTokenAddr)
 	if err != nil {
 		return err
 	}
@@ -396,13 +404,6 @@ func (sb *SubBridgeAPI) DeregisterToken(cBridgeAddr, pBridgeAddr, cTokenAddr, pT
 
 	if !cExist || !pExist {
 		return ErrNoBridgeInfo
-	}
-
-	pTokenAddrCheck := cBi.GetCounterPartToken(cTokenAddr)
-	cTokenAddrCheck := pBi.GetCounterPartToken(pTokenAddr)
-
-	if pTokenAddr != pTokenAddrCheck || cTokenAddr != cTokenAddrCheck {
-		return errors.New("invalid toke pair")
 	}
 
 	cBi.DeregisterToken(cTokenAddr, pTokenAddr)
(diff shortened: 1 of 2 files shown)


node/sc/api_bridge.go
```diff
@@ -398,13 +406,6 @@ func (sb *SubBridgeAPI) DeregisterToken(cBridgeAddr, pBridgeAddr, cTokenAddr, pT
 		return ErrNoBridgeInfo
 	}
 
-	pTokenAddrCheck := cBi.GetCounterPartToken(cTokenAddr)
-	cTokenAddrCheck := pBi.GetCounterPartToken(pTokenAddr)
-
-	if pTokenAddr != pTokenAddrCheck || cTokenAddr != cTokenAddrCheck {
-		return errors.New("invalid toke pair")
-	}
-
```
hqjang-pepper, line 407: I think removing this code is unnecessary. Can you explain why you removed this code?

node/sc/bridge_manager.go
```diff
@@ -270,26 +271,41 @@ func (bi *BridgeInfo) UpdateInfo() error {
...
 }
 
-// handleRequestValueTransferEvent handles the given request value transfer event.
-func (bi *BridgeInfo) handleRequestValueTransferEvent(ev *RequestValueTransferEvent) error {
-	tokenType := ev.TokenType
-	tokenAddr := bi.GetCounterPartToken(ev.TokenAddress)
-	// TODO-Klaytn-Servicechain Add counterpart token address in requestValueTransferEvent
-	if tokenType != KLAY && tokenAddr == (common.Address{}) {
-		logger.Warn("Unregistered counter part token address.", "addr", tokenAddr.Hex())
-		ctTokenAddr, err := bi.counterpartBridge.RegisteredTokens(nil, ev.TokenAddress)
```
hqjang-pepper, line 280: Okay you removed the runtime token registering logic, and just return error when something goes wrong. But is it sure this error does not occur when we fix initial token register logic? I think we need to do some test for this.

node/sc/api_bridge.go
hyunsooda: The function `GetCounterPartToken()` lookup the `counterpartToken` property in `bridgeInfo` struct. The `counterpartToken` member is initialized when the function `RegisterToken()` is called. The `RegisterToken()` stands for recording the counterpart token address. If a node is rebooted, and no `RegisterToken()` is not called, `counterpartToken` becomes empty. Thus, the checker above is invalid.

node/sc/bridge_manager.go
hyunsooda: The returns that return errors are actually fatal errors, meaning that the handle value transfer for the two types of the token must not be correctly handled. Please compare the previous code and this new change. You can find the key difference that a handle token address was retrieved from where. The handle token address was retrieved by asking for the counterpart chain's contract, not the own chain's contract. It opens a gateway to be victim's chain handles unintended value transfer events, although the unknown token contract has not been registered before.

hqjang-pepper (approved):

hyunsooda:
An attacker can simply deploy the modified contract that emits the originally paired parent's contract by hardcoded address. A contract-based verification cannot be a solution to detect the integrity of emitted events.

PR #1367, opened 2022-05-11
Remove vote when checkVote() fails in HandleGovernanceVote

ian0371:
## Proposed changes

- Adding an existing validator via `governance.vote("governance.addvalidator", existing_val_addr)` has a bug. Refer to the log below.
- This is caused by https://github.com/klaytn/klaytn/blob/dev/governance/handler.go#L365 where `checkVote()` checks if the address to be added already exists in the validator set, in which case it returns WITHOUT REMOVING THE VOTE.
- This fix removes the vote when `checkVote()` fails. (remove in this context means setting Casted to true)

### Bug log
- Initial setting
For an initial council, we remove one validator so that we can add it back.

```
> klay.getCouncil()
["0x19c032589b76ac395d3a713088e67f9ce9edf1c6", "0x353a5fd7cd9b8d0fe3e7c65d6c3d229cd0f14533", "0xd4ca4f3f2d80fb5b2f569df45d4c28a98f48a289", "0xe0c91bc72396b6c1e937ef717bf59a90c558f285"]
> addr = klay.getCouncil()[0]
"0x19c032589b76ac395d3a713088e67f9ce9edf1c6"
> governance.vote("governance.removevalidator", addr)
"Your vote is prepared. It will be put into the block header or applied when your node generates a block as a proposer."
> governance.myVotes
[{
    BlockNum: 6,
    Casted: true,
    Key: "governance.removevalidator",
    Value: "0x19c032589b76ac395d3a713088e67f9ce9edf1c6"
}]
```

- add validator (1)

```
> governance.vote("governance.addvalidator", addr)
"Your vote is prepared. It will be put into the block header or applied when your node generates a block as a proposer."
> governance.myVotes
[{
    BlockNum: 6,
    Casted: true,
    Key: "governance.removevalidator",
    Value: "0x19c032589b76ac395d3a713088e67f9ce9edf1c6"
}, {
    BlockNum: 16,
    Casted: true,
    Key: "governance.addvalidator",
    Value: "0x19c032589b76ac395d3a713088e67f9ce9edf1c6"
}]
```

- add validator (2) with an existing address

```
> governance.vote("governance.addvalidator", addr)
"Your vote is prepared. It will be put into the block header or applied when your node generates a block as a proposer."
> governance.myVotes
[{
    BlockNum: 0,
    Casted: false,
    Key: "governance.addvalidator",
    Value: "0x19c032589b76ac395d3a713088e67f9ce9edf1c6"
}, {
    BlockNum: 6,
    Casted: true,
    Key: "governance.removevalidator",
    Value: "0x19c032589b76ac395d3a713088e67f9ce9edf1c6"
}]
```

We can see that the second vote is not cast and it keeps appearing in every block the voter becomes the proposer.

```
> klay.getBlock(16).voteData
"0xf84294353a5fd7cd9b8d0fe3e7c65d6c3d229cd0f1453397676f7665726e616e63652e61646476616c696461746f729419c032589b76ac395d3a713088e67f9ce9edf1c6"
> klay.getBlock(22).voteData
"0xf84294353a5fd7cd9b8d0fe3e7c65d6c3d229cd0f1453397676f7665726e616e63652e61646476616c696461746f729419c032589b76ac395d3a713088e67f9ce9edf1c6"
> klay.getBlock(25).voteData
"0xf84294353a5fd7cd9b8d0fe3e7c65d6c3d229cd0f1453397676f7665726e616e63652e61646476616c696461746f729419c032589b76ac395d3a713088e67f9ce9edf1c6"
> klay.getBlock(28).voteData
"0xf84294353a5fd7cd9b8d0fe3e7c65d6c3d229cd0f1453397676f7665726e616e63652e61646476616c696461746f729419c032589b76ac395d3a713088e67f9ce9edf1c6"
```

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/consensus/istanbul/backend/engine_test.go
+++ b/consensus/istanbul/backend/engine_test.go
@@ -75,6 +75,18 @@ func makeCommittedSeals(hash common.Hash) [][]byte {
 	return committedSeals
 }
 
+// Include a node from the global nodeKeys and addrs
+func includeNode(addr common.Address, key *ecdsa.PrivateKey) {
+	for _, a := range addrs {
+		if a.String() == addr.String() {
+			// already exists
+			return
+		}
+	}
+	nodeKeys = append(nodeKeys, key)
+	addrs = append(addrs, addr)
+}
+
 // Exclude a node from the global nodeKeys and addrs
 func excludeNodeByAddr(target common.Address) {
 	for i, a := range addrs {
@@ -1057,6 +1069,40 @@ func TestSnapshot_Validators_AddRemove(t *testing.T) {
 				checkpointInterval + 9: {[]int{0, 1, 2, 3}},
 			},
 		},
+		{ // multiple addvalidator & removevalidator
+			10,
+			map[int]vote{
+				0: {"governance.removevalidator", 3},
+				2: {"governance.addvalidator", 3},
+				4: {"governance.addvalidator", 3},
+				6: {"governance.removevalidator", 3},
+				8: {"governance.removevalidator", 3},
+			},
+			map[int]expected{
+				1: {[]int{0, 1, 2}},
+				3: {[]int{0, 1, 2, 3}},
+				5: {[]int{0, 1, 2, 3}},
+				7: {[]int{0, 1, 2}},
+				9: {[]int{0, 1, 2}},
+			},
+		},
+		{ // multiple addvalidators & removevalidators
+			10,
+			map[int]vote{
+				0: {"governance.removevalidator", []int{2, 3}},
+				2: {"governance.addvalidator", []int{2, 3}},
+				4: {"governance.addvalidator", []int{2, 3}},
+				6: {"governance.removevalidator", []int{2, 3}},
+				8: {"governance.removevalidator", []int{2, 3}},
+			},
+			map[int]expected{
+				1: {[]int{0, 1}},
+				3: {[]int{0, 1, 2, 3}},
+				5: {[]int{0, 1, 2, 3}},
+				7: {[]int{0, 1}},
+				9: {[]int{0, 1}},
+			},
+		},
 	}
 
 	var configItems []interface{}
@@ -1115,8 +1161,7 @@ func TestSnapshot_Validators_AddRemove(t *testing.T) {
 				}
 				if v.key == "governance.addvalidator" {
 					for _, i := range indices {
-						nodeKeys = append(nodeKeys, allNodeKeys[i])
-						addrs = append(addrs, allAddrs[i])
+						includeNode(allAddrs[i], allNodeKeys[i])
 					}
 				}
 				if v.key == "governance.removevalidator" {
--- a/consensus/istanbul/backend/engine_test.go
+++ b/consensus/istanbul/backend/engine_test.go
@@ -1086,6 +1086,21 @@ func TestSnapshot_Validators_AddRemove(t *testing.T) {
 				9: {[]int{0, 1, 2}},
 			},
 		},
+		{ // multiple removevalidator & addvalidator
+			10,
+			map[int]vote{
+				0: {"governance.removevalidator", 3},
+				2: {"governance.removevalidator", 3},
+				4: {"governance.addvalidator", 3},
+				6: {"governance.addvalidator", 3},
+			},
+			map[int]expected{
+				1: {[]int{0, 1, 2}},
+				3: {[]int{0, 1, 2}},
+				5: {[]int{0, 1, 2, 3}},
+				7: {[]int{0, 1, 2, 3}},
+			},
+		},
 		{ // multiple addvalidators & removevalidators
 			10,
 			map[int]vote{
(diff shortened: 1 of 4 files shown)


blukat29:
@ian0371 Can you add a scenario test case in `TestSnapshot_Validators_AddRemove`? (https://github.com/klaytn/klaytn/blob/dev/consensus/istanbul/backend/engine_test.go#L995)
For instance, addval -> addval -> removeval -> removeval case.

I see that unit test for HandleGovernanceVote checks crucial conditions. But having an intuitive scenario as a test would be more comprehensible.

governance/handler.go
```diff
@@ -360,17 +360,33 @@ func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes
...
+							return false, true
+						}
+					}
+				} else {
+					// invalid value type
+					return true, false
 				}
-			} else {
+				return true, true
+			}
```
blukat29, line 381: Can we keep the original if statement? It seems adding `checkValidator` does not add code clarity. It's rather overcomplicated to me.

governance/handler.go
```diff
@@ -360,17 +360,33 @@ func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes
...
+
+			if checkVotePass, typeAssertPass := checkValidator(); !checkVotePass {
+				// this vote is adding existing validator or removing nonexistent validator
+				if self == proposer {
+					gov.removeDuplicatedVote(gVote, header.Number.Uint64())
+				}
```
blukat29, line 387: If you intended to remove duplication, how about defining below function?

```
removeCastedMyVote := func() {
    if self == proposer {
        gov.removeDuplicatedVote(gVote,
    }
}
```

hyunsooda:
Let me ask a simple question. Is it invalid removing duplicated vote immediately from [Vote()](https://github.com/klaytn/klaytn/blob/dev/governance/api.go#L84)? Why did you delegate it to `HandleGovernanceVote()`?

hyunsooda:
Each independent test case seems separated by comment. It looks better to make an arbitrary scope case by case or split them into functions. You may consider this later.

ian0371:
@hyunsooda
> Let me ask a simple question. Is it invalid removing duplicated vote immediately from [Vote()](https://github.com/klaytn/klaytn/blob/dev/governance/api.go#L84)? Why did you delegate it to `HandleGovernanceVote()`?

Good question. Validator set can be fetched from istanbul.snapshot, which is inaccessible from governance package (istanbul/backend package imports governance, so the reverse cannot be done. `HandleGovernanceVote()` receives valset as a parameter for similar reasons).

> Each independent test case seems separated by comment. It looks better to make an arbitrary scope case by case or split them into functions. You may consider this later.

I agree that the current readability is not so great. I'll take this into account later :)

hyunsooda:
Okay. then, does not have a side-effect of this post-processing?

ian0371:
> does not have a side-effect of this post-processing?

No, it does not have a side effect. The removing vote is rarely executed in the real world (i.e., when adding an existing validator).

blukat29 (approved):

hyunsooda:
Okay.. thank you.

yoomee1313 (approved):

consensus/istanbul/backend/engine_test.go
```diff
@@ -1115,8 +1161,7 @@ func TestSnapshot_Validators_AddRemove(t *testing.T) {
 				}
 				if v.key == "governance.addvalidator" {
 					for _, i := range indices {
-						nodeKeys = append(nodeKeys, allNodeKeys[i])
-						addrs = append(addrs, allAddrs[i])
+						includeNode(allAddrs[i], allNodeKeys[i])
```
ehnuje, line 1164: Why are these lines changed?

governance/handler.go
```diff
@@ -362,11 +362,17 @@ func (gov *Governance) HandleGovernanceVote(valset istanbul.ValidatorSet, votes
 			authorize := key == params.AddValidator
 			if addr, ok := gVote.Value.(common.Address); ok {
 				if !gov.checkVote(addr, authorize, valset) {
+					if proposer == self {
+						gov.removeDuplicatedVote(gVote, header.Number.Uint64())
```
ehnuje, line 366: How about leaving a log when removing duplicated votes?

consensus/istanbul/backend/engine_test.go
ian0371: Without them, the same node ends up twice in `nodeKeys`/`addrs`, resulting in "invalid committed seals".

governance/handler.go
ian0371: Added.

ian0371:
Rebased to the latest dev

governance/default_test.go
```diff
@@ -803,6 +803,23 @@ func TestGovernance_HandleGovernanceVote_None_mode(t *testing.T) {
 	}
 	gov.voteMap.Clear()
 
+	//////////////////////////////////////////////////////////////////////////////////////////////////////////
+	// Test adding an existing validator
```
zacscoding, line 807: How about adding remove not exist validator test?

ian0371: Added. PTAL

consensus/istanbul/backend/engine_test.go
ehnuje: @ian0371 What I meant to say was that changing the test code because of the core code change looks strange.
Do we need to sync the test code logic and the core code logic manually for this test?

ian0371: @ehnuje  Changing these lines is inevitable because the existing code cannot handle the new test cases. You can consider it "fixing a bug". 
I could add a new test function to verify them, but instead, I decided to reuse the existing function because most parts (i.e., except for these lines) will be reused. You might not want to go down that road.

blukat29 (approved):

zacscoding (approved):
Thanks @ian0371  👍

PR #1369, opened 2022-05-12
[RPC] Privileged APIs

hyunsooda:
## Proposed changes
- A feature, Privileged APIs(service), that does not expose specific APIs is implemented in this change.
- Privilege APIs are only available on IPC.
## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


blukat29:
How about saying "Protected API" or "Privileged API" instead of "Banned API"?

blukat29:
Is the purpose of new `Server.scheme` field is to selectively apply the ban list?
Why don't you let the callers of `RegisterName()` determine that?

For instance, in HTTPServer, the ban list must be applied. Therefore,

```go
func StartHTTPEndpoint(...) {
    handler := NewServer()
    for _, api := range apis {
        handler.RegisterName(api.Namespace, api.Service, api.Ban) // enforce api ban
    }
}
```

In InProc (IPC) server, the ban list can be ignored. Therefore,

```go
const NoBan = []uintptr{}

func (n *Node) startInProc(apis []rpc.API) error {
    handler := rpc.NewServer()
    for _, api := range apis {
        logger.Warn("Opening privileged APIs for IPC endpoint")
        handler.RegisterName(api.Namespace, api.Service, rpc.NoBan) // Explicitly does not ban
    }
}
```

This way, you don't have to add a field in `Server` struct. Please consider this direction.
If there is no other usage of `Server.scheme`, then adding the field could be an overkill.

networks/rpc/server.go
```diff
@@ -145,6 +146,17 @@ func (s *Server) RegisterName(name string, rcvr interface{}) error {
...
 
+	if s.scheme != IPCServer {
+		for _, api := range apiBans {
+			for name, m := range methods {
+				if api == m.method.Func.Pointer() {
+					logger.Trace("The banned API is not registered", "API", name, "scheme", SchemeToString(s.scheme))
+					delete(methods, name)
+				}
+			}
+		}
```
blukat29, line 157: Can we use a map instead of double for-loops?

hyunsooda:
> How about saying "Protected API" or "Privileged API" instead of "Banned API"?
> 
> "Banned" sounds like some users are banned from an API. What you want here is preventing remote access to some important RPCs in security viewpoint.

Cool. Applied.

networks/rpc/server.go
hyunsooda: Refactored. PTAL again.

node/sc/subbridge.go
```diff
@@ -300,6 +302,13 @@ func (sb *SubBridge) APIs() []rpc.API {
 			Service:   sb.APIBackend,
 			Public:    true,
 		},
+		{
+			Namespace:  "subbridge",
+			Version:    "1.0",
+			Service:    sb.PrivilegedAPIBackend,
+			Privileged: true,
+			Public:     true,
```
blukat29, line 310: What is the difference between `Public` and `Privileged`?
It seems there is a feature where an API is only exposed via IPC if `api.Public == false`.
https://github.com/klaytn/klaytn/blob/dev/networks/rpc/endpoints.go

hyunsooda:
> Is the purpose of new `Server.scheme` field is to selectively apply the ban list? Why don't you let the callers of `RegisterName()` determine that?
> 
> For instance, in HTTPServer, the ban list must be applied. Therefore,
> 
> ```go
> func StartHTTPEndpoint(...) {
>     handler := NewServer()
>     for _, api := range apis {
>         handler.RegisterName(api.Namespace, api.Service, api.Ban) // enforce api ban
>     }
> }
> ```
> 
> In InProc (IPC) server, the ban list can be ignored. Therefore,
> 
> ```go
> const NoBan = []uintptr{}
> 
> func (n *Node) startInProc(apis []rpc.API) error {
>     handler := rpc.NewServer()
>     for _, api := range apis {
>         logger.Warn("Opening privileged APIs for IPC endpoint")
>         handler.RegisterName(api.Namespace, api.Service, rpc.NoBan) // Explicitly does not ban
>     }
> }
> ```
> 
> This way, you don't have to add a field in `Server` struct. Please consider this direction. If there is no other usage of `Server.scheme`, then adding the field could be an overkill.

`Ban` is now removed. Say we pass an arbitrary identifier of server scheme instead of writing it to struct, we should keep track of this hardcoded value. Another problem is that there's no relationship between the caller's scope of `RegisterName` and the arbitrarily hardcoded value. It may cause error-prone writing of wrong value and must check the context near that caller. Instead, I delegated this record to the time of server initialization.

node/sc/subbridge.go
hyunsooda: There's another condition near that `api.Public`. See (https://github.com/klaytn/klaytn/blob/dev/networks/rpc/endpoints.go#L37). The `whitelist` value must be empty not to register the APIs. The `whitelist` becomes empty if no API is given in the option of `RPC_LIST` in `*.conf`. Thus, this intention is to enable the namespace of APIs that sets `Public` to true if there's no any given pre-set RPCs.

blukat29: Understood. Node operator can choose to expose non-`Public` RPCs with `RPC_LIST`. But we want to absolutely prohibit exposing `Privileged` RPCs whatsoever.

networks/rpc/endpoints.go
```diff
@@ -32,10 +32,10 @@ func StartHTTPEndpoint(endpoint string, apis []API, modules []string, cors []str
 		whitelist[module] = true
 	}
 	// Register all the APIs exposed by the services
-	handler := NewServer()
+	handler := NewServer(HTTPServer)
 	for _, api := range apis {
 		if whitelist[api.Namespace] || (len(whitelist) == 0 && api.Public) {
-			if err := handler.RegisterName(api.Namespace, api.Service); err != nil {
```
blukat29, line 38: I'm still not convinced about the `s.scheme != IPCServer` check in RegisterName.
Can't we simply add `api.Privileged` condition to the existing if statement?

```go
handler := NewServer()
for _, api := range apis {
    if !api.Privileged && ( whitelist[api.Namespace] || (len(whitelist) == 0 && api.Public) ) {
        if err := handler.RegisterName(api.Namespace, api.Service); err != nil {
            // ...
```

node/sc/subbridge.go
hyunsooda: Yes. The privileged APIs in the specific namespace of RPC are not exposed to HTTP scheme, FastHTTP scheme, gRPC scheme, WS scheme, and FastWS scheme, but are allowed to only IPC scheme. 

* RPC_API (Not `RPC_LIST`. I'm sorry I confused the option name)

networks/rpc/endpoints.go
hyunsooda: For the `s.scheme != IPCServer`, please refer to the description of the PR. It is for the description. It is to allow the privileged RPC to only IPC. 

Now the condition for privileged value is located inside of the `RegisterName`. This is to delegate that mandatory check to the function, not the caller's scope. Please retake a look at the new changes. I hope to hear your feedback again about the new changes.

blukat29: If I understand correctly, it boils down to "if not InProc && api.Privileged, skip RegisterName". My conjecture is that the logic can be implemented by conditionally not calling `RegisterName`.

We can insert `if !api.Privileged` condition (1) only in StartHTTPEndpoint, StartFastHTTPEndpoint, StartWSEndpoint, StartFastWSEndpoint, startgRPC, (2) but not in StartIPCEndpoint, startInProc. Just like how `whitelist + api.Public` check is done now.

I think the `api.Public` check and `api.Privileged` check are very similar access controls - the names, purposes. Is there a reason why those two should be treated differently?

- Do you think the `Privileged` checks must be done inside `RegisterName`? 
- Do you think the `whitelist + Public` checks must be done inside `RegisterName`?
- Is there any other `RegisterName` call sites, where server scheme cannot be determined?
  - `MainBridge.SetComponents` seems to be the case. But then you set `mb.rpcServer = rpc.NewServer(rpc.InProcServer)`. Which means `MainBridge.SetComponents` is safe to ignore `api.Privileged` flag.

ian0371: I second blukat. I sense that you're worried about developers forgetting to check api.Privileged flag, but this seems like the tail is wagging the dog. A warning comment at `RegisterName` would suffice.

networks/rpc/utils.go
```diff
@@ -247,3 +247,38 @@ func sanitizeTimeouts(timeouts HTTPTimeouts) HTTPTimeouts {
...
+		return "WS"
+	case FastWSServer:
+		return "FastWS"
+	case GRPCServer:
+		return "GRPC"
+	case InProcServer:
+		return "InProc"
+	case IPCServer:
+		return "IPC"
+
```
ian0371, line 267: remove the line plz

networks/rpc/utils.go
```diff
@@ -247,3 +247,38 @@ func sanitizeTimeouts(timeouts HTTPTimeouts) HTTPTimeouts {
 
 	return timeouts
 }
+
+func SchemeToString(scheme ServerScheme) string {
```
ian0371, line 251: How about managing this with map[ServerScheme]string?

hqjang-pepper (approved):

networks/rpc/endpoints.go
hyunsooda: Before putting my comment, let me correct your repeated incorrect name. 

- `InProc` is not `IPC`. `IPC` is to communicate via file scoket (`unix` or `window`).  `InProc` uses `process pipe` between two channels(pipe)`. 
-  Your condition is correct (`IPC && api.Privileged`).
- `Public` is imported from Ethereum, not me. The purpose of this property is described above in my comment. Thus, two properties are definitely different semantics.

Okay, now let me answer one by one:
> Do you think the Privileged checks must be done inside RegisterName?

The refactored code removes a few of lines that must appear before calling the `RegisterName()` in every call sites. Think logically, the privileged option was created to filter unprivileged RPCs. I think it should be handled by registration phase, which is a `RegisterName`. The validation in the caller seems not correct logically.

> Do you think the whitelist + Public checks must be done inside RegisterName? 

Sure, but I didn't touch the existing code for now.

> Is there any other RegisterName call sites, where server scheme cannot be determined?

The server scheme can be determined since the scheme parameter is passed to server initialization (`NewServer()`). The server scheme and privileged option belong to a different contexts. The scheme is checked inside of `RegisterName()`, and the privileged option is passed to `RegisterName()`. Please refer [1](https://github.com/klaytn/klaytn/pull/1369/files#diff-a0f8913808afa1e438d2664d0dbb0929a1ebc18873fb323b51f1f4a39e1d51c9R151) and [2](https://github.com/klaytn/klaytn/pull/1369/files#diff-a0f8913808afa1e438d2664d0dbb0929a1ebc18873fb323b51f1f4a39e1d51c9R242).

networks/rpc/utils.go
hyunsooda: Is there any good stuff by creating a global map instead of a function call?

networks/rpc/utils.go
hyunsooda: Do you mean the below?
```
case HTTPServer:
    return "HTTP"
...
```
If so, the line insert seems to be enforced by gofmt.

hyunsooda: Oh, I checked what you mean. Removed. Thank you

networks/rpc/utils.go
ian0371: I guess performance? (function call overhead vs. map lookup) But I'm not sure if it's significant, so if you don't agree, you can disregard this comment.

hyunsooda: Okay. the performance seems not to matter in my view. Function seems more flexible to extend (e.g., log invalid input if the parameter is not an expected type or value). Let me keep it as function.

networks/rpc/endpoints.go
blukat29: > The refactored code removes a few of lines that must appear before calling the RegisterName() in every call sites. Think logically, the privileged option was created to filter unprivileged RPCs. I think it should be handled by registration phase, which is a RegisterName. The validation in the caller seems not correct logically.

If you want to prevent someone accidentally missing the `Privileged` check, then I understand why the checks should be inside `RegisterName`.

> Sure, but I didn't touch the existing code for now.

Okay, I agree that we'd better keep the existing code for now.

> The server scheme can be determined since the scheme parameter is passed to server initialization (NewServer()). The server scheme and privileged option belong to a different contexts. The scheme is checked inside of RegisterName(), and the privileged option is passed to RegisterName(). Please refer [1](https://github.com/klaytn/klaytn/pull/1369/files#diff-a0f8913808afa1e438d2664d0dbb0929a1ebc18873fb323b51f1f4a39e1d51c9R151) and [2](https://github.com/klaytn/klaytn/pull/1369/files#diff-a0f8913808afa1e438d2664d0dbb0929a1ebc18873fb323b51f1f4a39e1d51c9R242).

I wanted to say if the check is outside `RegisterName`, then we won't have to pass the `scheme` to `NewServer()`. But for the check to be inside `RegisterName`, of course I see that `scheme` should be passed. 

Personally I would prefer conciseness in this particular case. I think the risk of missing the check is small because there are only handful number of call sites and we can review the changes. To me, changes are too large compared to the feature.

However, being careful is also important. Therefore I am okay with your changes.

blukat29 (dismissed):

networks/rpc/endpoints.go
hyunsooda: Yes. Of course, the accidentally missing condition is also one of the concerns of my changes. However, the main purpose is to decide the logically correct place where the `priviliged` flag is looked up. By delegating to the function, the caller does not take care of any consideration. The function takes care of validation.

nohkwak:
Please understand that I may ask questions because my understanding is not deep. 
I wonder why you implement startInProc to cmd/kscn/node.go, but to cmd/kbn/node.go.

networks/rpc/types.go
```diff
@@ -34,12 +34,26 @@ import (
...
 type API struct {
-	Namespace string      // namespace under which the rpc methods of Service are exposed
-	Version   string      // api version for DApp's
-	Service   interface{} // receiver instance which holds the methods
-	Public    bool        // indication if the methods must be considered safe for public use
+	Namespace  string      // namespace under which the rpc methods of Service are exposed
+	Version    string      // api version for DApp's
+	Service    interface{} // receiver instance which holds the methods
+	Public     bool        // indication if the methods must be considered safe for public use
+	Privileged bool
```
nohkwak, line 56: If you could comment simply what privilieged means, it would be nice for people looking at the code later.

hyunsooda: Oops, Thank you. Added

blukat29 (approved):

hqjang-pepper (approved):

hqjang-pepper:
@hyunsooda LGTM, but you need to rebase your branch from klaytn:dev in order to resolve linter issue.

hyunsooda:
pending

Issue #1364, opened 2022-05-10
tx pending on fee delegate valueTransfer( when sender = fee payer)


JayChoi1736:
**Describe the bug**
tx is stuck in tx pool ( pending state)

**How to reproduce**
send fee delegate tx ( sender :=A , receiver :=B, fee payer :=A, gas limit := x, transferred value := y, A's balance := z)
when x+y > z and x,y < z 

**Expected behavior**
print insufficient error

**Attachments**
If applicable, add data, logs or screenshots to help explain your problem.

**Environment (please complete the following information)**
 - Klaytn version or git revision that exhibits the issue `$ ken version`
 - go version `$ go version`
 - OS and its version [e.g. macOS/10.14]

**Additional context**

PR #1372, opened 2022-05-13
add check logic on delegate tx

JayChoi1736:
## Proposed changes

- add check logic on txpool when sender = feepayer

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

- issue #1364 

## Further comments


hqjang-pepper (dismissed):

aidan-kwon (commented):
Great!
Could you consider adding the same validation in https://github.com/klaytn/klaytn/blob/9f2f3f9c12f243b714a9c05cfbe727736d9a6c50/blockchain/tx_list.go#L313 ?
The balance shoulde be checked affter the txpool additon. With each block insertion, txpool examine all transactions in pending/queue txs again. Check https://github.com/klaytn/klaytn/blob/9f2f3f9c12f243b714a9c05cfbe727736d9a6c50/blockchain/tx_pool.go#L362

blockchain/tx_pool.go
```diff
@@ -739,6 +739,11 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 				return ErrInsufficientFundsFeePayer
 			}
 		}
+		// balance check in case of sender = feepayer
```
yoomee1313, line 742: It would be helpful if you add a reason why this additional sender balance check is needed. What about this?
```suggestion
		// additional sender balance check in case of sender = feepayer
		// since a single account has to bear the both cost(feepayer_cost + sender_cost),
		// it is necessary to check whether the balance is equal to the sum of the cost.
```

JayChoi1736: Thx!

ehnuje (commented):
How about adding a test case for your change?

blockchain/tx_list.go
```diff
@@ -310,9 +310,10 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
 // a point in calculating all the costs or if the balance covers all. If the threshold
 // is lower than the costcap, the caps will be reset to a new high after removing
 // the newly invalidated transactions.
-func (l *txList) Filter(senderBalance *big.Int, pool *TxPool) (types.Transactions, types.Transactions) {
+func (l *txList) Filter(sender common.Address, pool *TxPool) (types.Transactions, types.Transactions) {
 	// Filter out all the transactions above the account's funds
 	removed := l.txs.Filter(func(tx *types.Transaction) bool {
+		senderBalance := pool.getBalance(sender)
```
kjhman21, line 316: Need to check performance. I think this is called every block head is received.

kjhman21: Oh, it is already called every time. no issue.

blockchain/tx_list.go
```diff
@@ -326,7 +327,10 @@ func (l *txList) Filter(senderBalance *big.Int, pool *TxPool) (types.Transaction
 			feePayer, _ := tx.FeePayer()
 			feePayerBalance := pool.getBalance(feePayer)
 			feeRatio, isRatioTx := tx.FeeRatio()
-			if isRatioTx {
+
+			if sender == feePayer {
```
kjhman21, line 331: hmm? If it is a ratio tx, the comparison is right? I think this is to be included in the case that it is not ratio tx.

JayChoi1736: i think checking sender==feePayer condition is necessary for both fee delegate valueTransfer & ~ with ratio.
whether or not the fee delegated Transaction is ratioTx, the Tx should be rejected if tx.Cost() > sender balance and vice versa.

kjhman21: @JayChoi1736 What if the below case?
sender == feePayer && ratio == 10% ? With this code, it fallbacks the first condition. Is this your intention?

JayChoi1736: @kjhman21 yes. 
when the ratio is 10% and sender == feePayer,
check logic from prior code is
senderBalance < (tx.value + tx.fee&ast;0.9) || senderBalance < (tx.fee&ast;0.1)
and changed logic is 
senderBalance < tx.Cost()

1) obviously if this statement is **TRUE**, changed logic will return **TRUE** because tx.Cost is always bigger than (tx.value + tx.fee&ast;0.9) , tx.fee&ast;0.1

2) but when the statement is **FALSE**, there are two cases. 
2-1) if senderBalance > tx.Cost(), changed logic will also return **FALSE**
2-2) if senderBalance < tx.Cost(), changed logic will return **TRUE** (and the filter will remove the Tx) 
   - but this one is intended because in case of senderBalance > (tx.value + tx.fee&ast;0.9) && senderBalance > (tx.fee&ast;0.1) && senderBalance < tx.Cost() , the Tx will not be executed and be pended in tx pool.

Please let me know if I've missed out anything.

kjhman21: Oh, now I understood what problem you were trying to solve. Okay. So, this is kind of a shortcut before checking sender and fee payer balance check.

I rather suggest that do not add 'else' to reduce confusion in the next conditional statement.

blockchain/tx_list.go
```diff
@@ -326,7 +327,10 @@ func (l *txList) Filter(senderBalance *big.Int, pool *TxPool) (types.Transaction
...
+				return senderBalance.Cmp(tx.Cost()) < 0
+			} else if isRatioTx {
```
kjhman21, line 333:
```suggestion
			}
			if isRatioTx {
```

yoomee1313 (approved):

aidan-kwon (approved):

blockchain/tx_pool_test.go
```diff
@@ -115,6 +115,39 @@ func dynamicFeeTx(nonce uint64, gaslimit uint64, gasFee *big.Int, tip *big.Int,
...
 }
 
+func feeDelegatedTx(nonce uint64, gaslimit uint64, gasPrice *big.Int, amount *big.Int, senderPrvKey *ecdsa.PrivateKey, feePayerPrvKey *ecdsa.PrivateKey) *types.Transaction {
+	delegatedTx := types.NewTx(&types.TxInternalDataFeeDelegatedValueTransfer{
+		AccountNonce: nonce,
+		Price:        gasPrice,
+		GasLimit:     gaslimit,
+		Recipient:    common.Address{},
+		Amount:       amount,
+		From:         crypto.PubkeyToAddress(senderPrvKey.PublicKey),
```
tenderly44, line 125: meaning of senderPrvkey?

kjeom: @tenderly44 No, It is the sender's address.

Issue #1245, opened 2022-03-17
Implemented dynamic gas fee pricing mechanism(KIP-71)


kjhman21:
Implemented [KIP-71 : dynamic gas fee pricing mechanism](https://github.com/klaytn/kips/issues/71)
- [x] Implemented newly governance parameter to manage base fee parameter through governance. (@kjeom) 
  - #1379 
- [x] Dynamic base fee mechanism (@kjeom)
  - #1409 
  - Implemented base fee calculation logic and apply it. (@kjeom)
  - Implemented transaction fee burning according to KIP-71. (@kjeom)
  - Implemented transaction fee refund according to KIP-71. (@kjeom)
- [x] Implemented transaction pool policy according to KIP-71. (@sirano11)
  - #1382  
  - Hardfork branch logic for txPool. (@kjeom)
- [x] Implemented `baseFee` opcode. (@kjeom)
  - Need to test `baseFee` opcode
- [x] Porting RPC package from go-ethereum.
  -  #1385 
- [x] Implemented JSON/RPC API related KIP-71. (second phase)
  - #1452 
- [x] Separated logic before and after hardfork. (second phase)
  - #1442

PR #1379, opened 2022-05-16
Dynamic gas price

kjeom:
## Proposed changes

- Governance parameters added for dynamic gas price policy
  - LowerBoundBaseFee, UpperBoundBaseFee, GasTarget, BlockGasLimit, BaseFeeDenominator
- Once a related vote submitted, the changed parameter value is written in blockchain chainConfig by callback function after epoch

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- #1245


params/config.go
```diff
@@ -186,6 +186,7 @@ type GovernanceConfig struct {
 	GoverningNode  common.Address `json:"governingNode"`
 	GovernanceMode string         `json:"governanceMode"`
 	Reward         *RewardConfig  `json:"reward,omitempty"`
+	KIP71          *KIP71Config   `json:kip71,omitempty`
```
sirano11: Does It needs to double quote?
`json:"kip71,omitempty"`

work/work/mocks/txpool_mock.go
```diff
@@ -1,230 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
sirano11, line 1: Why did you remove this file?

params/config.go
kjeom: yes right, thanks

work/work/mocks/txpool_mock.go
kjeom: I have to add new setter function for mutex based update in blockchain and It also need to regenerate `blockchain_mock` and `txpool_mock`.
I figured it out a duplicated file, work/work/mocks/txpool_mock.go, work/mocks/txpool_mock.go.
Do you know where It can be used?

params/config.go
sirano11: I think it need to add test case that  encode / decode GovernanceConfig variable to json.

kjeom: the last commit changed the double quote.

work/work/mocks/txpool_mock.go
sirano11: Okay. I agreed this file is duplicated.

params/config.go
kjeom: You looked at the 2 steps previous commit.

kjeom: I also check the json encode/decode test by (read/write) (from/to) DB for `GovernanceConfig` by `NewGovernanceInitialize( )`
```go
func NewGovernanceInitialize(chainConfig *params.ChainConfig, dbm database.DBManager) *Governance {
	ret := NewGovernance(chainConfig, dbm)
	// nil is for testing or simple function usage
	if dbm != nil {
		ret.ReadGovernanceState()
		if err := ret.initializeCache(); err != nil {
			// If this is the first time to run, store governance information for genesis block on database
			cfg := GetGovernanceItemsFromChainConfig(chainConfig)
			if err := ret.WriteGovernance(0, cfg, NewGovernanceSet()); err != nil {
				logger.Crit("Error in writing governance information", "err", err)
			}
			// If failed again after writing governance, stop booting up
			if err = ret.initializeCache(); err != nil {
				logger.Crit("No governance cache index found in a database", "err", err)
			}
		}
	}
	return ret
}
```

blockchain/blockchain.go
```diff
@@ -454,6 +454,36 @@ func (bc *BlockChain) SetProposerPolicy(val uint64) {
 	bc.chainConfig.Istanbul.ProposerPolicy = val
 }
 
+func (bc *BlockChain) SetLowerBoundBaseFee(val uint64) {
+	bc.chainConfigMu.RLock()
+	defer bc.chainConfigMu.RUnlock()
+	bc.chainConfig.Governance.KIP71.LowerBoundBaseFee = val
```
blukat29, line 460: `chainConfig.Governance.KIP71` can be `nil`, if it wasn't set at the first place.
For instance, when `genesis.json` file don't contain the "kip71" at all.

Please check if the new Set...() methods work using homi or relevant unit tests.

governance/default.go
```diff
@@ -54,6 +54,11 @@ var (
 		"istanbul.policy":               params.Policy,
 		"istanbul.committeesize":        params.CommitteeSize,
 		"governance.unitprice":          params.UnitPrice,
+		"kip71.lowerboundbasefee":       params.LowerBoundBaseFee,
+		"kip71.gastarget":               params.GasTarget,
+		"kip71.blockgaslimit":           params.BlockGasLimit,
+		"kip71.basefeedenominator":      params.BaseFeeDenominator,
+		"kip71.upperboundbasefee":       params.UpperBoundBaseFee,
```
aidan-kwon, line 61: Keys of this map are exposed to node operators, and they should type them to vote. In this context, `kip71` is not intuitive. I think `fee`, `transaction`, `txpool`, or `governance` are more intuitive for those values. How about changing the names?

aidan-kwon (commented):
You will add trigger functions for the new governance items, right? Please, note the following works in the code with `TODO` comments or in the PR description. If not, it is hard for reviewers to find out what you are missing.

blockchain/blockchain.go
```diff
@@ -454,6 +454,36 @@ func (bc *BlockChain) SetProposerPolicy(val uint64) {
 	bc.chainConfig.Istanbul.ProposerPolicy = val
 }
 
+func (bc *BlockChain) SetLowerBoundBaseFee(val uint64) {
+	bc.chainConfigMu.RLock()
```
kjhman21, line 458: @kjeom Isn't RLock() for reader lock? It seems you update the value, so you need Lock() instead of RLock().

params/config.go
```diff
@@ -170,6 +173,7 @@ type ChainConfig struct {
 	IstanbulCompatibleBlock  *big.Int `json:"istanbulCompatibleBlock,omitempty"`  // IstanbulCompatibleBlock switch block (nil = no fork, 0 = already on istanbul)
 	LondonCompatibleBlock    *big.Int `json:"londonCompatibleBlock,omitempty"`    // LondonCompatibleBlock switch block (nil = no fork, 0 = already on london)
 	EthTxTypeCompatibleBlock *big.Int `json:"ethTxTypeCompatibleBlock,omitempty"` // EthTxTypeCompatibleBlock switch block (nil = no fork, 0 = already on ethTxType)
+	KIP71CompatibleBlock     *big.Int `json:"kip71CompatibleBlock,omitempty"`     // KIP71Compatible switch block (nil = no fork, 0 already on ethTxType)
```
kjhman21, line 176:
```suggestion
	KIP71CompatibleBlock     *big.Int `json:"kip71CompatibleBlock,omitempty"`     // KIP71Compatible switch block (nil = no fork, 0 already on KIP71)
```

blockchain/blockchain.go
kjeom: Thanks, It should have Lock to write

cmd/homi/setup/cmd.go
```diff
@@ -196,6 +196,22 @@ func genRewardConfig(ctx *cli.Context) *params.RewardConfig {
 	}
 }
 
+func genKIP71Config(ctx *cli.Context) *params.KIP71Config {
+	lowerBoundBaseFee := ctx.Uint64(kip71LowerBoundBaseFeeFlag.Name)
```
kjhman21, line 200: `.Name` is correct? might be `.Value`?

consensus/istanbul/backend/engine_test.go
```diff
@@ -1272,6 +1273,91 @@ func TestGovernance_Votes(t *testing.T) {
 				{vote{"governance.unitprice", uint64(2)}, 0}, // check on current
 			},
 		},
+		{
+			votes: []vote{
+				{"kip71.lowerboundbasefee", uint64(25000000000)},  // voted on block 1
```
kjhman21, line 1278: Should we vote to every block? it is not common practice in real use case.

kjeom: Yes, It is actually dummy data to create more blocks. It only checks data voted on block 2, 5.
votes on block 6~9 can be deleted.

blukat29: @kjhman21 Apparentely this test creates a block per each element in `votes`.
I have plan to simplify/refactor to [using map](https://github.com/klaytn/klaytn/blob/bd9db8914fdbf389f6fcc49fa79cc2173e3908f0/consensus/istanbul/backend/engine_test.go#L1009) some time later. But I'd say we keep it this way for now.

@kjeom not the best way, but you can use an empty `vote` to create a block without vote.
It works because return value of `engine.governance.AddVote(v.key, v.value)` is not checked.
It should make your intention more clear.

```
            votes: []vote{
                {"governance.governancemode", "none"},     // voted on block 1
                {"istanbul.committeesize", uint64(4)},     // voted on block 2
                {"governance.unitprice", uint64(2000000)}, // voted on block 3
                {"governance.governancemode", "single"},   // voted on block 4
                {},                                        // no votes on block 5
                {"governance.unitprice", uint64(2)},       // voted on block 6
                {"governance.governancemode", "none"},     // voted on block 7

```

governance/default.go
```diff
@@ -421,6 +436,15 @@ func NewGovernance(chainConfig *params.ChainConfig, dbm database.DBManager) *Gov
...
 	}
+	// TODO-klaytn KIP71 default setting
+	// If DB doesn't have the KIP71 state, default values will be used
+	if g.ChainConfig.Governance == nil {
+		g.ChainConfig.Governance = &params.GovernanceConfig{}
+	}
+	if g.ChainConfig.Governance.KIP71 == nil {
+		g.ChainConfig.Governance.KIP71 = params.GetDefaultKip71Config()
+	}
+	return g
```
blukat29, line 447: While it solves the problem, It's a little hard to expect that a constructor `NewGovernance` has a side-effect. How about moving this code to `cn.New()`?

```go
chainConfig, ... := blockchain.SetupGenesisBlock(...)

// ...

// Fill defaults to missing configs
if chainConfig.Governance == nil {
    chainConfig.Governance = &params.GovernanceConfig{}
}
if chainConfig.Governance.KIP71 == nil {
      chainConfig.Governance.KIP71 = params.GetDefaultKip71Config()
}

governance := governance.NewMixedEngine(chainConfig, chainDB)

// at this point, chainConfig = (DB or genesis.json) + defaults
logger.Info("Initialised chain configuration", "config", chainConfig)
```

kjeom: I think so, but governance initialization update the params from database. so cn.New( ) always update the values to default. How do you think refactoring to split the governance initialization and update code?

consensus/istanbul/backend/engine_test.go
kjeom: Empty vote entries looks better. thanks

governance/default.go
blukat29: @kjeom You're right about `Governance` updating params from database e.g. in [Governance.UnmarshalJSON](https://github.com/klaytn/klaytn/blob/v1.8.3/governance/default.go#L962) via [Governance.ReadGovernanceState](https://github.com/klaytn/klaytn/blob/v1.8.3/governance/default.go#L1006). But it does not affect other chainConfig objects held by BlockChain and TxPool. That's why we should use `SetGasPrice()`, etc.

Here, new ChainConfig is created by `json.Unmarshal` and replaces existing pointer `gov.ChainConfig`. But the contents of `bc.chainConfig` won't change. They are the different pointers  after executing the `gov.ChainConfig = j.ChainConfig` line.

In contrast, adding fields to the `chainConfig` field at `NewGovernance()` do affect other code, because `g.chainConfig` and `bc.chainConfig` are the same pointer.

blukat29: I'd love to refactor `default.go` but I'm not sure how. There are so many things going on here. Instead, I would recommend to minimize modifying the code until we have a clear plan to fix it.

kjeom: Ok I got it, I will consider more about the point with no side effect.

ian0371:
I couldn't find any issue, but I have a small question. Will the burn ratio not be added in this PR, but separately?

kjeom:
Yes it is. the logic for BaseFee will be added in another PR.

governance/default.go
kjeom: @blukat29 @ian0371 How do you think `governance.GetGovernance( )` to get the context in anywhere?

blukat29: Do you mean using `governance.GetGovernanceValue()` everywhere instead of pushing new values via `SetGasTarget()`?

That sounds good approach. That is similar to what I am [trying to do](https://github.com/klaytn/klaytn/pull/1381). Note that the governance part requires a lot of effort to make it pretty. Me and @ian0371 is [working on it](https://github.com/klaytn/klaytn/projects/2).

If we are going to change the way parameters are propagated, we'll have to fix for other params as well, at once. (UnitPrice, ProposerUpdateInterval, StakeUpdateInterval, ...). That should be another PR itself.

How about fitting into current structure for the sake of dynamic gas price in this PR?

hqjang-pepper:
Is there any error handling or code that ensures `LowerBoundBaseFee < UpperBoundBaseFee` ?

governance/default.go
ian0371: I agree that your suggestion is a better design.
However, I think it's better to leave it as is for the moment and reconsider this later.

kjeom:
I think we need the check routine because It can be modified by governance voting. Thanks pepper.

governance/default.go
kjeom: Thanks for your opinions. then I think it's better to leave as it is until more clear.

kjeom: @blukat29  @ian0371  I tested `cn.chainConfig's governance fields` by voting and confirmed that it works as intended (modify, persist).

1. After voting a lowerboundbasefee parameter, it exactly changed in the next epoch cycle as following capture
2. After rebooting the node, I was able to check the persistence of the parameter

- test epoch : 30
- index cache : 300 (voting confirmed)
- lowerboundbasefee : 30ston -> 250ston
image

Following is my opinion that can be used as references for governance design.
- Using the `governance.chainConfig API`, we can check the chainConfig of governance. 
When the node is first booted, chainConfig instances are shared as references with `cn.chainConfig` and `governance.chainConfig`, but after rebooting, a different instance is created between the two so API results and cn.chainConfig has a difference context.  I **think we will need a clear policy on this part later.**

- For example, if `governance.chainConfig API` is used to show the value of the genesis boot, cn and governance should not share the same instance even at the genesis.
If you are looking for the context of the current node by the API, not the value of the first boot, you should always share the same instance.

blukat29: @kjeom Glad to hear that it works!

- Yes the `governance.chainConfig` is clearly flawed. There are inconsistencies among the documentation (initial chain configuration), the implementation (updated every 1024 blocks), the expectation (updated every block).
- In your use case, as you may know, be sure the use other methods including
    - console command `governance.itemsAt(klay.blockNumber)`
    - directly printing in-memory `chainConfig` variable like you did.

There are some ways to fix it.

1. Make sure `governance.chainConfig == cn.chainConfig` at all times
  a. Always share the `ChainConfig` instance throughout all codes including BlockChain, TxPool, and Governance. 
  b. Make sure Governance do not replace its `ChainConfig` pointer field.
  c. At an epoch, the changes are directly reflected to `ChainConfig`. No explicit setters (i.e. no need to `SetGasPrice`).
  d. Fix Klaytn docs to say it returns "latest chain config"

2. Or just fix `governance.chainConfig` API to return the latest value
  a. Implement `API.ChainConfig() = API.ItemsAt("latest")`
  b. Fix Klaytn docs to say it returns "latest chain config"

I'm not sure if the "initial chain config" API is really necessary. We can use `itemsAt(0)` for that.

This discussion is definitely important. I think we can continue on this topic at https://github.com/klaytn/klaytn/issues/1391.

node/cn/backend.go
```diff
@@ -221,6 +221,17 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 	config.GasPrice = new(big.Int).SetUint64(chainConfig.UnitPrice)
 	logger.Info("Initialised chain configuration", "config", chainConfig)
 
+	// TODO-klaytn set KIP71 forked block number
+	// It's intended not to modify something like genesis in SetupGenesisBlock()
+	chainConfig.KIP71CompatibleBlock = params.DefaultKIP71CompatibleBlockNum
```
blukat29, line 226: Do you mean to always overwrite genesis's KIP71CompatibleBlock value?
It might overwrite the value from homi-generated genesis.json if any.

I'm asking because `CompatibleBlock = ` pattern does not appear in anywhere in our codebase except `cmd/homi/setup/cmd.go`.

params/config.go
```diff
@@ -143,6 +143,9 @@ var (
 	//  - 0x1: file (DATADIR/logs/vm.log)
 	//  - 0x2: stdout (like logger.DEBUG)
 	VMLogTarget = 0x0
+
+	// TODO-klaytn temporal number for test
```
blukat29, line 147: Did you mean temporary?

node/cn/backend.go
kjeom: You are right. It's our first hard fork by hard coded block number as I know.

It can be resolved by following
```go
if chainConfig.KIP71CompatibleBlock == nil {
  chainConfig.KIP71CompatibleBlock = params.DefaultKIP71CompatibleBlockNum
}
```
I reviewed the code about `cmd/homi/setup/cmd.go` for KIP71 config but didn't fine any use case.
I think the homi support for KIP71 is another issue. Do you think it's better to remove the code about homi in this pr?

blukat29: - If it's our first hard fork by hardcoded block number, then

  ```
  if chainConfig.KIP71CompatibleBlock == nil {
    chainConfig.KIP71CompatibleBlock = params.DefaultKIP71CompatibleBlockNum
  }
  ```
  seems good to me.

- No I didn't mean to say we should modify homi. Sorry for the confusion.

blukat29 (approved):

sirano11 (approved):

cmd/homi/setup/cmd.go
```diff
@@ -196,6 +196,22 @@ func genRewardConfig(ctx *cli.Context) *params.RewardConfig {
...
+	gasTarget := ctx.Uint64(kip71GasTarget.Name)
+	blockGasLimit := ctx.Uint64(kip71BlockGasLimit.Name)
+	baseFeeDenominator := ctx.Uint64(kip71BaseFeeDenominator.Name)
+
+	return &params.KIP71Config{
+		LowerBoundBaseFee:  lowerBoundBaseFee,
+		UpperBoundBaseFee:  upperBoundBaseFee,
+		GasTarget:          gasTarget,
+		BlockGasLimit:      blockGasLimit,
+		BaseFeeDenominator: baseFeeDenominator,
```
hyunsooda, line 211: Could you describe what `BaseFeeDenominator` value stands for? 

Furthermore, adding a brief description of every new property would be nice.

consensus/istanbul/backend/engine_test.go
```diff
@@ -1327,3 +1393,181 @@ func TestGovernance_Votes(t *testing.T) {
...
+		case "kip71.upperboundbasefee":
+			assert.Equal(t, tc.expected.value, govConfig.KIP71.UpperBoundBaseFee)
+		case "kip71.gastarget":
+			assert.Equal(t, tc.expected.value, govConfig.KIP71.GasTarget)
+		case "kip71.blockgaslimit":
+			assert.Equal(t, tc.expected.value, govConfig.KIP71.BlockGasLimit)
+		case "kip71.basefeedenominator":
+			assert.Equal(t, tc.expected.value, govConfig.KIP71.BaseFeeDenominator)
+		default:
+			assert.Error(t, nil)
```
hyunsooda, line 1479: It seems redundant

consensus/istanbul/backend/engine_test.go
```diff
@@ -1327,3 +1393,181 @@ func TestGovernance_Votes(t *testing.T) {
...
+		case "kip71.upperboundbasefee":
+			assert.Equal(t, tc.expected.value, gov.ChainConfig.Governance.KIP71.UpperBoundBaseFee)
+		case "kip71.gastarget":
+			assert.Equal(t, tc.expected.value, gov.ChainConfig.Governance.KIP71.GasTarget)
+		case "kip71.blockgaslimit":
+			assert.Equal(t, tc.expected.value, gov.ChainConfig.Governance.KIP71.BlockGasLimit)
+		case "kip71.basefeedenominator":
+			assert.Equal(t, tc.expected.value, gov.ChainConfig.Governance.KIP71.BaseFeeDenominator)
+		default:
+			assert.Error(t, nil)
```
hyunsooda, line 1568: It seems redundant.

node/cn/backend.go
hyunsooda: In my view, hardcoding the forked block number seems not bad. Any block number can be interpreted as a KIP17-compatable block number If the number is grabbed from `genesis.json`. I'm not sure the source code is a reasonable place to record the history. I would appreciate letting me know if there's a traditional process.

params/governance_params.go
```diff
@@ -90,6 +95,11 @@ var (
 	DefaultProposerPolicy          = uint64(RoundRobin)
 	DefaultSubGroupSize            = uint64(21)
 	DefaultUnitPrice               = uint64(250000000000)
+	DefaultLowerBoundBaseFee       = uint64(25000000000)
+	DefaultUpperBoundBaseFee       = uint64(750000000000)
+	DefaultGasTarget               = uint64(30000000)
+	DefaultBlockGasLimit           = uint64(84000000)
+	DefaultBaseFeeDenominator      = uint64(36)
```
hyunsooda, line 102: Please describe how these values were derived into the mainboard of the PR.

governance/api.go
```diff
@@ -96,6 +98,22 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
 			return "", errRemoveSelf
 		}
 	}
+	if strings.ToLower(key) == "kip71.lowerboundbasefee" {
+		if _, ok := api.governance.ValidateVote(&GovernanceVote{Key: key, Value: val}); !ok {
+			return "", errInvalidKeyValue
+		}
+		if val.(uint64) > api.governance.UpperBoundBaseFee() {
```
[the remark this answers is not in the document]
kjeom: You are right. It can be run with fixed base fee policy as well.

params/governance_params.go
kjeom: You can find KIP-71 history in https://github.com/klaytn/kips/pull/72

cmd/homi/setup/flags.go
```diff
@@ -294,6 +294,33 @@ var (
...
+	kip71BlockGasLimit = cli.Uint64Flag{
+		Name:  "block-gas-limit",
+		Usage: "block gas limit flag",
+		Value: params.DefaultBlockGasLimit,
+	}
+	kip71BaseFeeDenominator = cli.Uint64Flag{
+		Name:  "base-fee-denominator",
+		Usage: "base fee denominator flag",
+		Value: params.DefaultLowerBoundBaseFee,
+	}
```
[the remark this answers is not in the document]
kjeom: Actually I didn't find any use case for this parameters in `service chain` by `homi` so I'd like to remain this another issue.
```go
        if cypressTest {
		genesisJson = genCypressTestGenesis(validatorNodeAddrs, testAddrs)
	} else if cypress {
		genesisJson = genCypressGenesis(validatorNodeAddrs, testAddrs)
	} else if baobabTest {
		genesisJson = genBaobabTestGenesis(validatorNodeAddrs, testAddrs)
	} else if baobab {
		genesisJson = genBaobabGenesis(validatorNodeAddrs, testAddrs)
	} else if cliqueFlag {
		genesisJson = genCliqueGenesis(ctx, validatorNodeAddrs, testAddrs, chainid)
	} else if serviceChain {
		genesisJson = genServiceChainGenesis(validatorNodeAddrs, testAddrs)
	} else if serviceChainTest {
		genesisJson = genServiceChainTestGenesis(validatorNodeAddrs, testAddrs)
	} else {
		genesisJson = genIstanbulGenesis(ctx, validatorNodeAddrs, testAddrs, chainid)
	}
```
`genSetviceChainGenesis( )` doesn't have Governance parameter configuration in ` cmd/homi/setup/cmd.go`
Only `genIstanbulGenesis( )` has configuration for governance parameters like KIP-71. Could you explain if we need governance configuration or not for service chain?

params/config.go
```diff
@@ -143,6 +143,9 @@ var (
...
+	DefaultKIP71CompatibleBlockNum = big.NewInt(95000000)
```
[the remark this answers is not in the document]
kjeom: It's better. Thanks

blukat29:
@hqjang-pepper @kjeom

I have one small concern about the pre-voting check (https://github.com/klaytn/klaytn/pull/1379/files#diff-3cfb26c3349a61b610e0dd42d41eee189b302f233fe8a67bd707f5005ae21b62R101)
- Suppose current LowerBound=25, UpperBound=750.
- Suppose we want to vote for LowerBound=2500, UpperBound=7500 in an epoch.
- This scenario is conceptually valid because
  - Now.LowerBound < Now.UpperBound
  - Future.LowerBound < Future.UpperBound
- But the pre-voting check fails because
  - Future.LowerBound > Now.UpperBound

But I don't think we have to fix it somehow. I just want you to recognize this rare corner case.
In contrast, the pre-voting check is useful since it detects insane values. Therefore I approve the change.

governance/api.go
```diff
@@ -58,6 +58,8 @@ var (
 	errPermissionDenied       = errors.New("You don't have the right to vote")
 	errRemoveSelf             = errors.New("You can't vote on removing yourself")
 	errInvalidKeyValue        = errors.New("Your vote couldn't be placed. Please check your vote's key and value")
+	errInvalidLowerBound      = errors.New("lowerboundbasefee can not be over upperboundbasefee")
```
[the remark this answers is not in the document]
kjeom: I think "`cannot be set more than upperboundbasefee`" looks better.

governance/api.go
```diff
@@ -58,6 +58,8 @@ var (
...
+	errInvalidUpperBound      = errors.New("upperboundbasefee can not be under lowerboundbasefee")
```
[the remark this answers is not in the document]
kjeom: I think "cannot be set less than lowerboundbasefee" looks better.

governance/api_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+func newTestGovernanceApi() *PublicGovernanceAPI {
+	config := params.CypressChainConfig
+	config.Governance.KIP71 = params.GetDefaultKip71Config()
+	govApi := NewGovernanceAPI(NewMixedEngine(config, database.NewMemoryDBManager()))
+	govApi.governance.SetNodeAddress(common.HexToAddress("0x52d41ca72af615a1ac3301b0a93efa222ecc7541"))
+	return govApi
+}
+
+func TestUpperBoundUnderLowerBound(t *testing.T) {
```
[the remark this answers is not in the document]
kjeom: Yes it modified.

consensus/istanbul/backend/engine_test.go
kjeom: You are right but I'd like to see It as a secure coding

consensus/istanbul/backend/engine_test.go
```diff
@@ -1327,3 +1393,181 @@ func TestGovernance_Votes(t *testing.T) {
...
+
+		engine.governance.SetBlockchain(chain)
+		engine.governance.AddVote(tc.voting.key, tc.voting.value)
+		previousBlock = currentBlock
+		currentBlock = makeBlockWithSeal(chain, engine, previousBlock)
+		_, err = chain.InsertChain(types.Blocks{currentBlock})
+		assert.NoError(t, err)
+
+		// insert blocks until the vote is applied
+		for i := 0; i < 1023; i++ {
```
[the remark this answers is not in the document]
kjeom: looks more clear :)

governance/api_test.go
```diff
@@ -0,0 +1,36 @@
...
+	govApi := newTestGovernanceApi()
+
+	curLowerBoundBaseFee := govApi.governance.LowerBoundBaseFee()
+	invalidUpperBoundBaseFee := curLowerBoundBaseFee - 100
+	_, err := govApi.Vote("kip71.upperboundbasefee", invalidUpperBoundBaseFee)
+	assert.Equal(t, err, errInvalidUpperBound)
+}
+
+func TestLowerBoundOverUpperBound(t *testing.T) {
```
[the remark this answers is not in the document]
kjeom: Of course, `TestUpperBoundFeeSet_UnderLowerBound` looks more clear.

params/governance_params.go
hyunsooda: The proposal seems to lack grounds for those initial values. Anyway, thank you.

params/config.go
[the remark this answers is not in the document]
hyunsooda: Please mark it as resolved when you are done.

consensus/istanbul/backend/engine_test.go
hyunsooda: Why do you think that's secure handling? The test code has definitely five topics and the switch expression handles all that cases.

governance/api_test.go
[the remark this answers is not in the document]
hyunsooda: Seems not committed yet.

governance/api.go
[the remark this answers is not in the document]
hyunsooda: Got it. Thank you

governance/api.go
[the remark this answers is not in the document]
hyunsooda: "less then" seems awkward when you say it with a countable noun. Please refer to https://dictionary.cambridge.org/ko/%EC%82%AC%EC%A0%84/%EC%98%81%EC%96%B4/countable-noun

governance/api.go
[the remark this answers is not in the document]
hyunsooda: In my experience, "exceeding" is widely used when representing the numbers low and high. "More than" seems awkward.

consensus/istanbul/backend/engine_test.go
```diff
@@ -1272,6 +1273,71 @@ func TestGovernance_Votes(t *testing.T) {
 				{vote{"governance.unitprice", uint64(2)}, 0}, // check on current
 			},
 		},
+		{
+			votes: []vote{
+				{}, // voted on block 1
+				{"kip71.lowerboundbasefee", uint64(750000000000)}, // voted on block 2
+				{}, // voted on block 4
+				{}, // voted on block 3
```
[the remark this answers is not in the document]
hyunsooda: I meant a block number. PTAL.

cmd/homi/setup/flags.go
```diff
@@ -294,6 +294,33 @@ var (
 		Value: "2000000",
 	}
 
+	kip71LowerBoundBaseFeeFlag = cli.Uint64Flag{
+		Name:  "lower-bound-base-fee",
+		Usage: "lower bound base fee flag",
+		Value: params.DefaultLowerBoundBaseFee,
+	}
+
```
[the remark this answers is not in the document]
hyunsooda: Please mark it as resolved when you commit the suggested comment. Or, please add a comment if you disagree.

cmd/homi/setup/flags.go
```diff
@@ -294,6 +294,33 @@ var (
...
+	kip71UpperBoundBaseFeeFlag = cli.Uint64Flag{
+		Name:  "upper-bound-base-fee",
+		Usage: "upper bound base fee flag",
+		Value: params.DefaultUpperBoundBaseFee,
+	}
+
```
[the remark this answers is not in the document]
hyunsooda: Please mark it as resolved when you commit the suggested comment. Or, please add a comment if you disagree.

cmd/homi/setup/flags.go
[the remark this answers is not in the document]
hyunsooda: I'm sorry. I misunderstood those new values were not committed to `genesis.json`. Don't mind it. Thank you.

consensus/istanbul/backend/engine_test.go
kjeom: @hyunsooda What I mean is that It is my habit. I think It's safe for `switch` to have `default:` case.
I agree that the code is redundant and doesn't have any security issue in this case.
If someone need more testcase and put invalid key in it, the code would cover the case.

governance/api.go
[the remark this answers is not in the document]
kjeom: I got it. Thanks for the reference.

consensus/istanbul/backend/engine_test.go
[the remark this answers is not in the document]
kjeom: ahh, Thanks

cmd/homi/setup/flags.go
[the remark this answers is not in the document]
kjeom: I rather added the new line because other flags have the new line.

params/governance_params.go
kjeom: @hyunsooda  If you want to know why we derive the initial values like that in detail, please invite me for meeting.
It is derived from some historical back data and a little bit heuristic experience in klaytn.

governance/api.go
```diff
@@ -96,6 +98,22 @@ func (api *PublicGovernanceAPI) Vote(key string, val interface{}) (string, error
...
+		}
+	}
+	if strings.ToLower(key) == "kip71.upperboundbasefee" {
+		if _, ok := api.governance.ValidateVote(&GovernanceVote{Key: key, Value: val}); !ok {
+			return "", errInvalidKeyValue
+		}
+		if val.(uint64) < api.governance.LowerBoundBaseFee() {
+			return "", errInvalidUpperBound
+		}
+	}
```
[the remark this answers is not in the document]
kjeom: I think you are considering something like https://github.com/klaytn/klaytn/pull/1379#issuecomment-1138077733
If the `kip71.gastarget` is zero or very small number, then It can always cause the `basefee` to be set in `kip71.upperboundbasefee`. It's uneasy to constrain the parameters. Do you have any idea for this?

governance/api_test.go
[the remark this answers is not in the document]
kjeom: You can see in https://github.com/klaytn/klaytn/pull/1379/commits/67f414026028e6ab341a6e5a655c08c5c28ff779

hqjang-pepper (approved):

cmd/homi/setup/cmd.go
kjeom: I added some description in https://github.com/klaytn/klaytn/pull/1379/commits/46a7adc63c50c6704f656229e3e0942cb37ed5bc because the parameters are not intuitive respectively

governance/api.go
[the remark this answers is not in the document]
hyunsooda: `feedominator`:
- Must not be zero
- Must be `e` to 1, where `e` > 0
`blockgaslimit`: 
- Do not need to provide setter API. `Denomiator` already stands for suppressing an unexpectedly surprising change of the baseFee. I would appreciate it if let me know the need for this value. 
gastarget
- Do not need to provide setter API since it is currently a fixed value and not related to increase or decrease by the new policy. I recommend reconsidering the alternative way. For example, the `gastarget` can be an average gasUsed of near `n` blocks, rather fixes static value.

consensus/istanbul/backend/engine_test.go
hyunsooda: This is a unit test. When it comes to making a new unit test, creating a separate function for the corresponding unit test is reasonable.

cmd/homi/setup/cmd.go
hyunsooda: I meant a brief description of what each property stands for.

e.g.,
&MyStruct {
    age: uint8,           // "the period of time someone has been alive or something has existed"
    height: uint8,       //  "the distance from the top to the bottom of something, or the quality of being tall"
    ....
}

Moreover, the equations seem wrong. The proposal has two equations for basefee calculation. Also, division by `GasTarget` before division by `ParentBaseFee` seems absent in your equation. I recommend not writing equations in the comment before the proposal is stable and PoC launches. It may create wrong interpretations for careful readers.

governance/api_test.go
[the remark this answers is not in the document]
hyunsooda: Please reconsider again. You don't need to represent all the function's behavior to the function name. That's the beauty of naming that represents cores. If you want to describe in detail (but it's unnecessary cuz the code is clear as it is), leverage writing comments.

governance/api_test.go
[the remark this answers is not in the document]
hyunsooda: Please reconsider again. You don't need to represent all the function's behavior to the function name. That's the beauty of naming that represents cores. If you want to describe in detail (but it's unnecessary cuz the code is clear as it is), leverage writing comments.

params/governance_params.go
hyunsooda: Is there a public link for that history? If so, please put the link into the description of the PR so that anyone can figure it out and make sure to understand it.

hyunsooda:
@blukat29,

I think it depends on which vote is appeared first. It's valid If the vote for uppuerbound value comes first. However, the case of voting for lowerbound comes first is illegal in the current change and it seems reasonable. Why do you concern about this condition for local voting?

blukat29:
@hyunsooda
Even if UpperBound vote comes first, you cannot immediately vote for LowerBound because of the voting epoch.
You have to wait for 1~2 weeks so that UpperBound value actually changes to 7500. Then you can vote for LowerBound=2500.

As you said, in the hypothetical scenario where we change params from (25,750) to (2500,7500), we have to vote UpperBound first and then vote for LowerBound. But I wanted to point out that the pre-voting check is not 100% correct because of the delay in voting process.

hyunsooda:
@blukat29, I had no idea how the epoch value was set. The test simply makes three epochs basically. Let me ask a question. Is it illegal to vote on two topics at the same epoch although voting on two lower and upper bounds are different topics?  I just checked giving two votes on different topics, but seems to work well.

blukat29:
@hyunsooda You can vote two items in the same epoch. It is legal. But both will be applied after some time.

Have you checked the values from `governance.itemsAt(num)` [API](https://docs.klaytn.foundation/dapp/json-rpc/api-references/governance#governance_itemsat)? Try it with different `num` after you vote.

The voting epoch works as follows.
- An epoch is a range of blocks. Each epoch is 604800 blocks in Cypress.
- You place a vote in epoch 'X' (i.e. somewhere between the blocks `Epoch*X` - `Epoch*(X+1)`)
- The vote is finalized (i.e. written to the DB)
    at the end of the epoch (the block `Epoch*(X+1)`)
- The value is applied (i.e. retrieved from getters like `governance.UpperBoundBaseFee()`)
    at the end of next epoch (the block `Epoch*(X+2)`)

Therefore it takes at least 1 full epoch for a vote to take effect.

hyunsooda:
@blukat29,  You said:
> Even if UpperBound vote comes first, you cannot immediately vote for LowerBound because of the voting epoch.

and

> You can vote two items in the same epoch. It is legal. But both will be applied after some time.

I got the late effect of the vote, but you were concerned about the condition of vote parameters. It seems not related to the effect of the time, rather it still depends on which governance item is updated first at the pre-voting. Thus, the condition seems not to create an edge case at the time of pre-vote.

Hope to hear your comment again.

governance/api.go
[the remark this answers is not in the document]
kjeom: I'd like to remain it another issue. We need to discuss it after forward. Thanks.

params/governance_params.go
kjeom: I'm so sorry there are no documents to show you summarized history. I think we can treat this issue by meeting.

governance/api.go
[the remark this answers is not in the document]
hyunsooda: Okay. Let me note another comment before resolving it. 
- The operation of `feedominator` seems better to substitute from division to multiply. If the operation is 'division', the decision of its value seems not easy. If the operator is 'multiply', the value can be obviously zero to one.
- The name `feedominaotr` seems awkward. Refer to other notations for the scaling factor in well-known standards.

params/governance_params.go
hyunsooda: Okay. Next time, please describe these surprising changes such as these static values, because the new readers to your change are unaware of the behind the history of this decision. If the code is not obvious itself, please describe them in the description of the PR.

params/governance_params.go
```diff
@@ -84,20 +89,25 @@ const (
...
+	DefaultGovernanceMode            = "none"
+	DefaultGoverningNode             = "0x0000000000000000000000000000000000000000"
+	DefaultEpoch                     = uint64(604800)
+	DefaultProposerPolicy            = uint64(RoundRobin)
+	DefaultSubGroupSize              = uint64(21)
+	DefaultUnitPrice                 = uint64(250000000000)
+	DefaultLowerBoundBaseFee         = uint64(25000000000)
+	DefaultUpperBoundBaseFee         = uint64(750000000000)
+	DefaultGasTarget                 = uint64(30000000)
+	DefaultMaxBlockGasUsedForBaseFee = uint64(84000000)
```
aidan-kwon, line 101: @kjeom  As far as I know, this value is changed to 60000000. You gonna update this value to 60000000 later, right?

cmd/homi/setup/cmd.go
kjeom: I added the brief description in 0e464f4

governance/api_test.go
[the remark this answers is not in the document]
kjeom: I fixed it in 0e464f4

params/governance_params.go
```diff
@@ -53,6 +53,11 @@ const (
 	Policy
 	CommitteeSize
 	UnitPrice
+	LowerBoundBaseFee
+	UpperBoundBaseFee
+	GasTarget
+	MaxBlockGasUsedForBaseFee
+	BaseFeeDenominator
```
aidan-kwon, line 60: We should add them at the end of the itoa list. 
Those values are `iota` type and used for governance voting. It means that the itoa values of previous votings are contained in previous block headers. So, this change can cause chain forks during the synchronizing of old blocks.

params/config.go
```diff
@@ -464,6 +487,7 @@ type Rules struct {
 	ChainID    *big.Int
 	IsIstanbul bool
 	IsLondon   bool
+	IsKIP71    bool
```
aidan-kwon, line 490: Perhaps, we don't need to distinguish whether KIP71 hard-fork is enabled or not. How about setting `baseFee` of EVM to `0` if block header doesn't contains the `baseFee` field? Then, EVM doens't change its semantic after KIP71 hard-fork.

params/governance_params.go
blukat29: @aidan-kwon Is it really? As far as I know the string names are written into headers, not integer keys. Database stores string names too. Anyway I agree that enums should be appended at the end as a general caution.

params/config.go
kjeom: I think that's good idea. but the mechanism only works outside EVM.
In the first version I developed, It seems to be necessary for EVM internal like following.

in blockchain/state_transition.go
```go
func (st *StateTransition) buyGas() error {
	mgval := new(big.Int).Mul(new(big.Int).SetUint64(st.msg.Gas()), st.gasPrice)
	mgval := new(big.Int)
	if st.evm.ChainConfig().IsKIP71ForkEnabled(st.evm.BlockNumber) {
		mgval = mgval.Mul(new(big.Int).SetUint64(st.msg.Gas()), st.evm.BaseFee)
	} else {
		mgval = mgval.Mul(new(big.Int).SetUint64(st.msg.Gas()), st.gasPrice)
	}
```
we have already included the baseFee field for ethereum equivalence. Therefore we need to distinguish hark-forked block before/after.

params/governance_params.go
kjeom: Ahh, It could have been very critical.
I fixed it, thanks.

kjeom: Oh @blukat29 your review appeared lately. I think we need to discuss more this.

aidan-kwon: @blukat29 Uhm.. You're right. Votes in the header use plain string instead of a number. If Klaytn doesn't store governance data with this itoa values, it's gonna be okay. It is safe to add in the end, though.

kjeom: I got it, Thanks any way.

params/governance_params.go
kjeom: @aidan-kwon I modified the default values to reduce the amount of reviews in following PR, `dynamic base fee logic`
It will remove the already reviewed files (engine_test.go, governance_params.go, ...) in this PR.

params/config.go
aidan-kwon: @kjeom  I still think that we don't need to distinguish whether hard-forked or not. We can set the common policy for `buyGas` and `EVM` with the assumption of "baseFee will not be missing nor zero after the KIP71 hardfork". 
I think we can have a better discussion with the implementation code of KIP71 hard-fork application. Let's continue the discussion on the next PR.

aidan-kwon (approved):
I left some comments, but we can discuss them in the next PR. Let' move on.

PR #1382, opened 2022-05-18
[KIP-71] Implemented txPool policy logic.

sirano11:
## Proposed changes

- This PR implemented KIP-71 txPool policy. 
 - The baseFee will be set as a gasPrice of txpool when occurring  'ChainHeadEvent`.
   - The baseFee will be set next blocks baseFee. 
 - When the transaction income  tx pool through receiving message from peer or calling API
   - Tx pool only accepted if `gasPrice` of tx is bigger than or equal to `gasPrice` of txpool.
   - In case of Dynamic Transaction type, only check `gasFeeCap` field of tx and `gasTipCap` field of tx will be ignored.
 - When transactions in queue promoted pending.
   - Transactions greater than or equal to baseFee are promoted by sorting in nonce order.
   - If a specific transaction has a small baseFee, subsequent transactions (larger nonce) are not promoted.
 - When transactions in pending demoted queue.
   - Transaction lower than baseFee are demoted. 
 - When choosing transactions in pending for committing a new block, 
   - It also check that gasPrice of transaction bigger than or equal to baseFee. 
   - Transactions sorted by time ascending order.(Implemented `TxByTime`)
- These changes need to be hardfork, i commented existing code. `TODO : Need to KIP-71 hardfork`.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- Related to #1245

## Further comments

### DIFF
--- a/blockchain/tx_list.go
+++ b/blockchain/tx_list.go
@@ -195,6 +195,37 @@ func (m *txSortedMap) Ready(start uint64) types.Transactions {
 	return ready
 }
 
+// ReadyWithGasPrice retrieves a sequentially increasing list of transactions that greater than or
+// equal to the given baseFee starting at the provided nonce that is ready for processing.
+// If there is a transaction lower than the baseFee during the search, only previously collected transactions
+// are returned. The returned transactions will be removed from the list.
+//
+// Note, all transactions with nonces lower than start will also be returned to
+// prevent getting into and invalid state. This is not something that should ever
+// happen but better to be self correcting than failing!
+func (m *txSortedMap) ReadyWithGasPrice(start uint64, baseFee *big.Int) types.Transactions {
+	// Short circuit if no transactions are available
+	if m.index.Len() == 0 || (*m.index)[0] > start {
+		return nil
+	}
+	// Otherwise start accumulating incremental transactions
+	var ready types.Transactions
+	for next := (*m.index)[0]; m.index.Len() > 0 && (*m.index)[0] == next; next++ {
+		if m.items[next].GasPrice().Cmp(baseFee) < 0 {
+			break
+		}
+		ready = append(ready, m.items[next])
+		delete(m.items, next)
+		heap.Pop(m.index)
+	}
+	// If we had a cached order, shift the front
+	if m.cache != nil {
+		m.cache = m.cache[len(ready):]
+	}
+
+	return ready
+}
+
 // Len returns the length of the transaction map.
 func (m *txSortedMap) Len() int {
 	return len(m.items)
@@ -408,6 +439,18 @@ func (l *txList) Remove(tx *types.Transaction) (bool, types.Transactions) {
 // happen but better to be self correcting than failing!
 func (l *txList) Ready(start uint64) types.Transactions {
 	return l.txs.Ready(start)
+}
+
+// ReadyWithGasPrice retrieves a sequentially increasing list of transactions that greater than or
+// equal to the given baseFee starting at the provided nonce that is ready for processing.
+// If there is a transaction lower than the baseFee during the search, only previously collected transactions
+// are returned. The returned transactions will be removed from the list.
+//
+// Note, all transactions with nonces lower than start will also be returned to
+// prevent getting into and invalid state. This is not something that should ever
+// happen but better to be self correcting than failing!
+func (l *txList) ReadyWithGasPrice(start uint64, baseFee *big.Int) types.Transactions {
+	return l.txs.ReadyWithGasPrice(start, baseFee)
 }
 
 // Len returns the length of the transaction list.
--- a/blockchain/tx_list.go
+++ b/blockchain/tx_list.go
@@ -209,7 +209,7 @@ func (m *txSortedMap) ReadyWithGasPrice(start uint64, baseFee *big.Int) types.Tr
 		return nil
 	}
 	// Otherwise start accumulating incremental transactions
-	var ready types.Transactions
+	ready := make(types.Transactions, 0, m.index.Len())
 	for next := (*m.index)[0]; m.index.Len() > 0 && (*m.index)[0] == next; next++ {
 		if m.items[next].GasPrice().Cmp(baseFee) < 0 {
 			break
--- a/blockchain/tx_list.go
+++ b/blockchain/tx_list.go
@@ -326,7 +326,7 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
 			return false, nil
 		} else {
 			// Otherwise overwrite the old transaction with the current one
-			logger.Trace("Gas price of new tx is bigger than older tx!", "old", old.String(), "new", tx.String())
+			logger.Trace("The transaction was substituted by competitive gas price", "old", old.String(), "new", tx.String())
 		}
 	}
 
--- a/blockchain/tx_list.go
+++ b/blockchain/tx_list.go
@@ -312,14 +312,24 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
 	// If there's an older better transaction, abort
 	old := l.txs.Get(tx.Nonce())
 	if old != nil {
-		if tx.Type().IsCancelTransaction() {
-			logger.Trace("New tx is a cancel transaction. replace it!", "old", old.String(), "new", tx.String())
-		} else {
+		// TODO : Need to KIP-71 hardfork.
+		//if tx.Type().IsCancelTransaction() {
+		//	logger.Trace("New tx is a cancel transaction. replace it!", "old", old.String(), "new", tx.String())
+		//} else {
+		//	logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
+		//	return false, nil
+		//}
+
+		// If gas price of older is bigger than newer, abort.
+		if old.GasPrice().Cmp(tx.GasPrice()) > 0 {
 			logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
 			return false, nil
+		} else {
+			// Otherwise overwrite the old transaction with the current one
+			logger.Trace("Gas price of new tx is bigger than older tx!", "old", old.String(), "new", tx.String())
 		}
 	}
-	// Otherwise overwrite the old transaction with the current one
+
 	l.txs.Put(tx)
 
 	return true, old
--- a/blockchain/tx_list.go
+++ b/blockchain/tx_list.go
@@ -321,7 +321,7 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
 		//}
 
 		// If gas price of older is bigger than newer, abort.
-		if old.GasPrice().Cmp(tx.GasPrice()) > 0 {
+		if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
 			logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
 			return false, nil
 		} else {
--- a/blockchain/tx_list_test.go
+++ b/blockchain/tx_list_test.go
@@ -21,11 +21,12 @@
 package blockchain
 
 import (
-	"gotest.tools/assert"
 	"math/big"
 	"math/rand"
 	"reflect"
 	"testing"
+
+	"gotest.tools/assert"
 
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/crypto"
--- a/blockchain/tx_list_test.go
+++ b/blockchain/tx_list_test.go
@@ -21,6 +21,7 @@
 package blockchain
 
 import (
+	"gotest.tools/assert"
 	"math/big"
 	"math/rand"
 	"reflect"
@@ -151,5 +152,51 @@ func TestTxListReadyWithGasPricePartialFilter(t *testing.T) {
 		if result := reflect.DeepEqual(ready[i], txs[i]); result == false {
 			t.Error("ready hash : ", ready[i].Hash(), "tx[i] hash : ", txs[i].Hash())
 		}
+	}
+}
+
+// TestReplaceTransaction tests if the new tx has been successfully replaced
+// if it has the same nonce and greater gas price as the old tx.
+func TestReplaceTransaction(t *testing.T) {
+	// Generate a list of transactions to insert
+	key, _ := crypto.GenerateKey()
+	txList := newTxList(false)
+
+	oldTx := pricedTransaction(0, 21000, big.NewInt(50), key)
+	newTx := pricedTransaction(0, 21000, big.NewInt(60), key)
+
+	if result, _ := txList.Add(oldTx, DefaultTxPoolConfig.PriceBump); !result {
+		t.Error("it cannot add tx in tx list.")
+	}
+
+	result, replaced := txList.Add(newTx, DefaultTxPoolConfig.PriceBump)
+	if !result {
+		t.Error("it cannot replace tx in tx list.")
+	}
+
+	assert.Equal(t, replaced, oldTx)
+}
+
+// TestReplaceTransactionAbort checks if a new tx aborts a new transaction
+// if it has the same nonce and lower gas price as the old tx.
+func TestReplaceTransactionAbort(t *testing.T) {
+	// Generate a list of transactions to insert
+	key, _ := crypto.GenerateKey()
+	txList := newTxList(false)
+
+	oldTx := pricedTransaction(0, 21000, big.NewInt(50), key)
+	newTx := pricedTransaction(0, 21000, big.NewInt(40), key)
+
+	if result, _ := txList.Add(oldTx, DefaultTxPoolConfig.PriceBump); !result {
+		t.Error("it cannot add tx in tx list.")
+	}
+
+	result, replaced := txList.Add(newTx, DefaultTxPoolConfig.PriceBump)
+	if result {
+		t.Error("it replaced to newly tx in tx list.")
+	}
+
+	if replaced != nil {
+		t.Error("it replaced to newly tx in tx list.")
 	}
 }
--- a/blockchain/tx_list_test.go
+++ b/blockchain/tx_list_test.go
@@ -60,7 +60,7 @@ func TestStrictTxListAdd(t *testing.T) {
 
 // TestTxListReadyWithGasPrice check whether ReadyWithGasPrice() works well.
 // It makes a slice of 10 transactions and executes ReadyWithGasPrice() with baseFee 30.
-func TestTxListReadyWithGasPrice(t *testing.T) {
+func TestTxListReadyWithGasPriceBasic(t *testing.T) {
 	// Start nonce : 3
 	// baseFee : 30
 	// Transaction[0:9] has gasPrice 50
--- a/blockchain/tx_list_test.go
+++ b/blockchain/tx_list_test.go
@@ -21,7 +21,9 @@
 package blockchain
 
 import (
+	"math/big"
 	"math/rand"
+	"reflect"
 	"testing"
 
 	"github.com/klaytn/klaytn/blockchain/types"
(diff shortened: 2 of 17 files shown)


blockchain/tx_list_test.go
```diff
@@ -53,3 +57,147 @@ func TestStrictTxListAdd(t *testing.T) {
 		}
 	}
 }
+
+// TestTxListReadyWithGasPriceALL check whether ReadyWithGasPrice() works well.
```
kjhman21, line 61: which function is correct?

* TestTxListReadyWithGasPriceALL
* TestTxListReadyWithGasPrice

blockchain/tx_list_test.go
```diff
@@ -53,3 +57,147 @@ func TestStrictTxListAdd(t *testing.T) {
...
+// It makes a slice has 10 transactions and executes ReadyWithGasPrice() with baseFee 30.
```
kjhman21, line 62:
```suggestion
// It makes a slice of 10 transactions and executes ReadyWithGasPrice() with baseFee 30.
```

blockchain/tx_pool.go
```diff
@@ -655,24 +658,38 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
...
-			return ErrInvalidGasTipCap
-		}
-
-		if pool.gasPrice.Cmp(tx.GasFeeCap()) != 0 {
-			logger.Trace("fail to validate maxFeePerGas", "unitprice", pool.gasPrice, "maxFeePerGas", tx.GasTipCap())
-			return ErrInvalidGasFeeCap
+		//if tx.GasFeeCap().Cmp(tx.GasTipCap()) < 0 {
+		//	return ErrTipAboveFeeCap
+		//}
+		//
```
kjhman21, line 666: Do we need to remove this condition?

blockchain/tx_pool.go
```diff
@@ -655,24 +658,38 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
...
-			return ErrInvalidUnitPrice
+		// TODO : Need to KIP-71 hardfork.
+		//if pool.gasPrice.Cmp(tx.GasPrice()) != 0 {
+		//	logger.Trace("fail to validate unitprice", "unitPrice", pool.gasPrice, "txUnitPrice", tx.GasPrice())
+		//	return ErrInvalidUnitPrice
+		//}
+
+		// Ensure transaction's gasPrice is greater than or equal to transaction pool's gasPrice(baseFee).
+		if pool.gasPrice.Cmp(tx.GasPrice()) > 0 {
+			logger.Trace("fail to validate unitprice", "unitPrice", pool.gasPrice, "txUnitPrice", tx.GasPrice())
```
kjhman21, line 691:
```suggestion
			logger.Trace("fail to validate gasprice", "pool.gasPrice", pool.gasPrice, "tx.gasPrice", tx.GasPrice())
```

blockchain/tx_list.go
```diff
@@ -195,6 +195,37 @@ func (m *txSortedMap) Ready(start uint64) types.Transactions {
...
+// Note, all transactions with nonces lower than start will also be returned to
+// prevent getting into and invalid state. This is not something that should ever
+// happen but better to be self correcting than failing!
+func (m *txSortedMap) ReadyWithGasPrice(start uint64, baseFee *big.Int) types.Transactions {
+	// Short circuit if no transactions are available
+	if m.index.Len() == 0 || (*m.index)[0] > start {
+		return nil
+	}
+	// Otherwise start accumulating incremental transactions
+	var ready types.Transactions
```
kjhman21, line 212:
```suggestion
	ready := make(types.Transactions, 0, m.index.Len())
```
to avoid unnecessary memory allocation.

blockchain/tx_pool.go
```diff
@@ -1531,6 +1550,22 @@ func (pool *TxPool) demoteUnexecutables() {
...
+		if list.Len() > 0 {
+			needRemoved := false
+			for _, tx := range list.Flatten() {
+				hash := tx.Hash()
+				if needRemoved || tx.GasPrice().Cmp(pool.gasPrice) < 0 {
+					needRemoved = true
+					logger.Trace("Demoting transaction that has lower gas price than baseFee", "hash", hash)
+					list.Remove(tx)
+					pool.enqueueTx(hash, tx)
+				}
```
kjhman21, line 1565: I think we could print the exact log.
```suggestion
				if needRemoved || tx.GasPrice().Cmp(pool.gasPrice) < 0 {
					if needRemoved {
						logger.Trace("Demoting the transaction due to the previous transaction that has lower gas price than baseFee", "txhash", hash)
					} else {
						needRemoved = true
						logger.Trace("Demoting the transaction that has lower gas price than baseFee", "txhash", hash)
					}
					list.Remove(tx)
					pool.enqueueTx(hash, tx)
				}
```

blockchain/types/transaction.go
```diff
@@ -829,6 +829,26 @@ func TxDifference(a, b Transactions) (keep Transactions) {
...
+	for addr, list := range pending {
+		txs := list
+		for i, tx := range list {
+			if tx.GasPrice().Cmp(baseFee) < 0 {
+				txs = list[:i]
+				break
+			}
+		}
+
+		if len(txs) > 0 {
```
kjhman21, line 845: What if the first transaction is not satisfied in the above condition?
In this case len(txs)==0, then txMap is not updated. Is this the intended behavior?

blockchain/tx_list_test.go
sirano11: `TestTxListReadyWithGasPrice` is correct. I modified mistypo.

blockchain/types/transaction.go
sirano11: Yes, I intended this behavior. 
Since this list is sorted by nonce(only call `commitNewWork()`), if the first transaction does not satisfy the above conditions, the next transactions also must not be included.

blockchain/tx_pool.go
sirano11: I also removed this condition because `GasTipCap` field will be ignored.
Would you explain to me why this condition is needed?

blockchain/tx_pool_test.go
```diff
@@ -1897,59 +1909,367 @@ func TestDynamicFeeTransactionVeryHighValues(t *testing.T) {
...
+	t.Parallel()
+	baseFee := big.NewInt(30)
+
+	pool, key := setupTxPoolWithConfig(eip1559Config)
+	defer pool.Stop()
+	pool.SetGasPrice(baseFee)
+
+	testAddBalance(pool, crypto.PubkeyToAddress(key.PublicKey), big.NewInt(10000000000))
+
+	tx1 := pricedTransaction(0, 21000, big.NewInt(20), key)
```
ian0371, line 2051: tx1 -> tx

node/cn/handler.go
```diff
@@ -1117,8 +1117,8 @@ func (pm *ProtocolManager) BroadcastTxs(txs types.Transactions) {
 	// This function calls sendTransaction() to broadcast the transactions for each peer.
 	// In that case, transactions are sorted for each peer in sendTransaction().
 	// Therefore, it prevents sorting transactions by each peer.
-	if !sort.IsSorted(types.TxByPriceAndTime(txs)) {
-		sort.Sort(types.TxByPriceAndTime(txs))
+	if !sort.IsSorted(types.TxByTime(txs)) {
```
hyunsooda, line 1120: 1. `gasPrice` check first, and early arrived tx first if the `gasPrice` is the same in the prev implementation.
2. Early arrived tx first in this change.

This change seems not related to the new policy of the tx pool. Lack of motivation for this change in the description of the PR.

hyunsooda (commented):
1. There's no implementation of the increase and decrease in the gasPrice of the tx pool by the network congestion.
2. PR's description:
> The baseFee will be set as a gasPrice of txpool when occurring 'ChainHeadEvent`.
The baseFee will be set next blocks baseFee.

I couldn't find this implementation. Could you link me to that commit? or, did you forget to push that commit?

blockchain/tx_list.go
```diff
@@ -281,14 +312,24 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
 	// If there's an older better transaction, abort
 	old := l.txs.Get(tx.Nonce())
 	if old != nil {
-		if tx.Type().IsCancelTransaction() {
-			logger.Trace("New tx is a cancel transaction. replace it!", "old", old.String(), "new", tx.String())
-		} else {
```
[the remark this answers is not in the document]
kjeom: You can see there's new version for `cancelTransaction` below.
We added the replace mechanism for Tx with extra fee than older one.
think of the situation that cannot be dealt with due to an increase in price and the user want to pay more than changed `basefee`
This code will be fixed after the governance parameters PR is merged.

blockchain/tx_list_test.go
```diff
@@ -53,3 +56,147 @@ func TestStrictTxListAdd(t *testing.T) {
...
+// It makes a slice has 10 transactions and executes ReadyWithGasPrice() with baseFee 20.
+//
+// It checks whether filtering works well if there is a transaction
+// with a gasPrice lower than the baseFee among Tx.
+func TestTxListReadyWithGasPricePartialFilter(t *testing.T) {
+	// Start nonce : 3
+	// baseFee : 20
+	// Transaction[0:6] has gasPrice 30
+	// Transaction[7] has gasPrice 10
+	// Transaction[8:9] has gasPrice 50
```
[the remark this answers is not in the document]
sirano11: No, Transaction[8:9] are filtered because Transaction[7] has gas price 10(lower than baseFee).

blockchain/types/transaction.go
```diff
@@ -889,18 +930,18 @@ func (t *TransactionsByPriceAndNonce) Count() (int, int) {
...
 }
 
-func (t *TransactionsByPriceAndNonce) Txs() map[common.Address]Transactions {
+func (t *TransactionsByTimeAndNonce) Txs() map[common.Address]Transactions {
 	return t.txs
 }
 
-// NewTransactionsByPriceAndNonce creates a transaction set that can retrieve
+// NewTransactionsByTimeAndNonce creates a transaction set that can retrieve
 // price sorted transactions in a nonce-honouring way.
```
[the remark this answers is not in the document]
sirano11: Thanks. I fixed it.

blockchain/tx_pool_test.go
sirano11: I fixed it.

blockchain/tx_list.go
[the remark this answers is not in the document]
sirano11: This is because the behavior before and after the hardfork must be separated. Therefore, I wrote the following at the end of the PR explanation.
`These changes need to be hardfork, i commented existing code. TODO : Need to KIP-71 hardfork.`

blockchain/tx_list.go
```diff
@@ -281,14 +312,24 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
...
+		//	return false, nil
+		//}
+
+		// If gas price of older is bigger than newer, abort.
+		if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
 			logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
 			return false, nil
+		} else {
+			// Otherwise overwrite the old transaction with the current one
+			logger.Trace("Gas price of new tx is bigger than older tx!", "old", old.String(), "new", tx.String())
```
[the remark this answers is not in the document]
sirano11: Please see above comments.

blockchain/tx_list.go
```diff
@@ -281,14 +312,24 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
 	// If there's an older better transaction, abort
```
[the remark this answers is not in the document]
sirano11: You are right. 
`priceBump` parameter has been used in geth currently, but It seems to removed logic related `priceBump` in klaytn.

blockchain/tx_pool_test.go
```diff
@@ -1897,59 +1909,367 @@ func TestDynamicFeeTransactionVeryHighValues(t *testing.T) {
...
+		t.Error("error", "got", err)
 	}
 
-	if err := pool.AddLocal(tx); err != ErrTxTypeNotSupported {
-		t.Error("expected", ErrTxTypeNotSupported, "got", err)
+	// The transaction's gasPrice bigger than baseFee.
+	tx2 := pricedTransaction(1, 21000, big.NewInt(40), key)
+	if err := pool.AddRemote(tx2); err != nil {
+		t.Error("error", "got", err)
+	}
```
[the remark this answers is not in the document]
sirano11: I don't think this block don't need to loop.

blockchain/tx_list_test.go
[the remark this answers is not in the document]
ian0371: I was wondering about the mechanism as to why they're dropped. You mean when Transaction[7] is dropped, the remaining txs in the txlist are dropped simultaneously, right?

blockchain/tx_pool_test.go
```diff
@@ -1897,59 +1909,367 @@ func TestDynamicFeeTransactionVeryHighValues(t *testing.T) {
...
+	tx2 := dynamicFeeTx(1, 21000, big.NewInt(40), big.NewInt(30), key)
 	if err := pool.AddRemote(tx2); err != nil {
 		t.Error("error", "got", err)
 	}
+
+	// The GasFeeCap greater than baseFee and gasTipCap is bigger than baseFee(ignored).
+	tx3 := dynamicFeeTx(2, 21000, big.NewInt(50), big.NewInt(50), key)
+	if err := pool.AddRemote(tx3); err != nil {
+		t.Error("error", "got", err)
+	}
```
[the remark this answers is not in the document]
hyunsooda: Refer to below.
```
fees := []*big.Int{big.NewInt(30), big.NewInt(40), big.NewInt(50)}
tips := []*big.Int{big.NewInt(1), big.NewInt(30), big.NewInt(50)}
for i := 0; i < len(fees); i++ {
    tx := dynamicFeeTx(uint64(i), 21000, fees[i], tips[i], key)
    if err := pool.AddRemote(tx); err != nil {
        t.Error("error", "got", err)
    }
}
```

The redundant code (sequential paste) makes readers read a lot of stuff, even not easy to fix or extend.

blockchain/tx_pool_test.go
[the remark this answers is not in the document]
hyunsooda: Please take into consideration the comment above.

blockchain/tx_list.go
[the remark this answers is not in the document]
hyunsooda: Now I see. If I correctly understand it, the type of cancel transaction should be removed as well as related functionalities. Is it Correct?

blockchain/tx_list.go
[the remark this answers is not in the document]
hyunsooda: What do you mean? I'm pointing out the semantics of the logging. The new tx will substitute the older tx if the gas price is higher which is verified with that logging.

blockchain/tx_list.go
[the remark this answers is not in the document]
sirano11: No, Cancel transaction also remained as a transaction type. It just intended to separate logic before and after hardfork.

blockchain/tx_list.go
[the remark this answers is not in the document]
sirano11: I just intended to maintain an existed code and seperate logic before and after hardfork.
It will be uncommented and modified after merging governance related PR.

blockchain/tx_pool_test.go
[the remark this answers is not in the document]
sirano11: If you do it the way you suggested, the code will be concise, but I think it will reduce readability. 
I still don't think this code needs to use a loop, and I think it's more intuitive to write it like this for others to figure out the code.

blockchain/tx_list.go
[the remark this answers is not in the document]
hyunsooda: The log is already uncommented and was created by your change too. Moreover, the semantics of the log would be preserved because it definitely substitutes the older tx. Very confusing.

blockchain/tx_pool_test.go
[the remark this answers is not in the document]
hyunsooda: If the code is as concise as possible, the readability increase. For example, if the readers figure the what loops are doing out, the only initialization part is the only one to be a key implication of the test. Otherwise like your current change, the readers must see all the code at all times.

blockchain/tx_list.go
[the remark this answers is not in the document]
sirano11: Sorry.. I just confused. Your suggestion is more correctly. Thanks for review.

blockchain/tx_pool_test.go
```diff
@@ -1897,59 +1909,367 @@ func TestDynamicFeeTransactionVeryHighValues(t *testing.T) {
...
+	txs = append(txs, pricedTransaction(3, 100000, big.NewInt(30), key))
+
+	pool.enqueueTx(txs[0].Hash(), txs[0])
+	pool.enqueueTx(txs[1].Hash(), txs[1])
+	pool.enqueueTx(txs[2].Hash(), txs[2])
+	pool.enqueueTx(txs[3].Hash(), txs[3])
+
+	// set baseFee to 20.
+	baseFee = big.NewInt(30)
+	pool.gasPrice = baseFee
```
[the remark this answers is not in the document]
sirano11: This is necessary code.

blockchain/types/transaction.go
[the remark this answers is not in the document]
ian0371: Please also fix old file's line 849, 902, 920.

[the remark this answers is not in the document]
sirano11: I fixed it.

kjhman21 (approved):
LGTM

blockchain/tx_list.go
```diff
@@ -195,6 +195,37 @@ func (m *txSortedMap) Ready(start uint64) types.Transactions {
 	return ready
 }
 
+// ReadyWithGasPrice retrieves a sequentially increasing list of transactions that greater than or
```
jeongkyun-oh, line 198: Any reason why `Ready` method remains? How about adding deprecated comment?

blockchain/tx_pool.go
```diff
@@ -1531,6 +1550,26 @@ func (pool *TxPool) demoteUnexecutables() {
 			}
 		}
 
+		// TODO : Need to KIP-71 hardfork.
+		// Enqueue transaction if gasPrice of transaction is lower than gasPrice of txPool.
+		// All transactions with a nonce greater than enqueued transaction also stored queue.
+		if list.Len() > 0 {
+			needRemoved := false
```
yoomee1313, line 1557: It seems that you use `needRemoved` variable to remove out the txs which has bigger nonce than the tx which has invalid basefee. I think this is similar concept to the `strict`. 
https://github.com/klaytn/klaytn/blob/ed593ea89f26856a77cf78379a91d7abefd933ed/blockchain/tx_pool.go#L942
Can this concept be used instead of implementing remove-out logic at here?

blockchain/types/transaction.go
```diff
@@ -829,6 +829,26 @@ func TxDifference(a, b Transactions) (keep Transactions) {
 	return keep
 }
 
+// FilterTransactionWithBaseFee returns a list of transactions for each account that filters transactions
+// that are greater than or equal to baseFee.
+func FilterTransactionWithBaseFee(pending map[common.Address]Transactions, baseFee *big.Int) (map[common.Address]Transactions, error) {
```
jeongkyun-oh, line 834: Where this method is used?

blockchain/tx_list.go
```diff
@@ -281,14 +312,24 @@ func (l *txList) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Tran
...
+		//	logger.Trace("New tx is a cancel transaction. replace it!", "old", old.String(), "new", tx.String())
+		//} else {
+		//	logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
+		//	return false, nil
+		//}
+
+		// If gas price of older is bigger than newer, abort.
+		if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
 			logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
 			return false, nil
```
aidan-kwon, line 326: @sirano11 @kjhman21 I want to confirm the policy of the cancel transaction. Which option are you thinking?
1. Cancel transaction can cancel a transaction having the same nonce regardless of gas price 
2. Cancel transaction is handled as other transaction types, so it can cancel a transaction only when it has more gas price than the predecessor.

aidan-kwon (commented):
@sirano11  Klaytn will not accept transactions having a lower gas price than baseFee during block generation and insertion also. Will you implement this in another PR?

blockchain/types/transaction.go
kjeom: @jeongkyun-oh It is used in `work/worker.go` where filtering txs lower than `basefee`

blockchain/tx_list.go
kjeom: @aidan-kwon Please remove a duplicated review.

blockchain/tx_list.go
jeongkyun-oh: I heard that this will be used with hardfork.

jeongkyun-oh (approved):

blockchain/tx_pool.go
kjeom: I think so. If `tx.GasPrice( )` < `pool.gasPrice`, you don't need to remove `txs` after the `tx's nonce`
Therefore the log, `Demotinc the transaction due to the previous transaction ...` will not appear.
@yoomee1313 @sirano11 How do you think about following code?
```go
if list.Len() > 0 {
	for _, tx := range list.Flatten() {
		hash := tx.Hash()
		if tx.GasPrice().Cmp(pool.gasPrice) < 0 {
		        logger.Trace("Demoting the tx that is lower than the baseFee and those greater than the nonce of the tx.", "txhash", hash)
			list.Remove(tx) // delete all transactions satisfying the nonce value > tx.Nonce( )
			pool.enqueueTx(hash, tx)
			break
		}
	}
}
```

aidan-kwon (approved):

blockchain/tx_pool.go
yoomee1313: I think if you want to fix the code like that, you need to enqueue the filtered-out txs(txs which nonce is bigger than a removed tx), too. It seems that the filtered-out txs are returned by the list.Remove(tx).

kjeom: Thanks, I edited the proposal code.

yoomee1313: It looks nice, could you add the checking logic `if invalids == nil` when `removed` is true? The Remove returns `true,nil` when strict mode is false. I think strict mode is always true at here because the txList is the element of `pending queue`, right?

kjeom: You are right. 
But I think `for _, invalidTx := range invalids` works like `if invalids == nil` when `invalids == nil`
Actually I don't prefer the implicit code above like remove( ) that is complicated. I think it is better to split remove( ) and removeGreaterThanNonce( ).  However this function is used in many place, so I'd like to postpone this refactoring issue later.

yoomee1313: @kjeom Oh, you are right. Thanks!

yoomee1313 (approved):

blockchain/tx_pool.go
kjeom: Thanks!, I'll resolve it and I'm looking forward to discuss it again later

Issue #1200, opened 2022-02-21
Node synchronization not started


aidan-kwon:
Environment
- 4 CNs are running in a private network

Issue
- CN0 doesn't increase blocknumber. It keet printing round change logs and doesn't get blocks from other CNs
- Other three CNs were working normally. They only have single round change periodically (Probably when CN0 was a proposer)

Action
- Even after a few CN0 manual restarts, the same issue was continuos 
- I called `debug.setHead` with (recent blockNumber - N). After tens of minutes, it start to sync and keep operating well. 

<img width="606" alt="image" src="https://user-images.githubusercontent.com/48199072/155039114-607a9096-d225-4c8b-9ad4-6689ebd55bb0.png">

aidan-kwon:
It seems that this issue is solved via https://github.com/klaytn/klaytn/pull/1389

PR #1389, opened 2022-05-20
peer: fix multiChannelPeer ReadMsg loop to read closed ch immediately

yoomee1313:
## Proposed changes
- In the case of a CN, since the node can mine a block, when it is determined that the sync is complete, it switches to the mining mode. However, if it is a multichannel<->multichannel connection, and if the sync is completed before reaching the latest block, the CN should be converted to the sync mode in a few seconds. However, it did not converted to the sync mode immediately. It seems solve next issue, but actually it is not. https://github.com/klaytn/klaytn/issues/1200.
- It seems next causes are the reason of this issue, and i think (2) cause is direct.
  - (1) A message is read from the p2p connection, and if it is determined that there is an error, the p2p connection is disconnected. Becuase istanbul engine is shutdown due to the syncing, the istanbul consensus msg can be received. However, it is considered an error, so p2p connection is disconnected.
  - (2) But, in the case of multichannel, the p2p connection is not properly disconnected, so the corrupted connection is lasts more than 30 seconds.
- Therefore, in this PR, I tried to solve the above two problems. I thought a lot about whether it is correct to solve (1) as well. While reviewing, I would like you to look on this part.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -33,12 +33,13 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/klaytn/klaytn/consensus/istanbul"
+
 	"github.com/klaytn/klaytn/accounts"
 	"github.com/klaytn/klaytn/blockchain"
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/consensus"
-	"github.com/klaytn/klaytn/consensus/istanbul"
 	"github.com/klaytn/klaytn/crypto"
 	"github.com/klaytn/klaytn/datasync/downloader"
 	"github.com/klaytn/klaytn/datasync/fetcher"
@@ -531,13 +532,13 @@ func (pm *ProtocolManager) processMsg(msgCh <-chan p2p.Msg, p Peer, addr common.
 func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, addr common.Address, errCh chan<- error) {
 	for msg := range msgCh {
 		if handler, ok := pm.engine.(consensus.Handler); ok {
-			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
-			if err == istanbul.ErrStoppedEngine {
+			if _, err := handler.HandleMsg(addr, msg); err != nil {
 				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization.", "msg", msg, "err", err)
-			} else if err != nil {
-				errCh <- err
-				return
+				if err != istanbul.ErrStoppedEngine {
+					errCh <- err
+					return
+				}
 			}
 		}
 		msg.Discard()
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -33,13 +33,12 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/klaytn/klaytn/consensus/istanbul"
-
 	"github.com/klaytn/klaytn/accounts"
 	"github.com/klaytn/klaytn/blockchain"
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/consensus"
+	"github.com/klaytn/klaytn/consensus/istanbul"
 	"github.com/klaytn/klaytn/crypto"
 	"github.com/klaytn/klaytn/datasync/downloader"
 	"github.com/klaytn/klaytn/datasync/fetcher"
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -531,13 +531,12 @@ func (pm *ProtocolManager) processMsg(msgCh <-chan p2p.Msg, p Peer, addr common.
 func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, addr common.Address, errCh chan<- error) {
 	for msg := range msgCh {
 		if handler, ok := pm.engine.(consensus.Handler); ok {
+			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
-			if _, err := handler.HandleMsg(addr, msg); err != nil {
+			if err != nil {
 				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization.", "msg", msg, "err", err)
-				if err != istanbul.ErrStoppedEngine {
-					errCh <- err
-					return
-				}
+				errCh <- err
+				return
 			}
 		}
 		msg.Discard()
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -533,8 +533,9 @@ func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, add
 		if handler, ok := pm.engine.(consensus.Handler); ok {
 			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
-			if err != nil {
+			if err == istanbul.ErrStoppedEngine {
 				p.GetP2PPeer().Log().Warn("ProtocolManager failed to handle consensus message. This can happen during block synchronization.", "msg", msg, "err", err)
+			} else if err != nil {
 				errCh <- err
 				return
 			}
--- a/node/cn/peer.go
+++ b/node/cn/peer.go
@@ -1000,6 +1000,7 @@ func (p *multiChannelPeer) ReadMsg(rw p2p.MsgReadWriter, connectionOrder int, er
 	}, channelSizePerPeer)
 	go func() {
 		for {
+			// TODO-klaytn: check 30-second timeout works
 			msg, err := rw.ReadMsg()
 			readMsgCh <- struct {
 				p2p.Msg
--- a/node/cn/peer.go
+++ b/node/cn/peer.go
@@ -1010,28 +1010,36 @@ func (p *multiChannelPeer) ReadMsg(rw p2p.MsgReadWriter, connectionOrder int, er
 	}()
 
 	for {
+		var (
+			msg p2p.Msg
+			err error
+		)
 		select {
 		case pair := <-readMsgCh:
-			msg, err := pair.Msg, pair.error
-			if err != nil {
-				p.GetP2PPeer().Log().Warn("ProtocolManager failed to read msg", "err", err)
-				errCh <- err
-				return
-			}
-			msgCh, err := p.chMgr.GetChannelWithMsgCode(connectionOrder, msg.Code)
-			if err != nil {
-				p.GetP2PPeer().Log().Warn("ProtocolManager failed to get msg channel", "err", err)
-				errCh <- err
-				return
-			}
+			msg, err = pair.Msg, pair.error
+		case <-closed:
+			return
+		}
 
-			if msg.Size > ProtocolMaxMsgSize {
-				err = errResp(ErrMsgTooLarge, "%v > %v", msg.Size, ProtocolMaxMsgSize)
-				p.GetP2PPeer().Log().Warn("ProtocolManager over max msg size", "err", err)
-				errCh <- err
-				return
-			}
-			msgCh <- msg
+		if err != nil {
+			p.GetP2PPeer().Log().Warn("ProtocolManager failed to read msg", "err", err)
+			errCh <- err
+			return
+		}
+		msgCh, err := p.chMgr.GetChannelWithMsgCode(connectionOrder, msg.Code)
+		if err != nil {
+			p.GetP2PPeer().Log().Warn("ProtocolManager failed to get msg channel", "err", err)
+			errCh <- err
+			return
+		}
+		if msg.Size > ProtocolMaxMsgSize {
+			err = errResp(ErrMsgTooLarge, "%v > %v", msg.Size, ProtocolMaxMsgSize)
+			p.GetP2PPeer().Log().Warn("ProtocolManager over max msg size", "err", err)
+			errCh <- err
+			return
+		}
+		select {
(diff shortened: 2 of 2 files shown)


hyunsooda:
Hi @yoomee1313, 

What difference was logically made by your change? The `readLoopEndCh` channel was introduced in this change but seemed not to make a logical difference in my view. Could you elaborate on what difference was created in your change? 

Furthermore, the reader thread may cause a resource leak (the thread never exits) if there's no error found in the reader thread, but the outer function (`ReadMsg()`) returns by error.

yoomee1313:
Hi, @hyunsooda Thanks for asking in detail. 
> What difference was logically made by your change? The readLoopEndCh channel was introduced in this change but seemed not to make a logical difference in my view. Could you elaborate on what difference was created in your change?

Looking at the code itself, the problem is that when a message is received from a `closed` channel, the reader threads(ReadMsg goroutine) must be terminated. However,  in certain circumstances, the thread is hung at `msg, err := rw.ReadMsg()`, and it was impossible to enter `select`. So, I changed the code not to be affected by `msg, err := rw.ReadMsg()`. In other words, it can receive the msg from `closed` loop without blocking.

> Furthermore, the reader thread may cause a resource leak (the thread never exits) if there's no error found in the reader thread, but the outer function (ReadMsg()) returns by error.

In the `Handle`, a connection is established between two peers. Looking at the last part, if an error is received through `errChannel` from processConsensusMsg loop or processMsg loop, the `closed` channel is closed and accordingly all reader goroutines are terminated. Then, `Handle` waits until all goroutines are finished.

To explain my understanding of this code, it seems to keep the threads as long as the p2p connection is persistent or there are no errors. I'm not sure if it was answered.

https://github.com/klaytn/klaytn/blob/0d5d00455d22b0b99ce7dc2fa237d29004c7c2be/node/cn/peer.go#L1080-L1101

hyunsooda:
Gottcha. Now I understood your change. 

Let me describe the comment below again:
> Furthermore, the reader thread may cause a resource leak (the thread never exits) if there's no error found in the reader thread, but the outer function (ReadMsg()) returns by error.

The go-routine (line 974 to 983) may not be exited if the for-loop (line 987 to 1003) exits by error or closed signal, meaning that the resource (thread) may not be reclaimed.

node/cn/peer.go
```diff
@@ -965,30 +965,42 @@ func (p *multiChannelPeer) UpdateRWImplementationVersion() {
...
+	)
 
-		if msg.Size > ProtocolMaxMsgSize {
-			err := errResp(ErrMsgTooLarge, "%v > %v", msg.Size, ProtocolMaxMsgSize)
-			p.GetP2PPeer().Log().Warn("ProtocolManager over max msg size", "err", err)
-			errCh <- err
-			return
+	go func() {
+		for {
+			msg, err := rw.ReadMsg()
```
ian0371, line 976: If `rw.ReadMsg()` hangs, then the goroutine never returns.
How about adding a timeout before calling `rw.ReadMsg()` in the old code?

yoomee1313: @ian0371 thanks, ian. I think about your comment a lot, but I can't imagine well.. It is hard to think up the possible case of passing through `rw.ReadMsg` with timer. 

First of all,`rw.ReadMsg` hanging will not be happened because it returns err as soon as the peer is disconnected. I have tested several times, and it always return the error.

Even assumed that the ReadMsg hangs, it has it's own timer which is set to 30s, so it can't be hung forever.
https://github.com/klaytn/klaytn/blob/0326446c5273f50aa4cd3e1ddb03d755c48d2310/networks/p2p/rlpx.go#L101

btw, this PR contains the change about avoiding the unnecessary wating when error is not related to the timeout(ex. if timeout, err would be EOF).

@hyunsooda I think this can be the answer to your comment, too. thanks for the great advice, and Is there any other concern? btw, sorry for the late answer.

ian0371: To sum up,
1. rw.ReadMsg() waits at most 30 seconds
2. If the channel is closed, it waits until the timeout
3. However, if the channel is closed, there's no need to wait. We want to execute `case <-closed` right away.

This seems like a clever way to solve the problem. What do you think @hyunsooda ?

ian0371: Or is this problematic?

```diff
diff --git a/node/cn/peer.go b/node/cn/peer.go
index f0983313..271199ae 100644
--- a/node/cn/peer.go
+++ b/node/cn/peer.go
@@ -966,6 +966,12 @@ func (p *multiChannelPeer) UpdateRWImplementationVersion() {
 func (p *multiChannelPeer) ReadMsg(rw p2p.MsgReadWriter, connectionOrder int, errCh chan<- error, wg *sync.WaitGroup, closed <-chan struct{}) {
        defer wg.Done()
        for {
+               select {
+               case <-closed:
+                       return
+               default:
+               }
+
                msg, err := rw.ReadMsg()
                if err != nil {
                        p.GetP2PPeer().Log().Warn("ProtocolManager failed to read msg", "err", err)
@@ -989,8 +995,6 @@ func (p *multiChannelPeer) ReadMsg(rw p2p.MsgReadWriter, connectionOrder int, er

                select {
                case msgCh <- msg:
-               case <-closed:
-                       return
                }
        }
 }
```

yoomee1313: I have tested with the above code, and it is hung at `case <-closed:` because it waits for the closed channel.

node/cn/handler.go
```diff
@@ -522,8 +523,9 @@ func (pm *ProtocolManager) processConsensusMsg(msgCh <-chan p2p.Msg, p Peer, add
 		if handler, ok := pm.engine.(consensus.Handler); ok {
 			_, err := handler.HandleMsg(addr, msg)
 			// if msg is istanbul msg, handled is true and err is nil if handle msg is successful.
-			if err != nil {
+			if err == istanbul.ErrStoppedEngine {
```
aidan-kwon, line 526: Why don't you send an error to `errCh` for this?

node/cn/peer.go
```diff
@@ -965,30 +965,42 @@ func (p *multiChannelPeer) UpdateRWImplementationVersion() {
...
+			if err != nil {
+				p.GetP2PPeer().Log().Warn("ProtocolManager failed to read msg", "err", err)
+				errCh <- err
+				close(readLoopEndCh)
```
hyunsooda, line 980:
```suggestion
```
@yoomee1313, Good to know a read timeout was already placed there. Okay looks fine. Another concern from here is closing the closed channel. The possible scenario is that `readLoopEndCh` may be already closed, but the late closing again by the timeout can encounter runtime panic. Removal of this code looks fine in my view. if the right above line `errCh <- err` is executed, the loop (line 987 ~ 1007) will be exited itself by line 1111.

node/cn/handler.go
yoomee1313: This error is printed at here when shutting down the istanbul engine, and it is lead to the disconnection. I think disconnection is not proper. The reason is because the shutdown of an istanbul engine is caused by sync. During sync, it still needs connection. Rather, the two peer is disconnected.

As always, my thought can be wrong, so if you have any concern, i can revert back to disconnect the connection.

node/cn/peer.go
yoomee1313: @hyunsooda yes, you're right. I removed the chennel. Thanks!

hyunsooda (dismissed):

node/cn/peer.go
```diff
@@ -965,30 +965,33 @@ func (p *multiChannelPeer) UpdateRWImplementationVersion() {
...
+			msgCh, err := p.chMgr.GetChannelWithMsgCode(connectionOrder, msg.Code)
+			if err != nil {
+				p.GetP2PPeer().Log().Warn("ProtocolManager failed to get msg channel", "err", err)
+				errCh <- err
+			} else if msg.Size > ProtocolMaxMsgSize {
+				err := errResp(ErrMsgTooLarge, "%v > %v", msg.Size, ProtocolMaxMsgSize)
+				p.GetP2PPeer().Log().Warn("ProtocolManager over max msg size", "err", err)
+				errCh <- err
+			}
+			msgCh <- msg
```
aidan-kwon, line 994: Is is okay to send msg to msgCh when `err != nil` ?

node/cn/handler.go
aidan-kwon: In the currency full mesh CN network, your opinion is right. I am a little bit afraid of unexpected situations, but we can evaluate them in the QA process. 
Thank you!

node/cn/peer.go
```diff
@@ -965,30 +965,33 @@ func (p *multiChannelPeer) UpdateRWImplementationVersion() {
...
-			return
-		}
-
-		if msg.Size > ProtocolMaxMsgSize {
-			err := errResp(ErrMsgTooLarge, "%v > %v", msg.Size, ProtocolMaxMsgSize)
-			p.GetP2PPeer().Log().Warn("ProtocolManager over max msg size", "err", err)
-			errCh <- err
-			return
+	readMsgCh := make(chan p2p.Msg, channelSizePerPeer)
+	go func() {
```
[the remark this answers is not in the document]
yoomee1313: @aidan-kwon Thanks for letting me know about those side effects.
> This logic will generate a goroutine when it receives a message. Is it okay in the performance aspect?

-> The goroutine is generated per peer connection, so i think performance is not affected easily. However, the goroutine communicates with errCh.. I move the error check logic back to main loop. https://github.com/klaytn/klaytn/pull/1389/commits/9dbccc9b7b52b22dcfbcde282bb29ce0e33bceeb.
I conclusion, the goroutine only receives the msg, and it is passed to the main loop and processed.

> This goroutine isn't terminated gracefully when a node stopped, is it?

-> I have checked there is a 30 second timeout, but I'm not sure it really work as i expect. I'll figure it out later. https://github.com/klaytn/klaytn/pull/1389/commits/14d2345403c1a8ba4b56db62204464a3a38408ea

node/cn/peer.go
yoomee1313: Thanks, `return` is ommitted. I added at here https://github.com/klaytn/klaytn/pull/1389/commits/e0c0d77c53127e786fdb20b172d1db60f3a9b74b.

aidan-kwon: When `msgCh` is blocked, it cannot reach `case <-closed:` line. Is it okay?

blukat29: @aidan-kwon @yoomee1313 How about this?
This way, we keep the original `select case msgCh <- msg`.

```go
for {
    var msg Message
    var err error
    select {
    case pair := <-readMsgCh:
        msg, err = pair.Msg, pair.Error
    case <-closed:
        return
    }
    
    if err != nil { ... return } 

    select {
    case msgCh <- msg:
    case <-closed:
        return
    }
}
```

node/cn/peer.go
```diff
@@ -993,28 +993,51 @@ func (p *multiChannelPeer) UpdateRWImplementationVersion() {
...
+		for {
+			// TODO-klaytn: check 30-second timeout works
+			msg, err := rw.ReadMsg()
+			readMsgCh <- struct {
+				p2p.Msg
+				error
+			}{msg, err}
+		}
+	}()
```
jeongkyun-oh, line 1010: Will this goroutine be terminated?

node/cn/peer.go
yoomee1313: @aidan-kwon yes, as you said, it can be blocked..
@blukat29 it helps a lot! It solves the problem, thank you. https://github.com/klaytn/klaytn/pull/1389/commits/8161c47eb91fbb02cf4df6fc6977ade028b65c45.

aidan-kwon (approved):

node/cn/peer.go
yoomee1313: @jeongkyun-oh 
> Will this goroutine be terminated?

The child goroutine is expected to be terminated when the ReadMsg function is exited.
But to make sure, i changed the code for child goroutine to wait `closed`, also. If `closed` is close, the child goroutine can also receive that signal, and exit.
Can you take a look? https://github.com/klaytn/klaytn/pull/1389/commits/8f988f8458dc0ea3ed7c3d0a79de1eee65233605

> It looks better to launch a goroutine out of ReadMsg

Can I ask the reason why you think that launching this goroutine outside better? In my opinion, inside the ReadMsg looks better because the goroutine is dependent at ReadMsg.

node/cn/peer.go
```diff
@@ -993,28 +993,51 @@ func (p *multiChannelPeer) UpdateRWImplementationVersion() {
...
-		msg, err := rw.ReadMsg()
+		var (
+			msg p2p.Msg
+			err error
+		)
+		select {
+		case pair := <-readMsgCh:
+			msg, err = pair.Msg, pair.error
+		case <-closed:
+			return
```
hqjang-pepper, line 1025:
```suggestion
		case <-closed:
		        close(readMsgCh)
			return
```

How about adding a close channel logic when for loop is closed?

hqjang-pepper: Oh you added a close logic at https://github.com/klaytn/klaytn/pull/1389/commits/8f988f8458dc0ea3ed7c3d0a79de1eee65233605.

yoomee1313: thanks for the comment, i changed the child loop to receive close signal when closing `closed` channel.

hqjang-pepper (approved):

jiseongnoh (approved):

node/cn/peer.go
jeongkyun-oh: > The child goroutine is expected to be terminated when the ReadMsg function is exited.
But to make sure, i changed the code for child goroutine to wait closed, also. If closed is close, the child goroutine can also receive that signal, and exit.
Can you take a look? https://github.com/klaytn/klaytn/commit/8f988f8458dc0ea3ed7c3d0a79de1eee65233605

I think still the goroutine should be terminated before the main function terminated. But, it seems there is no panic or hang in the goroutine for now, so resolved this comment.

> Can I ask the reason why you think that launching this goroutine outside better? In my opinion, inside the ReadMsg looks better because the goroutine is dependent at ReadMsg.

Okay.

jeongkyun-oh (approved):

hyunsooda (approved):

ian0371 (approved):

PR #1394, opened 2022-05-23
Implemented staking information downloader

jeongkyun-oh:
## Proposed changes

Staking information is about the data which is needed to calculate proposer list as well as validators. The data is stored in a contract which can be retrieved from a contract method. However, while snap syncing, staking information cannot be retrieved since state tries do not exist until a pivot block. Remote peers store those data in `misc` db for state-db migration feature, so this PR added a process to download staking information from remote peers.

### Protocol
- Staking information download is needed so that new message types (`StakingInfoRequestMsg`, `StakingInfoMsg`) are added
- The common messages took from `0x00` to `0x0f` and istanbul engine took `0x11`, but not for gxhash and clique.
- Also, the protocol is updated to `klay/64` with two dummy messages (`0x10`, `0x11`) and added two staking info messages (0x12, 0x13) with `klay/65`.

### Staking info download process

- Klaytn supports not fast but only full synchronization. Since full sync doesn't need to download staking info download, it is added when fast sync is on.
- The sync process is like described below and `StakingInfoQueue` and `fetchStakingInfo` are added on this PR.

스크린샷 2022-05-24 오전 10 46 39

### TODO

- Staking information is fetched from remote peers, but its consistency cannot be proved. So, the more proof is needed
  - staking information hashes should be compared with more than 2 peers
  - another staking information hash chain to prove the consistency after pivot block
- This will be discussed and implemented later

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

### DIFF
--- a/datasync/downloader/downloader.go
+++ b/datasync/downloader/downloader.go
@@ -28,14 +28,13 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/klaytn/klaytn/reward"
-
 	"github.com/klaytn/klaytn"
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/event"
 	"github.com/klaytn/klaytn/log"
 	"github.com/klaytn/klaytn/params"
+	"github.com/klaytn/klaytn/reward"
 	"github.com/klaytn/klaytn/storage/database"
 	"github.com/klaytn/klaytn/storage/statedb"
 	"github.com/rcrowley/go-metrics"
--- a/datasync/downloader/downloader.go
+++ b/datasync/downloader/downloader.go
@@ -28,6 +28,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/klaytn/klaytn/reward"
+
 	"github.com/klaytn/klaytn"
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
@@ -40,13 +42,14 @@ import (
 )
 
 var (
-	MaxHashFetch    = 512 // Amount of hashes to be fetched per retrieval request
-	MaxBlockFetch   = 128 // Amount of blocks to be fetched per retrieval request
-	MaxHeaderFetch  = 192 // Amount of block headers to be fetched per retrieval request
-	MaxSkeletonSize = 128 // Number of header fetches to need for a skeleton assembly
-	MaxBodyFetch    = 128 // Amount of block bodies to be fetched per retrieval request
-	MaxReceiptFetch = 256 // Amount of transaction receipts to allow fetching per request
-	MaxStateFetch   = 384 // Amount of node state values to allow fetching per request
+	MaxHashFetch        = 512 // Amount of hashes to be fetched per retrieval request
+	MaxBlockFetch       = 128 // Amount of blocks to be fetched per retrieval request
+	MaxHeaderFetch      = 192 // Amount of block headers to be fetched per retrieval request
+	MaxSkeletonSize     = 128 // Number of header fetches to need for a skeleton assembly
+	MaxBodyFetch        = 128 // Amount of block bodies to be fetched per retrieval request
+	MaxReceiptFetch     = 256 // Amount of transaction receipts to allow fetching per request
+	MaxStakingInfoFetch = 128 // Amount of staking information to allow fetching per request
+	MaxStateFetch       = 384 // Amount of node state values to allow fetching per request
 
 	MaxForkAncestry  = 3 * params.EpochDuration // Maximum chain reorganisation
 	rttMinEstimate   = 2 * time.Second          // Minimum round-trip time to target for download requests
@@ -124,12 +127,14 @@ type Downloader struct {
 	committed       int32
 
 	// Channels
-	headerCh      chan dataPack        // [klay/62] Channel receiving inbound block headers
-	bodyCh        chan dataPack        // [klay/62] Channel receiving inbound block bodies
-	receiptCh     chan dataPack        // [klay/63] Channel receiving inbound receipts
-	bodyWakeCh    chan bool            // [klay/62] Channel to signal the block body fetcher of new tasks
-	receiptWakeCh chan bool            // [klay/63] Channel to signal the receipt fetcher of new tasks
-	headerProcCh  chan []*types.Header // [klay/62] Channel to feed the header processor new tasks
+	headerCh          chan dataPack        // [klay/62] Channel receiving inbound block headers
+	bodyCh            chan dataPack        // [klay/62] Channel receiving inbound block bodies
+	receiptCh         chan dataPack        // [klay/63] Channel receiving inbound receipts
+	stakingInfoCh     chan dataPack        // [klay/64] Channel receiving inbound staking infos
+	bodyWakeCh        chan bool            // [klay/62] Channel to signal the block body fetcher of new tasks
+	receiptWakeCh     chan bool            // [klay/63] Channel to signal the receipt fetcher of new tasks
+	stakingInfoWakeCh chan bool            // [klay/64] Channel to signal the staking info fetcher of new tasks
+	headerProcCh      chan []*types.Header // [klay/62] Channel to feed the header processor new tasks
 
 	// for stateFetcher
 	stateSyncStart chan *stateSync
@@ -146,10 +151,11 @@ type Downloader struct {
 	quitLock sync.RWMutex  // Lock to prevent double closes
 
 	// Testing hooks
-	syncInitHook     func(uint64, uint64)  // Method to call upon initiating a new sync run
-	bodyFetchHook    func([]*types.Header) // Method to call upon starting a block body fetch
-	receiptFetchHook func([]*types.Header) // Method to call upon starting a receipt fetch
-	chainInsertHook  func([]*fetchResult)  // Method to call upon inserting a chain of blocks (possibly in multiple invocations)
+	syncInitHook         func(uint64, uint64)  // Method to call upon initiating a new sync run
+	bodyFetchHook        func([]*types.Header) // Method to call upon starting a block body fetch
+	receiptFetchHook     func([]*types.Header) // Method to call upon starting a receipt fetch
+	stakingInfoFetchHook func([]*types.Header) // Method to call upon starting a staking info fetch
+	chainInsertHook      func([]*fetchResult)  // Method to call upon inserting a chain of blocks (possibly in multiple invocations)
 }
 
 // LightChain encapsulates functions required to synchronise a light chain.
@@ -206,26 +212,28 @@ func New(mode SyncMode, stateDB database.DBManager, stateBloom *statedb.SyncBloo
 	}
 
 	dl := &Downloader{
-		mode:           uint32(mode),
-		stateDB:        stateDB,
-		stateBloom:     stateBloom,
-		mux:            mux,
-		queue:          newQueue(blockCacheMaxItems, blockCacheInitialItems),
-		peers:          newPeerSet(),
-		rttEstimate:    uint64(rttMaxEstimate),
-		rttConfidence:  uint64(1000000),
-		blockchain:     chain,
-		lightchain:     lightchain,
-		dropPeer:       dropPeer,
-		headerCh:       make(chan dataPack, 1),
-		bodyCh:         make(chan dataPack, 1),
-		receiptCh:      make(chan dataPack, 1),
-		bodyWakeCh:     make(chan bool, 1),
-		receiptWakeCh:  make(chan bool, 1),
-		headerProcCh:   make(chan []*types.Header, 1),
-		quitCh:         make(chan struct{}),
-		stateCh:        make(chan dataPack),
-		stateSyncStart: make(chan *stateSync),
+		mode:              uint32(mode),
+		stateDB:           stateDB,
+		stateBloom:        stateBloom,
+		mux:               mux,
+		queue:             newQueue(blockCacheMaxItems, blockCacheInitialItems),
+		peers:             newPeerSet(),
+		rttEstimate:       uint64(rttMaxEstimate),
+		rttConfidence:     uint64(1000000),
+		blockchain:        chain,
+		lightchain:        lightchain,
+		dropPeer:          dropPeer,
+		headerCh:          make(chan dataPack, 1),
+		bodyCh:            make(chan dataPack, 1),
+		receiptCh:         make(chan dataPack, 1),
+		stakingInfoCh:     make(chan dataPack, 1),
+		bodyWakeCh:        make(chan bool, 1),
+		receiptWakeCh:     make(chan bool, 1),
+		stakingInfoWakeCh: make(chan bool, 1),
+		headerProcCh:      make(chan []*types.Header, 1),
+		quitCh:            make(chan struct{}),
+		stateCh:           make(chan dataPack),
+		stateSyncStart:    make(chan *stateSync),
 		syncStatsState: stateSyncStats{
 			processed: stateDB.ReadFastTrieProgress(),
 		},
@@ -376,13 +384,13 @@ func (d *Downloader) synchronise(id string, hash common.Hash, td *big.Int, mode 
 	d.queue.Reset(blockCacheMaxItems, blockCacheInitialItems)
 	d.peers.Reset()
 
-	for _, ch := range []chan bool{d.bodyWakeCh, d.receiptWakeCh} {
+	for _, ch := range []chan bool{d.bodyWakeCh, d.receiptWakeCh, d.stakingInfoWakeCh} {
 		select {
 		case <-ch:
 		default:
 		}
 	}
-	for _, ch := range []chan dataPack{d.headerCh, d.bodyCh, d.receiptCh} {
+	for _, ch := range []chan dataPack{d.headerCh, d.bodyCh, d.receiptCh, d.stakingInfoCh} {
 		for empty := false; !empty; {
 			select {
 			case <-ch:
@@ -483,6 +491,7 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td *big.I
 		func() error { return d.fetchHeaders(p, origin+1, pivot) }, // Headers are always retrieved
 		func() error { return d.fetchBodies(origin + 1) },          // Bodies are retrieved during normal and fast sync
 		func() error { return d.fetchReceipts(origin + 1) },        // Receipts are retrieved during fast sync
+		func() error { return d.fetchStakingInfos(origin + 1) },    // StakingInfos are retrieved during fast sync
 		func() error { return d.processHeaders(origin+1, pivot, td) },
 	}
 	if mode == FastSync {
@@ -605,6 +614,7 @@ func (d *Downloader) fetchHeight(p *peerConnection) (*types.Header, error) {
 
 		case <-d.bodyCh:
 		case <-d.receiptCh:
+		case <-d.stakingInfoCh:
 			// Out of bounds delivery, ignore
 		}
 	}
@@ -703,6 +713,7 @@ func (d *Downloader) findAncestor(p *peerConnection, height uint64) (uint64, err
 
 		case <-d.bodyCh:
 		case <-d.receiptCh:
+		case <-d.stakingInfoCh:
 			// Out of bounds delivery, ignore
 		}
 	}
@@ -767,6 +778,7 @@ func (d *Downloader) findAncestor(p *peerConnection, height uint64) (uint64, err
 
 			case <-d.bodyCh:
 			case <-d.receiptCh:
+			case <-d.stakingInfoCh:
 				// Out of bounds delivery, ignore
 			}
 		}
@@ -896,7 +908,7 @@ func (d *Downloader) fetchHeaders(p *peerConnection, from uint64, pivot uint64) 
 			d.dropPeer(p.id)
 
 			// Finish the sync gracefully instead of dumping the gathered data though
-			for _, ch := range []chan bool{d.bodyWakeCh, d.receiptWakeCh} {
+			for _, ch := range []chan bool{d.bodyWakeCh, d.receiptWakeCh, d.stakingInfoWakeCh} {
 				select {
 				case ch <- false:
 				case <-d.cancelCh:
@@ -1000,6 +1012,33 @@ func (d *Downloader) fetchReceipts(from uint64) error {
 		d.receiptFetchHook, fetch, d.queue.CancelReceipts, capacity, d.peers.ReceiptIdlePeers, setIdle, "receipts")
 
 	logger.Debug("Transaction receipt download terminated", "err", err, "elapsed", time.Since(start))
+	return err
+}
+
+// fetchStakingInfos iteratively downloads the scheduled staking information, taking any
+// available peers, reserving a chunk of staking information for each, waiting for delivery
+// and also periodically checking for timeouts.
+func (d *Downloader) fetchStakingInfos(from uint64) error {
+	logger.Debug("Downloading staking information", "origin", from)
+
+	start := time.Now()
+	var (
+		deliver = func(packet dataPack) (int, error) {
+			pack := packet.(*stakingInfoPack)
+			return d.queue.DeliverStakingInfos(pack.peerId, pack.stakingInfos)
+		}
+		expire   = func() map[string]int { return d.queue.ExpireStakingInfos(d.requestTTL()) }
+		fetch    = func(p *peerConnection, req *fetchRequest) error { return p.FetchStakingInfo(req) }
+		capacity = func(p *peerConnection) int { return p.StakingInfoCapacity(d.requestRTT()) }
+		setIdle  = func(p *peerConnection, accepted int, deliveryTime time.Time) {
+			p.SetStakingInfoIdle(accepted, deliveryTime)
+		}
+	)
+	err := d.fetchParts(d.stakingInfoCh, deliver, d.stakingInfoWakeCh, expire,
+		d.queue.PendingStakingInfos, d.queue.InFlightStakingInfos, d.queue.ReserveStakingInfos,
+		d.stakingInfoFetchHook, fetch, d.queue.CancelStakingInfo, capacity, d.peers.StakingInfoIdlePeers, setIdle, "stakingInfos")
+
+	logger.Debug("Staking information download terminated", "err", err, "elapsed", time.Since(start))
 	return err
 }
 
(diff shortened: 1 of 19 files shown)


reward/staking_manager.go
```diff
@@ -103,14 +103,29 @@ func GetStakingManager() *StakingManager {
...
+// GetStakingInfo returns a stakingInfo on the staking block of the given block number.
+// Note that staking block is the block on which the associated staking information is stored and used during an interval.
 func GetStakingInfo(blockNum uint64) *StakingInfo {
+	stakingBlockNumber := params.CalcStakingBlockNumber(blockNum)
+	return GetStakingInfoOnStakingBlock(stakingBlockNumber)
+}
+
+// GetStakingInfoOnStakingBlock returns a corresponding StakingInfo for a staking block number.
+// If the given number is not on the staking block, it returns nil.
+func GetStakingInfoOnStakingBlock(stakingBlockNumber uint64) *StakingInfo {
```
blukat29, line 115: What's the difference between `GetStakingInfoOnStakingBlock(sBN)` and `getStakingInfo(0, sBN)`?

reward/staking_manager.go
```diff
@@ -103,14 +103,29 @@ func GetStakingManager() *StakingManager {
...
 	if stakingManager == nil {
 		logger.Error("unable to GetStakingInfo", "err", ErrStakingManagerNotSet)
 		return nil
 	}
 
-	stakingBlockNumber := params.CalcStakingBlockNumber(blockNum)
+	return getStakingInfo(0, stakingBlockNumber)
+}
+
+func getStakingInfo(blockNum, stakingBlockNumber uint64) *StakingInfo {
```
blukat29, line 124: Can we remove the `blockNum` argument? It seems that `blockNum` argument is not used except logging.

reward/staking_manager.go
jeongkyun-oh: @blukat29 `GetStakingInfoOnStakedBlock` is introduced here to differentiate `GetStakingInfo` without calling `CalcStakingBlockNumber`. As you see, it checks `stakingManager`is nil and it is public method. Any suggestion?

reward/staking_manager.go
jeongkyun-oh: @blukat29 Ture, but it is used in `GetStakingInfo` as well and it may not be a 0. Do you think still it is not necessary?

reward/staking_manager.go
blukat29: @jeongkyun-oh The nil check is also in `getStakingInfo`. So `GetStakingInfoOnStakedBlock` is simply a wrapper that adds the '0' argument.

I'd say we can only leave the public method unless there are other uses of `getStakingInfo`.

reward/staking_manager.go
blukat29: - As per your change,
  - `GetStakingInfo(n)` -> `GetStakingInfoOnStakingBlock(m)` -> `getStakingInfo(0, m)`
  - `GetStakingInfoOnStakingBlock(m)` -> `getStakingInfo(0,m)`
  - The private function `getStakingInfo` not used anywhere else.
- So I thought we can simplify to
  - `GetStakingInfo` -> `GetStakingInfoOnStakingBlock`
  - `GetStakingInfoOnStakingBlock`

- But if you expect more use cases of `getStakingInfo`, (especially with nonzero blockNum) then I am okay with this change.

- Regarding the blockNum,
  - blockNum=0 can be misleading in the logs. Isn't it supposed to be like blockNum=172877, staking block number=86400? logging blockNum=0 conveys no meaning. If my understanding about blockNum is incorrect, please tell me.
  - Or we can call `getStakingInfo(stakingBlockNumber, stakingBlockNumber)` for clarity

reward/staking_manager.go
jeongkyun-oh: @blukat29 I see. I have updated. Thanks :)

reward/staking_manager.go
jeongkyun-oh: @blukat29 I agreed with you. It would be misleading the logs and the same as `getStakingInfo(stakingBlockNumber, stakingBlockNumber)`. I added a line of log for the original block number when calling `GetStakingInfo` only. Please take another look.

blukat29 (dismissed):
Thanks @jeongkyun-oh!

node/cn/protocol.go
```diff
@@ -74,7 +77,15 @@ const (
 	ReceiptsRequestMsg = 0x0e
 	ReceiptsMsg        = 0x0f
 
-	MsgCodeEnd = 0x10
+	// Protocol messages belonging to klay/64
+	DummyMsg1 = 0x10
+	DummyMsg2 = 0x11 // it is used for consensus message for istanbul engine
```
yoomee1313, line 82: I can't understand why `DummyMsg2` is added..  it seems duplicated, because `0x11` is already defined at istanbul package. Could you explain what DummyMsg1 & DummyMsg2's role? It would be nice to add a comment about the role of DummyMsgs.

jeongkyun-oh: So, I wanted to add new messages (staking info messages), but the `0x11` was taken in istanbul engine. That is why I added two dummy messages in `klay` protocol. Please welcome any better suggestion. I will add some comments though.

yoomee1313: @jeongkyun-oh Oh, I'm just curious that dummyMsgs have some special role.

yoomee1313 (dismissed):

datasync/downloader/downloader_test.go
```diff
@@ -575,6 +579,10 @@ func (dlp *downloadTesterPeer) RequestReceipts(hashes []common.Hash) error {
 	return nil
 }
 
+func (dlp *downloadTesterPeer) RequestStakingInfo(hashes []common.Hash) error {
+	panic("implement me")
```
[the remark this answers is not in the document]
jeongkyun-oh: @ian0371 I have fixed and updated the existing testcases. Please take another look if you have time.

jeongkyun-oh:
@blukat29 @ian0371 @yoomee1313 Please take another look

datasync/downloader/queue.go
```diff
@@ -102,6 +108,13 @@ func (f *fetchResult) SetReceiptsDone() {
 	}
 }
 
+// SetStakingInfoDone flags the receipts as finished.
+func (f *fetchResult) SetStakingInfoDone() {
+	if v := atomic.LoadInt32(&f.pending); (v & (1 << stakingInfoType)) != 0 {
+		atomic.AddInt32(&f.pending, -4)
```
blukat29, line 114: Maybe `-(1 << stakingInfoType)` for clarity?

datasync/downloader/queue.go
```diff
@@ -805,6 +886,24 @@ func (q *queue) DeliverReceipts(id string, receiptList [][]*types.Receipt) (int,
 	return q.deliver(id, q.receiptTaskPool, q.receiptTaskQueue, q.receiptPendPool, receiptReqTimer, len(receiptList), validate, reconstruct)
 }
 
+// DeliverStakingInfos injects a receipt retrieval response into the results queue.
```
blukat29, line 889:
```suggestion
// DeliverStakingInfos injects a stakinginfo retrieval response into the results queue.
```

datasync/downloader/downloader_test.go
```diff
@@ -75,6 +80,8 @@ type downloadTester struct {
 func newTester() *downloadTester {
 	testdb := database.NewMemoryDBManager()
 	genesis := blockchain.GenesisBlockForTesting(testdb, testAddress, big.NewInt(1000000000))
+	params.SetStakingUpdateInterval(4)
+	reward.SetTestStakingManagerWithDB(testDB)
```
blukat29, line 84: Can we rollback both StakingUpdateInterval and StakingManager to original values after testing?  e.g. in `tester.terminate()`

sample: https://github.com/klaytn/klaytn/blob/v1.8.4/consensus/istanbul/backend/engine_test.go#L755

node/cn/protocol.go
blukat29, line 82: Is 0x10 skipped because there's 0x11? If so, can we make it more explicit?

```suggestion
        Unused10 = 0x10 // Skipped a number because 0x11 is already taken
	Unused11 = 0x11 // Already used by consensus (IstanbulMsg)
```

datasync/downloader/queue.go
blukat29: Nevermind. Existing code is just using -1, -2, etc.

blukat29 (approved):
LGTM other than minor comments.

node/cn/protocol.go
jeongkyun-oh: I have updated it as you suggested.

datasync/downloader/queue.go
jeongkyun-oh: Updated, thanks :)

datasync/downloader/downloader_test.go
jeongkyun-oh: Hmm.. I have updated for setting and rollback governance for testing purpose. Since staking manager is global variable, I have updated differently. Please take another look.

blukat29: @jeongkyun-oh Nice job handling concurrent tests!

datasync/downloader/downloader.go
```diff
@@ -1578,6 +1624,14 @@ func (d *Downloader) commitFastSyncData(results []*fetchResult, stateSync *state
 func (d *Downloader) commitPivotBlock(result *fetchResult) error {
 	block := types.NewBlockWithHeader(result.Header).WithBody(result.Transactions)
 	logger.Debug("Committing fast sync pivot as new head", "number", block.Number(), "hash", block.Hash())
+	if result.StakingInfo != nil {
+		if err := reward.AddStakingInfoToDB(result.StakingInfo); err != nil {
+			logger.Error("Inserting downloaded staking info is failed", "err", err)
+			return fmt.Errorf("failed to insert the downloaded staking information: %v", err)
+		} else {
+			logger.Info("Imported new staking information", "number", result.StakingInfo.BlockNum)
```
kjhman21, line 1632: If the string is exactly the same, later it is not easy which statement is executed. Could you little bit differentiate the log string between commitPivotBlock() and commitFastSyncData()?

datasync/downloader/downloader_test.go
```diff
@@ -46,6 +49,48 @@ func init() {
...
+	defer lock.Unlock()
+	if setter == nil {
+		setter = &govSetter{
+			numTesting:          0,
+			origStakingInterval: params.StakingUpdateInterval(),
+			origStakingManager:  reward.GetStakingManager(),
+		}
+
+		reward.SetTestStakingManagerWithDB(database.NewMemoryDBManager())
+		params.SetStakingUpdateInterval(4)
```
kjhman21, line 76: this 4 should be a variable instead of hard-coded number.

datasync/downloader/downloader_test.go
```diff
@@ -46,6 +49,48 @@ func init() {
...
+	}
+	setter.numTesting += 1
```
kjhman21, line 78: I think we can use a waitgroup instead of just using a counter.
https://gobyexample.com/waitgroups

datasync/downloader/downloader_test.go
```diff
@@ -617,12 +742,14 @@ func assertOwnForkedChain(t *testing.T, tester *downloadTester, common int, leng
 		headers += length - common
 		blocks += length - common
 		receipts += length - common - fsMinFullBlocks
+		stakingInfos += length - common - fsMinFullBlocks
 	}
+	stakingInfos = stakingInfos / 4 // assuming that staking information update interval is 4
```
kjhman21, line 747: also here. use variable instead of `4`

datasync/downloader/peer.go
```diff
@@ -484,7 +534,7 @@ func (ps *peerSet) HeaderIdlePeers() ([]*peerConnection, int) {
 		defer p.lock.RUnlock()
 		return p.headerThroughput
 	}
-	return ps.idlePeers(62, 64, idleCheck, throughput)
+	return ps.idlePeers(62, 65, idleCheck, throughput)
```
kjhman21, line 537: Can we use Klay65 instead of a number?

datasync/downloader/queue.go
```diff
@@ -326,6 +363,15 @@ func (q *queue) Schedule(headers []*types.Header, from uint64) []*types.Header {
 				q.receiptTaskQueue.Push(header, -int64(header.Number.Uint64()))
 			}
 		}
+
+		if q.mode == FastSync && params.IsStakingUpdateInterval(header.Number.Uint64()) {
```
kjhman21, line 367: It seems I don't know well about this, but I wonder that it is okay that we test the fixed staking interval as `4` in the previous downloader test.

node/cn/channel_manager_test.go
```diff
@@ -55,6 +55,10 @@ func testChannelManager(t *testing.T, chSize int) {
 		// Before calling RegisterChannelWithIndex,
 		// calling GetChannelWithMsgCode with registered MsgCode should return no channel and no error.
 		for i := StatusMsg; i < MsgCodeEnd; i++ {
+			if i == DummyMsg1 || i == DummyMsg2 {
```
kjhman21, line 58: Where is the definition of `DummyMsg1` ?

kjhman21: maybe Unused10? 
https://github.com/klaytn/klaytn/pull/1394/files?w=0#diff-102c0c38488b22c8d83cc296c1d1bd21330d33fdf8ba3483c2464c0883fec602R81

reward/staking_manager.go
```diff
@@ -103,38 +103,45 @@ func GetStakingManager() *StakingManager {
...
+// Note that staking block is the block on which the associated staking information is stored and used during an interval.
 func GetStakingInfo(blockNum uint64) *StakingInfo {
+	stakingBlockNumber := params.CalcStakingBlockNumber(blockNum)
+	logger.Debug("Staking information is requested", "blockNum", blockNum, "staking block number", stakingBlockNumber)
+	return GetStakingInfoOnStakingBlock(stakingBlockNumber)
+}
+
+// GetStakingInfoOnStakingBlock returns a corresponding StakingInfo for a staking block number.
+// If the given number is not on the staking block, it returns nil.
+func GetStakingInfoOnStakingBlock(stakingBlockNumber uint64) *StakingInfo {
```
kjhman21, line 116: hmm the function name is little bit weird...

1. Doesn't it need to check whether the block number is on the staking interval or not?
2. What about the function name of `GetStakingInfoOnStakingInterval`? `GetStakingInfoStrict`? Let me think little bit more...

reward/staking_manager.go
```diff
@@ -235,6 +242,14 @@ func StakingManagerUnsubscribe() {
 	stakingManager.chainHeadSub.Unsubscribe()
 }
 
+// SetTestStakingManagerWithDB sets the staking manager with the given database.
+// Note that this method is used only for testing purpose.
+func SetTestStakingManagerWithDB(testDB stakingInfoDB) {
```
kjhman21, line 247: Hmm.. is it okay to use like this? this seems for testing, but it is in real code?

jeongkyun-oh: Okay, but the there is another function to set it for testing. I will move these methods to test file in another PR.

datasync/downloader/downloader_test.go
```diff
@@ -655,12 +785,12 @@ func assertOwnForkedChain(t *testing.T, tester *downloadTester, common int, leng
 // binary searching.
 func TestCanonicalSynchronisation62(t *testing.T)     { testCanonicalSynchronisation(t, 62, FullSync) }
 func TestCanonicalSynchronisation63Full(t *testing.T) { testCanonicalSynchronisation(t, 63, FullSync) }
-func TestCanonicalSynchronisation63Fast(t *testing.T) { testCanonicalSynchronisation(t, 63, FastSync) }
```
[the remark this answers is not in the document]
jeongkyun-oh: Fast sync not working on protocol 63 and it won't fail after introducing staking info downloader.

reward/staking_manager.go
jeongkyun-oh: 1. As the comment is described, it returns `nil` if the number is not on staking block. It should be added if we need to handle the related error.

datasync/downloader/downloader.go
jeongkyun-oh: Okay. I have updated it.

datasync/downloader/downloader_test.go
```diff
@@ -99,11 +149,25 @@ func newTester() *downloadTester {
...
 // makeChain creates a chain of n blocks starting at and including parent.
 // the returned hash chain is ordered head->parent. In addition, every 3rd block
-// contains a transaction to allow testing correct block
-// reassembly.
-func (dl *downloadTester) makeChain(n int, seed byte, parent *types.Block, parentReceipts types.Receipts, heavy bool) ([]common.Hash, map[common.Hash]*types.Header, map[common.Hash]*types.Block, map[common.Hash]types.Receipts) {
+// contains a transaction to allow testing correct block reassembly.
+// On every 4th block, staking information is updated to allow testing staking info
+// downloading as well.
+func (dl *downloadTester) makeChain(n int, seed byte, parent *types.Block, parentReceipts types.Receipts, heavy bool) ([]common.Hash, map[common.Hash]*types.Header, map[common.Hash]*types.Block, map[common.Hash]types.Receipts, map[common.Hash]*reward.StakingInfo) {
+	stakingInterval := uint64(4)
```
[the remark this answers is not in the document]
jeongkyun-oh: Updated :)

datasync/downloader/downloader_test.go
jeongkyun-oh: Updated :)

datasync/downloader/downloader_test.go
jeongkyun-oh: Updated :)

datasync/downloader/downloader_test.go
jeongkyun-oh: Is this necessary? We don't need to wait for all goroutines to be terminated, but only to set governance before testing. I think also this is good to go..

node/cn/handler.go
```diff
@@ -901,6 +912,59 @@ func handleReceiptsMsg(pm *ProtocolManager, p Peer, msg p2p.Msg) error {
...
+			break
+		} else if err != nil {
+			return errResp(ErrDecode, "msg %v: %v", msg, err)
+		}
+
+		// Retrieve the requested block's staking information, skipping if unknown to us
+		header := pm.blockchain.GetHeaderByHash(hash)
+		result := reward.GetStakingInfoOnStakingBlock(header.Number.Uint64())
+		if result == nil {
+			logger.Error("Failed to get staking information on a specific block", "number", header.Number.Uint64())
```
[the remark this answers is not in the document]
jeongkyun-oh: Added as you suggested :)

reward/staking_manager.go
kjhman21: If you already consider this in your mind, could you add "TODO" here not to forget?

reward/staking_manager.go
kjhman21: But I couldn't find the check logic. Where is it checked?

datasync/downloader/queue.go
jeongkyun-oh: According to *fixed* staking interval, AFAIK the two interval (staking update interval, proposer update interval) cannot be updated using governance vote.

reward/staking_manager.go
jeongkyun-oh: I have added TODO comment for this as well as protocol number.

reward/staking_manager.go
jeongkyun-oh: Oh, I said if we need to handle it. I think we don't need to handle it. Just returning nil is fine IMO.

ethan-kr (dismissed):

datasync/downloader/peer.go
```diff
@@ -407,19 +455,21 @@ func (ps *peerSet) Register(p *peerConnection) error {
...
 			p.blockThroughput += peer.blockThroughput
 			p.receiptThroughput += peer.receiptThroughput
+			p.stakingInfoThroughput += peer.stakingInfoThroughput
 			p.stateThroughput += peer.stateThroughput
 			peer.lock.RUnlock()
 		}
 		p.headerThroughput /= float64(len(ps.peers))
 		p.blockThroughput /= float64(len(ps.peers))
 		p.receiptThroughput /= float64(len(ps.peers))
+		p.stakingInfoThroughput /= float64(len(ps.peers))
```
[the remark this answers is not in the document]
jeongkyun-oh: @ethan-kr `len(ps.peers) > 0` looks more make sense. Also, all`float64(len(ps.peers))` should be changed. So, let's keep consistency at the moment.

reward/staking_manager.go
```diff
@@ -103,14 +103,29 @@ func GetStakingManager() *StakingManager {
 	return stakingManager
 }
 
-// GetStakingInfo returns a corresponding stakingInfo for a blockNum.
+// GetStakingInfo returns a stakingInfo on the staking block of the given block number.
+// Note that staking block is the block on which the associated staking information is stored and used during an interval.
 func GetStakingInfo(blockNum uint64) *StakingInfo {
+	stakingBlockNumber := params.CalcStakingBlockNumber(blockNum)
```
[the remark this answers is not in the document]
jeongkyun-oh: @ethan-kr Actually, the parameter is removed since it is not used in the method except for logs. Please check out the comment https://github.com/klaytn/klaytn/pull/1394#discussion_r883321856.

reward/staking_manager.go
kjhman21: @jeongkyun-oh I couldn't find the explicit condition check below. which function call treats that?

datasync/downloader/peer.go
```diff
@@ -210,6 +220,28 @@ func (p *peerConnection) FetchReceipts(request *fetchRequest) error {
 	return nil
 }
 
+// FetchStakingInfo sends a staking information retrieval request to the remote peer.
+func (p *peerConnection) FetchStakingInfo(request *fetchRequest) error {
+	// Sanity check the protocol version
+	if p.version < 65 {
+		panic(fmt.Sprintf("staking info fetch [klay/65+] requested on klay/%d", p.version))
```
kjhman21, line 227: If a peer does not support protocol/65, it panics. Is this right?

datasync/downloader/downloader.go
```diff
@@ -1567,6 +1605,14 @@ func (d *Downloader) commitFastSyncData(results []*fetchResult, stateSync *state
 	for i, result := range results {
 		blocks[i] = types.NewBlockWithHeader(result.Header).WithBody(result.Transactions)
 		receipts[i] = result.Receipts
+		if result.StakingInfo != nil {
```
aidan-kwon, line 1608: Isn't `AddStakingInfoToDB` necessary process? If this nil checking logic doesn't exit, `AddStakingInfoToDB` will return an error for `StakingInfo == nil` case. But, this if branch miss the error handling

datasync/downloader/downloader.go
```diff
@@ -1578,6 +1624,14 @@ func (d *Downloader) commitFastSyncData(results []*fetchResult, stateSync *state
 func (d *Downloader) commitPivotBlock(result *fetchResult) error {
 	block := types.NewBlockWithHeader(result.Header).WithBody(result.Transactions)
 	logger.Debug("Committing fast sync pivot as new head", "number", block.Number(), "hash", block.Hash())
+	if result.StakingInfo != nil {
```
aidan-kwon, line 1627: Isn't AddStakingInfoToDB necessary process? If this nil checking logic doesn't exit, AddStakingInfoToDB will return an error for StakingInfo == nil case. But, this if branch miss the error handling

datasync/downloader/downloader.go
aidan-kwon: It seems that `stakingManager` exists only for `WeightedRandom` proposer policy. Is it okay that the info is missing when the proposer policy is used

datasync/downloader/peer.go
```diff
@@ -210,6 +220,28 @@ func (p *peerConnection) FetchReceipts(request *fetchRequest) error {
 	return nil
 }
 
+// FetchStakingInfo sends a staking information retrieval request to the remote peer.
+func (p *peerConnection) FetchStakingInfo(request *fetchRequest) error {
```
aidan-kwon, line 224: Please, check whether this method is called when p.version < 65

datasync/downloader/peer.go
jeongkyun-oh: @aidan-kwon @kjhman21 @ethan-kr 
I have checked the logic again. `FetchingStakingInfo` method is called to only idle peers which are satisfying the version match as well. So, it should be fine. Please take another look.

- `fetchParts`
```
// numTotalPeers means the number of peers satisfying the protocol requirement.
idlePeers, numTotalPeers := idlePeers()
pendCount := pending()

for _, peer := range idlePeers {
...
if err := fetch(peer, request); err != nil {
...
```

- `idlePeers`
```
func (ps *peerSet) idlePeers(minProtocol, maxProtocol int, idleCheck func(*peerConnection) bool, throughput func(*peerConnection) float64) ([]*peerConnection, int) {
	ps.lock.RLock()
	defer ps.lock.RUnlock()

	idlePeers, numTotalPeers := make([]*peerConnection, 0, len(ps.peers)), 0
	tps := make([]float64, 0, len(ps.peers))
	for _, p := range ps.peers {
		if p.version >= minProtocol && p.version <= maxProtocol {
			if idleCheck(p) {
				idlePeers = append(idlePeers, p)
				tps = append(tps, throughput(p))
			}
			numTotalPeers++
		}
	}
	// sort peers in the descending order of throughput
	sortPeers := &peerThroughputSort{idlePeers, tps}
	sort.Sort(sortPeers)
	return sortPeers.p, numTotalPeers
}
```

datasync/downloader/peer.go
jeongkyun-oh: Please take a look https://github.com/klaytn/klaytn/pull/1394#discussion_r901229502

datasync/downloader/peer.go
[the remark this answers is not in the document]
jeongkyun-oh: I introduced a new variable for `protocolLengthFloat` instead.

reward/staking_manager.go
jeongkyun-oh: I have updated the shortcut.

node/cn/handler.go
```diff
@@ -611,12 +611,12 @@ func (pm *ProtocolManager) handleMsg(p Peer, addr common.Address, msg p2p.Msg) e
...
 		}
 
-	case msg.Code == StakingInfoRequestMsg:
+	case p.GetVersion() >= klay65 && msg.Code == StakingInfoRequestMsg:
 		if err := handleStakingInfoRequestMsg(pm, p, msg); err != nil {
 			return err
 		}
 
-	case msg.Code == StakingInfoMsg:
+	case p.GetVersion() >= klay65 && msg.Code == StakingInfoMsg:
```
ethan-kr, line 618: Version check inside a function?

aidan-kwon (approved):

datasync/downloader/downloader.go
jeongkyun-oh: I am going to resolve this with another PR.

ethan-kr (approved):

jeongkyun-oh:
@kjhman21 Please let me know if you have other opinions.

PR #1402, opened 2022-05-30
Add log.EnableLogForTest for all tests

blukat29:
## Proposed changes

- Replace `tests.enableLog()` by `log.EnableLogForTest()`. Allows using this function throughout the codebase.
- Clears a [TODO](https://github.com/klaytn/klaytn/blob/v1.8.4/tests/smartcontract_creation_test.go#L41)

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Further comments

### DIFF
--- a/log/test_util.go
+++ b/log/test_util.go
@@ -1,0 +1,35 @@
+package log
+
+import (
+	"io"
+	"os"
+	"testing"
+
+	"github.com/klaytn/klaytn/log/term"
+	"github.com/mattn/go-colorable"
+)
+
+// Enable logging to STDERR
+// Exmaple use
+//     log.EnableLogForTest(log.LvlCrit, log.LvlTrace)
+// `normalLvl` is used in most cases
+// `verboseLvl` is used if `go test -v` flag is given
+func EnableLogForTest(normalLvl, verboseLvl Lvl) {
+	lvl := Lvl(normalLvl)
+	if testing.Verbose() {
+		lvl = Lvl(verboseLvl)
+	}
+
+	usecolor := term.IsTty(os.Stderr.Fd()) && os.Getenv("TERM") != "dumb"
+	output := io.Writer(os.Stderr)
+	if usecolor {
+		output = colorable.NewColorableStderr()
+	}
+
+	glogger := NewGlogHandler(StreamHandler(output, TerminalFormat(usecolor)))
+	PrintOrigins(true)
+	ChangeGlobalLogLevel(glogger, lvl)
+	glogger.Vmodule("")
+	glogger.BacktraceAt("")
+	Root().SetHandler(glogger)
+}
--- a/snapshot/generate_test.go
+++ b/snapshot/generate_test.go
@@ -22,9 +22,7 @@ package snapshot
 
 import (
 	"fmt"
-	"io"
 	"math/big"
-	"os"
 	"testing"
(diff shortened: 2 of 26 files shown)


ian0371 (approved):

log/test_util.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/klaytn/klaytn/log/term"
+	"github.com/mattn/go-colorable"
+)
+
+// Enable logging to STDERR
+// Exmaple use
+//     log.EnableLogForTest(log.LvlCrit, log.LvlTrace)
+// `normalLvl` is used in most cases
+// `verboseLvl` is used if `go test -v` flag is given
+func EnableLogForTest(normalLvl, verboseLvl Lvl) {
```
ehnuje, line 17: Do you think we need both level parameters?

ehnuje (approved):
Other than a comment, LGTM. Thanks!

jeongkyun-oh (approved):

log/test_util.go
blukat29: - I added two levels because there was one case which I thought both are needed - https://github.com/klaytn/klaytn/blob/v1.8.4/storage/database/dynamodb_test.go#L200
- Also in general having at least `Crit` level is useful because otherwise all we'd see is 'exit 1' message.
- It does seem little unnecessary but it's just for testing, so should be no big deal.

snapshot/generate_test.go
```diff
@@ -64,27 +60,9 @@ func genSmartContractAccount(nonce uint64, balance *big.Int, storageRoot common.
...
-	glogger.BacktraceAt("")
-	log.Root().SetHandler(glogger)
-}
-
 // Tests that snapshot generation from an empty database.
 func TestGeneration(t *testing.T) {
-	if false {
-		enableLog()
-	}
+	log.EnableLogForTest(log.LvlCrit, log.LvlTrace)
```
yoomee1313, line 65: In this kind of tests, it seems that `enableLog()` is not called unless we change the test code.. however, it seems you changed the code to call `enableLog()`. Is this intended?

blukat29: Yes you understood correctly. It is intended based on an assumption.

I guessed original author of the test wanted to do something like `if testing.Verbose() { enableLog() }`. That should be the purpose of the if statement.

Therefore I've replaced it with the equivalent that does not require manual code editing. I think adding `-v` flag is more convenient because it is quicker and creates zero git diff.

hqjang-pepper:
From my understanding, this PR is written to apply the logic to output logs at the verbose level consistantly. Is that correct?

blukat29:
@hqjang-pepper I wanted a consistent way to turn on/off logs in tests.
There will be no logs if you run with `go test` or `make test`.
Logs will appear if you run with `go test -v`.

With the `-v` flag, you don't have to edit the code to turn on/off the logs.
Below are actual codes test writers used.
```go
// enableLog()
```
```go
if false {
    enableLog()
}
```

snapshot/generate_test.go
yoomee1313: Oh i see, it's definitely better than before..

yoomee1313 (approved):

hqjang-pepper (approved):
Got it. LGTM

PR #1409, opened 2022-06-02
Implement dynamic base fee mechanism

kjeom:
## Proposed changes
Added new features
- `BaseFee` field in block header
- Base fee calculation dynamically
- Header verification for base fee
- 50% tx fee burned
- Hard forked by `BaseFee` field

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- https://github.com/klaytn/klaytn/issues/1245

Issue #1429, opened 2022-06-10
Question about governance function on hard-forked chain by KIP71


hyunsooda:
As the txPool's gas price is determined by `calcBaseFee()` call in KIP71 hardfork, I think the API [updateUnitPrice](https://github.com/klaytn/klaytn/blob/dynamicGasPrice/governance/handler.go#L112) should be removed. 

cc. @kjeom, @blukat29

blukat29:
@hyunsooda It should not be removed. The code will be used by
- chains with KIP71Hardfork = nil 
- chains with KIP71Hardfork block number, but before the block number.

I think it's txpool's responsibility to ignore the `SetGasPrice()` call after hardfork block number.

- https://github.com/klaytn/klaytn/blob/v1.8.4/blockchain/tx_pool.go#L490:
```go
func (pool *TxPool) SetGasPrice(price *big.Int) {
    if pool.chainconfig.IsKIP71ForkEnabled(pool.currentBlockNumber) {
        return
    }
    // existing code..
}
```

hyunsooda:
My question was whether to remove `updateUnitPrice()`, not the `SetGasPrice()`. Or maybe I'm missing your point?

kjeom:
@hyunsooda updateUnitPrice() is a callback function when updating governance parameter so it has been used before hardfork.

hyunsooda:
Yes. It must not be removed for now. I was considering this functionality should be removed once the hardfork is done.

blukat29:
> My question was whether to remove `updateUnitPrice()`, not the `SetGasPrice()`. Or maybe I'm missing your point?

@hyunsooda I meant to keep both `updateUnitPrice()` and `SetGasPrice()`. We can support the use case where KIP71 fork is not enabled forever. For instance, a hypothetical company may want to run a v1.9.0 service chain without KIP71 fork for the fixed gas price experience.

Is there a reason why `updateUnitPrice` must be removed?

hyunsooda:
@blukat29, 

Thanks for the good explanation. The fixed gas price seems still preserved if the GCs vote the `UpperBoundBaseFee` and `LowerBoundBaseFee` values equally. Thus, the guide to teaching for the case seems better instead of holding the ambiguous function.

blukat29:
@hyunsooda Good point. I see that `UpperBoundBaseFee == LowerBoundBaseFee` is equivalent to setting `UnitPrice`.

## When can we remove it

[...]

PR #1442, opened 2022-06-17
 Dynamic gas price : BaseFee for generating a new block in worker and txPool policy

kjeom:
## Proposed changes

- Added KIP-71 hardfork branch in worker and txPool
- Modified unit tests for the hardfork

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- https://github.com/klaytn/klaytn/issues/1245

## Further comments

- This PR is an extension of the PRs, #1409 and #1382
- The main changes here are calling `consensus/misc/kip71.go/NextBlockBaseFee( )` with KIP-71 hardfork branch in worker and txPool
- In worker, BaseFee field in block header is filled with newly calculated `base fee` when generating a new block
- In txPool, gasPrice field in txPool is filled with that also for txPool policy like dropping txs with under the gasPrice


blukat29:
@kjeom Are we going to disable `TxPool.SetGasPrice()` after hardfork? this function is quite powerful. It flushes all pools upon UnitPrice change. After hardfork we don't want the UnitPrice change to affect txpool. How about adding a guard condition like https://github.com/klaytn/klaytn/issues/1429#issuecomment-1152102291?

consensus/istanbul/backend/engine.go
```diff
@@ -172,7 +172,8 @@ func (sb *backend) computeSignatureAddrs(header *types.Header) error {
 // given engine. Verifying the seal may be done optionally here, or explicitly
 // via the VerifySeal method.
 func (sb *backend) VerifyHeader(chain consensus.ChainReader, header *types.Header, seal bool) error {
-	return sb.verifyHeader(chain, header, nil)
+	parents := []*types.Header{chain.GetHeader(header.ParentHash, header.Number.Uint64())}
```
blukat29, line 175: According to [verifyCascadingFields()](https://github.com/klaytn/klaytn/blob/v1.8.4/consensus/istanbul/backend/engine.go#L206-L218), you should use `number-1` to find parent header.

```suggestion
        parents := []*types.Header{}
        number := header.Number.Uint64()
        if number > 0 {
                parents = append(parents, chain.GetHeader(header.ParentHash, number-1))
        }
```

work/worker.go
```diff
@@ -538,9 +538,10 @@ func (self *worker) commitNewWork() {
 	// Create the current work task
 	work := self.current
 	if self.nodetype == common.CONSENSUSNODE {
-		// TODO : Need to KIP-71 hardfork.
-		// TODO : It need to uncomment after implemented baseFee logic.
-		// pending = types.FilterTransactionWithBaseFee(pending, baseFee)
+		if self.config.IsKIP71ForkEnabled(header.Number) {
+			header.BaseFee = misc.NextBlockBaseFee(parent.Header(), self.config)
+			pending = types.FilterTransactionWithBaseFee(pending, header.BaseFee)
+		}
```
blukat29, line 544: `FilterTransactionWithBaseFee()` function has a heavy nested for-loop.
I think we should run heavy tasks before waiting for ideal time to not consume precious consensus time.
Can we move this if block to the start of `commitNewWork`? (where we've just obtained `pending` list)

consensus/istanbul/backend/engine.go
kjeom: @blukat29 Thanks! great code :)

work/worker.go
kjeom: Thanks for good Idea. I'll try it

blukat29:
Perhaps we can merge `TxPool.Pending()` and `FilterTransactionWithBaseFee()` into one function, for instance `TxPool.PendingAboveBaseFee()`. We can reduce a for-loop.

```go
// TxPool.Pending()
	pending := make(map[common.Address]types.Transactions)
	for addr, list := range pool.pending {
		pending[addr] = list.Flatten()
	}
// FilterTransactionWithBaseFee()
	txMap := make(map[common.Address]Transactions)
	for addr, list := range pending {
		txs := list
		for i, tx := range list {
			if tx.GasPrice().Cmp(baseFee) < 0 {
				txs = list[:i]
				break
			}
		}
		if len(txs) > 0 {
			txMap[addr] = txs
		}
	}
	return txMap
```

```go
// TxPool.PendingAboveBaseFee()
	pending := make(map[common.Address]types.Transactions)
	for addr, list := range pool.pending {
		txs := list
		for i, tx := range list {
			if tx.GasPrice().Cmp(baseFee) < 0 {
				txs = list[:i]
				break
			}
		}
		if len(txs) > 0 {
			pending[addr] = txs
		}
	}
```

work/worker.go
blukat29: On a second thought, we might be able to integrate FilterTransactionWithBaseFee into Pending(). Let me elaborate in another comment.

kjeom: Yeah, @blukat29 actually I also found many sites to be refactored. However I'd like to remain it `TODO`.
Because I think txPool code is complex and some fix can cause unexpected side effects. How about considering it separately later?

blukat29: That's fair. I see that performance is not the top priority.

blockchain/tx_list.go
```diff
@@ -308,30 +308,30 @@ func (l *txList) Overlaps(tx *types.Transaction) bool {
...
-		//	return false, nil
-		//}
-
-		// If gas price of older is bigger than newer, abort.
-		if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
-			logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
-			return false, nil
+		if kip71Hardforked {
+			// If gas price of older is bigger than newer, abort.
+			if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
```
jimni1222, line 317: Don't we need to add `tx.Type().IsCancelTransaction()` case when `kip71Hardforked` also?

blockchain/tx_pool.go
```diff
@@ -658,38 +670,42 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
...
-		//}
-
-		// Ensure transaction's gasFeeCap is greater than or equal to transaction pool's gasPrice(baseFee).
-		if pool.gasPrice.Cmp(tx.GasFeeCap()) > 0 {
-			logger.Trace("fail to validate maxFeePerGas", "unitPrice", pool.gasPrice, "maxFeePerGas", tx.GasFeeCap())
-			return ErrFeeCapBelowBaseFee
+		if isKIP71Forked {
+			// Ensure transaction's gasFeeCap is greater than or equal to transaction pool's gasPrice(baseFee).
+			if pool.gasPrice.Cmp(tx.GasFeeCap()) > 0 {
+				logger.Trace("fail to validate maxFeePerGas", "unitPrice", pool.gasPrice, "maxFeePerGas", tx.GasFeeCap())
```
jimni1222, line 676:
```suggestion
				logger.Trace("fail to validate maxFeePerGas", "pool.gasPrice", pool.gasPrice, "maxFeePerGas", tx.GasFeeCap())
```

blockchain/tx_pool.go
```diff
@@ -870,7 +886,7 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
 	from, _ := types.Sender(pool.signer, tx) // already validated
 	if list := pool.pending[from]; list != nil && list.Overlaps(tx) {
 		// Nonce already pending, check if required price bump is met
-		inserted, old := list.Add(tx, pool.config.PriceBump)
+		inserted, old := list.Add(tx, pool.config.PriceBump, pool.chainconfig.IsKIP71ForkEnabled(pool.chain.CurrentBlock().Number()))
```
jimni1222, line 889: Is this fine to use current block number? not current block number + 1?

kjeom: Yes, you are correct. Thanks!

blockchain/tx_pool.go
```diff
@@ -460,6 +461,7 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	// Enable Ethereum tx type transactions
 	pool.eip2718 = pool.chainconfig.IsEthTxTypeForkEnabled(next)
 	pool.eip1559 = pool.chainconfig.IsEthTxTypeForkEnabled(next)
+	pool.kip71 = pool.chainconfig.IsKIP71ForkEnabled(next)
```
jimni1222, line 464:
```suggestion
	// Enable dynamic base fee
	pool.kip71 = pool.chainconfig.IsKIP71ForkEnabled(next)
```

blockchain/tx_pool.go
```diff
@@ -511,6 +520,11 @@ func (pool *TxPool) SetGasPrice(price *big.Int) {
 	}
 }
 
+// only for test
+func (pool *TxPool) SetBaseFee(baseFee *big.Int) {
```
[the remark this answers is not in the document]
kjeom: Just for test. To test changing the `gasPrice` in txPool we need the setter like it because we added a guard if hardforked or not in legacy function, `SetGasPrice`

blockchain/tx_pool.go
```diff
@@ -1357,15 +1375,24 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) {
...
+					promoted = append(promoted, tx)
+				}
+			}
+		} else {
+			for _, tx := range list.Ready(pool.getPendingNonce(addr)) {
+				hash := tx.Hash()
+				if pool.promoteTx(addr, hash, tx) {
+					logger.Trace("Promoting queued transaction", "hash", hash)
+					promoted = append(promoted, tx)
+				}
```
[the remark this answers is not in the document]
kjeom: What you mean, a one-time iteration?

consensus/misc/kip71.go
```diff
@@ -24,14 +24,15 @@ func VerifyKIP71Header(config *params.ChainConfig, parentHeader, header *types.H
 }
 
 func NextBlockBaseFee(parentHeader *types.Header, config *params.ChainConfig) *big.Int {
```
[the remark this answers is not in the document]
hyunsooda: Let me fix ambiguous wording first. Did you mean `first hardforked block` as genesis block of KIP71 enabled?

1. The guard seems not for the only genesis block. There's another condition for the parent header's KIP71 status. If `self.config.IsKIP71ForkEnabled(num - 1)` is true, then `self.config.IsKIP71ForkEnabled(num)` is true too. So, I was confused about why the latter condition is located throughout? 
2. Make sense. Thank you.

blockchain/tx_pool_test.go
```diff
@@ -2083,10 +2097,12 @@ func TestTransactionsPromoteFull(t *testing.T) {
...
-	assert.Equal(t, pool.pending[from].Len(), 4)
-
-	for i, tx := range txs {
-		assert.True(t, reflect.DeepEqual(tx, pool.pending[from].txs.items[uint64(i)]))
+	r := assert.NotEqual(t, len(pool.pending), 0)
+	if r {
+		assert.Equal(t, pool.pending[from].Len(), 4)
+		for i, tx := range txs {
+			assert.True(t, reflect.DeepEqual(tx, pool.pending[from].txs.items[uint64(i)]))
+		}
```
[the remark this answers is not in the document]
hyunsooda: Ideally and as preceding expected, I thought `pool.pending[from].Len()` must be four. Why may it panic by your change?

blockchain/tx_pool.go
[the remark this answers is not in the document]
hyunsooda: Let me fix it. One-time iteration --> one iteration. Still confused?

blockchain/tx_pool.go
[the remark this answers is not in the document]
hyunsooda: No, I meant why did you place it here, not into a test file.

blockchain/tx_pool.go
```diff
@@ -1550,22 +1577,23 @@ func (pool *TxPool) demoteUnexecutables() {
...
+				for _, tx := range list.Flatten() {
+					hash := tx.Hash()
+					if tx.GasPrice().Cmp(pool.gasPrice) < 0 {
+						logger.Trace("Demoting the tx that is lower than the baseFee and those greater than the nonce of the tx.", "txhash", hash)
+						removed, invalids := list.Remove(tx) // delete all transactions satisfying the nonce value > tx.Nonce()
+						if removed {
+							for _, invalidTx := range invalids {
+								pool.enqueueTx(invalidTx.Hash(), invalidTx)
+							}
+							pool.enqueueTx(hash, tx)
```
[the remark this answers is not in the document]
hyunsooda: Got it. Thank you.

blockchain/tx_pool.go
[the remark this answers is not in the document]
kjeom: It needs to set the `gasPrice` field in txPool.

blockchain/tx_pool_test.go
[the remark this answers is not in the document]
kjeom: It is a trivial issue, not critical. When I test it for kip71 hardfork, It returns the error.
I just added the guard because I don't want to see the panic. 
The error fixed now. So if you think this code is unnecessary, you can delete it.

consensus/misc/kip71.go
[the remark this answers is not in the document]
kjeom: - Right, first hardforked block means a KIP71 enabled genesis block.
- Right, but we have an assumption that once hardfork is activated, there is no going back before the hardfork.

blockchain/tx_pool.go
```diff
@@ -419,9 +421,10 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	pool.currentState = stateDB
 	pool.pendingNonce = make(map[common.Address]uint64)
 	pool.currentBlockNumber = newHead.Number.Uint64()
-	// TODO :  Need to KIP-71 hardfork
-	// TODO : It need to update gas price of tx pool after implemnted baseFee calculation logic.
-	// pool.gasPrice = misc.calculateBaseFee()
+	// It need to update gas price of tx pool after kip71 hardfork
+	if pool.kip71 {
```
jimni1222, line 425: Is this okay to put here? How about moving this logic after [L465](https://github.com/klaytn/klaytn/pull/1442/files#diff-b4c0a52e4fd03bcfee19698d39bf4eb83280db74b214427a02752f49cac20757R317)?

kjeom: It would have been shifted.. Thanks!!

blockchain/tx_list.go
```diff
@@ -308,30 +308,35 @@ func (l *txList) Overlaps(tx *types.Transaction) bool {
...
-			logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
-			return false, nil
+		if kip71Hardforked {
+			// If tx is CancelTransaction, replace it even thought tx has lower gasPrice than previous tx.
+			if tx.Type().IsCancelTransaction() {
+				logger.Trace("New tx is a cancel transaction. replace it!", "old", old.String(), "new", tx.String())
+			}
+
+			// If gas price of older is bigger than newer, abort.
+			if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
```
jimni1222, line 322: I think this should be `else if` now to connect with above if.

kjeom: Thanks! It is fixed.

blockchain/tx_pool.go
```diff
@@ -1550,22 +1578,23 @@ func (pool *TxPool) demoteUnexecutables() {
...
-			needRemoved := false
-			for _, tx := range list.Flatten() {
-				hash := tx.Hash()
-				if needRemoved || tx.GasPrice().Cmp(pool.gasPrice) < 0 {
-					if needRemoved {
-						logger.Trace("Demoting the transaction due to the previous transaction that has lower gas price than baseFee", "txhash", hash)
-					} else {
-						needRemoved = true
-						logger.Trace("Demoting the transaction that has lower gas price than baseFee", "txhash", hash)
+		if pool.kip71 {
```
jimni1222, line 1583: Can you explain why `if pool.kip71` condition check should be here?
Is it not necessary to execute the logic below before the hard fork?

kjeom: Yes, It is added in https://github.com/klaytn/klaytn/pull/1382/files#diff-4a20f160f5182ec35d76a3566352906f577b84fbf3b242b922c4d2fc9c219cd7R1553.

blockchain/tx_pool_test.go
```diff
@@ -65,6 +68,15 @@ func init() {
...
 	fork.SetHardForkBlockNumberConfig(eip1559Config)
+
+	cpy = *params.TestChainConfig
+	kip71Config = &cpy
+	kip71Config.KIP71CompatibleBlock = common.Big0
+	kip71Config.IstanbulCompatibleBlock = common.Big0
+	kip71Config.LondonCompatibleBlock = common.Big0
+	kip71Config.EthTxTypeCompatibleBlock = common.Big0
+	kip71Config.Governance = &params.GovernanceConfig{KIP71: params.GetDefaultKip71Config()}
+	fork.SetHardForkBlockNumberConfig(eip1559Config)
```
jimni1222, line 79:
```suggestion
	fork.SetHardForkBlockNumberConfig(kip71Config)
```

kjeom: I think we need to delete it. kip71 doesn't use `fork` instance. Thanks!

tests/klay_test_blockchain_test.go
```diff
@@ -166,6 +166,9 @@ func (bcdata *BCData) prepareHeader() (*types.Header, error) {
 		Governance: common.Hex2Bytes("b8dc7b22676f7665726e696e676e6f6465223a22307865373333636234643237396461363936663330643437306638633034646563623534666362306432222c22676f7665726e616e63656d6f6465223a2273696e676c65222c22726577617264223a7b226d696e74696e67616d6f756e74223a393630303030303030303030303030303030302c22726174696f223a2233342f33332f3333227d2c22626674223a7b2265706f6368223a33303030302c22706f6c696379223a302c22737562223a32317d2c22756e69745072696365223a32353030303030303030307d"),
 		Vote:       common.Hex2Bytes("e194e733cb4d279da696f30d470f8c04decb54fcb0d28565706f6368853330303030"),
 	}
+	if bcdata.bc.Config().IsKIP71ForkEnabled(num) {
```
jimni1222, line 169: Is this fine to check fork enabled with parent block number here?

kjeom: Ahh, child block number should be checked here. Thanks!

tests/tx_cancel_test.go
```diff
@@ -16,482 +16,486 @@
...
+
+		txpool.AddRemotes(txs)
+
+		pending, queued := txpool.Content()
+		assert.Equal(t, 0, len(queued))
+		assert.Equal(t, 1, len(pending))
+		assert.True(t, tx.Equal(pending[*bcdata.addrs[0]][0]))
+	}
+
+	// 4. Insert a TxCancel transaction with nonce 0 and different gas limit. This should replace the tx with the same nonce.
```
jimni1222, line 152: Can you add a test case with TxTypeCancel that has smaller gas limit and same nonce?

jimni1222: sorry please ignore this

blockchain/tx_list_test.go
```diff
@@ -195,11 +195,11 @@ func TestSubstituteTransactionAbort(t *testing.T) {
 	oldTx := pricedTransaction(0, 21000, big.NewInt(50), key)
 	newTx := pricedTransaction(0, 21000, big.NewInt(40), key)
 
-	if result, _ := txList.Add(oldTx, DefaultTxPoolConfig.PriceBump); !result {
+	if result, _ := txList.Add(oldTx, DefaultTxPoolConfig.PriceBump, true); !result {
 		t.Error("it cannot add tx in tx list.")
 	}
 
-	if result, replaced := txList.Add(newTx, DefaultTxPoolConfig.PriceBump); result || replaced != nil {
+	if result, replaced := txList.Add(newTx, DefaultTxPoolConfig.PriceBump, true); result || replaced != nil {
```
jimni1222, line 202: Could you add test code for cancel transaction (Whatever the gas price is, if it is greater than the base rate, replace it with the old tx)?
Or is there any test code for that?

blockchain/tx_pool.go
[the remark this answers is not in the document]
jimni1222: I think what his meaning is that define `var txs types.Transactions` and depends on `pool.kip71`, use `txs = list. ReadyWithGasPrice` or `txs = list.Ready`.
Then you can remove redundant code below.

blockchain/tx_pool_test.go
```diff
@@ -309,19 +320,21 @@ func TestInvalidTransactions(t *testing.T) {
...
-		t.Error("expected", ErrGasPriceBelowBaseFee, "got", err)
+	// regardless of local or remote.
+	if pool.chainconfig.IsKIP71ForkEnabled(pool.chain.CurrentBlock().Number()) {
+		if err := pool.AddRemote(tx); err != ErrGasPriceBelowBaseFee {
+			t.Error("expected", ErrGasPriceBelowBaseFee, "got", err)
+		}
+		if err := pool.AddLocal(tx); err != ErrGasPriceBelowBaseFee {
+			t.Error("expected", ErrGasPriceBelowBaseFee, "got", err)
+		}
+	} else {
```
jimni1222, line 331: I don't think this else logic will be operated.
How about set kip71 value of the tx pool like `pool.kip71 = true/false` and then run `AddRemote` and `AddLocal`?

blockchain/tx_pool.go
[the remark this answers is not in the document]
kjeom: I think It would be better. Thanks!

blockchain/tx_pool_test.go
kjeom: Right, we can leave this code for before kip71 hardfork because `TestTransactionNotAcceptedWithLowerGasPrice` can cover this situation.

blockchain/tx_list_test.go
kjeom: Sure, I'll try it

blockchain/tx_pool_test.go
```diff
@@ -1970,9 +1983,9 @@ func TestDynamicFeeTransactionAccepted(t *testing.T) {
 	t.Parallel()
 	baseFee := big.NewInt(30)
 
-	pool, key := setupTxPoolWithConfig(eip1559Config)
```
kjhman21, line 1973: Don't we need to have a test case for eip1559 enabled and kip71 is disabled?

consensus/misc/kip71.go
```diff
@@ -24,14 +24,15 @@ func VerifyKIP71Header(config *params.ChainConfig, parentHeader, header *types.H
...
+	// governance parameters
+	lowerBoundBaseFee := new(big.Int).SetUint64(config.Governance.KIP71.LowerBoundBaseFee)
+	upperBoundBaseFee := new(big.Int).SetUint64(config.Governance.KIP71.UpperBoundBaseFee)
+
 	// If the parent is the kip71 disabled block or genesis, then return default base fee (250ston)
 	if !config.IsKIP71ForkEnabled(parentHeader.Number) || parentHeader.Number.Cmp(new(big.Int).SetUint64(0)) == 0 {
-		return new(big.Int).SetUint64(config.UnitPrice)
+		return lowerBoundBaseFee
```
kjhman21, line 33: Can we say that lowerBoundBaseFee == 250 ston?

kjeom: Ahh, the comment should be edited by 25ston, Thanks!

blockchain/tx_pool_test.go
kjeom: It could be done with `setupTxPoolWithConfig(eip1559Config)` that means eip1559 enabled and kip71 disabled.

blockchain/tx_list_test.go
kjeom: The testcase is added in `TestCancelTransactionAccepted`

consensus/misc/kip71.go
blukat29: It's a little confusing to me what the 'default base fee' refers to.

- If you intention is to return lowerBoundBaseFee (config.Governance.KIP71.UpperBoundBaseFee), then the comment can be
  ```go
  // If the parent is the kip71 disabled block or genesis, then return the LowerBoundBaseFee (default 25 ston)
  ```
- If your intention is to return default base fee (params.DefaultLowerBoundBaseFee = 25ston) then the code should be
  ```go
  return params.DefaultLowerBoundBaseFee
  ```

consensus/misc/kip71.go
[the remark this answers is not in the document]
hyunsooda: The necessary condition is already located in a specific function, but you made redundant guards that are already placed inside of it. This was my question. So, I'm still confused about which correlation is there between the preceding and subsequent hardfork and redundant guards?

blockchain/tx_pool.go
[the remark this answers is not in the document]
hyunsooda: I'm not asking about its functionality, asked where the test function is located.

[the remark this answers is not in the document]
kjeom: Are you asking where is the location to use `SetBaseFee`?

blockchain/tx_pool_test.go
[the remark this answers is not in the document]
kjeom: I deleted it.

consensus/misc/kip71.go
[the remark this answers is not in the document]
kjeom: I think you are confused. these are different each other.
callsite condition : config.IsKIP71ForkEnabled(parentHeader.Number)
condition inside : !config.IsKIP71ForkEnabled(parentHeader.Number)

consensus/misc/kip71.go
kjeom: Ok, I intended the former. fixed it.

jimni1222 (approved):

consensus/misc/kip71.go
kjhman21: @kjeom Hmm? really? anyway current gas price is 250 ston and before KIP-17 is enabled, we should return 250 ston, shouldn't we?

blockchain/tx_pool_test.go
kjhman21: You mean we have other test cases that having setupTxPoolWithConfig(eip1559Config)?

I'm asking because you replaced it with setupTxPoolWithConfig(kip71Config). It seems like you just removed the previous test cases and replaced it with a new test case with KIP-71.

blockchain/tx_pool.go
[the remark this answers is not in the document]
hyunsooda: The second one. The function can be located in `tx_pool_test.go` and there's no constraint to be located in `tx_pool.go`. If you think it's trivial, how about skipping it?

consensus/misc/kip71.go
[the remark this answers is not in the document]
hyunsooda: They seem the same.. no? The callsite condition should be true to call `NextBaseFee()`. The condition of `NextBaseFee()` should be true to calculate the next base fee. Then, two conditions are equivalent. If you think this redundant check is trivial, you can resolve it.

blockchain/tx_list.go
```diff
@@ -308,30 +308,33 @@ func (l *txList) Overlaps(tx *types.Transaction) bool {
...
-		//	return false, nil
-		//}
-
-		// If gas price of older is bigger than newer, abort.
-		if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
-			logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
-			return false, nil
+		if kip71Hardforked {
+			// If tx is CancelTransaction, replace it even thought tx has lower gasPrice than previous tx.
+			if tx.Type().IsCancelTransaction() {
```
jimni1222, line 317: This is not important, but how about making `if tx.Type().IsCancelTransaction()` as a common condition like below?

```
// If tx is CancelTransaction, replace it even thought tx has lower gasPrice than previous tx.
if tx.Type().IsCancelTransaction() {
    logger.Trace("New tx is a cancel transaction. replace it!", "old", old.String(), "new", tx.String())
} else if kip71Hardforked {
    if old.GasPrice().Cmp(tx.GasPrice()) >= 0 {
        // If gas price of older is bigger than newer, abort.
        logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
        return false, nil
    }
    // Otherwise overwrite the old transaction with the current one.
    logger.Trace("The transaction was substituted by competitive gas price", "old", old.String(), "new", tx.String())
} else {
    logger.Trace("already nonce exist", "nonce", tx.Nonce(), "with gasprice", old.GasPrice(), "priceBump", priceBump, "new tx.gasprice", tx.GasPrice())
    return false, nil
}
```

kjeom: It is more clear. refactored it.

blockchain/tx_pool_test.go
```diff
@@ -52,18 +52,29 @@ var (
...
 
+
 	eip1559Config = params.TestChainConfig.Copy()
 	eip1559Config.IstanbulCompatibleBlock = common.Big0
 	eip1559Config.LondonCompatibleBlock = common.Big0
 	eip1559Config.EthTxTypeCompatibleBlock = common.Big0
-	fork.SetHardForkBlockNumberConfig(eip1559Config)
+	fork.SetHardForkBlockNumberConfig(&eip1559Config)
+
+	kip71Config = *params.TestChainConfig
```
kjhman21, line 72:
```suggestion
	kip71Config = params.TestChainConfig.Copy()
```

I think this would be better?

blockchain/tx_pool_test.go
```diff
@@ -341,19 +370,12 @@ func TestInvalidTransactions(t *testing.T) {
...
-	// TODO : Need to KIP-71 hardfork.
-	//if err := pool.AddRemote(tx); err != ErrInvalidUnitPrice {
-	//	t.Error("expected", ErrInvalidUnitPrice, "got", err)
-	//}
-	//if err := pool.AddLocal(tx); err != ErrInvalidUnitPrice {
-	//	t.Error("expected", ErrInvalidUnitPrice, "got", err)
-	//}
-	if err := pool.AddRemote(tx); err != ErrGasPriceBelowBaseFee {
-		t.Error("expected", ErrGasPriceBelowBaseFee, "got", err)
+	// regardless of local or remote.
```
kjhman21, line 373: Don't we need to have test cases for both before KIP-71 and after KIP-71?

blockchain/tx_pool_test.go
```diff
@@ -115,6 +130,20 @@ func dynamicFeeTx(nonce uint64, gaslimit uint64, gasFee *big.Int, tip *big.Int,
...
+		types.TxValueKeyGasLimit: gasLimit,
+		types.TxValueKeyGasPrice: gasPrice,
+		types.TxValueKeyFrom:     from,
+	})
+	if err != nil {
+		// Since we do not have testing.T here, call panic() instead of t.Fatal().
+		panic(err)
+	}
+	cancelTx := types.NewTx(d)
+	signedTx, _ := types.SignTx(cancelTx, types.LatestSignerForChainID(params.TestChainConfig.ChainID), key)
```
jimni1222, line 145:
```suggestion
	signedTx, _ := types.SignTx(cancelTx, types.LatestSignerForChainID(params.TestChainConfig.ChainID), key)
	return signedTx
}
```

consensus/misc/kip71.go
kjhman21: @kjeom Then, why we need to check `config.IsKIP71ForkEnabled(parentHeader.Number)` if you're sure that this will not be called if KIP-71 is not enabled?

blockchain/tx_pool.go
```diff
@@ -658,38 +668,42 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
...
-			return ErrFeeCapBelowBaseFee
+		if pool.kip71 {
+			// Ensure transaction's gasFeeCap is greater than or equal to transaction pool's gasPrice(baseFee).
+			if pool.gasPrice.Cmp(tx.GasFeeCap()) > 0 {
+				logger.Trace("fail to validate maxFeePerGas", "pool.gasPrice", pool.gasPrice, "maxFeePerGas", tx.GasFeeCap())
+				return ErrFeeCapBelowBaseFee
+			}
+		} else {
+			// Ensure gasFeeCap is greater than or equal to gasTipCap.
+			if tx.GasFeeCap().Cmp(tx.GasTipCap()) < 0 {
```
jimni1222, line 679: I think we can make this condition as a common condition (not only when `!pool.kip71`).

blockchain/tx_pool.go
```diff
@@ -658,38 +668,42 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
...
-		//	return ErrInvalidUnitPrice
-		//}
-
-		// Ensure transaction's gasPrice is greater than or equal to transaction pool's gasPrice(baseFee).
-		if pool.gasPrice.Cmp(tx.GasPrice()) > 0 {
-			logger.Trace("fail to validate gasprice", "pool.gasPrice", pool.gasPrice, "tx.gasPrice", tx.GasPrice())
-			return ErrGasPriceBelowBaseFee
+		if pool.kip71 {
+			// Ensure transaction's gasPrice is greater than or equal to transaction pool's gasPrice(baseFee).
+			if pool.gasPrice.Cmp(tx.GasPrice()) > 0 && tx.Type() != types.TxTypeCancel {
```
jimni1222, line 697: If tx is a transaction cancel type, shouldn't it be necessary to satisfy `pool.gasPrice.Cmp(tx.GasPrice()) <= 0`?
I think cancel tx should replace without comparison with gas price of old tx, but still should satisfy `pool.gasPrice.Cmp(tx.GasPrice()) <= 0`.

blockchain/tx_pool.go
```diff
@@ -1557,22 +1576,23 @@ func (pool *TxPool) demoteUnexecutables() {
...
-			needRemoved := false
-			for _, tx := range list.Flatten() {
-				hash := tx.Hash()
-				if needRemoved || tx.GasPrice().Cmp(pool.gasPrice) < 0 {
-					if needRemoved {
-						logger.Trace("Demoting the transaction due to the previous transaction that has lower gas price than baseFee", "txhash", hash)
-					} else {
-						needRemoved = true
-						logger.Trace("Demoting the transaction that has lower gas price than baseFee", "txhash", hash)
+		if pool.kip71 {
```
jimni1222, line 1581: How about changing this to `if pool.kip71 && list.Len() > 0`?

blockchain/tx_pool_test.go
```diff
@@ -1941,45 +1963,45 @@ func TestDynamicFeeTransactionVeryHighValues(t *testing.T) {
...
-//	if err := pool.AddRemote(tx); err != nil {
-//		t.Error("error", "got", err)
-//	}
-//
-//	tx2 := dynamicFeeTx(1, 21000, big.NewInt(1), big.NewInt(1), key)
-//	if err := pool.AddRemote(tx2); err != nil {
-//		t.Error("error", "got", err)
-//	}
-//}
+// func TestDynamicFeeTransactionHasNotSameGasPrice(t *testing.T) {
```
jimni1222, line 1966: Why don't we test those?

blockchain/tx_pool_test.go
kjeom: It is deleted while merging with Dev branch. I added it. Thanks!

blockchain/tx_pool_test.go
```diff
@@ -2231,9 +2340,10 @@ func TestTransactionsPromoteMultipleAccount(t *testing.T) {
 func TestTransactionsDemotionMultipleAccount(t *testing.T) {
 	t.Parallel()
 
-	pool, _ := setupTxPoolWithConfig(eip1559Config)
+	pool, _ := setupTxPoolWithConfig(&kip71Config)
 	defer pool.Stop()
-	pool.SetGasPrice(big.NewInt(10))
+	// pool.SetBaseFee(big.NewInt(10))
+	pool.gasPrice = big.NewInt(10)
```
jimni1222, line 2346: why not use `pool.SetBaseFee`?

blockchain/tx_pool_test.go
```diff
@@ -2277,11 +2387,12 @@ func TestTransactionsDemotionMultipleAccount(t *testing.T) {
 	// tx[6] : queue[from[2]]
 	// tx[7] : queue[from[2]]
 	// tx[7] : queue[from[2]]
+	// pool.SetBaseFee(big.NewInt(35))
 	pool.gasPrice = big.NewInt(35)
```
jimni1222, line 2391: here also !!! why not use `pool.SetBaseFee`? 
I think use one way to set gasPrice of the tx pool better!! (If you prefer `pool.gasPrice = `, you can change to that)

kjeom: I agree. Thansk!

blockchain/tx_pool_test.go
kjeom: Actually I just wanted to test these in a function. But It needs to be forked in test function and I don't want to test like that. I would rather check internally doing hardfork well, but can't. So I added `TestDynamicFeeTransactionAcceptedKip71` test case and enabled above function you suggested in this review.
Thanks!

blockchain/tx_pool_test.go
kjeom: It is refactored. Thanks!

blockchain/tx_pool.go
kjeom: I tested the case in `TestCancelTransactionAcceptedKip71`. If `tx == TxTypeCancel`, it just pass the condition. Why are you thinking `TxTypeCancel` should satisfy `pool.gasPrice.Cmp(tx.GasPrice()) <= 0`?

blockchain/tx_pool.go
kjeom: You mean It's better to make a function to check this condition?

blockchain/tx_pool_test.go
kjeom: I agree. It is added in `TestInvalidTransactionsKip71`

blockchain/tx_pool_test.go
```diff
@@ -52,18 +52,29 @@ var (
...
 	eip1559Config = params.TestChainConfig.Copy()
 	eip1559Config.IstanbulCompatibleBlock = common.Big0
 	eip1559Config.LondonCompatibleBlock = common.Big0
 	eip1559Config.EthTxTypeCompatibleBlock = common.Big0
-	fork.SetHardForkBlockNumberConfig(eip1559Config)
+	fork.SetHardForkBlockNumberConfig(&eip1559Config)
+
+	kip71Config = *params.TestChainConfig
+	kip71Config.KIP71CompatibleBlock = common.Big0
+	kip71Config.IstanbulCompatibleBlock = common.Big0
```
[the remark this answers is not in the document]
kjeom: `func init() {
	testTxPoolConfig = DefaultTxPoolConfig
	testTxPoolConfig.Journal = ""

	eip1559Config = params.TestChainConfig.Copy()
	eip1559Config.IstanbulCompatibleBlock = common.Big0
	eip1559Config.LondonCompatibleBlock = common.Big0
	eip1559Config.EthTxTypeCompatibleBlock = common.Big0
	fork.SetHardForkBlockNumberConfig(eip1559Config)

	kip71Config = params.TestChainConfig.Copy()
	kip71Config.KIP71CompatibleBlock = common.Big0
	kip71Config.IstanbulCompatibleBlock = common.Big0
	kip71Config.LondonCompatibleBlock = common.Big0
	kip71Config.EthTxTypeCompatibleBlock = common.Big0
	kip71Config.Governance = &params.GovernanceConfig{KIP71: params.GetDefaultKip71Config()}
}`
This code is recent version. where did you find the code?

blockchain/tx_pool.go
```diff
@@ -493,6 +499,10 @@ func (pool *TxPool) GasPrice() *big.Int {
 
 // SetGasPrice updates the gas price of the transaction pool for new transactions, and drops all old transactions.
 func (pool *TxPool) SetGasPrice(price *big.Int) {
+	if pool.kip71 {
+		logger.Warn("Ignoring SetGasPrice after KIP71 fork")
```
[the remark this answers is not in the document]
kjeom: You mean It's better to delete the logging or change log level something like `Info`?

work/worker.go
```diff
@@ -495,6 +495,14 @@ func (self *worker) commitNewWork() {
 	defer self.currentMu.Unlock()
 
 	parent := self.chain.CurrentBlock()
+	nextBlockNum := new(big.Int).Add(parent.Number(), common.Big1)
+	var nextBaseFee *big.Int
+	if self.nodetype == common.CONSENSUSNODE {
+		if self.config.IsKIP71ForkEnabled(nextBlockNum) {
+			nextBaseFee = misc.NextBlockBaseFee(parent.Header(), self.config)
+			pending = types.FilterTransactionWithBaseFee(pending, nextBaseFee)
```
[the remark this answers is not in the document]
kjeom: You mean It's better to merge the two logics?

blockchain/tx_pool.go
```diff
@@ -1557,22 +1576,21 @@ func (pool *TxPool) demoteUnexecutables() {
 			}
 		}
 
-		// TODO : Need to KIP-71 hardfork.
 		// Enqueue transaction if gasPrice of transaction is lower than gasPrice of txPool.
 		// All transactions with a nonce greater than enqueued transaction also stored queue.
-		if list.Len() > 0 {
-			needRemoved := false
+		if pool.kip71 && list.Len() > 0 {
```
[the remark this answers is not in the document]
kjeom: It's okay. This logic is newly added for kip71.

blockchain/tx_pool.go
jimni1222: I thought we will keep this condition after hard fork also.. maybe i misunderstood?

blockchain/tx_pool.go
kjeom: If any cancelTx can be accepted, an attacker could fill the txPool with unprocessed transactions.
It should be fixed. Thanks!

kjeom: fixed it.

jimni1222 (approved):
LGTM! Thank you for hard working!

blockchain/tx_pool.go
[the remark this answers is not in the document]
yoomee1313: Yes, because it is a normal behavior, not a warning thing. As you said, Info level seems better.

blockchain/tx_pool_test.go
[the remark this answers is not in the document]
yoomee1313: Oh, there's a updated version..

work/worker.go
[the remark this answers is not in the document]
yoomee1313: Is there a difference? As far as i understand, those lines are just allocating `pending`. Anyway, is it okay to allocate pending under self.mu & self.currentMu critical section? Because, before hard-fork, the pending is allocated at outside of the section.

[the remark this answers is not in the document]
blukat29: @yoomee1313 Your question may be related to this discussion (https://github.com/klaytn/klaytn/pull/1442#discussion_r899950415). I thought it's inefficient to traverse the txpool twice. So I suggested to merge two loops (https://github.com/klaytn/klaytn/pull/1442#issuecomment-1158703822). But I also think optimization can be done later.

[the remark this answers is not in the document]
blukat29: Oh it's not related to performance.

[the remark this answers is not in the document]
yoomee1313: @blukat29 thanks for letting me know. Anyway, it looks strange because pending is allocated at different section depending on hard-fork.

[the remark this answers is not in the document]
blukat29: - It seems the mission of `self.mu` to protect fields under `self`, not local variables.
  Allocation time of `pending` doesn't harm concurrency I guess?
- It's not trivial to move the filtering code above mutex after all. It requires `parent := self.chain.CurrentBlock()` which must be protected by `self.mu`.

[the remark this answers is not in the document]
kjeom: @blukat29 Thanks for explaining.
@yoomee1313 I think It's rather concerned about txPool mutex, not worker's mutex.
 - txPool.pending() copy all the transactions in pending of txPool and worker use these txs for filtering by BaseFee.
 - the `FilterTransactionWithBaseFee()` has no side effect except its arguments so If we can assure the argument is concurrency safe by worker, I think there's no problem.

blockchain/tx_pool.go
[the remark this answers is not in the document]
kjeom: It is changed, Thanks!

work/worker.go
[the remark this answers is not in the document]
yoomee1313: @blukat29 Oh That's why the pending allocation logic is seperated between critical section. It seems okay to me, however, i'm not sure, too.
@kjeom Thanks for the explanation I understand now.. You mean second allocation is better located in the critical section, right? What about adding a comment about this? It would be helpful to understand why moving is difficult!

blockchain/tx_pool_test.go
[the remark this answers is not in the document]
yoomee1313: Nevermind.. You don't have to do this :) Just ignore this comment!

blockchain/tx_pool_test.go
```diff
@@ -1997,14 +2061,40 @@ func TestDynamicFeeTransactionNotAcceptedNotEnableHardfork(t *testing.T) {
...
-// TestDynamicFeeTransactionAccepted tests that pool accept the transaction which has gasFeeCap bigger than or equal to baseFee.
-func TestDynamicFeeTransactionAccepted(t *testing.T) {
+func TestDynamicFeeTransactionAcceptedEip1559(t *testing.T) {
 	t.Parallel()
 	baseFee := big.NewInt(30)
 
 	pool, key := setupTxPoolWithConfig(eip1559Config)
 	defer pool.Stop()
-	pool.SetGasPrice(baseFee)
+	pool.SetBaseFee(baseFee)
```
kjhman21, line 2070: If this test case is for EIP-1559, don't  we need to call pool.SetGasPrice()?

consensus/misc/kip71.go
kjhman21: Okay. got it. Thanks.

tests/tx_validation_test.go
```diff
@@ -119,7 +118,7 @@ func TestValidationPoolInsert(t *testing.T) {
 		fn   func(types.TxType, txValueMap, common.Address) (txValueMap, error)
 	}{
 		{"invalidNonce", decreaseNonce},
-		{"invalidGasLimit", decreaseGasLimit},
+		{"invalidGasLimit", decreaseGasPrice},
```
kjhman21, line 121: 1. Name is inconsistent with the function name.
2. Also a test case for decreaseGasPriceKip71 should be added.

kjeom: 1. Right, Thansk!
2. TestValidationPoolInsertKip71 test case is what you intended? or you said this name needs to be renamed?

blockchain/tx_pool_test.go
kjeom: That's better, thanks!

tests/tx_validation_test.go
kjhman21: @kjeom We have a function named TestValidationPoolInsertKip71, but I couldn't find the test case calling that function. So, I wanted to add the test case calling the function.

kjeom: Ahh, `TestValidationPoolInsertKip71` calls the `decreaseGasPriceKip71`.

work/worker.go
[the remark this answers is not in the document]
kjeom: @yoomee1313 added comments. Please review it again.

blukat29 (approved):
Thank you for your efforts!

tests/tx_validation_test.go
kjhman21: got it.

kjhman21 (approved):
LGTM.

yoomee1313 (approved):

PR #1445, opened 2022-06-20
[SC] Unintended gas exhaustion attack

hyunsooda:
## Proposed changes

As the handle of value transfer transaction is created by operators on parent/child chains, the contract call must not execute any other unintended code. The `safeTransferFrom()` function calls `onERC721Received()` if the `_to` parameter is a contract address which can be attacker's contract. This change replaced `safeTransferFrom()` with `transferFrom()`.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


hyunsooda:
cc. @2dvorak, @hqjang-pepper, @nohkwak, @ian0371

contracts/bridge/BridgeTransferERC721.sol
```diff
@@ -66,7 +66,7 @@ contract BridgeTransferERC721 is BridgeTokens, IERC721BridgeReceiver, BridgeTran
         if (modeMintBurn) {
             require(ERC721MetadataMintable(_tokenAddress).mintWithTokenURI(_to, _tokenId, _tokenURI), "mint failed");
         } else {
-            IERC721(_tokenAddress).safeTransferFrom(address(this), _to, _tokenId);
+            IERC721(_tokenAddress).transferFrom(address(this), _to, _tokenId);
```
hqjang-pepper, line 69: If your change point is applied, onERC721Received() won't be executed even for a normal contract. So I'm concerned that there will be a side effect. Because if we simply use transferfrom(), then we can't check whether `to` address can receive token or not. 
Is it intended?

hyunsooda: I know that's why the `onERC721Received()` stands for further treatment when a token receives. This further treatment is not available in the receiver address is not a token contract address. If the transaction signer is the sender of that transaction, we do not need to make this change. However, the problem of the current value transfer design is that the handle transaction of value transfer is signed by the operator on each chain. So, it seems inevitable change.

hyunsooda:
@hyunsooda, TODO: Doc update

contracts/bridge/BridgeTransferERC721.sol
hqjang-pepper: Okay. There's another way to confirm whether we received ERC721 well or not, 
But we never know which contract is malicious or not. 
So this can be the proper solution

hqjang-pepper (approved):

2dvorak (approved):

hyunsooda:
Added doc: https://github.com/klaytn/klaytn-docs/pull/474

PR #1460, opened 2022-06-24
[SC] Temporal locking of token ID

hyunsooda:
## Proposed changes

The value transfer between two chains is done by contract calls. As the two chains are separate networks, the sender does not know the state of the counterpart chain. It was an implicit delegation that checks the state before issuing value transfer request transactions. Fortunately, ERC20 token is a just number and not unique. Thus, the caller of value transfer request does not need to check any precondition. Considering the NFT, however, the handle of value transfer may not succeed if a corresponding tokenID is already owned by someone. The problem is the sender's token is already burned or owned by the bridge contract address and does not be refined to senders.

Considering the scope of value transfer, anybody must be able to make a value transfer request transaction at any time. However, any transactions may not be aware of the implicit mandatory of checking a registered counterpart chain's token. This change removes this implicit mandatory and is delegated to the platform level.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


hyunsooda:
To be reimplemented following #1565

Issue #1448, opened 2022-06-21
GetStakingInfo RPC returns invalid Gini value


blukat29:
**Describe the bug**
The `governance_getStakingInfo` RPC result sometimes contains `Gini: -1`. Gini coefficients is defined between 0 and 1 and `-1` is an invalid value.
Request to "latest" block returns the correct value, but request to an old block returns `-1`. 

**How to reproduce**
Attach to an EN, call `governance.getStakingInfo()` with (1) an arbitrary block (2) "latest".

**Expected behavior**
If there are nonempty `CouncilStakingAmounts`, `Gini` value must be between 0 and 1.

**Attachments**
```javascript
> governance.getStakingInfo(91234567)
{
  BlockNum: 91065600,
  CouncilNodeAddrs: ["0x99fb17d324fa0e07f23b49d09028ac0919414db6", "0x571e53df607be97431a5bbefca1dffe5aef56f4d", "0xb74ff9dea397fe9e231df545eb53fe2adf776cb2", "0x5cb1a7dccbd0dc446e3640898ede8820368554c8", "0x776817c0ef3d06d794cf01ae9afa33d7397b9b40", "0xc180ca565b34b5b63877674f5fe647e7da079022", "0x03497f51c31fe8b402df0bde90fd5a85f87aa943", "0xfb416e2a714a5d1e034beacf0646a36876b72765", "0xd6d7762dd7fadf0512d83d0143685fc85488b64e", "0xe36ea146c05757a38ec1493eea8c878677a9c19a"],
  CouncilRewardAddrs: ["0xb2bd3178affccd9f9f5189457f1cad7d17a01c9d", "0x6559a7b6248b342bc11fbcdf9343212bbc347edc", "0x82829a60c6eac4e3e9d6ed00891c69e88537fd4d", "0xa86fd667c6a340c53cc5d796ba84dbe1f29cb2f7", "0x6e22cbe2b8bbd1df9f1d3c8ebae6d7ff5414a734", "0x24e593fb29731e54905025c230727dc28d229f77", "0x2b2a7a1d29a203f60e0a964fc64231265a49cd97", "0x37865a44b8cb07b00b29474758bf5b2dc823d3da", "0x9fe4ddee729d50e21e4ebe470092d61501d80a74", "0x7db2dbb1321611be0b7e5a145f3a5a95e3713c1d"],
[...]

PR #1464, opened 2022-06-24
Fill missing Gini field to the GetStakingInfo result

blukat29:
## Proposed changes

- The output `*StakingInfo` of `GetStakingInfo(num)` now have valid `Gini` field whenever applicable.
  - Previously the `StakingInfo.Gini` field was sometimes `-1`.
  - Now correct `StakingInfo.Gini` value is stored if `UseGini` is true.

- The proposer selection code does not update the `StakingInfo.Gini` in global cache.
  - Previously `weightedValidator.Refresh()` has updated the `StakingInfo.Gini` field stored in the StakingManager cache, affecting the global function `GetStakingInfo()`.
  - Now proposer selection code calculates its own `gini` value and use it.

- Note that the `StakingInfo.Gini` is different from `gini` used in proposer selection.
  - The `StakingInfo.Gini` is calculated among all CNs in the AddressBook, except the ones staked less than `minStake`.
  - The `gini` in proposer selection is calculated among validators (i.e. `ValSet`), a subset of all CNs.

- Add `ConsolidatedStakingInfo` to simplify gini calculation.
  - `ConsolidatedStakingInfo` is a set of `consolidatedNode`s.
  - A `consolidatedNode` is a collection of AddressBook entries under the same `RewardAddr`. Such entries have different `nodeId`s in AddressBook, but they constitute one effective Consensus Node.
  - This new data structure can aid simplifying weightedValidator.Refresh(). But further refactoring is outside the scope of this PR.

- With this PR, I expect the `governance_getStakingInfo` RPC to return meaningful Gini coefficient values.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

https://github.com/klaytn/klaytn/issues/1448

### DIFF
--- a/consensus/istanbul/validator/multi_staking_test.go
+++ b/consensus/istanbul/validator/multi_staking_test.go
@@ -170,7 +170,9 @@ func TestCalcTotalAmount(t *testing.T) {
 		stakingAmounts := testCase.stakingAmounts
 		totalAmount := calcTotalAmount(testCase.weightedValidators, testCase.stakingInfo, stakingAmounts)
 
-		assert.Equal(t, testCase.expectedGini, testCase.stakingInfo.Gini)
+		// Since calcTotalAmount does not update stakingInfo.Gini, we cannot assert expectedGini.
+		// However asserting expectedStakingAmounts proves that calcTotalAmount() used correct Gini.
+		// assert.Equal(t, testCase.expectedGini, testCase.stakingInfo.Gini)
 		assert.Equal(t, testCase.expectedTotalAmount, totalAmount)
 		assert.Equal(t, testCase.expectedStakingAmounts, stakingAmounts)
 	}
--- a/consensus/istanbul/validator/multi_staking_test.go
+++ b/consensus/istanbul/validator/multi_staking_test.go
@@ -115,28 +115,28 @@ func TestWeightedCouncil_getStakingAmountsOfValidators(t *testing.T) {
 // if UseGini is true, gini is calculated and reflected to stakingAmounts.
 func TestCalcTotalAmount(t *testing.T) {
 	testCases := []struct {
-		weightedValidators     []*weightedValidator
-		stakingInfo            *reward.StakingInfo
-		stakingAmounts         []float64
-		expectedGini           float64
-		expectedTotalAmount    float64
-		expectedStakingAmounts []float64
+		weightedValidators     []*weightedValidator // Produced by getStakingAmountsOfValidators()
+		stakingInfo            *reward.StakingInfo  // Produced by GetStakingInfo()
+		stakingAmounts         []float64            // Produced by getStakingAmountsOfValidators()
+		expectedGini           float64              // Gini among the []weightedValidators which is a subset of CouncilNodeAddrs
+		expectedTotalAmount    float64              // Sum of Gini-adjusted amounts
+		expectedStakingAmounts []float64            // Gini-adjusted amounts
 	}{
-		{
+		{ // Gini disabled
 			[]*weightedValidator{
 				{address: common.StringToAddress("101")}, {address: common.StringToAddress("102")}, {address: common.StringToAddress("103")},
 			},
 			&reward.StakingInfo{
 				CouncilNodeAddrs: []common.Address{common.StringToAddress("101"), common.StringToAddress("102"), common.StringToAddress("103")},
 				UseGini:          false,
-				Gini:             reward.DefaultGiniCoefficient,
+				Gini:             reward.DefaultGiniCoefficient, // To see if calcTotalAmount modifies Gini field, set it to -1.
 			},
 			[]float64{5000000, 5000000, 5000000},
 			reward.DefaultGiniCoefficient,
 			15000000,
 			[]float64{5000000, 5000000, 5000000},
 		},
-		{
+		{ // Gini enabled but equal amounts.
 			[]*weightedValidator{
 				{address: common.StringToAddress("101")}, {address: common.StringToAddress("102")}, {address: common.StringToAddress("103")},
 			},
@@ -150,8 +150,7 @@ func TestCalcTotalAmount(t *testing.T) {
 			15000000,
 			[]float64{5000000, 5000000, 5000000},
 		},
-
-		{
+		{ // Gini enabled and unequal amounts.
 			[]*weightedValidator{
 				{address: common.StringToAddress("101")}, {address: common.StringToAddress("102")}, {address: common.StringToAddress("103")}, {address: common.StringToAddress("104")}, {address: common.StringToAddress("105")},
 			},
@@ -168,11 +167,10 @@ func TestCalcTotalAmount(t *testing.T) {
 	}
 	for _, testCase := range testCases {
 		stakingAmounts := testCase.stakingAmounts
-		totalAmount := calcTotalAmount(testCase.weightedValidators, testCase.stakingInfo, stakingAmounts)
+		totalAmount, giniUsed := calcTotalAmount(testCase.weightedValidators, testCase.stakingInfo, stakingAmounts)
 
-		// Since calcTotalAmount does not update stakingInfo.Gini, we cannot assert expectedGini.
-		// However asserting expectedStakingAmounts proves that calcTotalAmount() used correct Gini.
-		// assert.Equal(t, testCase.expectedGini, testCase.stakingInfo.Gini)
+		assert.Equal(t, testCase.expectedGini, giniUsed)                          // calcTotalAmount computes gini among validators on its own
+		assert.Equal(t, reward.DefaultGiniCoefficient, testCase.stakingInfo.Gini) // stakingInfo.Gini left untouched
 		assert.Equal(t, testCase.expectedTotalAmount, totalAmount)
 		assert.Equal(t, testCase.expectedStakingAmounts, stakingAmounts)
 	}
@@ -312,7 +310,7 @@ func TestWeightedCouncil_validatorWeightWithStakingInfo(t *testing.T) {
 		candidates := append(council.validators, council.demotedValidators...)
 		weightedValidators, stakingAmounts, err := getStakingAmountsOfValidators(candidates, testCase.stakingInfo)
 		assert.NoError(t, err)
-		totalStaking := calcTotalAmount(weightedValidators, testCase.stakingInfo, stakingAmounts)
+		totalStaking, _ := calcTotalAmount(weightedValidators, testCase.stakingInfo, stakingAmounts)
 		calcWeight(weightedValidators, stakingAmounts, totalStaking)
 
 		for i, weight := range testCase.expectedWeights {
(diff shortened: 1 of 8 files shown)


consensus/istanbul/validator/weighted.go
```diff
@@ -801,10 +806,10 @@ func calcTotalAmount(weightedValidators []*weightedValidator, stakingInfo *rewar
 				tempStakingAmounts = append(tempStakingAmounts, stakingAmounts[vIdx])
 			}
 		}
-		stakingInfo.Gini = reward.CalcGiniCoefficient(tempStakingAmounts)
+		gini := reward.CalcGiniCoefficient(tempStakingAmounts)
```
jeongkyun-oh, line 809: Is local variable intended?

consensus/istanbul/validator/weighted.go
```diff
@@ -813,7 +818,7 @@ func calcTotalAmount(weightedValidators []*weightedValidator, stakingInfo *rewar
 		}
 	}
 
-	logger.Debug("calculate totalStaking", "UseGini", stakingInfo.UseGini, "Gini", stakingInfo.Gini, "totalStaking", totalStaking, "stakingAmounts", stakingAmounts)
+	logger.Debug("calculate totalStaking", "UseGini", stakingInfo.UseGini, "Gini", gini, "totalStaking", totalStaking, "stakingAmounts", stakingAmounts)
```
jeongkyun-oh, line 821: Doesn't this `gini` always print `-1`?

reward/reward_config_cache_test.go
```diff
@@ -92,6 +92,10 @@ func (governance *testGovernance) StakingUpdateInterval() uint64 {
 	return governance.stakingInterval
 }
 
+func (governance *testGovernance) GetMinimumStakingAtNumber(num uint64) (uint64, error) {
+	return 5000000, nil
```
jeongkyun-oh, line 96: It is minor, but how about using `DefaultMinimumStake` instead of magic number?

consensus/istanbul/validator/weighted.go
blukat29: Fixed in https://github.com/klaytn/klaytn/commit/4992fab050621b1ff3261c4d810e291b521c83c7.

reward/reward_config_cache_test.go
blukat29: Fixed in c12e0861.

consensus/istanbul/validator/multi_staking_test.go
```diff
@@ -170,7 +170,9 @@ func TestCalcTotalAmount(t *testing.T) {
 		stakingAmounts := testCase.stakingAmounts
 		totalAmount := calcTotalAmount(testCase.weightedValidators, testCase.stakingInfo, stakingAmounts)
 
-		assert.Equal(t, testCase.expectedGini, testCase.stakingInfo.Gini)
+		// Since calcTotalAmount does not update stakingInfo.Gini, we cannot assert expectedGini.
+		// However asserting expectedStakingAmounts proves that calcTotalAmount() used correct Gini.
+		// assert.Equal(t, testCase.expectedGini, testCase.stakingInfo.Gini)
```
jeongkyun-oh, line 175: I didn't get why this testcase is failed. Shouldn't this pass?

reward/staking_info.go
```diff
@@ -174,6 +200,77 @@ func (s *StakingInfo) DecodeRLP(st *rlp.Stream) error {
...
+func (c *ConsolidatedStakingInfo) GetAllNodes() []consolidatedNode {
+	return c.nodes
+}
+
+func (c *ConsolidatedStakingInfo) GetConsolidatedNode(nodeAddr common.Address) *consolidatedNode {
+	if idx, ok := c.nodeIndex[nodeAddr]; ok {
+		return &c.nodes[idx]
+	} else {
+		return nil
+	}
```
jeongkyun-oh, line 246: It is minor, though.

```suggestion
	if idx, ok := c.nodeIndex[nodeAddr]; ok {
		return &c.nodes[idx]
	}
	return nil
```

consensus/istanbul/validator/multi_staking_test.go
blukat29: Before this PR, this test expects `calcTotalAmount()` to fill in the `stakingInfo.Gini`.
After this PR, `calcTotalAmount()` should not change the `stakingInfo.Gini`.

The tests actually fail because `testCase.expectedGini = 0.27` but `testCase.stakingInfo.Gini = -1`.

reward/staking_info.go
```diff
@@ -174,6 +200,77 @@ func (s *StakingInfo) DecodeRLP(st *rlp.Stream) error {
...
+}
+
+// Calculate Gini coefficient of the StakingAmounts.
+// Only amounts greater or equal to `minStake` are included in the calculation.
+// Set `minStake` to 0 to calculate Gini coefficient of all amounts.
+func (c *ConsolidatedStakingInfo) CalcGiniCoefficient(minStake uint64) float64 {
```
jeongkyun-oh, line 252: We have the cases where all nodes doesn't have enough amount of staking. Do you think we should take consider the case?

consensus/istanbul/validator/multi_staking_test.go
blukat29: @jeongkyun-oh To make it more clear, I will modify `calcTotalAmount` to return the `gini` (`giniUsed` might be more appropriate naming) as well, and then compare with `testCase.expectedGini`.

```go
totalAmount, giniUsed := calcTotalAmount(...)
assert.Equal(t, testCase.expectedGini, giniUsed) // calcTotalAmount computes gini on its own among validators
assert.Equal(t, params. DefaultGiniCoefficient, testCase.stakingInfo.Gini) // stakingInfo.Gini left untouched
```

jeongkyun-oh (dismissed):
LGTM except for minors.

consensus/istanbul/validator/multi_staking_test.go
jeongkyun-oh: @blukat29 Wouldn't it `calcTotalAmount` change the normal behaviour to calculate weight of validators?

reward/staking_info.go
jiseongnoh, line 252: The function name `CalcGiniCoefficient` is the same as the function name below.

reward/staking_info.go
blukat29: Fixed in fc5dba2e.

consensus/istanbul/validator/multi_staking_test.go
blukat29: @jeongkyun-oh Does changing `calcTotalAmount()` affect the behavior of `calcWeight()`? No because `calcWeight()` does not depend on `stakingInfo`.

As long as `calcTotalAmount()` returns correct `stakingAmounts` and `totalStaking`, `calcWeight()` should keep working correctly.

jeongkyun-oh: Oh, never mind. Applying your suggestion is good enough. Thanks.

reward/staking_info.go
jeongkyun-oh: Currently it returns `-1`, right? Both makes sense to me, but if `NaN` makes inefficient then I think the original change seems better to me. I resolve this comment, but you may be able to update this.

jeongkyun-oh (approved):

reward/staking_manager.go
```diff
@@ -161,6 +177,11 @@ func updateStakingInfo(blockNum uint64) (*StakingInfo, error) {
 		return nil, err
 	}
 
+	// Fill in Gini coeff before adding to cache and DB.
+	if err := fillMissingGiniCoefficient(stakingInfo, blockNum); err != nil {
+		logger.Warn("Cannot fill in gini coefficient", "blockNum", blockNum, "err", err)
+	}
+
```
ian0371, line 184: This could update existing DB's stakingInfo due to the following `AddStakingInfoToDB()`, resulting in non-consistency among multiple nodes. Is this ok?

blukat29: 1. Before this PR, DB always stores `{Gini: -1}`.
2. After this PR, DB sometimes stores `{Gini: 0.xx}`.
    ```
    //   If cache hit     -> fillMissingGini -> modifies cached in-memory object
    //   If db hit        -> fillMissingGini -> write to cache
    //   If read contract -> fillMissingGini -> write to db -> write to cache
    ```
    In this case, DB content may differ among nodes depending on when they have upgraded to v1.9.0.

How about I change the fixup logic like below @ian0371?

```diff
  //   If cache hit     -> fillMissingGini -> modifies cached in-memory object
  //   If db hit        -> fillMissingGini -> write to cache
- //   If read contract -> fillMissingGini -> write to db -> write to cache
+ //   If read contract -> write to db (-1) -> fillMissingGini -> write to cache (0.xx)
```

ian0371: LGTM :)

reward/staking_info.go
blukat29: - I meant to say `NaN` makes it more efficient. (by distinguishing "never calculated" and "calculated but empty array" in fillMissingGiniCoefficient).
- But after some research I found that having `NaN` in JSON is problematic in [Java](https://stackoverflow.com/questions/15228651/how-to-parse-json-string-containing-nan-in-node-js). So let's not use `NaN` anyway.

reward/staking_manager.go
blukat29: Fixed in 4032a013.

reward/staking_info.go
blukat29: Fixed in [3a45fdf](https://github.com/klaytn/klaytn/pull/1464/commits/3a45fdf6d0badef06f06466355b8520f54457cc5).

ian0371 (approved):

jiseongnoh (approved):
Looks good to me

aidan-kwon (approved):

consensus/istanbul/validator/weighted.go
```diff
@@ -665,7 +665,7 @@ func (valSet *weightedCouncil) Refresh(hash common.Hash, blockNum uint64, config
 		return nil
 	}
 
-	totalStaking := calcTotalAmount(weightedValidators, newStakingInfo, stakingAmounts)
+	totalStaking, _ := calcTotalAmount(weightedValidators, newStakingInfo, stakingAmounts)
```
yoomee1313, line 668: This is just a minor question.. the returned `gini` is not used. Can I ask the reason where the returned `gini` is used?

yoomee1313 (approved):

jeongkyun-oh (approved):

consensus/istanbul/validator/weighted.go
blukat29: It is used for teting only. See https://github.com/klaytn/klaytn/pull/1464#discussion_r913382015.

PR #1465, opened 2022-06-24
add checking z sign in affineFromJacobian

bobpkr:
## Proposed changes

When the value of `ModInverse(z, BitCurve.P)` which is `zinv`  is 0, Regardless of the arguments ​​of `x` and `y`, the return values ​​of `xOut` and `yOut` are both 0. Therefore, when `zinv == nil`, there is no need to reset `zinv = 0` and calculate any more, just return `0, 0`. When the argument `z` is 0. `zinv`, which is the value of `ModInverse(z, BitCurve.P)` is `nil`.

When the argument `z` is 0,  just return  `0, 0`. [go-ethereum](https://github.com/ethereum/go-ethereum/blob/master/crypto/secp256k1/curve.go#L108-L115) and [golang](https://github.com/golang/go/blob/master/src/crypto/elliptic/params.go#L76-L82) are already doing that.

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/crypto/secp256k1/curve.go
+++ b/crypto/secp256k1/curve.go
@@ -139,7 +139,18 @@ func (BitCurve *BitCurve) affineFromJacobian(x, y, z *big.Int) (xOut, yOut *big.
 
 // Add returns the sum of (x1,y1) and (x2,y2)
 func (BitCurve *BitCurve) Add(x1, y1, x2, y2 *big.Int) (*big.Int, *big.Int) {
+	// If one point is at infinity, return the other point.
+	// Adding the point at infinity to any point will preserve the other point.
+	if x1.Sign() == 0 && y1.Sign() == 0 {
+		return x2, y2
+	}
+	if x2.Sign() == 0 && y2.Sign() == 0 {
+		return x1, y1
+	}
 	z := new(big.Int).SetInt64(1)
+	if x1.Cmp(x2) == 0 && y1.Cmp(y2) == 0 {
+		return BitCurve.affineFromJacobian(BitCurve.doubleJacobian(x1, y1, z))
+	}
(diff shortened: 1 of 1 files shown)


blukat29:
Good spot @bobpkr!

This PR made me to run a quick diff against go-ethereum, in which I've found there are additional checks in `Add()`. It is related to `affineFromJacobian()`. Can we fix it as well?

```sh
git diff klaytn/crypto/secp256k1/ go-ethereum/crypto/secp256k1/
```

```diff
 // Add returns the sum of (x1,y1) and (x2,y2)
 func (BitCurve *BitCurve) Add(x1, y1, x2, y2 *big.Int) (*big.Int, *big.Int) {
+       // If one point is at infinity, return the other point.
+       // Adding the point at infinity to any point will preserve the other point.
+       if x1.Sign() == 0 && y1.Sign() == 0 {
+               return x2, y2
+       }
+       if x2.Sign() == 0 && y2.Sign() == 0 {
+               return x1, y1
+       }
        z := new(big.Int).SetInt64(1)
+       if x1.Cmp(x2) == 0 && y1.Cmp(y2) == 0 {
+               return BitCurve.affineFromJacobian(BitCurve.doubleJacobian(x1, y1, z))
+       }
        return BitCurve.affineFromJacobian(BitCurve.addJacobian(x1, y1, z, x2, y2, z))
 }
 ```

blukat29 (approved):
Changes in `affineFromJacobian` looks good.

bobpkr:
@blukat29 Good! The point at infinity of elliptic curve is 0(zero) which is identity element. When one of them is 0, adding two points will return the other one. When the two points are same, we can call a function of double a point.

blukat29 (approved):

JayChoi1736 (approved):

PR #1473, opened 2022-06-30
implemented snapshot synchronization

jeongkyun-oh:
## Proposed changes

snapsync is a faster synchronisation feature derived from Ethereum. This PR ported the snapsync related PR from Ethereum. 

### Synchronisation methods from Ethereum
1. `full` sync: fetch headers and bodies (transactions) only, but reexecute the downloaded transactions to output transaction receipts and world states. Klaytn supports only full sync for now.
2. `fast` sync: fetch headers, bodies, and receipts until pivot block (no trie verification), and download all the trie nodes on the pivot.
3. `snap` sync: fetch headers, bodies, and receipts until pivot block and download snapshot data (trie leaf nodes) on the pivot after which the entire trie is reconstructed.

We had to fetch staking information as well to verify headers.

### Snap sync components
1. *snap protocol*: a message protocol to send the snapshot data including accounts, storages, bytecodes as well as trie nodes data. Basically, it sends all the data which is needed to reconstruct a state trie on a specific pivot block. Snap protocol is independent to the original `klay`/`istanbul` protocol.
2. *snap syncer*: snap syncer schedules the snapshot data fetching and healing the reconstructed trie. The healing phase is explained later.

### Snap sync process
1. Assuming that the remote peers have the snapshot data which is qualified to serve snap protocol.
2. Fetch a pivot block first. Pivot block is a specific block where the state trie is reconstructed. Pivot block is dynamically moved forward by 64 blocks. This is because a remote peer hold snapshot data about recent 128 blocks. 
3. Snap sync is started. Snap syncer scheduling the data retrieval. At the beginning, It divides hash space (`0x000...000` ... `0xfff...fff`) to 16 spaces, and it requests the range to retrieve the snapshot account data whose key is within that range. So, max 16 snap peers can be served to send those data.
4. The remote peers iterates the snapshot data in the increasing order of the keys, and pack the data to send back to the requested peer. At this moment, the merkle proof is added to the packet. 
5. After receiving the account data, it stores the data to database. If the account is smart contract, schedule code and storage retrieval as well. If all related data is downloaded, it processes the data such that creating partial state trie and persisting the data (range of accounts, and partial trie nodes) to database.
6. The syncer synchronises the snapshot data based on a state root hash. However, the root would be changed every time the pivot moves. So, the downloaded snapshot data might be out-dated. We considered this data as correct even though it is not and will be fixed in heal stage.
7. After all account data is downloaded, the state trie has to be healed since the out-dated accounts were inserted to the trie. In this stage, heal task is scheduled. It first fetch root and then children recursively. If it meets a subtrie root, then it doesn't go into deeper trie. Likewise it can heal the entire trie from root node.
8. After reconstructing the state trie, the sync mode is changed to full, and it will creates block as we know.

### TODO
- log enhancement
- add metrics
- stack trie update

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -123,7 +123,7 @@ var (
 	defaultSyncMode = cn.GetDefaultConfig().SyncMode
 	SyncModeFlag    = TextMarshalerFlag{
 		Name:  "syncmode",
-		Usage: `Blockchain sync mode (only "full" is supported)`,
+		Usage: `Blockchain sync mode ("full" or "snap")`,
 		Value: &defaultSyncMode,
 	}
 	GCModeFlag = cli.StringFlag{
@@ -1586,8 +1586,11 @@ func SetKlayConfig(ctx *cli.Context, stack *node.Node, cfg *cn.Config) {
 
 	if ctx.GlobalIsSet(SyncModeFlag.Name) {
 		cfg.SyncMode = *GlobalTextMarshaler(ctx, SyncModeFlag.Name).(*downloader.SyncMode)
-		if cfg.SyncMode != downloader.FullSync {
-			log.Fatalf("only syncmode=full can be used for syncmode!")
+		if cfg.SyncMode == downloader.SnapSync {
+			logger.Info("Snap sync requested, enabling --snapshot")
+			ctx.Set(SnapshotFlag.Name, "true")
+		} else {
+			cfg.SnapshotCacheSize = 0 // Disabled
 		}
 	}
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1586,6 +1586,9 @@ func SetKlayConfig(ctx *cli.Context, stack *node.Node, cfg *cn.Config) {
 
 	if ctx.GlobalIsSet(SyncModeFlag.Name) {
 		cfg.SyncMode = *GlobalTextMarshaler(ctx, SyncModeFlag.Name).(*downloader.SyncMode)
+		if cfg.SyncMode != downloader.FullSync && cfg.SyncMode != downloader.SnapSync {
+			log.Fatalf("Full Sync or Snap Sync (prototype) is supported only!")
+		}
 		if cfg.SyncMode == downloader.SnapSync {
 			logger.Info("Snap sync requested, enabling --snapshot")
 			ctx.Set(SnapshotFlag.Name, "true")
--- a/datasync/downloader/downloader.go
+++ b/datasync/downloader/downloader.go
@@ -1455,6 +1455,7 @@ func (d *Downloader) processHeaders(origin uint64, td *big.Int) error {
 				// Terminate if something failed in between processing chunks
 				select {
 				case <-d.cancelCh:
+					rollbackErr = errCanceled
 					return errCanceled
 				default:
 				}
--- a/datasync/downloader/downloader.go
+++ b/datasync/downloader/downloader.go
@@ -405,8 +405,8 @@ func (d *Downloader) synchronise(id string, hash common.Hash, td *big.Int, mode 
 			// sync completely heals and finishes. Pause snapshot maintenance in the mean
 			// time to prevent access.
 			if snapshots := d.blockchain.Snapshots(); snapshots != nil { // Only nil in tests
-				logger.Warn("State snapshot is disabled")
 				snapshots.Disable()
+				logger.Warn("State snapshot is disabled")
 			}
 			logger.Warn("Enabling snapshot sync prototype")
 			d.snapSync = true
--- a/datasync/downloader/downloader.go
+++ b/datasync/downloader/downloader.go
@@ -33,8 +33,10 @@ import (
 	"github.com/klaytn/klaytn/common"
 	"github.com/klaytn/klaytn/event"
 	"github.com/klaytn/klaytn/log"
+	"github.com/klaytn/klaytn/node/cn/snap"
 	"github.com/klaytn/klaytn/params"
 	"github.com/klaytn/klaytn/reward"
+	"github.com/klaytn/klaytn/snapshot"
 	"github.com/klaytn/klaytn/storage/database"
 	"github.com/klaytn/klaytn/storage/statedb"
 	"github.com/rcrowley/go-metrics"
@@ -135,6 +137,10 @@ type Downloader struct {
 	stakingInfoWakeCh chan bool            // [klay/65] Channel to signal the staking info fetcher of new tasks
 	headerProcCh      chan []*types.Header // [klay/62] Channel to feed the header processor new tasks
 
+	// for snapsyncer
+	snapSync   bool         // Whether to run state sync over the snap protocol
+	SnapSyncer *snap.Syncer // TODO-Klaytn-Snapsyncer make private! hack for now
+
 	// for stateFetcher
 	pivotHeader *types.Header
 	pivotLock   sync.RWMutex
@@ -205,6 +211,9 @@ type BlockChain interface {
 
 	// InsertReceiptChain inserts a batch of receipts into the local chain.
 	InsertReceiptChain(types.Blocks, []types.Receipts) (int, error)
+
+	// Snapshots returns the blockchain snapshot tree.
+	Snapshots() *snapshot.Tree
 }
 
 // New creates a new downloader to fetch hashes and blocks from remote peers.
@@ -235,6 +244,7 @@ func New(mode SyncMode, stateDB database.DBManager, stateBloom *statedb.SyncBloo
 		headerProcCh:      make(chan []*types.Header, 1),
 		quitCh:            make(chan struct{}),
 		stateCh:           make(chan dataPack),
+		SnapSyncer:        snap.NewSyncer(stateDB),
 		stateSyncStart:    make(chan *stateSync),
 		syncStatsState: stateSyncStats{
 			processed: stateDB.ReadFastTrieProgress(),
@@ -263,6 +273,8 @@ func (d *Downloader) Progress() klaytn.SyncProgress {
 	switch mode {
 	case FullSync:
 		current = d.blockchain.CurrentBlock().NumberU64()
+	case SnapSync:
+		current = d.blockchain.CurrentFastBlock().NumberU64()
 	case FastSync:
 		current = d.blockchain.CurrentFastBlock().NumberU64()
 	case LightSync:
@@ -319,6 +331,10 @@ func (d *Downloader) UnregisterPeer(id string) error {
 	d.queue.Revoke(id)
 
 	return nil
+}
+
+func (d *Downloader) GetSnapSyncer() *snap.Syncer {
+	return d.SnapSyncer
 }
 
 // Synchronise tries to sync up our local block chain with a remote peer, both
@@ -381,6 +397,20 @@ func (d *Downloader) synchronise(id string, hash common.Hash, td *big.Int, mode 
 	// when the user attempts to fast sync a new empty network.
 	if mode == FullSync && d.stateBloom != nil {
 		d.stateBloom.Close()
+	}
+
+	if mode == SnapSync {
+		if !d.snapSync {
+			// Snap sync uses the snapshot namespace to store potentially flakey data until
+			// sync completely heals and finishes. Pause snapshot maintenance in the mean
+			// time to prevent access.
+			if snapshots := d.blockchain.Snapshots(); snapshots != nil { // Only nil in tests
+				logger.Warn("State snapshot is disabled")
+				snapshots.Disable()
+			}
+			logger.Warn("Enabling snapshot sync prototype")
+			d.snapSync = true
+		}
 	}
 	// Reset the queue, peer set and wake channels to clean any internal leftover state
 	d.queue.Reset(blockCacheMaxItems, blockCacheInitialItems)
@@ -454,7 +484,7 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td *big.I
 	if err != nil {
 		return err
 	}
-	if mode == FastSync && pivot == nil {
+	if (mode == FastSync || mode == SnapSync) && pivot == nil {
 		// If no pivot block was returned, the head is below the min full block
 		// threshold (i.e. new chain). In that case we won't really fast sync
 		// anyway, but still need a valid pivot block to avoid some code hitting
@@ -475,7 +505,7 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td *big.I
 	d.syncStatsLock.Unlock()
 
 	// Ensure our origin point is below any fast sync pivot point
-	if mode == FastSync {
+	if mode == FastSync || mode == SnapSync {
 		if height <= uint64(fsMinFullBlocks) {
 			origin = 0
 		} else {
@@ -486,7 +516,7 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td *big.I
 		}
 	}
 	d.committed = 1
-	if mode == FastSync && pivot.Number.Uint64() != 0 {
+	if (mode == FastSync || mode == SnapSync) && pivot.Number.Uint64() != 0 {
 		d.committed = 0
 	}
 	// Initiate the sync using a concurrent header and content retrieval algorithm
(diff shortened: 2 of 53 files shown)


cmd/utils/flags.go
```diff
@@ -123,7 +123,7 @@ var (
 	defaultSyncMode = cn.GetDefaultConfig().SyncMode
 	SyncModeFlag    = TextMarshalerFlag{
 		Name:  "syncmode",
-		Usage: `Blockchain sync mode (only "full" is supported)`,
+		Usage: `Blockchain sync mode ("full" or "snap")`,
 		Value: &defaultSyncMode,
```
kjhman21, line 127: Don't we need to change default sync mode?

networks/p2p/tracker/tracker.go
```diff
@@ -0,0 +1,215 @@
...
+	if req.expire.Prev() == nil {
+		if t.wake.Stop() {
+			t.schedule()
+		}
+	}
+	g := fmt.Sprintf("%s/%s/%d/%#02x", trackedGaugeName, t.protocol, req.version, req.reqCode)
+	gauge := metrics.GetOrRegisterGauge(g, nil)
+	gauge.Update(gauge.Value() - 1)
+
+	//h := fmt.Sprintf("%s/%s/%d/%#02x", waitHistName, t.protocol, req.version, req.reqCode)
```
kjhman21, line 208: Why these are commented?

node/cn/peer_set.go
```diff
@@ -469,3 +510,67 @@ func (peers *peerSet) UpdateTypePeersWithoutTxs(tx *types.Transaction, nodeType
...
+// WaitSnapExtension blocks until all satellite protocols are connected and tracked
+// by the peerset.
+func (ps *peerSet) WaitSnapExtension(peer Peer) (*snap.Peer, error) {
+	// If the peer does not support a compatible `snap`, don't wait
+	if !peer.RunningCap(snap.ProtocolName, snap.ProtocolVersions) {
+		return nil, nil
+	}
+	// Ensure nobody can double connect
+	ps.lock.Lock()
+
```
kjhman21, line 553: It would be good to make an internal function and use `defer` for better code maintenance.

node/cn/handler.go
```diff
@@ -429,17 +498,34 @@ func (pm *ProtocolManager) handle(p Peer) error {
...
 		p.GetP2PPeer().Log().Debug("Klaytn peer handshake failed", "err", err)
 		return err
 	}
+	reject := false
+	if atomic.LoadUint32(&pm.snapSync) == 1 {
+		if snap == nil {
+			// If we are running snap-sync, we want to reserve roughly half the peer
+			// slots for peers supporting the snap protocol.
+			// The logic here is; we only allow up to 5 more non-snap peers than snap-peers.
+			if all, snp := pm.peers.Len(), pm.peers.SnapLen(); all-snp > snp+5 {
```
kjhman21, line 511: It would be good to have a variable instead of using hard-coded value 5.

node/cn/snap/handler.go
```diff
@@ -0,0 +1,524 @@
...
+		if err := msg.Decode(&req); err != nil {
+			return fmt.Errorf("%w: message %v: %v", errDecode, msg, err)
+		}
+		// Service the request, potentially returning nothing in case of errors
+		accounts, proofs := ServiceGetAccountRangeQuery(reader, &req)
+
+		// Send back anything accumulated (or empty in case of errors)
+		return p2p.Send(peer.rw, AccountRangeMsg, &AccountRangePacket{
+			ID:       req.ID,
+			Accounts: accounts,
```
kjhman21, line 118: `accounts` can be nil. Is it okay?

kjhman21: Is there any test case to cover this case?

kjhman21:
@jeongkyun-oh Is it possible to turn off this feature? For example, CNs do not want to provide snapshot request to avoid performance degradation.

node/cn/snap/handler.go
```diff
@@ -0,0 +1,524 @@
...
+		return nil, nil
+	}
+	// Iterate over the requested range and pile accounts up
+	var (
+		accounts []*AccountData
+		size     uint64
+		last     common.Hash
+	)
+	for it.Next() {
+		hash, account := it.Hash(), common.CopyBytes(it.Account())
```
kjhman21, line 251: Does this pollute cache? It seems like it reads only once to respond the request. In this case we don't need to update cache.

node/cn/snap/handler.go
```diff
@@ -0,0 +1,524 @@
...
+		size += uint64(common.HashLength + len(account))
+		accounts = append(accounts, &AccountData{
+			Hash: hash,
+			Body: account,
+		})
+		// If we've exceeded the request threshold, abort
+		if bytes.Compare(hash[:], req.Limit[:]) >= 0 {
+			break
+		}
+		if size > req.Bytes {
```
kjhman21, line 266: In this case size can be larger than req.Bytes because anyway we append the account data before comparison.

node/cn/snap/handler.go
```diff
@@ -0,0 +1,524 @@
...
+		logger.Warn("Failed to prove account range", "origin", req.Origin, "err", err)
+		return nil, nil
+	}
+	if last != (common.Hash{}) {
+		if err := tr.Prove(last[:], 0, proof); err != nil {
+			logger.Warn("Failed to prove account range", "last", last, "err", err)
+			return nil, nil
+		}
+	}
+	proofs := make([][]byte, len(proof.NodeList()))
```
kjhman21, line 284:
```suggestion
	proofs := make([][]byte, len(proof.NodeList()), 0)
```

If we omit the last parameter, len(proofs) == len(proof.NodeList()). Is this your intention?

kjhman21: https://go.dev/play/p/xOowbGEYSHy

node/cn/snap/handler.go
```diff
@@ -0,0 +1,524 @@
...
+				return nil, nil
+			}
+			serializer := account.NewAccountSerializer()
+			if err := rlp.DecodeBytes(accTrie.Get(accountHash[:]), serializer); err != nil {
+				return nil, nil
+			}
+			acc := serializer.GetAccount()
+			pacc := account.GetProgramAccount(acc)
+			if pacc == nil {
+				return nil, nil
```
kjhman21, line 375: Do we need to return? I think we can skip this account if it is not a program account.

node/cn/snap/nodeset.go
```diff
@@ -0,0 +1,174 @@
...
+
+// NewNodeSet creates an empty node set
+func NewNodeSet() *NodeSet {
+	return &NodeSet{
+		nodes: make(map[string][]byte),
+	}
+}
+
+func (db *NodeSet) ReadCachedTrieNode(hash common.Hash) ([]byte, error) {
+	return db.Get(hash.Bytes())
```
kjhman21, line 51: Don't we need to first search db.nodes?

node/cn/snap/nodeset.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Put stores a new node in the set
+func (db *NodeSet) Put(key []byte, value []byte) error {
+	db.lock.Lock()
+	defer db.lock.Unlock()
+
+	if _, ok := db.nodes[string(key)]; ok {
+		return nil
+	}
+	keystr := string(key)
```
kjhman21, line 66:
```suggestion
	keystr := string(key)
	if _, ok := db.nodes[keystr]; ok {
		return nil
	}
```

node/cn/snap/protocol.go
```diff
@@ -0,0 +1,241 @@
...
+	Proof    [][]byte       // List of trie nodes proving the account range
+}
+
+// AccountData represents a single account in a query response.
+type AccountData struct {
+	Hash common.Hash  // Hash of the account
+	Body rlp.RawValue // Account body in slim format
+}
+
+type accountData struct {
```
kjhman21, line 97: Do we need really to have `accountData`? It seems like if we AccountData.Body is a type of `[]byte`, then we don't need it.

node/cn/snap/protocol.go
```diff
@@ -0,0 +1,241 @@
...
+// to not double the memory use.
+func (p *AccountRangePacket) Unpack() ([]common.Hash, [][]byte, error) {
+	var (
+		hashes   = make([]common.Hash, len(p.Accounts))
+		accounts = make([][]byte, len(p.Accounts))
+	)
+	for i, acc := range p.Accounts {
+		hashes[i], accounts[i] = acc.Hash, acc.Body
+	}
+	return hashes, accounts, nil
```
kjhman21, line 136: Is there any case that we return error? In the sequence of this function, there is no case to return an error.

node/cn/snap/sync.go
```diff
@@ -0,0 +1,2884 @@
...
+
+	// RequestTrieNodes fetches a batch of account or storage trie nodes rooted in
+	// a specificstate trie.
+	RequestTrieNodes(id uint64, root common.Hash, paths []TrieNodePathSet, bytes uint64) error
+
+	// Log retrieves the peer's own contextual logger.
+	Log() log.Logger
+}
+
+// Syncer is an Ethereum account and storage trie syncer based on snapshots and
```
kjhman21, line 388:
```suggestion
// Syncer is an Klaytn account and storage trie syncer based on snapshots and
```

node/cn/snap/sync.go
```diff
@@ -0,0 +1,2884 @@
...
+	targetTTL := s.rates.TargetTimeout()
+	for id := range s.accountIdlers {
+		if _, ok := s.statelessPeers[id]; ok {
+			continue
+		}
+		idlers.ids = append(idlers.ids, id)
+		idlers.caps = append(idlers.caps, s.rates.Capacity(id, AccountRangeMsg, targetTTL))
+	}
+	if len(idlers.ids) == 0 {
+		return
```
kjhman21, line 904: Don't we need to print a log when there is no available peer to fetch accounts in snapshot?

networks/p2p/tracker/tracker.go
jeongkyun-oh: The metrics wasn't updated yet, so I added TODO comment for this.

cmd/utils/flags.go
jeongkyun-oh: Let's update this to default after stabilization.

node/cn/peer_set.go
jeongkyun-oh: I have refactored this on 154e3e65c76c7d0b50f7c35b431768505ceb6104. Please take another look.

node/cn/snap/handler.go
jeongkyun-oh: My bad. I have updated this on c4d46e4f3d4b2ed6613a793974f222e244ae2489.

node/cn/snap/nodeset.go
jeongkyun-oh: Please take another look. Related change: c4d46e4f3d4b2ed6613a793974f222e244ae2489

node/cn/snap/protocol.go
jeongkyun-oh: Please take another look. Related change: c4d46e4f3d4b2ed6613a793974f222e244ae2489

node/cn/snap/protocol.go
jeongkyun-oh: Please take another look. Related change: c4d46e4f3d4b2ed6613a793974f222e244ae2489

node/cn/snap/sync.go
jeongkyun-oh: Please take another look. Related change: c4d46e4f3d4b2ed6613a793974f222e244ae2489

node/cn/snap/handler.go
jeongkyun-oh: This may be considered later for optimization.

node/cn/snap/handler.go
jeongkyun-oh: As we talked let's keep this for now, and it can be enhanced later if necessary.

node/cn/snap/handler.go
kjhman21: @jeongkyun-oh Could you please add a comment around the code so that later we can check it again?

node/cn/snap/handler.go
kjhman21: @jeongkyun-oh The commit you referred does not include the change.

node/cn/snap/sync.go
kjhman21: got it. Thanks.

node/cn/snap/handler.go
jeongkyun-oh: Oh, I already updated on 34aa99c1d0d90695a9f316f96447113175481825.
`// TODO-Klaytn-SnapSync investigate the cache pollution`

node/cn/snap/handler.go
kjhman21: https://github.com/klaytn/klaytn/pull/1473/commits/15770956e225f0d9c3c5a5ea2753b2b720011677#diff-99aff59e7985f48e09f17fbf0adbba2fb8a1d83b2c057142d41c76313b642bf5R285

Anyway I found that it was fixed.

cmd/utils/flags.go
```diff
@@ -1586,8 +1586,11 @@ func SetKlayConfig(ctx *cli.Context, stack *node.Node, cfg *cn.Config) {
 
 	if ctx.GlobalIsSet(SyncModeFlag.Name) {
 		cfg.SyncMode = *GlobalTextMarshaler(ctx, SyncModeFlag.Name).(*downloader.SyncMode)
-		if cfg.SyncMode != downloader.FullSync {
-			log.Fatalf("only syncmode=full can be used for syncmode!")
+		if cfg.SyncMode == downloader.SnapSync {
+			logger.Info("Snap sync requested, enabling --snapshot")
+			ctx.Set(SnapshotFlag.Name, "true")
+		} else {
+			cfg.SnapshotCacheSize = 0 // Disabled
```
aidan-kwon, line 1593: In the previous code, the process is terminated if "syncmode" is an unexpected value. However, this code doesn't terminate the process anymore.

node/cn/snap/sync.go
```diff
@@ -0,0 +1,2885 @@
...
+
+	needCode  []bool // Flags whether the filling accounts need code retrieval
+	needState []bool // Flags whether the filling accounts need storage retrieval
+	needHeal  []bool // Flags whether the filling accounts's state was chunked and need healing
+
+	codeTasks  map[common.Hash]struct{}    // Code hashes that need retrieval
+	stateTasks map[common.Hash]common.Hash // Account hashes->roots that need full state retrieval
+
+	// TODO-Klaytn-Snapsync consider to use stack trie
+	genTrie *statedb.Trie // Node generator from storage slots
```
aidan-kwon, line 303: When we adopt stacktrie, the new client will not be compatible with the previous snap data, right?

datasync/downloader/downloader.go
```diff
@@ -382,6 +398,20 @@ func (d *Downloader) synchronise(id string, hash common.Hash, td *big.Int, mode
...
 	}
+
+	if mode == SnapSync {
+		if !d.snapSync {
+			// Snap sync uses the snapshot namespace to store potentially flakey data until
+			// sync completely heals and finishes. Pause snapshot maintenance in the mean
+			// time to prevent access.
+			if snapshots := d.blockchain.Snapshots(); snapshots != nil { // Only nil in tests
+				logger.Warn("State snapshot is disabled")
+				snapshots.Disable()
```
aidan-kwon, line 409:
```suggestion
				snapshots.Disable()
				logger.Warn("State snapshot is disabled")
```

node/cn/snap/handler.go
jeongkyun-oh: I have added testcases, but encoding/decoding `nil` accounts were okay. Am I missing something? Please take another look.

datasync/downloader/downloader.go
jeongkyun-oh: updated :)

storage/database/batch.go
```diff
@@ -48,3 +48,28 @@ type Batcher interface {
 	// until a final write is called.
 	NewBatch() Batch
 }
+
+// HookedBatch wraps an arbitrary batch where each operation may be hooked into
+// to monitor from black box code.
+type HookedBatch struct {
```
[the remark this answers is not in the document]
jeongkyun-oh: removed :)

cmd/utils/flags.go
jeongkyun-oh: Right. I have added a guard to support only `full` and `snap` sync.

node/cn/snap/handler.go
kjhman21: https://github.com/klaytn/klaytn/pull/1473/files#diff-7598d07f999ff8aea6077cbbbf66d0557e29530315933b1a03ba84c8111b75c6R46

If you are using testDownloader, DeliverSnapPacket() does nothing.

kjhman21: @jeongkyun-oh I added my test case with the name of handler2_test.go

ethan-kr (approved):

PR #1490, opened 2022-07-06
fixed error using before initilizing governance when making genesis b…

kjeom:
## Proposed changes

- When activating from genesis block, It should have baseFee in header.
- It is fixed by changing from governance parameter to hard coded `lowerboundbasefee` because the governance parameter is not initialized at that time.
- NOTE : the `baseFee` of the next block will be calculated within normal range of the governance parameter (lower - upper) even if the `baseFee` of genesis block is out of range (lower - upper) by calculation logic of `NextBlockBaseFee`.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/genesis.go
+++ b/blockchain/genesis.go
@@ -329,7 +329,7 @@ func (g *Genesis) ToBlock(baseStateRoot common.Hash, db database.DBManager) *typ
 	}
 	// TODO-klaytn should set genesis block's base fee
 	if g.Config != nil && g.Config.IsKIP71ForkEnabled(common.Big0) {
-		head.BaseFee = new(big.Int).SetUint64(g.Config.Governance.KIP71.LowerBoundBaseFee)
+		head.BaseFee = new(big.Int).SetUint64(params.DefaultLowerBoundBaseFee)
 	}
 	stateDB.Commit(false)
 	stateDB.Database().TrieDB().Commit(root, true, g.Number)


blockchain/genesis.go
```diff
@@ -329,7 +329,7 @@ func (g *Genesis) ToBlock(baseStateRoot common.Hash, db database.DBManager) *typ
 	}
 	// TODO-klaytn should set genesis block's base fee
 	if g.Config != nil && g.Config.IsKIP71ForkEnabled(common.Big0) {
-		head.BaseFee = new(big.Int).SetUint64(g.Config.Governance.KIP71.LowerBoundBaseFee)
+		head.BaseFee = new(big.Int).SetUint64(params.DefaultLowerBoundBaseFee)
```
jimni1222, line 332: So if activate KIP-71 hard fork, then baseFee of genesis block will be always `params.DefaultLowerBoundBaseFee` right?
But when calculate next base fee, governance lower/upper bound will be used, so even `params.DefaultLowerBoundBaseFee` is not in lower/upper bound range, baseFee will be valid (inside of that range). 
Did i understand correctly?

jimni1222: Sounds good!

jimni1222 (approved):

bobpkr (approved):

hqjang-pepper (approved):

PR #1385, opened 2022-05-19
Porting go-ethereum/rlp package.

sirano11:
## Proposed changes

For using "optional" struct tag, ported latest go-ethereum/rlp package.
  - Among the packages imported and referenced by the rlp package, Changed the ethereum package that referenced from rlp to the klaytn package.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1506, opened 2022-07-08
Use both 'go get' and 'go install' to install xgo

aidan-kwon:
## Proposed changes

From go1.18, go requires another way to install packages.
It registers a package in `go.mod` using`go get` and installs the package binary using `go install` 
(Reference: search 'go get' in https://tip.golang.org/doc/go1.18)

After merging https://github.com/klaytn/klaytn/pull/1385, `xgo` is removed from `go.mod`.
So, we need to call both `go get` and `go install` to install `xgo` with go1.18 or later. 

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/build/ci.go
+++ b/build/ci.go
@@ -1007,14 +1007,13 @@ func doXgo(cmdline []string) {
 	flag.CommandLine.Parse(cmdline)
 	env := build.Env()
 
-	subCmd := "get"
-	if strings.HasPrefix(runtime.Version(), "go1.18") {
-		subCmd = "install"
-	}
-
 	// Make sure xgo is available for cross compilation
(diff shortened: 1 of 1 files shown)


henry-will (approved):
How about using semantic version comparing code, like https://github.com/hashicorp/go-version, next time?
This time is important to take release tag.
 
It is ok in cases : go1.18 <  go1.19 < go1.20 < go2.0.
It is not ok in cases : go1.18 < go1.19 > go1.100, but it not bad because such a case will not happen until version 2.0.

aidan-kwon:
@henry-will I tried to use https://github.com/hashicorp/go-version,  but the return format (v1.18.0) of the library is different from `runtime.Version()` (go1.18.0). So, I just take a simple and quick way for now

PR #1534, opened 2022-07-19
changed suggestGasPrice to return baseFee * 2

kjeom:
## Proposed changes

- This change makes the `eth_gasPrice`, `klay_gasPrice` API to return current `baseFee * 2`
- It also makes users to use this API to set the gasPrice field in transaction.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -1113,14 +1113,8 @@ func (args *EthTransactionArgs) setDefaults(ctx context.Context, b Backend) erro
 				}
 				args.MaxFeePerGas = (*hexutil.Big)(gasFeeCap)
 			}
-			if isMagma {
-				if args.MaxFeePerGas.ToInt().Cmp(gasPrice) < 0 {
-					return fmt.Errorf("maxFeePerGas (%v) < BaseFee (%v)", args.MaxFeePerGas, gasPrice)
-				}
-			} else {
-				if args.MaxPriorityFeePerGas.ToInt().Cmp(gasPrice) != 0 || args.MaxFeePerGas.ToInt().Cmp(gasPrice) != 0 {
-					return fmt.Errorf("only %s is allowed to be used as maxFeePerGas and maxPriorityPerGas", gasPrice.Text(16))
-				}
+			if !isMagma && (args.MaxPriorityFeePerGas.ToInt().Cmp(gasPrice) != 0 || args.MaxFeePerGas.ToInt().Cmp(gasPrice) != 0) {
+				return fmt.Errorf("only %s is allowed to be used as maxFeePerGas and maxPriorityPerGas", gasPrice.Text(16))
 			}
 			if args.MaxFeePerGas.ToInt().Cmp(args.MaxPriorityFeePerGas.ToInt()) < 0 {
 				return fmt.Errorf("maxFeePerGas (%v) < maxPriorityFeePerGas (%v)", args.MaxFeePerGas, args.MaxPriorityFeePerGas)
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -1113,8 +1113,14 @@ func (args *EthTransactionArgs) setDefaults(ctx context.Context, b Backend) erro
 				}
 				args.MaxFeePerGas = (*hexutil.Big)(gasFeeCap)
 			}
-			if !isMagma && (args.MaxPriorityFeePerGas.ToInt().Cmp(gasPrice) != 0 || args.MaxFeePerGas.ToInt().Cmp(gasPrice) != 0) {
-				return fmt.Errorf("only %s is allowed to be used as maxFeePerGas and maxPriorityPerGas", gasPrice.Text(16))
+			if isMagma {
+				if args.MaxFeePerGas.ToInt().Cmp(new(big.Int).Div(gasPrice, common.Big2)) < 0 {
+					return fmt.Errorf("maxFeePerGas (%v) < BaseFee (%v)", args.MaxFeePerGas, gasPrice)
+				}
+			} else {
+				if args.MaxPriorityFeePerGas.ToInt().Cmp(gasPrice) != 0 || args.MaxFeePerGas.ToInt().Cmp(gasPrice) != 0 {
+					return fmt.Errorf("only %s is allowed to be used as maxFeePerGas and maxPriorityPerGas", gasPrice.Text(16))
+				}
 			}
 			if args.MaxFeePerGas.ToInt().Cmp(args.MaxPriorityFeePerGas.ToInt()) < 0 {
 				return fmt.Errorf("maxFeePerGas (%v) < maxPriorityFeePerGas (%v)", args.MaxFeePerGas, args.MaxPriorityFeePerGas)
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -1109,7 +1109,7 @@ func (args *EthTransactionArgs) setDefaults(ctx context.Context, b Backend) erro
 				)
 				if isMagma {
 					// After Magma hard fork, `gasFeeCap` was set to `baseFee*2` by default.
-					gasFeeCap = new(big.Int).Mul(gasPrice, big.NewInt(2))
+					gasFeeCap = gasPrice
 				}
 				args.MaxFeePerGas = (*hexutil.Big)(gasFeeCap)
 			}
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -1113,8 +1113,10 @@ func (args *EthTransactionArgs) setDefaults(ctx context.Context, b Backend) erro
 				}
 				args.MaxFeePerGas = (*hexutil.Big)(gasFeeCap)
 			}
-			if isMagma && args.MaxFeePerGas.ToInt().Cmp(new(big.Int).Div(gasPrice, common.Big2)) < 0 {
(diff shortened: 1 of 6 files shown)


api/api_public_klay.go
```diff
@@ -42,9 +44,10 @@ func NewPublicKlayAPI(b Backend) *PublicKlayAPI {
 	return &PublicKlayAPI{b}
 }
 
-// GasPrice returns a suggestion for a gas price.
+// GasPrice returns a suggestion for a gas price (baseFee * 2).
 func (s *PublicKlayAPI) GasPrice(ctx context.Context) (*hexutil.Big, error) {
 	price, err := s.b.SuggestPrice(ctx)
```
jimni1222, line 49: How about chainging logic in `s.b.SuggestPrice(ctx)`?

kjeom: Eth namespace API, `GasPrice`, just call the `PublicKlayAPI.GasPrice`. So I think we need to change some logic only about setDefaults.

jimni1222: Oh okayokay.
I think `SuggestPrice` function name includes `"Suggest"`, so changing logic in `SuggestPrice` function might be better in my opinion. What's your thinking with this?

jimni1222: And shouldn't we also check if the current block is a hard fork or not?

kjeom: - I agree what the 'Suggest' means.
- Right, we should check the hard fork.
  - Before hard fork, It just returns the unitPrice.
  - After hard fork, It returns the BaseFee * 2.

api/api_ethereum.go
```diff
@@ -1109,7 +1109,7 @@ func (args *EthTransactionArgs) setDefaults(ctx context.Context, b Backend) erro
 				)
 				if isMagma {
 					// After Magma hard fork, `gasFeeCap` was set to `baseFee*2` by default.
-					gasFeeCap = new(big.Int).Mul(gasPrice, big.NewInt(2))
+					gasFeeCap = gasPrice
```
jimni1222, line 1112: I think L1117 validation also no more a thing.

api/tx_args.go
```diff
@@ -157,15 +157,12 @@ func (args *SendTxArgs) setDefaults(ctx context.Context, b Backend) error {
...
+		//                = baseFee * 2,   for after Magma
 		price, err := b.SuggestPrice(ctx)
 		if err != nil {
 			return err
 		}
-		if isMagma {
-			// we need to set baseFee * 2 after Magma hard fork
-			price = new(big.Int).Mul(price, common.Big2)
-		}
+		price = new(big.Int).Mul(price, common.Big2)
```
jimni1222, line 165:
```suggestion
```

node/cn/api_backend.go
```diff
@@ -278,8 +278,14 @@ func (b *CNAPIBackend) ProtocolVersion() int {
 	return b.cn.ProtocolVersion()
 }
 
+// If current block is magma hard forked, it returns the baseFee * 2.
+// Other cases, it returns the unitPrice.
 func (b *CNAPIBackend) SuggestPrice(ctx context.Context) (*big.Int, error) {
-	return b.gpo.SuggestPrice(ctx)
+	suggestedPrice, _ := b.gpo.SuggestPrice(ctx)
+	if b.ChainConfig().IsMagmaForkEnabled(b.CurrentBlock().Number()) {
```
jimni1222, line 285: How about using `current block number + 1`?

kjeom: Great! Thanks

node/cn/api_backend.go
```diff
@@ -278,8 +278,14 @@ func (b *CNAPIBackend) ProtocolVersion() int {
...
+		return new(big.Int).Mul(suggestedPrice, common.Big2), nil
```
aidan-kwon, line 286: It should be equal or smaller than Upperbound

node/cn/api_backend.go
```diff
@@ -278,8 +278,14 @@ func (b *CNAPIBackend) ProtocolVersion() int {
 	return b.cn.ProtocolVersion()
 }
 
+// If current block is magma hard forked, it returns the baseFee * 2.
+// Other cases, it returns the unitPrice.
 func (b *CNAPIBackend) SuggestPrice(ctx context.Context) (*big.Int, error) {
-	return b.gpo.SuggestPrice(ctx)
+	suggestedPrice, _ := b.gpo.SuggestPrice(ctx)
```
jimni1222, line 284: May I ask why you added this logic here rather than `b.gpo.SuggestPrice(ctx)` ?

node/cn/api_backend.go
kjeom: It doesn't matter because there is no condition to check tx.gasPrice >= upperBound internally.

node/cn/api_backend.go
```diff
@@ -278,8 +278,14 @@ func (b *CNAPIBackend) ProtocolVersion() int {
...
+	if b.ChainConfig().IsMagmaForkEnabled(new(big.Int).Add(b.CurrentBlock().Number(), common.Big1)) {
+		return new(big.Int).Mul(suggestedPrice, common.Big2), nil
+	}
+	return suggestedPrice, nil
```
ian0371, line 288: This function now always returns error=nil. Is this ok?

kjeom: Yes, actually gpo.SuggestPrice always return nil as well.

node/cn/api_backend.go
jimni1222: If so, maybe we can add one more parameter like `isMagmaForkEnabled` so pass whether after Magma or not?

kjeom: That's possible. Actually gpo has backend instance by interface so If we add the currentBlock() in the interface then we can get the block number.
How do you think about this?

jimni1222: Sounds good, or we can use txPool because tx pool also has fork indicator.

kjeom: Good idea! I think It's better to use the txPool's field.

kjeom: Ohh... but It needs an additional function like isMagma().
I think hard fork checking function, isMagmaCompatible(), is enough to avoid confusion so we can just use the currentBlock().

jimni1222: Okay~ whatever you think is better, i'm down for it!

api/api_ethereum.go
```diff
@@ -1109,18 +1109,12 @@ func (args *EthTransactionArgs) setDefaults(ctx context.Context, b Backend) erro
...
 				}
 				args.MaxFeePerGas = (*hexutil.Big)(gasFeeCap)
 			}
-			if isMagma {
```
jimni1222, line 1116: Please apply those modifications to `setDefaults` function in `tx_args.go`.

kjeom: I think these cases are different.
In api_ethereum.go, `args.MaxFeePerGas` cannot be a nil because If it is nil, it is assigned some value.
In tx_args.go, if `args.MaxFeePerGas` is not nil, in other words user assigned some value, we need to check args.MaxFeePerGas < baseFee, not gasPrice(baseFee * 2)

`args.MaxFeePerGas < gasPrice` ==> `args.MaxFeePerGas < currentBlockHeader.BaseFee`

Anyway I think we need to refactor these functions...

jimni1222: Um.. still i'm confused...
I can't understand why we don't need to check in api_ethereum.go, but have to check in tx_args.go.
I think still possible to set `args.MaxFeePerGas` value by user in api_ethereum.go also.

kjeom: Ohh.. right then we need to remain the `isMagma` condition in `api_ethereum.go`, right?

jimni1222: Yes, I also agree with doing the same (or similar) validation check will be better

kjeom: Thansk!

api/api_ethereum.go
```diff
@@ -1113,7 +1113,9 @@ func (args *EthTransactionArgs) setDefaults(ctx context.Context, b Backend) erro
 				}
 				args.MaxFeePerGas = (*hexutil.Big)(gasFeeCap)
 			}
-			if !isMagma && (args.MaxPriorityFeePerGas.ToInt().Cmp(gasPrice) != 0 || args.MaxFeePerGas.ToInt().Cmp(gasPrice) != 0) {
+			if isMagma && args.MaxFeePerGas.ToInt().Cmp(new(big.Int).Div(gasPrice, common.Big2)) < 0 {
```
jimni1222, line 1116: I think you should separate the 'isMagma' check. 
Otherwise still the 'else if' check logic can be operated when it's a magma hard fork.

jimni1222 (dismissed):

bobpkr (approved):

node/cn/api_backend.go
```diff
@@ -278,6 +278,8 @@ func (b *CNAPIBackend) ProtocolVersion() int {
 	return b.cn.ProtocolVersion()
 }
 
+// If current block is magma hard forked, it returns the baseFee * 2.
```
aidan-kwon, line 281: For the convention of goDoc comments https://go.dev/blog/godoc
```suggestion
// SuggestPrice returns the baseFee * 2 if the current block is magma hard forked.
```

jimni1222 (approved):

aidan-kwon (approved):
It looks okay. We need to refind `setDefaults` later, though

PR #1549, opened 2022-07-22
More validations for tracer output converting logic

aidan-kwon:
## Proposed changes

`InternalTxTracer.result` is only called by fastCallTracer now. 
(fastCallTracer can be used for debug APIs or chaindatafetcher)

The tracer result parses an error string when a transaction is reverted.
Since current logic doesn't have enough validation logic about slice range or overflow/underflow, the tracer can be panic when an abnormal contract executed. 

This change introduces only small parts of validation for recent invalid cases. 
After v1.9.0 QA, more concrete validation will be introduced later. 

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/vm/internaltx_trace_json_test.go
+++ b/blockchain/vm/internaltx_trace_json_test.go
@@ -258,3 +258,14 @@ func TestInternalTxTrace_MarshalJSON(t *testing.T) {
 		})
 	}
 }
+
+func TestInternalTxTracer_result_invalidOutput(t *testing.T) {
+	tracer := NewInternalTxTracer()
+	tracer.ctx["error"] = errEvmExecutionReverted
+	tracer.ctx["output"] = "0x08c379a0000000000000000000000000000000000000000000000000ffffffffffffffff000000000000000000000000000000000000000000000000ffffffffffffffff"
+
+	_, err := tracer.result()
+	if err != nil {
+		t.Fatal(err)
+	}
+}
--- a/blockchain/vm/internaltx_trace_json_test.go
+++ b/blockchain/vm/internaltx_trace_json_test.go
@@ -262,6 +262,7 @@ func TestInternalTxTrace_MarshalJSON(t *testing.T) {
 func TestInternalTxTracer_result_invalidOutput(t *testing.T) {
 	tracer := NewInternalTxTracer()
 	tracer.ctx["error"] = errEvmExecutionReverted
+	// An 8 bytes value that might cause a string version error or out-of-range error.
 	tracer.ctx["output"] = "0x08c379a0000000000000000000000000000000000000000000000000ffffffffffffffff000000000000000000000000000000000000000000000000ffffffffffffffff"
 
 	_, err := tracer.result()
--- a/blockchain/vm/internaltx_tracer.go
+++ b/blockchain/vm/internaltx_tracer.go
@@ -515,24 +515,57 @@ func (this *InternalTxTracer) result() (*InternalTxTrace, error) {
 	}
 	if err := this.ctx["error"]; err != nil && err.(error).Error() == errEvmExecutionReverted.Error() {
 		outputHex := this.ctx["output"].(string) // it is already a hex string
-		if len(outputHex) >= 11 && outputHex[2:10] == "08c379a0" {
+		outputHexLength := len(outputHex)
+
+		// outputHexLength should be equal for larger than 138 (10+32*2+32*2) to parse a revert string
+		// outputHex[:10]: "0x08c379a0"
+		// outputHex[10:10+32*2]: a revert string
+		// outputHex[10+32*2:10+32*2+32*2]: the length of a revert string
+		if outputHexLength >= 138 && outputHex[2:10] == "08c379a0" {
 			defaultOffset := 10
 
-			stringOffset, err := strconv.ParseInt(outputHex[defaultOffset:defaultOffset+32*2], 16, 64)
+			// TODO-Klaytn: introduce error handling logic for the case the parsing data is bigger than math.MaxUint64
+			stringOffset, err := strconv.ParseUint(outputHex[defaultOffset:defaultOffset+32*2], 16, 64)
 			if err != nil {
 				logger.Error("failed to parse hex string to get stringOffset",
 					"err", err, "outputHex", outputHex)
 				return nil, err
 			}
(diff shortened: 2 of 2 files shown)


blockchain/vm/internaltx_tracer.go
```diff
@@ -515,24 +515,57 @@ func (this *InternalTxTracer) result() (*InternalTxTrace, error) {
...
 			end := start + int(stringLength*2)
-			asciiInBytes, err := hex.DecodeString(outputHex[start:end])
+
+			if start < 0 {
+				start = 0
+			}
+			if end < 0 {
+				end = 0
+			}
+			// return nothing if end is out of the range
```
kjhman21, line 557: @aidan-kwon What if start > outputHexLength ?

blockchain/vm/internaltx_trace_json_test.go
```diff
@@ -258,3 +258,14 @@ func TestInternalTxTrace_MarshalJSON(t *testing.T) {
 		})
 	}
 }
+
+func TestInternalTxTracer_result_invalidOutput(t *testing.T) {
+	tracer := NewInternalTxTracer()
+	tracer.ctx["error"] = errEvmExecutionReverted
+	tracer.ctx["output"] = "0x08c379a0000000000000000000000000000000000000000000000000ffffffffffffffff000000000000000000000000000000000000000000000000ffffffffffffffff"
```
jeongkyun-oh, line 265: Would you mind to describe what is the meaning of this output?

blockchain/vm/internaltx_tracer.go
```diff
@@ -515,24 +515,57 @@ func (this *InternalTxTracer) result() (*InternalTxTrace, error) {
 	}
 	if err := this.ctx["error"]; err != nil && err.(error).Error() == errEvmExecutionReverted.Error() {
 		outputHex := this.ctx["output"].(string) // it is already a hex string
-		if len(outputHex) >= 11 && outputHex[2:10] == "08c379a0" {
+		outputHexLength := len(outputHex)
+
+		// outputHexLength should be equal for larger than 138 (10+32*2+32*2) to parse a revert string
+		// outputHex[:10]: "0x08c379a0"
```
jeongkyun-oh, line 521:
```suggestion
		// outputHex[:10]: "0x08c379a0" = crypto.Keccak256([]byte("Error(string)"))[:4]
```

blockchain/vm/internaltx_tracer.go
```diff
@@ -515,24 +515,57 @@ func (this *InternalTxTracer) result() (*InternalTxTrace, error) {
...
+				end = 0
+			}
+			// return nothing if end is out of the range
+			if end > outputHexLength {
+				start = outputHexLength
+				end = outputHexLength
+			}
+
+			// left-padding with 0 for hex decoding
+			outputHexSlice := outputHex[start:end]
```
jeongkyun-oh, line 564: Is it possible `start > end`?

jeongkyun-oh:
How about `callTracer`? Should we align `callTracer.js` with this change?

blockchain/vm/internaltx_tracer.go
aidan-kwon: `end := start + int(stringLength*2)` and stringLength is always positive or zero. So, end > outputHexLength if start > outputHexLength

blockchain/vm/internaltx_tracer.go
aidan-kwon: end := start + int(stringLength*2) and stringLength is always positive or zero. So, end >= start

aidan-kwon:
> How about `callTracer`? Should we align `callTracer.js` with this change?

We should. But, not for this time. Since we are about to release v1.9.0, I don't want to put large changes. This is a kind of hot fixes for recent issues. We need to investigate more about tracers for v1.9.1.

blockchain/vm/internaltx_trace_json_test.go
aidan-kwon, line 265:
```suggestion
	// A 8 bytes value that might cause a string version error or out-of-range error. 
	tracer.ctx["output"] = "0x08c379a0000000000000000000000000000000000000000000000000ffffffffffffffff000000000000000000000000000000000000000000000000ffffffffffffffff"
```

blockchain/vm/internaltx_tracer.go
aidan-kwon: 1. If an overflow occurs with `int(stringOffset*2)`, `start` will be less than zero in the most cases. So, `start` will be set to zero. 
3. If an overflowed value is in `[-1, -72]`. `start` could be an invalid value in `[0, 71]`. However, this value cannot make a panic because the length of outputHex is larger than 138. Though it could returns a weird value for revertString, but i think it is okay because if can happend with wired revert message only.

blockchain/vm/internaltx_tracer.go
```diff
@@ -515,24 +515,57 @@ func (this *InternalTxTracer) result() (*InternalTxTrace, error) {
...
+			if start < 0 {
+				start = 0
+			}
+			if end < 0 {
+				end = 0
+			}
+			// return nothing if end is out of the range
+			if end > outputHexLength {
+				start = outputHexLength
+				end = outputHexLength
```
kjhman21, line 560:
```suggestion
				start = outputHexLength-1
				end = outputHexLength-1
```

blockchain/vm/internaltx_tracer.go
kjhman21: int(stringLength*2) can be negative.

aidan-kwon: stringLength is equal or smaller than outputHexLength. outputHexLength is a slice size, and the size cannot be bigger than `MaxInt/2`

blockchain/vm/internaltx_tracer.go
aidan-kwon: Done

blockchain/vm/internaltx_tracer.go
kjhman21: okay. got it. Thanks.

blockchain/vm/internaltx_tracer.go
jeongkyun-oh: Ah, I got it. Thanks.

jeongkyun-oh (approved):

kjhman21 (approved):

PR #1527, opened 2022-07-17
Return nil if from == to

kjhman21:
## Proposed changes

- Fix a bug when from == to

## Types of changes

Please put an x in the boxes related to your change.

- [x] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1576, opened 2022-08-03
Safely handle corner cases in Tracers

blukat29:
## Proposed changes

- Harden fastCallTracer (`internaltx_tracer.go`)
  - Use `abi.UnpackRevert()` to parse revert output.
- Harden callTracer (`call_tracer.js`)
  - Try-catch `bigInt()`
- Harden revertTracer (`revert_tracer.js`)
  - Try-catch `bigInt()`
- Update tracers test code
  - Respect the chainConfig in the testdata json
  - New helper scripts to generate testdata json
- Add testdata related to this fix and previous fixes ([#1527](https://github.com/klaytn/klaytn/pull/1527), [#1549](https://github.com/klaytn/klaytn/pull/1549))

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist


blockchain/vm/internaltx_tracer.go
```diff
@@ -515,64 +514,13 @@ func (this *InternalTxTracer) result() (*InternalTxTrace, error) {
...
-			// left-padding with 0 for hex decoding
-			outputHexSlice := outputHex[start:end]
-			if len(outputHexSlice)%2 == 1 {
-				outputHexSlice = "0" + outputHexSlice
-			}
-			asciiInBytes, err := hex.DecodeString(outputHexSlice)
-			if err != nil {
-				logger.Error("failed to parse hex string to get ASCII representation",
-					"err", err, "outputHex", outputHex)
-				return nil, err
```
jeongkyun-oh, line 572: It seems that there is no error case in this method. How about removing `error` to be returned?

blukat29: I am okay with not returning errors from the private function `result()`. I will add a commit.

@jeongkyun-oh But should we stop returning errors from the public function `GetResult()`?
I'd say we keep returning errors for the following reasons:
- If we add another Go tracer, the tracer might return errors.
- For consistency with JS tracer https://github.com/klaytn/klaytn/blob/v1.9.0/node/cn/tracers/tracer.go#L631
- For compatibility with Eth https://github.com/ethereum/go-ethereum/blob/v1.10.21/eth/tracers/tracers.go#L36

jeongkyun-oh: @blukat29 It makes sense. In my side, a block should be processed without returning an error because this causes bad block. I think updating only private function is enough in this PR. I will update the outer function if you agree.

jeongkyun-oh (approved):

blockchain/vm/internaltx_tracer.go
blukat29: @jeongkyun-oh To avoid bad block we can simply modify `ApplyTransaction()`? https://github.com/klaytn/klaytn/blob/v1.9.0/blockchain/blockchain.go#L2628.

```diff
	var internalTrace *vm.InternalTxTrace
	if vmConfig.EnableInternalTxTracing {
		internalTrace, err = GetInternalTxTrace(vmConfig.Tracer)
		if err != nil {
			logger.Error("failed to get tracing result from a transaction", "txHash", tx.Hash().String(), "err", err)
-			return nil, 0, nil, err
		}
	}
```

I think the caller should deal with it. Block processors will ignore the error and debug APIs will return the error.

jeongkyun-oh: @blukat29 Yes, you are right. This is simplest one. How about panic? Do you think that we need to recover the panic?

blukat29: @jeongkyun-oh Methods like `CaptureState()` and `GetResult()` is good places to recover.
But after the previous and present fixes, I don't see any possible panic in InternalTxTracer.
While recovering from the panic is no harm, I am also fine without recovery for the sake of simplicity.

jeongkyun-oh: @blukat29 Thanks for sharing your idea.

ian0371 (approved):

blockchain/vm/internaltx_tracer.go
hyunsooda: Then,` this.err` did not have effect until now? Which means every callers `GetResult()` does not need to take and consider the error (this [PR](https://github.com/klaytn/klaytn/pull/1584) as well)

hyunsooda (approved):
LGTM

blockchain/vm/internaltx_tracer.go
blukat29: @hyunsooda `this.err` is often populated when (1) tracing interrupted or (2) `this.step()` fails, (3) execution faults. 
Among those, case (1) can happen and is relevant to debug APIs ([code](https://github.com/klaytn/klaytn/blob/v1.9.0/node/cn/api_tracer.go#L782)), case (2) never happens in the current code, case (3) can happen when executing a malformed contract (e.g. ErrInvalidJump).
Therefore `this.err` is necessary even after this PR.

hyunsooda: Uh oh, I was confused that the `this.err` and returned error of the private function `result()` are the same thing. Oops.. I'm so sorry for the noise. Thank you.

aidan-kwon (approved):

Issue #1494, opened 2022-07-07
Configuration based on environment variables as well as file.


odg0318:
**Is your request related to a problem? Please describe.**
Currently, configuration can be passed only by command line arguments.
It seems like there is a file such as `/etc/kcnd/conf/kcnd.conf` but the file cannot be directly used with klaytn binary.
To configure klaytn binary using conf file, startup script like `kcnd` is required.
To execute klaytn binary, three files are required; `kcnd`, `kcn`, `kcnd.conf`, which makes flexibility down.

`kcnd` is working like the followings:

#### /etc/kcnd/conf/kcnd.conf
```bash
DATA_DIR=/var/kcnd/data
```
#### /usr/bin/kcnd
```bash
source /etc/kcnd/conf/kcnd.conf

__check_option() {
    if [[ ! -z $DATA_DIR ]]; then 
         OPTIONS="--datadir $DATA_DIR
    fi
}

kcn $OPTIONS
```
#### Finally execute
```
$ kcnd
```

I think that the configuration way can be improved.

**Describe the solution you'd like**
When I checked the code, https://github.com/urfave/cli is used to implement flag. `urfavce/cli` provides a feature to configure using environment variables or yaml files. The code can be refactored using them.

```golang
var (
        DataDirFlag = DirectoryFlag{
                Name:  "datadir",
                Usage: "Data directory for the databases and keystore. This value is only used in local DB.",
                Value: DirectoryString{node.DefaultDataDir()},
                EnvVar: "KLAYTN_DATADIR",
        }
)
```

Let's go to https://cli.urfave.org/v1/#values-from-the-environment in detail.

Now, klaytn binary can be executed only with executable binary and environment variables.
```
$ export KLAYTN_DATADIR=/var/kcnd/conf/kcnd
$ kcn
```

Additionally, the arguments can be also passed using files.
```golang
  flags := []cli.Flag{
        DataDirFlag,
        cli.StringFlag{Name: "conf"},
  }

   app.Before = altsrc.InitInputSourceWithContext(flags, altsrc.NewYamlSourceFromFlagFunc("conf"))
```
#### /etc/kcnd/conf/kcnd.yaml
```yaml
datadir: /var/kcnd/conf/kcnd.conf
```
```bash
$ kcn --conf /etc/kcnd/conf/kcnd.yaml
```

With the all above, the configuration way can be flexible much more and it's more helpful to run klaytn inside container.

**Describe alternatives you've considered**
N/A

**Additional context**
[...]

PR #1580, opened 2022-08-09
added to accept environment variable and yaml file for configuration

kjeom:
## Proposed changes

**- Issues**
  - It is needed to execute the klaytn binaries (kcn, kpn, ken, ...) with environment variable or configuration file.
  - With this resolution, you can use legacy config file(kcnd.conf, kpnd.conf, ...) with environment variable.

**- Proposed change pattern (environment variable)**
```go
// legacy flag
verbosityFlag = cli.IntFlag{
	Name:   "verbosity",
	Usage:  "Logging verbosity: 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=detail",
	Value:  3,
}
```
is changed to following for applying environment variable that is top priority.
```go
verbosityFlag = cli.IntFlag{
	Name:   "verbosity",
	Usage:  "Logging verbosity: 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=detail",
	Value:  3,
	EnvVar: "KLAYTN_VERBOSITY",
}
```
**- Proposed change pattern (yaml file)**
```go
// legacy flag
verbosityFlag = cli.IntFlag{ ... }
```
is changed to following for applying yaml configuration file to the flags
```go
verbosityFlag = cli.IntFlag{ ... }
verbosityFlagsAddedYaml = altsrc.NewIntFlag(verbosityFlag)

...

altsrc.InitInputSourceWithContext(
  allNodeFlags(), // flags to read from yaml
  altsrc.NewYamlSourceFromFlagFunc("conf") // file name entered via --conf option
)(ctx)
```
**- Proposed change pattern (flag arrange)**
```go
// legacy flag setting code
nodeFlags = append(nodecmd.CommonNodeFlags, nodecmd.KENFlags...)
rpcFlags = nodecmd.CommonRPCFlags
nodecmd.GetConsoleCommand(nodeFlags, rpcFlags),
```
is changed like following and we can see the overview of flag bundles being used for each binaries in the nodeflags.go.
```go
nodecmd.GetConsoleCommand(nodecmd.KenNodeFlags(), nodecmd.CommonRPCFlags),
// KenNodeFlags() is defined in cmd/utils/nodecmd/nodeflags.go (same every node flags)
```

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- #1494 

## furthur comments

- Some Env_Var and yaml configuration is tested on each binaries
- Flag test case will be added in next PR


api/debug/flags.go
```diff
@@ -32,73 +32,94 @@ import (
...
 
 var (
 	verbosityFlag = cli.IntFlag{
-		Name:  "verbosity",
-		Usage: "Logging verbosity: 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=detail",
-		Value: 3,
+		Name:   "verbosity",
+		Usage:  "Logging verbosity: 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=detail",
+		Value:  3,
+		EnvVar: "KLAYTN_VERBOSITY",
```
kjhman21, line 45: Do we need to add "KLAYTN" prefix for all env variables?

Is it possible to follow the previous convention we used in kcnd.conf?
https://github.com/klaytn/klaytn/blob/dev/build/packaging/linux/conf/kcnd.conf

cmd/utils/flags.go
```diff
@@ -81,1013 +81,1227 @@ func NewApp(gitCommit, usage string) *cli.App {
...
+		Usage:  "Directory for the keystore (default = inside the datadir)",
+		EnvVar: "KLAYTN_KEYSTORE",
 	}
 	// TODO-Klaytn-Bootnode: redefine networkid
 	NetworkIdFlag = cli.Uint64Flag{
-		Name:  "networkid",
-		Usage: "Network identifier (integer, 1=MainNet (Not yet launched), 1000=Aspen, 1001=Baobab)",
-		Value: cn.GetDefaultConfig().NetworkId,
+		Name:   "networkid",
+		Usage:  "Network identifier (integer, 1=MainNet (Not yet launched), 1000=Aspen, 1001=Baobab)",
```
kjhman21, line 119:
```suggestion
		Usage:  "Network identifier (integer, 8217= Cypress (Mainnet) , 1000=Aspen, 1001=Baobab)",
```

api/debug/flags.go
kjeom: I think It can cause unintended configuration.
It will be failed to execute the node in some environment where has `export VERBOSITY="some value"` because `EnvVar` has top priority.
So I just added the prefix, KLAYTN.

kjhman21: hmm... Okay. I think there are pros and cons to use prefix "KLAYTN"

pros
- no unintended conflict
cons
- brakes the previous convention
- long environment variable

Maybe there are other pros and cons, let's hear others opinion.

odg0318: @kjhman21 I think that the prefix is required because of the same reason @kjeom suggested.

I provide another project like the followings:
* https://www.runatlantis.io/docs/server-configuration.html#environment-variables
* https://docs.datadoghq.com/agent/guide/environment-variables/

To avoid the duplicated environment' names, most projects are using the prefix.

cmd/kpn/main.go
```diff
@@ -63,42 +55,28 @@ func init() {
...
-		logDir := (&node.Config{DataDir: utils.MakeDataDir(ctx)}).ResolvePath("logs")
-		debug.CreateLogDir(logDir)
-		if err := debug.Setup(ctx); err != nil {
-			return err
-		}
-		metricutils.StartMetricCollectionAndExport(ctx)
-		utils.SetupNetwork(ctx)
-		return nil
-	}
+	app.Before = nodecmd.BeforeRunNode
```
odg0318, line 79: As I see the implementation of any other cmds, this code looks different a little bit.
Is there any reason not to execute `nodecmd.CheckCommand` here like the below?
```golang
app.Before = func(ctx *cli.Context) error {
		if err := nodecmd.CheckCommands(ctx); err != nil {
			return err
		}
		return nodecmd.BeforeRunNode(ctx)
	}
```

cmd/kspn/main.go
```diff
@@ -63,38 +55,25 @@ func init() {
...
-		logDir := (&node.Config{DataDir: utils.MakeDataDir(ctx)}).ResolvePath("logs")
-		debug.CreateLogDir(logDir)
-		if err := debug.Setup(ctx); err != nil {
-			return err
-		}
-		metricutils.StartMetricCollectionAndExport(ctx)
-		utils.SetupNetwork(ctx)
-		return nil
-	}
+	app.Before = nodecmd.BeforeRunNode
```
odg0318, line 76: Same question as the above
https://github.com/klaytn/klaytn/pull/1580/files#r944160746

odg0318:
Where or how can we write document for this configuration way?
I think that more tasks remain even though this PR is merged. And the tasks can be handled in other PRs.
* Documentation
* Provide an example yaml file for configuration
* Re-write packaging for distribution with yaml file

cmd/kpn/main.go
kjeom: You are right.
- I have seen that each node has difference codes for app.Before
- And I analyzed it so that would like to modify it to like following in every nodes.
```go
app.Before = nodecmd.BeforeRunNode(ctx)
```
- But I couldn't assure 100% if It has no side-effect when operating the CN, PN, EN and tried to keep it same logic.
- So I hope we can discuss it together the next time we sort out unused flags.

api/debug/flags.go
kjhman21: @kjeom @odg0318 Okay got it. Please make sure not to break compatibility

cmd/kpn/main.go
kjeom: @kjhman21 @aidan-kwon Could you comment on this?

api/debug/flags.go
aidan-kwon, line 45: All flags have the same prefix `KLAYTN_`, so it seems meaningless

kjeom: Yes, but I think it can be distinguished from environment variables for other applications.

odg0318: This was already discussed in [the above.](https://github.com/klaytn/klaytn/pull/1580#discussion_r941351200)

cmd/kpn/main.go
aidan-kwon: I think this logic could be the same with other binaries.  It seems that this logic is missed during the refactoring

aidan-kwon: You can refer old repository https://github.com/Krustuniverse-Klaytn-Group/klaytn/pull/3300/files

aidan-kwon (dismissed):

cmd/kpn/main.go
kjeom: Okay, thanks for the refference. I added the checkCommand to the BeforeRunNode in common

aidan-kwon (approved):

aidan-kwon:
Reviewers, PTAL

cmd/utils/customflags.go
```diff
@@ -51,12 +54,58 @@ func (self *DirectoryString) Set(value string) error {
...
+}
+
+func (f *WrappedDirectoryFlag) Apply(set *flag.FlagSet) {
+	f.set = set
+	f.DirectoryFlag.Apply(set)
+}
+
+func (f *WrappedDirectoryFlag) ApplyInputSourceValue(context *cli.Context, isc altsrc.InputSourceContext) error {
+	if f.set != nil {
+		// if !isEnvVarSet(f.EnvVar) {
```
kjhman21, line 73: Is this intentional?

cmd/utils/flags.go
```diff
@@ -81,1013 +81,1227 @@ func NewApp(gitCommit, usage string) *cli.App {
...
+		Usage:  "Directory for the keystore (default = inside the datadir)",
+		EnvVar: "KLAYTN_KEYSTORE",
 	}
 	// TODO-Klaytn-Bootnode: redefine networkid
 	NetworkIdFlag = cli.Uint64Flag{
-		Name:  "networkid",
-		Usage: "Network identifier (integer, 1=MainNet (Not yet launched), 1000=Aspen, 1001=Baobab)",
-		Value: cn.GetDefaultConfig().NetworkId,
+		Name:   "networkid",
+		Usage:  "Network identifier (integer, 8217= Cypress (Mainnet) , 1000=Aspen, 1001=Baobab)",
```
kjhman21, line 119:
```suggestion
		Usage:  "Network identifier (integer, 8217=Cypress (Mainnet) , 1000=Aspen, 1001=Baobab)",
```

cmd/utils/flags.go
```diff
@@ -81,1013 +81,1227 @@ func NewApp(gitCommit, usage string) *cli.App {
...
+		EnvVar: "KLAYTN_VM_INTERNALTX",
 	}
 
 	// Logging and debug settings
 	MetricsEnabledFlag = cli.BoolFlag{
-		Name:  metricutils.MetricsEnabledFlag,
-		Usage: "Enable metrics collection and reporting",
+		Name:   metricutils.MetricsEnabledFlag,
+		Usage:  "Enable metrics collection and reporting",
+		EnvVar: "KLAYTN_ETRICUTILS_METRICSENABLEDFLA",
```
kjhman21, line 509:
```suggestion
		EnvVar: "KLAYTN_METRICUTILS_METRICSENABLEDFLA",
```
Seems like `M` is missing?

cmd/utils/flags.go
```diff
@@ -81,1013 +81,1227 @@ func NewApp(gitCommit, usage string) *cli.App {
...
 	}
 	PrometheusExporterFlag = cli.BoolFlag{
-		Name:  metricutils.PrometheusExporterFlag,
-		Usage: "Enable prometheus exporter",
+		Name:   metricutils.PrometheusExporterFlag,
+		Usage:  "Enable prometheus exporter",
+		EnvVar: "KLAYTN_ETRICUTILS_PROMETHEUSEXPORTERFLA",
```
kjhman21, line 514:
```suggestion
		EnvVar: "KLAYTN_METRICUTILS_PROMETHEUSEXPORTERFLAG",
```

cmd/utils/flags.go
```diff
@@ -81,1013 +81,1227 @@ func NewApp(gitCommit, usage string) *cli.App {
...
 	}
 	PrometheusExporterPortFlag = cli.IntFlag{
-		Name:  metricutils.PrometheusExporterPortFlag,
-		Usage: "Prometheus exporter listening port",
-		Value: 61001,
+		Name:   metricutils.PrometheusExporterPortFlag,
+		Usage:  "Prometheus exporter listening port",
+		Value:  61001,
+		EnvVar: "KLAYTN_ETRICUTILS_PROMETHEUSEXPORTERPORTFLA",
```
kjhman21, line 520:
```suggestion
		EnvVar: "KLAYTN_METRICUTILS_PROMETHEUSEXPORTERPORTFLAG",
```

cmd/utils/flags.go
```diff
@@ -81,1013 +81,1227 @@ func NewApp(gitCommit, usage string) *cli.App {
...
-		Usage: "The maximum difference between current block and event block. 0 means off",
-		Value: 0,
+		Name:   "dbsyncer.max.block.diff",
+		Usage:  "The maximum difference between current block and event block. 0 means off",
+		Value:  0,
+		EnvVar: "KLAYTN_DBSYNCER_MAX_BLOCK_DIFF",
 	}
 	AutoRestartFlag = cli.BoolFlag{
 		Name:  "autorestart.enable",
 		Usage: "Node can restart itself when there is a problem in making consensus",
```
kjhman21, line 1199: Don't we need to have EnvVar for this flag?

cmd/utils/nodecmd/defaultcmd.go
```diff
@@ -162,7 +163,52 @@ func CheckCommands(ctx *cli.Context) error {
...
+	nodeFlags = union(nodeFlags, KPNFlags)
+	nodeFlags = union(nodeFlags, KENFlags)
+	nodeFlags = union(nodeFlags, KSCNFlags)
+	nodeFlags = union(nodeFlags, KSPNFlags)
+	nodeFlags = union(nodeFlags, KSENFlags)
+	return nodeFlags
+}
+
+func FlagsFromYaml(ctx *cli.Context) error {
+	if ctx.String("conf") != "" {
```
kjhman21, line 200: What about adding a global variable like this instead of using the hard-coded string.

`var confFile = "conf"`

cmd/utils/nodecmd/nodeflags.go
```diff
@@ -21,353 +21,432 @@
...
-	utils.AutoRestartFlag,
-	utils.RestartTimeOutFlag,
-	utils.DaemonPathFlag,
-	utils.ConfigFileFlag,
-	utils.APIFilterGetLogsMaxItemsFlag,
-	utils.APIFilterGetLogsDeadlineFlag,
-	utils.OpcodeComputationCostLimitFlag,
-	utils.SnapshotFlag,
-	utils.SnapshotCacheSizeFlag,
+	utils.ConfFlag,
```
kjhman21, line 110: Don't we need to call `altsrc.NewStringFlag(utils.ConfFlag)`?

Also I couldn't find ConfFlag.

cmd/utils/flags.go
odg0318: How about the below code? I think we must take care of a case when constant, `metricutils.MetricsEnabledFlag` is updated.

```suggestion
		EnvVar: "KLAYTN_ETRICUTILS_" + strings.ReplaceAll(strings.ReplaceAll(strings.ToUpper(metricutils.MetricsEnabledFlag), "-", "_"), ".", "_"),
```

odg0318:
@aidan-kwon @kjeom @kjhman21 
I suggest the below function to automatically generate envVar string from flag name. Even in this PR, we can find typo in envVar. 
* We can not only reduce total code lines but also expect all envVar have same rule.
* We don't touch any code even if some arguments are updated or added.

```golang
const (
       flagPrefixName = "KLAYTN"
       flagConfName   = "conf"
)

func generateEnvVar(v string) string {
        v = strings.ToUpper(v)
        v = strings.ReplaceAll(v, "-", "_")
        v = strings.ReplaceAll(v, ".", "_")
        v = flagPrefixName + "_" + v
        return v
}

func OnBefore(ctx *cli.Context) error {
        for _, v := range ctx.App.Flags {
                if v.GetName() == flagConfName {
                        continue
                }

                switch v.(type) {
                case *altsrc.StringFlag:
                        f := v.(*altsrc.StringFlag)
                        f.StringFlag.EnvVar = generateEnvVar(f.StringFlag.GetName())
                case *altsrc.BoolFlag:
                        f := v.(*altsrc.BoolFlag)
                        f.BoolFlag.EnvVar = generateEnvVar(f.BoolFlag.GetName())
                // All flag type must be implemented here.
                }
        }
}
```

cmd/utils/nodecmd/nodeflags.go
kjeom: `altsrc.NewStringFlag(utils.ConfFlag)` is for reading the flag in yaml file.
`--conf` flag option is a special case we have to put in command line like `./kcn --conf conf.yaml`

aidan-kwon:
> @aidan-kwon @kjeom @kjhman21
> 
> I suggest the below function to automatically generate envVar string from flag name. Even in this PR, we can find typo in envVar.

I agree on this suggestion. This auto-generation can reduce lots of human errors.

cmd/utils/flags.go
kjeom: Yes It's my mistake. Reflected, thanks

cmd/utils/nodecmd/defaultcmd.go
kjeom: Good, Thanks

cmd/utils/customflags.go
kjeom: It's my mistake. Thanks!

kjeom:
@odg0318 Thanks for the good suggestion.
@aidan-kwon Let me leave this issue for the next PR that include unit test for the flags.

kjhman21 (approved):

aidan-kwon (approved):

Issue #1579, opened 2022-08-08
Is there any method to remove pending subbridge txs?


egonspace:
We ran a service chain having subbridge with Baobab.

But every time a block is generated,  a following log is recorded.

```
ERROR[08/08,07:13:59 Z] [45] A bridge tx was not executed              err="nonce too low"              txHash=0x475a724407005e5bc9c424709a1e6a6718cc3593e8a4a07e42c107b3c8e6be0b txGasPrice=750000000000
ERROR[08/08,07:13:59 Z] [45] A bridge tx was not executed              err="nonce too low"              txHash=0x60f6b86f6b26cb6f9a5c5d0a1614fecf2ef8fc6c5bf36303e883d1cb0bf5af8e txGasPrice=750000000000
ERROR[08/08,07:13:59 Z] [45] A bridge tx was not executed              err="nonce too low"              txHash=0x5d09e1f89d5f1f2d69e1247f7c6a273a75dc7959606cc0a15358be8db65ef4e8 txGasPrice=750000000000
ERROR[08/08,07:13:59 Z] [45] A bridge tx was not executed              err="nonce too low"              txHash=0xad5daee7eb555f27d5f2d916334c2b28862e3019c9b6a1e0b83d0ad9e74b4e7f txGasPrice=750000000000
INFO[08/08,07:14:00 Z] [51] Commit new mining work                    number=345581 hash=13ab14…6583fc txs=0 elapsed=113.601µs commitTime=63.28µs  finalizeTime=45.31µs
```

So when I looked up the subbridge information, I could see that there are pending txs as follows.

```
txPending: {
    0xea3a0b918ff3cdd15c9b104e0f8c50d3b80dccde: [{
        from: "0xea3a0b918ff3cdd15c9b104e0f8c50d3b80dccde",
        gas: "0x186a0",
        gasPrice: "0xae9f7bcc00",
        hash: "0x475a724407005e5bc9c424709a1e6a6718cc3593e8a4a07e42c107b3c8e6be0b",
[...]

PR #1565, opened 2022-07-28
[SC] Value transfer rollback

hyunsooda:
## Proposed changes

Ideally, the value transfer from one chain to the other chain is initiated by end-users with enough balance to make a value transfer transaction. The current implementation, however, had been vulnerable to public usage. First of all, value transfer can be failed for a series of reasons of the other chain’s state, which results in the loss of the sender’s KLAY or tokens without return. Thus, we considered a rollback feature to refund for failed value transfers. Second, the value transfer recovery (VTR) was motivated to resend unhandled value transfer requests just in case of the absence of a connected bridge node. However, its implementation simply resends if a corresponding nonce is not marked (handled), including all unexecuted and failed cases. For the latter, the failed transactions must not be resent, reason its failure, and decide whether to refund or not. Eventually, this change implements value transfer rollback.

Changes are:
1. Added three database write and read operations for all the pair of request value transfer tx and event with handle value transfer tx and event, failed handle value transfer tx and event, refund event and tx. Also, read operations are available with APIs.
2. Implemented rollback. All the failed transactions are refunded. The rollback implementation is based on operator's message integrity and event integrity of the bridge contract.
3. Only unexecuted handle value transfer txs are resent by VTR.
4. As operators pay the value transfer services on each chain, the service fee to be paid to operators is reasonable. If the fee is costly, the service may not be attractive. The operator will make a loss if the fee is highly cheap or zero. This change also provides an API for the suggestion of the least fee to not have a deficit.
5. Operators can withdraw KLAY or tokens that a bridge contract holds by the value transfer request. The amount of KLAY or tokens is equally divided by each operator. Users can make a refund call if the reserved period is passed.
6. Refined the log of summarized transfer.
    - Before: `request=, handle=, lowerHandle=, pending=`
    - Now: `request=, handled=, numberOfFailed=, numberOfRefunds=, lowerHandle=, pending=`

## Types of changes

Please put an x in the boxes related to your change.

- [x] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
[...]

PR #1593, opened 2022-08-19
[Bridge handler] Added a reserved period for bridge tx resending

hyunsooda:
## Proposed changes

A value transfer direction from child to parent has a loop that repeatedly requests bridge transactions (value transfer and anchoring transactions) if the corresponding receipt does not arrive. Its duplicated transaction leads to pay unnecessary operator's gas costs. This PR added a reserved period (30s), where the resending should be triggered. Currently, the value transfer recovery implementation that has the wrong semantic can send one duplicated transaction. This duplicated sending is not associated to the bridge v2 implementation (#1565)

Issue:
- https://github.com/klaytn/klaytn/pull/1295#issuecomment-1177082342
- #1579 
- https://forum.klaytn.foundation/t/baobob-invalid-message/5813?u=jack_jin
## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


node/sc/sub_bridge_handler.go
```diff
@@ -383,26 +390,18 @@ func (sbh *SubBridgeHandler) handleParentChainInvalidTxResponseMsg(msg p2p.Msg)
 	if err := msg.Decode(&invalidTxs); err != nil && err != rlp.EOL {
 		return errResp(ErrDecode, "msg %v: %v", msg, err)
 	}
-	txPool := sbh.subbridge.GetBridgeTxPool()
 	for _, invalidTx := range invalidTxs {
-		if tx := txPool.Get(invalidTx.TxHash); tx != nil {
-			logger.Error("A bridge tx was not executed", "err", invalidTx.ErrStr,
+		if tx := sbh.subbridge.GetBridgeTxPool().Get(invalidTx.TxHash); tx != nil {
```
hqjang-pepper, line 394:
```suggestion
		if tx := txPool.Get(invalidTx.TxHash); tx != nil {
```

node/sc/sub_bridge_handler.go
```diff
@@ -383,26 +390,18 @@ func (sbh *SubBridgeHandler) handleParentChainInvalidTxResponseMsg(msg p2p.Msg)
 	if err := msg.Decode(&invalidTxs); err != nil && err != rlp.EOL {
 		return errResp(ErrDecode, "msg %v: %v", msg, err)
 	}
-	txPool := sbh.subbridge.GetBridgeTxPool()
 	for _, invalidTx := range invalidTxs {
```
hqjang-pepper, line 393:
```suggestion
        txPool := sbh.subbridge.GetBridgeTxPool()
	for _, invalidTx := range invalidTxs {
```
I think we don't have to get txPool newly when we loop everytime.

node/sc/sub_bridge_handler.go
hyunsooda: Thanks. changed.

node/sc/sub_bridge_handler.go
hyunsooda: Thanks. changed.

hyunsooda:
Fixed a logical error. Thank you, @2dvorak. PTAL again.

node/sc/sub_bridge_handler.go
```diff
@@ -600,3 +614,16 @@ func (sbh *SubBridgeHandler) WriteReceiptFromParentChain(blockHash common.Hash,
 func (sbh *SubBridgeHandler) GetReceiptFromParentChain(blockHash common.Hash) *types.Receipt {
 	return sbh.subbridge.chainDB.ReadReceiptFromParentChain(blockHash)
 }
+
+func (sbh *SubBridgeHandler) removeBridgeTx(tx *types.Transaction) {
+	if err := sbh.subbridge.GetBridgeTxPool().RemoveTx(tx); err != nil {
+		logger.Error("[SC][HandleResponse] Failed to remove bridge tx",
+			"txType", tx.Type(), "txNonce", tx.Nonce(), "txHash", tx.Hash().String())
```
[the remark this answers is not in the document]
hyunsooda: Resolved via another channel

2dvorak (commented):
Tested with the following steps:
1. Deploy parent and child chain
2. Deploy bridge & token contract and subscribe event
3. Stop CN
4. Request value transfer from child -> parent

Before this PR, I observed several "known transaction" error messages because the subbridge constantly resends the same tx until it gets the corresponding receipt. Due to inter-chain latency, those errors were produced even when I did not stopped CN in step 3.

After this PR, I could not see any "known transaction" error being produced, and even when I stopped the CN in step 3, I could observe only 1 "known transaction" error. That unhandled tx was handled in the next period of VT recovery (if enabled) ~~or next value transfer event.~~

Just one question - without VT recovery, if the CN is down, the value transfer event will be retried only once after resend interval. ~~If the CN is still down, that unhandled event will be retried when another value transfer request is made.~~ Is this intended?

**EDIT** Striked out some wrong comments

hyunsooda:
@2dvorak, Let me explain why you observed them. First, the only one-time resending is because of the `known transaction` by txpool policy. The `EN` already holds a transaction and if a duplicated transaction comes, that duplicated transaction is dropped (You can see from the `EN`s log) and sends an invalid tx message to the bridge node. The bridge node receives the invalid message and removes that transaction from bridge txpool. Thus, no resending is created from now on.

Could you recheck this one? `If the CN is still down, that unhandled event will be retried when another value transfer request is made`. The only new bridge transaction is sent if the previously unhandled transaction was removed.

2dvorak:
@hyunsooda OK, thank you for the explanation.

And I was wrong about `If the CN is still down, that unhandled event will be retried when another value transfer request is made`. I checked again and observed that the VT failed due to CN being down was handled after CN went alive, regardless of new VT request being made. Sorry for confusing you, I will edit my above comment.

2dvorak (approved):

hyunsooda:
@2dvorak, I slightly touched the resending logic. As the value transfer recovery (VTR) makes a new tx and sends it in an independent thread, I think we do not need to manage those txs for resending. However, the anchoring txs are not captured by the VTR and thus we should manage only anchoring txs for resending. The new change removes value transfer txs right after they are sent, but not anchoring txs.

hyunsooda:
pending

hyunsooda:
pending

PR #1594, opened 2022-08-19
Datadog integration

mjkim:
## Proposed changes

PR for datadog integration of RPC trace.

image

It is the same method as the existing newrelic integration, and it is controlled through environment variables.

If DD_TRACE_ENABLED is explicitly set to true, datadog trace works, and other environment variables supported by https://github.com/DataDog/dd-trace-go can be used as they are.

The additionally implemented DD_TRACE_HEADER_TAGS header is separated by , and the set header key and value are put into span.
(DD_TRACE_HEADER_TAGS is a header used in Python and Java implementations of datadog, but it is not implemented in Go, so it is additionally implemented.)

If DD_KLAYTN_RPC_RESPONSE is set to true, the response body is also added to the span in the form of a json string.
## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

The following packages are required for dependencies.
- gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http
- gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer


networks/rpc/http.go
```diff
@@ -263,6 +263,11 @@ func NewFastHTTPServer(cors []string, vhosts []string, timeouts HTTPTimeouts, sr
 		handler = newNewRelicHTTPHandler(nrApp, handler)
 	}
 
+	ddTracer := newDatadogTracer()
```
odg0318, line 266: It looks that `newDatadogTracer` must return `DatadogTracer` object because of the name. When I see your implementation, datadog sdk doesn't provide any struct or interface. And so I suggest the below codes to follow  the existing newrelic implementation.

```golang
ddTracer := newDatadogTracer()
if ddTracer != nil {
    handler = newDatadogHTTPHandler(ddTracer, handler)
}
```

```golang
type DatadogTracer struct {
    Tags                    []string
     KlaytnResponse bool
}

func newDatadogTracer() *DatadogTracer {
	ddTraceEnabled := os.Getenv("DD_TRACE_ENABLED")
	if strings.ToLower(ddTraceEnabled) != "true" {
		return nil
	}
        
        tags := strings.Split(os.Getenv("DD_TRACE_HEADER_TAGS"), ",")
	klaytnResponse := strings.ToLower(os.Getenv("DD_KLAYTN_RPC_RESPONSE")) == "true"
	tracer.Start()

	return &DatadogTracer{tags, klaytnResponse}
}

func newDatadogHTTPHandler(ddTracer *DatadogTracer, handler http.Handler) http.Handler {
...
}
```

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,130 @@
...
+		} else {
+			reqMethod = reqs[0].method
+			if isBatch {
+				reqMethod += "_batch"
+			}
+			encoded, _ := json.Marshal(reqs[0].params)
+			reqParam = string(encoded)
+		}
+
+		// new relic transaction name contains the first API method of the request
```
odg0318, line 53: Comment indicates newrelic but this code is related to datadog. Can you confirm this line?

mjkim: my mistake.. 😅

networks/rpc/http.go
mjkim: It seems your method is more clear.
It will be reflected in PR.

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,145 @@
...
+			reqMethod = reqs[0].method
+			if isBatch {
+				reqMethod += "_batch"
+			}
+			encoded, _ := json.Marshal(reqs[0].params)
+			reqParam = string(encoded)
+		}
+
+		// datadog transaction name contains the first API method of the request
+		resource := r.Method + " " + r.URL.String() + " " + reqMethod
```
odg0318, line 63: `fmt.Sprintf` looks better here.
```suggestion
		resource := fmt.Sprintf("%s %s %s", r.Method, r.URL.String(), reqMethod)
```

odg0318:
@aidan-kwon Can you review this PR? Good to go for me.

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,145 @@
...
+	ddTraceEnabled := os.Getenv("DD_TRACE_ENABLED")
+	if strings.ToLower(ddTraceEnabled) != "true" {
+		return nil
+	}
+
+	tracer.Start()
+
+	tags := strings.Split(os.Getenv("DD_TRACE_HEADER_TAGS"), ",")
+	service := os.Getenv("DD_SERVICE")
+	klaytnResponse := strings.ToLower(os.Getenv("DD_KLAYTN_RPC_RESPONSE")) == "true"
```
aidan-kwon, line 31: Just out of curiosity. Is there any reason to use "true" instead of 1?

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,145 @@
...
+			ResponseWriter: w,
+			body:           bytes.NewBufferString(""),
+		}
+
+		spanOpts := []ddtrace.StartSpanOption{
+			tracer.Tag("request.method", reqMethod),
+			tracer.Tag("request.params", reqParam),
+		}
+
+		for _, header := range ddTracer.Tags {
```
aidan-kwon, line 76: It seems that `ddTracer.Tags` has always the same value during the process lifecycle since the value is parsed from `DD_TRACE_HEADER_TAGS` os env.. Am I right?

networks/rpc/http_datadog.go
mjkim: All other datadog related environment variables have true/false values.
So I used true/false to go similar.

ex)
https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/option.go#L236 
https://ddtrace.readthedocs.io/en/stable/configuration.html

networks/rpc/http_datadog.go
mjkim: Yes. `ddTracer.Tags` has always the same value during the process lifecycle.

networks/rpc/http_datadog.go
mjkim: Thanks. changed.

networks/rpc/http_datadog.go
aidan-kwon: Then, we better move this logic to `newDatadogTracer`. This handler function will be executed on each API call.

networks/rpc/http_datadog.go
aidan-kwon: It seems that your reference is using a boolean value, but "true" is a string.

networks/rpc/http_datadog.go
mjkim: ddTracer.Tags has always same value. But, this is only used as the key of the tag, and the value is set differently for each request.

```golang
if header == "remote_addr" {
	tag = tracer.Tag(key, r.RemoteAddr)
} else {
	tag = tracer.Tag(key, r.Header.Get(header))
}
```

The value of tag is different for every request, depending on the headers value.

networks/rpc/http_datadog.go
mjkim: Oh. I know what you're saying. 
I Changed to use parseBool instead of string compare. Would this be okay?

aidan-kwon: Thanks!

networks/rpc/http_datadog.go
aidan-kwon: I agree. However, the following code can be moved
```
			header = strings.TrimSpace(header)
			header = strings.ToLower(header)
			key := fmt.Sprintf("http.%s", header)
```

odg0318: I agree with @aidan-kwon opinion. This loop is executed every request even though it returns the same result like constants. The lines can be moved in `newDatadogTracer` function.

mjkim: @aidan-kwon @odg0318 good point. I hadn't thought about that part. 
I'll change it.

odg0318:
@aidan-kwon It looks like that this PR is ready to be merged now. Can we go?

odg0318:
Addition to this implementation, I think that documentation is required for this feature. But we have no `/docs` directory yet in `main` branch so we need to touch it in another topic.

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,167 @@
...
+
+func newDatadogTracer() *DatadogTracer {
+	var err error
+	ddTraceEnabled := false
+	if v := os.Getenv("DD_TRACE_ENABLED"); v != "" {
+		ddTraceEnabled, err = strconv.ParseBool(v)
+		if err != nil || ddTraceEnabled == false {
+			return nil
+		}
+	}
```
aidan-kwon, line 35: Refining
```suggestion
	if v := os.Getenv("DD_TRACE_ENABLED"); v != "" {
		ddTraceEnabled, err := strconv.ParseBool(v)
		if err != nil || ddTraceEnabled == false {
			return nil
		}
	}
```

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,167 @@
...
+		key := fmt.Sprintf("http.%s", header)
+
+		tags[i].header = header
+		tags[i].key = key
+	}
+	service := os.Getenv("DD_SERVICE")
+
+	klaytnResponse := false
+	if v := os.Getenv("DD_KLAYTN_RPC_RESPONSE"); v != "" {
+		klaytnResponse, err = strconv.ParseBool(v)
```
aidan-kwon, line 51: Refining
```suggestion
	if v := os.Getenv("DD_KLAYTN_RPC_RESPONSE"); v != "" {
		klaytnResponse, err := strconv.ParseBool(v)
```

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,167 @@
...
+				logger.ErrorWithStack("Datadog http handler panic", "err", err)
+			}
+		}()
+
+		reqMethod := ""
+		reqParam := ""
+
+		// parse RPC requests
+		reqs, isBatch, err := getRPCRequests(r)
+
```
aidan-kwon, line 75:
```suggestion
```

aidan-kwon (approved):

aidan-kwon:
@odg0318 I think this PR is good to go. For the documentation, you can make an issue or PR in https://github.com/klaytn/klaytn-docs

networks/rpc/http_datadog.go
kjeom: suggest trivial thing : can remove err variable.

networks/rpc/http_datadog.go
```diff
@@ -0,0 +1,167 @@
...
+		ddTraceEnabled, err = strconv.ParseBool(v)
+		if err != nil || ddTraceEnabled == false {
+			return nil
+		}
+	}
+
+	headers := strings.Split(os.Getenv("DD_TRACE_HEADER_TAGS"), ",")
+	tags := make([]tagInfo, len(headers))
+	for i, header := range headers {
+		header = strings.TrimSpace(header)
```
kjeom, line 40: I think It's better to trim by comma(,) because of consistency with others.

kjeom: Ahh, got it thanks

odg0318 (approved):

aidan-kwon (dismissed):

kjhman21 (approved):

kjeom (approved):

kjeom:
I leave a comment to discuss later.
Datadog option is set by environment variables. I think we need a config structure like DataDogConfig and put it in the [NodeConfig](https://github.com/klaytn/klaytn/blob/dev/node/config.go#L54) to manage the config.
I'm analyzing and testing the klaytn configuration(yaml, toml, EnvVar, ...) now so I'd like to discuss in detail about this after it's done.

mjkim:
Sorry, I added a commit to apply gofumpt, but approved was removed.
Can you do it one more time?
@odg0318 @aidan-kwon @kjeom @kjhman21

odg0318 (approved):

aidan-kwon (approved):

kjhman21 (approved):

Issue #1206, opened 2022-02-22
A CN of the private network got panic after addValidator


aidan-kwon:
**Issue**
- #1600 

A CN of a v1.7.3 private network got panic. The network consisted of 3 CN nodes and another CN had been added as a validator. The newly added CN got panic and others generate blocks without the panic issue. 
After the CN restart, the panic still occurs at the same block number.
After the binary change to v1.7.2, the still occurs at the same block number.
After rewinding a few blocks with the `start-block-num` flag, it doesn't show the panic log and synchronization started. 

- The private network was using `RoundRobin` proposerPolicy instead of `WeightedRandom`.

```
INFO[02/22,02:40:20 Z] [5] Inserted a new block                      number=25918 hash=d595db…8d6913                                                      txs=0 gas=0 elapsed=1.997ms    processTxs=78ns  finalize=72.308µs  validateState=2.604µs totalWrite=1.84381ms  trieWrite=1.770379ms
panic: sync/atomic: store of nil value into Value

goroutine 2500 [running]:
sync/atomic.(*Value).Store(0xc01003e328, 0x0, 0x0)
	/usr/local/go/src/sync/atomic/value.go:47 +0x12c
github.com/klaytn/klaytn/consensus/istanbul/validator.(*defaultSet).Copy(0xc00fdf3f80, 0x0, 0x0)
	/go/src/github.com/klaytn/klaytn/kcn-linux-x86_64/rpmbuild/BUILD/kcnd-v1.7.3/consensus/istanbul/validator/default.go:362 +0x259
github.com/klaytn/klaytn/consensus/istanbul/backend.(*backend).verifyCommittedSeals(0xc00016e000, 0x1f9ba60, 0xc000203900, 0xc010036900, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/klaytn/klaytn/kcn-linux-x86_64/rpmbuild/BUILD/kcnd-v1.7.3/consensus/istanbul/backend/engine.go:307 +0x15c
github.com/klaytn/klaytn/consensus/istanbul/backend.(*backend).verifyCascadingFields(0xc00016e000, 0x1f9ba60, 0xc000203900, 0xc010036900, 0x0, 0x0, 0x0, 0xc010040000, 0x1)
	/go/src/github.com/klaytn/klaytn/kcn-linux-x86_64/rpmbuild/BUILD/kcnd-v1.7.3/consensus/istanbul/backend/engine.go:233 +0x2a6
[...]

PR #1600, opened 2022-08-23
[CN] Panic by nil access and nil store

hyunsooda:
### 1. Found a panic from nil access error as shown below.
```
DEBUG[08/22,15:58:26 +09] [28] Recalculated downloader QoS values        rtt=12.125s       confidence=0.618 ttl=58.812687048s
ERROR[08/22,15:58:27 +09] [24] Snapshot not found.                       err="unknown ancestor"
ERROR[08/22,15:58:27 +09] [26] Proposer is nil                           validators=[]
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x13f4d3e]

goroutine 58945 [running]:
github.com/klaytn/klaytn/consensus/istanbul/validator.(*defaultSet).Copy(0xc03a080840)
    klaytn/consensus/istanbul/validator/default.go:361 +0x11e
github.com/klaytn/klaytn/consensus/istanbul/core.(*core).handlePreprepare(0xc0002126e0, 0xc019f477a0, {0x1d40ab8?, 0xc027555de8})
    klaytn/consensus/istanbul/core/preprepare.go:73 +0x370
github.com/klaytn/klaytn/consensus/istanbul/core.(*core).handleCheckedMsg(0xc0002126e0, 0xc019f477a0, {0x1d40ab8?, 0xc027555de8})
    klaytn/consensus/istanbul/core/handler.go:202 +0x174
github.com/klaytn/klaytn/consensus/istanbul/core.(*core).handleMsg(0xc0002126e0, {0xc01a31a300, 0x2fd, 0x2fd})
    klaytn/consensus/istanbul/core/handler.go:185 +0x5e8
github.com/klaytn/klaytn/consensus/istanbul/core.(*core).handleEvents(0xc0002126e0)
    klaytn/consensus/istanbul/core/handler.go:103 +0x331
created by github.com/klaytn/klaytn/consensus/istanbul/core.(*core).Start
    klaytn/consensus/istanbul/core/handler.go:36 +0x6c
```

I'm not sure how it can be reproduced. As I correctly remember, I killed one `CN`(`SCN`) node. At that time, that node might be selected as a proposer probably. This PR logically fixes the absence of nil condition. Please feel free to add, remove, or modify if you are familiar with this package.

------------------------------------------------------------------ Added ------------------------------------------------------------------

### 2. Panic by nil store into atomic storage

This panic is created by putting nil into the underlying atomic object. (Known issues: #1206, #1612)
The panic path is initiated from fetcher and faces panic by a wrong caller of the atomically stored proposer. Also, this panic path is associated with only the round-robin policy.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


consensus/istanbul/validator/default.go
```diff
@@ -279,7 +279,7 @@ func (valSet *defaultSet) CalcProposer(lastProposer common.Address, round uint64
 	defer valSet.validatorMu.RUnlock()
 
 	if len(valSet.validators) == 0 {
-		logger.Error("len of validators is 0, Proposer is nil", "validators", valSet.validators)
+		logger.Error("List of validators is empty", "validators", len(valSet.validators))
```
henry-will, line 282:
```suggestion
		logger.Error("List of validators is empty, Proposer is nil")
```
I guess that this error print like "List of validators is empty validators 0". What about my suggestion?

hyunsooda: Just logging `Validiator list is empty` is not bad, but I thought showing the actual value would not be bad as well. As the validator list is zero, the suggested `Proposer is nil` makes sense. However, the logging of nil proposer is already located in other places. Either one is okay for me. The code owner may add a further suggestion.

consensus/istanbul/validator/default.go
```diff
@@ -358,8 +358,12 @@ func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
 	}
 
 	newValSet := NewSubSet(addresses, valSet.policy, valSet.subSize).(*defaultSet)
-	_, proposer := newValSet.GetByAddress(valSet.GetProposer().Address())
-	newValSet.proposer.Store(proposer)
+	if proposer := newValSet.GetProposer(); proposer != nil {
+		_, proposer := newValSet.GetByAddress(proposer.Address())
+		newValSet.proposer.Store(proposer)
```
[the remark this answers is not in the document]
hyunsooda: Already reported some times. Please refer to the [description](https://github.com/klaytn/klaytn/pull/1600#issue-1347472306). Also, you can reproduce the [same panic](https://github.com/klaytn/klaytn/issues/1206).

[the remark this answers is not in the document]
henry-will: In other words, the proposer on 361 line is not nil, but the next assigned proposer on 362 can be nil, it can cause panic on 363 line. So I commented on checking the second assigned proposer for nil one more time.

consensus/istanbul/validator/default.go
henry-will: I agree your opinion, too. How about add just a comma in "List of validators is empty validators 0" like "List of validators is empty , validators 0"?

consensus/istanbul/validator/default.go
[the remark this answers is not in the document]
hyunsooda: Oops, I'm sorry. I misinterpreted your comment. A necessary condition to be nil is that a snapshot should set wrong validators. This is impossible because the parent hash of the fetched block differs if the `extraData` differs and does not contribute to updating the snapshot.

consensus/istanbul/validator/default.go
hyunsooda: If the log does not contain the concrete value printing, simply `Validator list is empty` seems an appropriate candidate I think.

consensus/istanbul/validator/default.go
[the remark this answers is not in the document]
henry-will: Thank you for your kind explanation. I will take a closer look at the code referring your explanation.

consensus/istanbul/validator/default.go
henry-will: It's good to me, also.

consensus/istanbul/validator/default.go
```diff
@@ -358,8 +358,12 @@ func (valSet *defaultSet) Copy() istanbul.ValidatorSet {
...
+	} else {
+		logger.Error("Found nil proposer", "validators", len(valSet.validators))
+	}
```
yoomee1313, line 366: @hyunsooda In the known issues(https://github.com/klaytn/klaytn/issues/1206, https://github.com/klaytn/klaytn/issues/1612), The panic is caused because `NewValSet` does not contain original valSet's proposer (`valSet.proposer`). 

So, What about next code?
```
	idx, proposer := newValSet.GetByAddress(valSet.GetProposer().Address())
	if idx != -1 {
		newValSet.proposer.Store(proposer)
	}
	// The original proposer is removed but new proposer is not calculated
	// because it's syncing. Set default proposer as a proposer.
```
cc. I think setting random proposer as a newValSet.proposer works well because if the node enters mining mode, it calculates the original proposer anyway.. @jiseongnoh

hyunsooda: Your suggestion may still cause panic because `GetProposer()` may return nil and the code suggested does not have a guard whether it is nil or not.

yoomee1313: @hyunsooda Ok, then i think adding valSet.proposer nil checking logic. Was this what you pointed out? 
```
	if proposer := valSet.GetProposer(); proposer != nil {
		if idx, _ := newValSet.GetByAddress(proposer.Address()); idx != -1 {
			newValSet.proposer.Store(proposer)
		}
	}

	// The original proposer is removed but new proposer is not calculated
	// because it's syncing. Set default proposer as a proposer.
```

yoomee1313: btw, can i ask why you check,set,and logerror the newValSet.proposer here? because it is already set at NewSubSet.

hyunsooda: The originally proposed changes in `backend` package were rollbacked since those checks are unnecessary.
 
The panic by snapshot error was hard to reproduce because its context is too less. I think it’s better to wait until a new report is issued.

The panic is not observable now, but let me note one thing here that maybe you already know. The released validators cannot request block sync to current validators. To do that, (1) running `en` binary works once the existing chaindata migration is done. Also, (2) the `cn` binary can still sync with `en` process via peer connection without chaindata migration.

I found a vulnerability and posted its description. [You can see here](https://github.com/klaytn/klaytn/issues/1690).

yoomee1313: @hyunsooda Could you add `TODO` about this context?

mckim19 (approved):

yoomee1313 (approved):

Issue #1431, opened 2022-06-10
Develop a better way to load last state


jeongkyun-oh:
**Problem description**
Loading a last state may be failed when a last block hash is stored, but not block rlp itself. Currently, it resets the last state to genesis, but it can be restored in a better way (e.g. finding existing block while decreasing block number like `setHead` method). We may need to find a better way to resolve this problem.

```go
	// Restore the last known head block
	head := bc.db.ReadHeadBlockHash()
	if head == (common.Hash{}) {
		// Corrupt or empty database, init from scratch
		logger.Info("Empty database, resetting chain")
		return bc.Reset()
	}
	// Make sure the entire head block is available
	currentBlock := bc.GetBlockByHash(head)
	if currentBlock == nil {
		// Corrupt or empty database, init from scratch
		logger.Error("Head block missing, resetting chain", "hash", head.String())
		return bc.Reset()
	}
	// Everything seems to be fine, set as the head block
	bc.currentBlock.Store(currentBlock)
	bc.lastCommittedBlock = currentBlock.NumberU64()
```

cc @aidan-kwon 

relates to #1630

PR #1630, opened 2022-09-29
backup & load last state #1431

ethan-kr:
## Proposed changes

- This PR solves this issue https://github.com/klaytn/klaytn/issues/1431
- Resolves an issue where data sync does not match because header hash storage and misc storage are different. While writing the header hash, the header backup hash before 64blocks is recorded once more. If it fails to read the header hash, it reads the header backup hash and proceeds with the process. ( rollback to 128 blocks )

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


hqjang-pepper (dismissed):

storage/database/db_manager.go
```diff
@@ -42,6 +42,9 @@ var (
 	logger = log.NewModuleLogger(log.StorageDatabase)
 
 	errGovIdxAlreadyExist = errors.New("a governance idx of the more recent or the same block exist")
+
+	backupHashes [128]common.Hash
```
ian0371, line 46: In the PR description, backup stores the hash before 64 blocks, and yet this size is 128.
Doesn't this mean that when storing hash of block N, backup hash stores block N-128?

Also, is block N-128 guaranteed (or probably guaranteed) to have block contents? I wonder if the number has a rationale behind.

storage/database/db_manager.go
```diff
@@ -934,12 +939,31 @@ func (dbm *databaseManager) ReadHeadBlockHash() common.Hash {
...
+	data, _ := db.Get(headBlockBackupKey)
+	if len(data) == 0 {
+		return common.Hash{}
+	}
+	return common.BytesToHash(data)
+}
+
 // WriteHeadBlockHash stores the head block's hash.
 func (dbm *databaseManager) WriteHeadBlockHash(hash common.Hash) {
+	backupHashes[idx%128] = hash
```
jeongkyun-oh, line 954: It would be better to add a constant instead of a magic number 128.

storage/database/db_manager.go
```diff
@@ -934,12 +939,31 @@ func (dbm *databaseManager) ReadHeadBlockHash() common.Hash {
 	return common.BytesToHash(data)
 }
 
+// Block Backup Hash operations.
+func (dbm *databaseManager) ReadHeadBlockBackupHash() common.Hash {
```
jeongkyun-oh, line 943: How about combining this method with `ReadHeadBlockHash()`? So, we don't need to update the upper method.

governance/default.go
```diff
@@ -713,6 +713,10 @@ func (g *Governance) initializeCache() error {
 		if num := g.db.ReadHeaderNumber(headBlockHash); num != nil {
 			headBlockNumber = *num
 		}
+	} else if headBlockHash := g.db.ReadHeadBlockBackupHash(); !common.EmptyHash(headBlockHash) {
+		if num := g.db.ReadHeaderNumber(headBlockHash); num != nil {
+			headBlockNumber = *num
+		}
```
jeongkyun-oh, line 719: What if `HeadBlockHash` exist, but the associated block doesn't?

storage/database/db_manager.go
ethan-kr: My initial intention was 128 blocks as ian said. So I made it into an array of 128. However, since there are two places to store the headhash (WriteHeadBlockHash, WriteHeadFastBlockHash), the result is N-64 blocks

blockchain/headerchain.go
```diff
@@ -87,6 +88,10 @@ func NewHeaderChain(chainDB database.DBManager, config *params.ChainConfig, engi
 		if chead := hc.GetHeaderByHash(head); chead != nil {
 			hc.currentHeader.Store(chead)
 		}
+	} else if head := chainDB.ReadHeadBlockBackupHash(); head != (common.Hash{}) {
+		if chead := hc.GetHeaderByHash(head); chead != nil {
+			hc.currentHeader.Store(chead)
+		}
```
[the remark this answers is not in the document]
ethan-kr: you are right

governance/default.go
ethan-kr: you are right, too

storage/database/db_manager.go
ian0371: Oh, I see. I think it's quite hard to understand for readers.
What do you think about introducing a new object for backup hash?
I see it as a variant of _queue_ with unique IDs that can be only popped when it is full.
I think it's more readable this way.

```go
const backupHashCnt 64

struct backupHash {
    backupHashes [backupHashCnt]common.Hash
    idx int
}

func (b *backupHash) push(h common.Hash) { /* skip if duplicate; push to backupHashes */ }
func (b *backupHash) pop() common.Hash { /* if not full, return empty hash; otherwise, return N-64 */ }
```

storage/database/db_manager.go
ethan-kr: Changes to constant value

storage/database/db_manager.go
ethan-kr: Yes. Changes to queue structure

storage/database/db_manager.go
```diff
@@ -302,13 +307,31 @@ const (
 	databaseEntryTypeSize
 )
 
+type backupHash struct {
```
jeongkyun-oh, line 310: How about `backupHashList` or `backupHashQueue`?

jeongkyun-oh (approved):
LGTM except for minor

mckim19 (approved):
LGTM :)

jeongkyun-oh (approved):

PR #1704, opened 2022-11-16
Add block number parameter to governance.chainConfigAt() API

ian0371:
## Proposed changes

This PR
- introduces `governance.chainConfigAt(num)`
- changes the semantic of `governance.chainConfig` to return the latest chainConfig
- At num=0, ReadGovernance uses DefaultEpoch because istanbul.epoch might not exist at program startup
- Remove unused methods
- closes #1615

Before:
```js
> governance.chainConfig
{
  chainId: 1001,
  deriveShaImpl: 2,
  governance: {
    governanceMode: "single",
...
}
```

After:
```js
> governance.chainConfig // returns the latest
{
  chainId: 1001,
  deriveShaImpl: 0,
  governance: {
    governanceMode: "single",
...
}

> governance.chainConfig() // latest block
{
  chainId: 1001,
  deriveShaImpl: 0,
  governance: {
    governanceMode: "single",
...
}

> governance.chainConfig(0) // specified block
{
  chainId: 1001,
  deriveShaImpl: 2,
  governance: {
    governanceMode: "single",
...
}
```

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [ ] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

#1615

PR #1713, opened 2022-11-21
[API] `ItemsAt` returns error on future block number

hyunsooda:
## Proposed changes

As-Is: Returns the latest gov params
To-Be: Returns an error for the future block number

Considering the programming in service or API query, a definite error can teach the invalid input. For instance, the current block number is 100 and the user input number is 10000. Definitely, the user does not have the intention to query for the latest gov params from the current block 100. 

cc. @blukat29 

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/governance/api.go
+++ b/governance/api.go
@@ -18,6 +18,7 @@ package governance
 
 import (
 	"errors"
+	"fmt"
 	"math/big"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


blukat29:
@hyunsooda
`itemsAt` and `chainConfigAt` (#1704) both are able to calculate parameters at any given future block.
```diff
- As-Is: Returns the latest gov params
+ As-Is: Returns the expected gov params at given future block number
```

It is a useful feature to query parameters at future block numbers. To predict upcoming changes, to check if the votes are correctly processed, etc.

hyunsooda:
How the future governance can be predictable? Given cur blk=100, epoch=10, and input=1000, the block between 101 and 999 are unknown and future blocks that may have voting or not. It seems unpredictable.  Am I missing something?

blukat29:
> How the future governance can be predictable? Given cur blk=100, epoch=10, and input=1000, the block between 101 and 999 are unknown and future blocks that may have voting or not. It seems unpredictable. Am I missing something?

@hyunsooda You are not missing anything.
The future parameters are not final. It is a prediction based on current knowledge. The result can change by future votes.
`itemsAt` has been working like that until v1.9. `itemsAt` and `chainConfigAt` will be the same in v1.10.

But because it shows some prediction, network operators can see if the votes were successfully applied. If it only shows past and current parameters, there is no way to check if the vote has worked until the chain reaches the parameter-changing block.

I think the benefit of seeing uncertain future data outweighs the confusion of witnessing the change in future data.

hyunsooda:
The prediction seems misleading. To say a prediction, we should have confidence in the given ground knowledge (Typically requires mathematical confidence). However, the future blocks are unknown if they contain vote or not. Thus, prediction seems misleading word and may cause confusion to the developers. Returning simply an error that says "future block error" seems more reliable. Nop?

hyunsooda:
Converted as a draft for now.

blukat29:
Then let's find a way to explicitly tell API users that the `itemsAt(furtureBlock)` result is tentative data.

PR #1725, opened 2022-11-24
Adjust governance related logs

blukat29:
## Proposed changes

- Tone down to DEBUG "Could not find GovParam contract address" as it commonly happens.
  (When GovParamContract is not specified after Kore block number)
- Add INFO logs about notable governance header-related actions. 

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Further comments

### DIFF
--- a/governance/contract.go
+++ b/governance/contract.go
@@ -121,7 +121,7 @@ func (e *ContractEngine) contractAddrAt(num uint64) (common.Address, error) {
 	// this happens when GovParamContract has not been voted
 	param, ok := headerParams.Get(params.GovParamContract)
 	if !ok {
-		logger.Error("Could not find GovParam contract address")
+		logger.Debug("Could not find GovParam contract address")
 		return common.Address{}, nil
 	}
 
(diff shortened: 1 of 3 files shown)


yoomee1313 (approved):

governance/contract.go
```diff
@@ -121,7 +121,7 @@ func (e *ContractEngine) contractAddrAt(num uint64) (common.Address, error) {
 	// this happens when GovParamContract has not been voted
 	param, ok := headerParams.Get(params.GovParamContract)
 	if !ok {
-		logger.Error("Could not find GovParam contract address")
+		logger.Debug("Could not find GovParam contract address")
```
jiseongnoh, line 124: how about logger.Info? I believe that this kind of information is shown at the Info level.

blukat29: The log is printed when (1) the Kore block has passed AND (2) "governance.govparamcontract" is empty, which is a very ordinary situation. This log is printed about 5 times per block. It will quickly clog the log file.

In contrast, an Info log is printed if the AddressBook is not ready (https://github.com/klaytn/klaytn/blob/dev/reward/addressbook_connector.go#L205).
It is okay to use Info-level here because the log is only printed every 86400 blocks.

jiseongnoh: I see. My concern is when I run some nodes privately for testing purposes, I imagine just forgetting to add the contract address and I do not have any clue whether I did deploy it or not.

blukat29: In that case, you can verify the output of `governance.itemsAt` along with other test preconditions.

blukat29: If the "governance.govparamcontract" points to a broken contract address, other Warn or Error logs will be printed in the contractCaller.

jiseongnoh: Thank you for your detailed explanation!

PR #1733, opened 2022-12-02
tx_list: bump tx max size to 128KB

yoomee1313:
## Proposed changes
* it increases tx max size to 128KB
* slot is introduced. For example, if a slot size is 32KB, and the tx size is 96kB, when that tx is inserted into txpool, it adds 96+32-1/32 = 3 slots.
  * dev: 1 is added to the txpool tx count when a tx is inserted
  * PR: tx slot size is added to the txpool tx count(which is actually slots) when a tx is inserted. See numSlots.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/tx_pool.go
+++ b/blockchain/tx_pool.go
@@ -43,12 +43,23 @@ import (
 const (
 	// chainHeadChanSize is the size of channel listening to ChainHeadEvent.
 	chainHeadChanSize = 10
+
+	// txSlotSize is used to calculate how many data slots a single transaction
+	// takes up based on its size. The slots are used as DoS protection, ensuring
+	// that validating a new transaction remains a constant operation (in reality
+	// O(maxslots), where max slots are 4 currently).
+	txSlotSize = 32 * 1024
+
+	// MaxTxDataSize is the maximum size a single transaction can have. This field has
+	// non-trivial consequences: larger transactions are significantly harder and
+	// more expensive to propagate; larger transactions also take more resources
+	// to validate whether they fit into the pool or not.
+	MaxTxDataSize = 4 * txSlotSize // 128KB
+
 	// demoteUnexecutablesFullValidationTxLimit is the number of txs will be fully validated in demoteUnexecutables.
 	demoteUnexecutablesFullValidationTxLimit = 1000
 	// txMsgCh is the number of list of transactions can be queued.
 	txMsgChSize = 100
-	// MaxTxDataSize is a heuristic limit of tx data size, and txPool rejects transactions over 32KB to prevent DOS attacks.
-	MaxTxDataSize = 32 * 1024
 )
 
 var (
@@ -77,6 +88,7 @@ var (
 	invalidTxCounter     = metrics.NewRegisteredCounter("txpool/invalid", nil)
 	underpricedTxCounter = metrics.NewRegisteredCounter("txpool/underpriced", nil)
 	refusedTxCounter     = metrics.NewRegisteredCounter("txpool/refuse", nil)
+	slotsGauge           = metrics.NewRegisteredGauge("txpool/slots", nil)
 )
 
 // TxStatus is the current status of a transaction as seen by the pool.
@@ -870,7 +882,7 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
 	// (2) remove an old Tx with the largest nonce from queue to make a room for a new Tx with missing nonce
 	// (3) discard a new Tx if the new Tx does not have a missing nonce
 	// (4) discard underpriced transactions
-	if uint64(pool.all.Count()) >= pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll {
+	if uint64(pool.all.Slots()+numSlots(tx)) >= pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll {
 		// (1) discard a new Tx if there is no room for the account of the Tx
 		from, _ := types.Sender(pool.signer, tx)
 		if pool.queue[from] == nil {
@@ -899,7 +911,7 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
 			return false, ErrUnderpriced
 		}
 		// New transaction is better than our worse ones, make room for it
-		drop := pool.priced.Discard(pool.all.Count()-int(pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll-1), pool.locals)
+		drop := pool.priced.Discard(pool.all.Slots()-int(pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll)+numSlots(tx), pool.locals)
 		for _, tx := range drop {
 			logger.Trace("Discarding freshly underpriced transaction", "hash", tx.Hash(), "price", tx.GasPrice())
 			underpricedTxCounter.Inc(1)
@@ -1717,8 +1729,9 @@ func (as *accountSet) add(addr common.Address) {
 // peeking into the pool in TxPool.Get without having to acquire the widely scoped
 // TxPool.mu mutex.
 type txLookup struct {
-	all  map[common.Hash]*types.Transaction
-	lock sync.RWMutex
+	all   map[common.Hash]*types.Transaction
+	slots int
+	lock  sync.RWMutex
 }
 
 // newTxLookup returns a new txLookup structure.
@@ -1726,6 +1739,14 @@ func newTxLookup() *txLookup {
 	return &txLookup{
 		all: make(map[common.Hash]*types.Transaction),
 	}
+}
+
+// Slots returns the current number of slots used in the lookup.
+func (t *txLookup) Slots() int {
+	t.lock.RLock()
+	defer t.lock.RUnlock()
+
+	return t.slots
 }
 
 // Range calls f on each key and value present in the map.
@@ -1761,6 +1782,9 @@ func (t *txLookup) Add(tx *types.Transaction) {
 	t.lock.Lock()
 	defer t.lock.Unlock()
 
+	t.slots += numSlots(tx)
+	slotsGauge.Update(int64(t.slots))
+
 	t.all[tx.Hash()] = tx
 }
 
@@ -1769,5 +1793,13 @@ func (t *txLookup) Remove(hash common.Hash) {
 	t.lock.Lock()
 	defer t.lock.Unlock()
 
+	t.slots -= numSlots(t.all[hash])
+	slotsGauge.Update(int64(t.slots))
+
 	delete(t.all, hash)
+}
+
+// numSlots calculates the number of slots needed for a single transaction.
+func numSlots(tx *types.Transaction) int {
+	return int((tx.Size() + txSlotSize - 1) / txSlotSize)
 }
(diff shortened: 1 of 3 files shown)


kjhman21:
@yoomee1313 Change is too big to be included after the code freeze. Let's talk about this in our internal meeting.

yoomee1313:
This PR will be included in v1.10.1 or v1.11.0.

blockchain/tx_pool.go
```diff
@@ -43,12 +43,23 @@ import (
...
+	// more expensive to propagate; larger transactions also take more resources
+	// to validate whether they fit into the pool or not.
+	MaxTxDataSize = 4 * txSlotSize // 128KB
+
 	// demoteUnexecutablesFullValidationTxLimit is the number of txs will be fully validated in demoteUnexecutables.
 	demoteUnexecutablesFullValidationTxLimit = 1000
 	// txMsgCh is the number of list of transactions can be queued.
 	txMsgChSize = 100
-	// MaxTxDataSize is a heuristic limit of tx data size, and txPool rejects transactions over 32KB to prevent DOS attacks.
-	MaxTxDataSize = 32 * 1024
```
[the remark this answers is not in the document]
yoomee1313: @hyunsooda If we limit the tx_pool size by the slot, not the number of txs, we can keep tx_pool memory size same even we change the the max size of the transaction. I think it's the benefit of bringing the slot concept.

blockchain/tx_pool.go
```diff
@@ -870,13 +883,13 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
 	// (2) remove an old Tx with the largest nonce from queue to make a room for a new Tx with missing nonce
 	// (3) discard a new Tx if the new Tx does not have a missing nonce
 	// (4) discard underpriced transactions
-	if uint64(len(pool.all)) >= pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll {
+	if uint64(pool.all.Count()) >= pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll {
```
hyunsooda, line 886: [Based on your slot concept](https://github.com/klaytn/klaytn/pull/1733#discussion_r1044050420), the `pool.all.Count()` should be changed to `pool.all.Slots() + num Slots(tx)` ? this is because the `pool.all.Count()` does not consider the amount of total tx size.

yoomee1313: You're right, i have updated to add Slots size. By the way, this PR is too complicated, so i divided into two PRs. Could you review the next PR first? https://github.com/klaytn/klaytn/pull/1748

yoomee1313:
This PR is too complicated.. I divided it to two PRs. Please review next PR first. https://github.com/klaytn/klaytn/pull/1748

blockchain/tx_pool.go
kjeom: @yoomee1313 https://github.com/klaytn/klaytn/pull/1733#issuecomment-1367205381 You mean the #1748 need to be reviewed first?
I think It seems to be confused "Please review this PR first."

yoomee1313: @kjeom Oh, sorry. It is right that #1748 need to be reviewed first

blockchain/tx_pool.go
```diff
@@ -1761,6 +1782,9 @@ func (t *txLookup) Add(tx *types.Transaction) {
 	t.lock.Lock()
 	defer t.lock.Unlock()
 
+	t.slots += numSlots(tx)
+	slotsGauge.Update(int64(t.slots))
```
blukat29, line 1786: Can we update `slotsGauge` elsewhere? Preferrably after `pool.all.Add()` and `pool.all.Remove()`.
- I see that there is only one `txLookup` instance under normal circumstances (since `TxPool.all == TxPool.priced.all`). But updating a global counter inside a container's method seems odd to me.
- Other counters such as `pendingReplaceCounter` are updated in `TxPool` methods.

yoomee1313: If the update is done in `Txpool` methods, it might done in too many places and i think it could leads to a mistake such as omitting if the code changes. Even though it's a global counter, its scope seems restricted to the `txLookup.slots`. In my opinion, i think it's okay to update in `txLookup` methods which writes the `txLookup.slots`. 

Btw, i have updated the code since the initialization is missed when initializing txLookup. Could you take a look? https://github.com/klaytn/klaytn/pull/1733/commits/8ad2cc49608b4038a1a00653ad4e3a96de06b9ea

blukat29: That's fair enough. I'll be okay with a comment somewhere saying that `txLookup` is effectively a singleton.

blockchain/tx_pool_test.go
```diff
@@ -1303,6 +1312,62 @@ func TestTransactionPendingGlobalLimiting(t *testing.T) {
 	}
 }
 
+// Test the limit on transaction size is enforced correctly.
+// This test verifies every transaction having allowed size
+// is added to the pool, and longer transactions are rejected.
+func TestTransactionAllowedTxSize(t *testing.T) {
```
ian0371, line 1318: Can we make this test more readable? e.g.,
```go
	testcases := []struct {
		tx      *types.Transaction
		success bool
	}{
		{ // maximal allowed size
			pricedDataTransaction(0, 100000000, big.NewInt(1), key, dataSize),
			true,
		},
		// ...
		{ // random not allowed size
			pricedDataTransaction(2, 100000000, big.NewInt(1), key, dataSize+1+uint64(rand.Intn(int(10*MaxTxDataSize)))),
			false,
		},
		// ...
	}
```

yoomee1313: It looks better, thanks :) 

I applied at https://github.com/klaytn/klaytn/pull/1733/commits/718da50149264f5babf510525074b4ca3a2ae5e9. Please take a look.

ian0371: I think the second tc is wrong (it's not random).

Also, I thought "a transaction of minimal not allowed size" has the size of either `dataSize + 1` or `dataSize + txSlotSize`, since "a transaction with maximal allowed size" has the size of `dataSize`. Am I missing something??

yoomee1313: >I think the second tc is wrong (it's not random)

it was my mistake.. i fixed it thanks. https://github.com/klaytn/klaytn/pull/1733/commits/a1b1a6188be1d0a2e20be6fd2562f4363b33c719

You're right. Is something wrong?

blockchain/tx_pool.go
```diff
@@ -870,7 +882,7 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
 	// (2) remove an old Tx with the largest nonce from queue to make a room for a new Tx with missing nonce
 	// (3) discard a new Tx if the new Tx does not have a missing nonce
 	// (4) discard underpriced transactions
-	if uint64(pool.all.Count()) >= pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll {
+	if uint64(pool.all.Slots()+numSlots(tx)) >= pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll {
```
ian0371, line 885: I think the equality sign should be dropped.
If `pool.all.Slots()` is 4, `numSlots(tx)` is 4, and RHS is 8, then we don't need to discard txs, right?

blockchain/tx_pool_test.go
yoomee1313: Revised the comment. https://github.com/klaytn/klaytn/pull/1733/commits/c81cf24aeb94689e04b281e6580a08c39234a286

blockchain/tx_pool.go
yoomee1313: Thanks! Yes, it needs to be dropped. https://github.com/klaytn/klaytn/pull/1733/commits/9dc8ba1f4939fb4d289d5481222b16b5214a22a0

blukat29 (approved):

ian0371 (dismissed):

blockchain/tx_pool.go
yoomee1313: @hyunsooda would you check this?

blockchain/tx_pool.go
```diff
@@ -899,7 +912,7 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
 			return false, ErrUnderpriced
 		}
 		// New transaction is better than our worse ones, make room for it
-		drop := pool.priced.Discard(pool.all.Count()-int(pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll-1), pool.locals)
+		drop := pool.priced.Discard(pool.all.Slots()-int(pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll)+numSlots(tx), pool.locals)
```
hyunsooda, line 915: Arithmetic reordering seems to make it readable linearly in my view, but maybe not. 
```
drop := pool.priced.Discard(pool.all.Slots()+numSlots(tx)  -int(pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll), pool.locals)
```

Also, a shortcut for long wording `pool.config.ExecSlotsAll+pool.config.NonExecSlotsAll` can be used later.

blockchain/tx_pool.go
```diff
@@ -1717,17 +1730,27 @@ func (as *accountSet) add(addr common.Address) {
...
 // newTxLookup returns a new txLookup structure.
 func newTxLookup() *txLookup {
+	slotsGauge.Update(int64(0))
 	return &txLookup{
 		all: make(map[common.Hash]*types.Transaction),
 	}
 }
 
+// Slots returns the current number of slots used in the lookup.
+func (t *txLookup) Slots() int {
```
hyunsooda, line 1747: Seems unsafe because the returned integer may not be synched one if another thread modifies it at the same time. But, we may overlook it for now as its side-effect causes only wrong dropping. Any comment?

hyunsooda (approved):

blockchain/tx_pool.go
hyunsooda: Got it. thansk.

kjeom (approved):

blockchain/tx_pool.go
yoomee1313: @hyunsooda Does the lock protects the section, doesn't it? Is there any places which is not protected?

blockchain/tx_pool.go
yoomee1313: I'll apply it later, thanks!

blockchain/tx_pool.go
hyunsooda: Uh, the returned value of `Slots()` does not have a [lifetime](https://doc.rust-lang.org/book/ch10-03-lifetime-syntax.html) (the reference is Rust-based, but I believe the concept is not different) as it is not assigned to a variable. Thus, it seems not unsafe. I concerned if the value is assigned to the variable, then the variable (primitive value(integer))may have unsynched one with other contexts that may modifies it at the same time.

yoomee1313: What happens if there's lifetime?

yoomee1313: As we discussed in offline, it won't happen fundamentally. Only `Txpool.add` function calls the `Slots` function to access `t.slots`, but this `add` function would not be executed concurrently.

blukat29 (approved):

hyunsooda (approved):

PR #1735, opened 2022-12-05
increase log level on rpc call

JayChoi1736:
## Proposed changes

- increase log level on rpc call to logger.Trace

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [ ] New feature or enhancement
- [x] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1762, opened 2023-01-16
[RPC] Added timeout trigger for the default trace logger

hyunsooda:
## Proposed changes

The default trace logger (`debug.traceTransaction(txHash)`) aggregates all stack, memory and storage logs. The time that the log dump loop ends depends on the given set of instructions for the given transaction. It may crash the node through a huge loop that may arise OOM. The tracer has two parts, one for transaction executor and another one for log aggregation. A default timeout exists for the early exit of the former case but not for the latter. This PR proposes a default timeout trigger for the latter case, which is hard-coded 1s, but it can be overwritten with `loggerTimeout` option. For instance, `debug.traceTransaction(txHash, {loggerTimeout:"100ms"})` overwrites the timeout to 100ms.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/api/api_public_blockchain.go
+++ b/api/api_public_blockchain.go
@@ -504,9 +504,21 @@ type StructLogRes struct {
 }
 
 // formatLogs formats EVM returned structured logs for json output
-func FormatLogs(logs []vm.StructLog) []StructLogRes {
+func FormatLogs(timeout time.Duration, logs []vm.StructLog) ([]StructLogRes, error) {
+	logTimeout := false
+	deadlineCtx, cancel := context.WithTimeout(context.Background(), timeout)
+	go func() {
+		<-deadlineCtx.Done()
+		logger.Error("[Trace logger] Timeout reached", "timeout", timeout, "err", deadlineCtx.Err())
+		logTimeout = true
+	}()
+	defer cancel()
+
 	formatted := make([]StructLogRes, len(logs))
 	for index, trace := range logs {
+		if logTimeout {
+			return nil, fmt.Errorf("[Trace logger] Log truncated by timeout")
+		}
 		formatted[index] = StructLogRes{
 			Pc:      trace.Pc,
 			Op:      trace.Op.String(),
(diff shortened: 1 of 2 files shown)


blukat29:
@hyunsooda Hi, the `loggerTimeout` seems not working. For instance, `debug.traceTransaction(txid, {loggerTimeout:'10ms'})` does not immediately return, whereas `debug.traceTransaction(txid, {tracer:'callTracer', timeout: '10ms'})` does return quickly.

I think the tracer itself must be stopped to prevent OOM.
I have a WIP code (https://github.com/klaytn/klaytn/compare/dev...blukat29:klaytn:debug-trace-timeout). PTAL.

hyunsooda:
@blukat29, Another patch you made looks fine. Before uploading that PR, it seems good to consider whether it really makes OOM and whether that patch is necessary.

api/api_public_blockchain.go
```diff
@@ -504,9 +504,21 @@ type StructLogRes struct {
 }
 
 // formatLogs formats EVM returned structured logs for json output
-func FormatLogs(logs []vm.StructLog) []StructLogRes {
+func FormatLogs(timeout time.Duration, logs []vm.StructLog) ([]StructLogRes, error) {
+	logTimeout := false
+	deadlineCtx, cancel := context.WithTimeout(context.Background(), timeout)
+	go func() {
+		<-deadlineCtx.Done()
+		logger.Error("[Trace logger] Timeout reached", "timeout", timeout, "err", deadlineCtx.Err())
```
blukat29, line 512:
```suggestion
		logger.Debug("trace logger timeout", "timeout", timeout, "err", deadlineCtx.Err())
```
Error level is too verbose. For reference, failed RPC calls are logged at Debug level (https://github.com/klaytn/klaytn/pull/1735).

api/api_public_blockchain.go
```diff
@@ -504,9 +504,21 @@ type StructLogRes struct {
...
+		logTimeout = true
+	}()
+	defer cancel()
+
 	formatted := make([]StructLogRes, len(logs))
 	for index, trace := range logs {
+		if logTimeout {
+			return nil, fmt.Errorf("[Trace logger] Log truncated by timeout")
```
blukat29, line 520:
```suggestion
			return nil, fmt.Errorf("trace logger timeout")
```
Can we match the error style with others? (e.g. `execution timeout`)

hyunsooda: Thanks. I missed fixing the wrong msg. We may consider a new logging format that abstracts where its logging is involved by adding a prefix.

blukat29 (approved):

kjeom (approved):

PR #1765, opened 2023-01-17
Change ntp check logic

kjeom:
## Proposed changes

- Changed ntp time check with local to be more flexible (`stop server` -> `warning periodically`)
  - It checks the ntp time periodically until synced in ntp time and warn the status
- Changed it to use the `ntp` module in `discover` package and tuned ntp check count for average

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -814,6 +814,7 @@ func (n *Node) apis() []rpc.API {
 const (
 	ntpTolerance = time.Second
 	RFC3339Nano  = "2006-01-02T15:04:05.999999999Z07:00"
+	retry        = 10
 )
 
 func timeIsNear(lhs, rhs time.Time) bool {
@@ -840,14 +841,29 @@ func NtpCheckWithLocal(n *Node) error {
 		return err
 	}
 
+	ntpRetryTime := time.Duration(1)
+	var remote *time.Time
+	for i := 0; ; i++ {
+		time.Sleep(ntpRetryTime)
+		remote, err = ntpclient.GetNetworkTime(url, portNum)
+		if remote != nil {
+			break
+		}
+		// remote == nil && err != nil && i >= retry
+		if i >= retry {
+			return err
+		}
+		ntpRetryTime = ntpRetryTime * 2
+	}
 	local := time.Now()
-	remote, err := ntpclient.GetNetworkTime(url, portNum)
+
 	if err != nil {
 		return err
 	}
 	if !timeIsNear(local, *remote) {
 		errFormat := "System time is out of sync, local:%s remote:%s"
-		return fmt.Errorf(errFormat, local.UTC().Format(RFC3339Nano), remote.UTC().Format(RFC3339Nano))
+		usage := "You can use \"--ntp.disable\" option to disable ntp time checking"
+		return fmt.Errorf(errFormat+"\n"+usage, local.UTC().Format(RFC3339Nano), remote.UTC().Format(RFC3339Nano))
 	}
 	logger.Info("Ntp time check", "local", local.UTC().Format(RFC3339Nano), "remote", remote.UTC().Format(RFC3339Nano))
 	return nil
--- a/node/node.go
+++ b/node/node.go
@@ -843,23 +843,19 @@ func NtpCheckWithLocal(n *Node) error {
 
 	ntpRetryTime := time.Duration(1)
 	var remote *time.Time
-	for i := 0; ; i++ {
+	for i := 0; i < ntpMaxRetry; i++ {
 		time.Sleep(ntpRetryTime)
 		remote, err = ntpclient.GetNetworkTime(url, portNum)
 		if remote != nil {
 			break
 		}
-		// remote == nil && err != nil && i >= retry
-		if i >= retry {
-			return err
-		}
 		ntpRetryTime = ntpRetryTime * 2
 	}
-	local := time.Now()
-
 	if err != nil {
 		return err
 	}
+
+	local := time.Now()
 	if !timeIsNear(local, *remote) {
 		errFormat := "System time is out of sync, local:%s remote:%s"
 		usage := "You can use \"--ntp.disable\" option to disable ntp time checking"
--- a/node/node.go
+++ b/node/node.go
@@ -814,7 +814,7 @@ func (n *Node) apis() []rpc.API {
 const (
 	ntpTolerance = time.Second
 	RFC3339Nano  = "2006-01-02T15:04:05.999999999Z07:00"
-	retry        = 10
+	ntpMaxRetry  = 10
 )
 
 func timeIsNear(lhs, rhs time.Time) bool {


blukat29:
@kjeom Thanks for fixing this NTP problem. I have some questions.

1. Is `github.com/bt51/ntpclient` package no longer used? then please run `go mod tidy` to remove from dependencies.
2. How was the constant parameter `ntpChecksForBoot=5` chosen?
3. If there's reason to make it 5, then why not increase `ntpChecks=5` as well?

networks/p2p/discover/ntp.go
```diff
@@ -43,18 +44,38 @@ func (s durationSlice) Len() int           { return len(s) }
 func (s durationSlice) Less(i, j int) bool { return s[i] < s[j] }
 func (s durationSlice) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 
+// checkClockDrift queries an NTP server for clock drifts and warns the user if
+// one large enough is detected.
+func CheckClockDriftForBoot(url, port string) {
+	drift := driftThreshold * 2
+	for drift < -driftThreshold || drift > driftThreshold {
```
blukat29, line 51:
```suggestion
	var drift time.Duration
	for {
```
The stop condition already inside the loop should be enough.

networks/p2p/discover/ntp.go
```diff
@@ -43,18 +44,38 @@ func (s durationSlice) Len() int           { return len(s) }
...
+		time.Sleep(time.Second)
```
blukat29, line 52: Do you have to wait before the first try? Otherwise the Sleep can be moved to the end of this loop.

kjeom:
> 
1. resolved it.
2. At first considering the log of udp packet, I fixed it to check more than before. It can be done by `continue` when occured timeout and subtract 1 also to the divider for average. But now it doesn't matter because it restarts the ntp check when received the timeout.

networks/p2p/discover/ntp.go
```diff
@@ -43,18 +44,38 @@ func (s durationSlice) Len() int           { return len(s) }
 func (s durationSlice) Less(i, j int) bool { return s[i] < s[j] }
 func (s durationSlice) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 
+// checkClockDrift queries an NTP server for clock drifts and warns the user if
+// one large enough is detected.
+func CheckClockDriftForBoot(url, port string) {
```
blukat29, line 49: CheckClockDriftWithRetry? or something not specific to its use case?

networks/p2p/discover/ntp.go
kjeom: resolved it. Thanks!

networks/p2p/discover/ntp.go
kjeom: Good suggestion! thanks

networks/p2p/discover/ntp.go
```diff
@@ -43,18 +44,38 @@ func (s durationSlice) Len() int           { return len(s) }
...
+	for {
+		drift, err := sntpDrift(url, port, ntpChecksForBoot)
+		if err != nil {
+			// if received timeout, try to check again
+			continue
+		}
+		if drift >= -driftThreshold && drift <= driftThreshold {
+			break
+		}
+		logger.Warn(fmt.Sprintf("System clock seems shifted by %v, which can prevent network connectivity", drift))
```
aidan-kwon, line 60:
```suggestion
		logger.Warn(fmt.Sprintf("System clock seems shifted by %v, which may interfere with P2P connection", drift))
```

networks/p2p/discover/ntp.go
```diff
@@ -43,18 +44,38 @@ func (s durationSlice) Len() int           { return len(s) }
 func (s durationSlice) Less(i, j int) bool { return s[i] < s[j] }
 func (s durationSlice) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 
+// checkClockDrift queries an NTP server for clock drifts and warns the user if
+// one large enough is detected.
+func CheckClockDriftWithRetry(url, port string) {
+	var drift time.Duration
+	for {
```
aidan-kwon, line 51: How about adding a maximum trial limit for this loop?

networks/p2p/discover/ntp.go
```diff
@@ -31,8 +31,9 @@ import (
 )
 
 const (
-	ntpPool   = "pool.ntp.org" // ntpPool is the NTP server to query for the current time
-	ntpChecks = 3              // Number of measurements to do against the NTP server
+	ntpPool          = "pool.ntp.org" // ntpPool is the NTP server to query for the current time
+	ntpChecks        = 3              // Number of measurements to do against the NTP server
+	ntpChecksForBoot = 5
```
blukat29, line 36:
```suggestion
	ntpChecks        = 5              // Number of measurements to do against the NTP server
```
If we seek robustness, why not just use `5` in both cases?

networks/p2p/discover/ntp.go
kjeom: I think It must alerts the warning until the operator notices. Before this PR, we just stop the server but it would be more relaxed.

blukat29:
A continuous 1 burst per sec is too frequent (a burst is ~8 packets in a row - https://greencloud33.tistory.com/7)

Various usages suggest different frequencies. In general, they are at least above 0.1 packet/sec.
- Klaytn UDP stack's NTP poll is limited to 1 burst per 10min (https://github.com/klaytn/klaytn/blob/v1.10.1/networks/p2p/discover/udp.go#L516-522)
- An Oracle switch pauses at least 16sec before next poll (https://docs.oracle.com/cd/E19077-01/n1k.switch/819-3047-11/xdoc/switchservices_ntp_server_show.html)
- An NTP pool mail says around 1 burst per min is acceptable (https://www.mail-archive.com/pool@lists.ntp.org/msg01361.html)

We shouldn't flood NTP servers with Klaytn nodes. There should be either
- longer delay (at least 1 min)
- exponential backoff with upper limit (ex. wait 1, 2, 4, ..., 256, 512, 1024, 1024, 1024, ...)

blukat29:
> Becuase this notice pattern has been already used when if the discovery option flag is on.

Does "this notice pattern" refers to `checkClockDrift()`? Then this notice pattern looks good to me. Operators can go see the logs. However, it only happens after 32 consecutive UDP timeouts AND 10 minutes from the last notice. (https://github.com/klaytn/klaytn/blob/v1.10.1/networks/p2p/discover/udp.go#L516-522)

I guess the current notice method is okay because nodes can start anyway. It's just too frequent.

node/node.go
```diff
@@ -840,14 +841,29 @@ func NtpCheckWithLocal(n *Node) error {
 		return err
 	}
 
+	ntpRetryTime := time.Duration(1)
+	var remote *time.Time
+	for i := 0; ; i++ {
+		time.Sleep(ntpRetryTime)
```
blukat29, line 847:
```suggestion
	for i := 0; i < ntpMaxRetry; i++ {
```

node/node.go
```diff
@@ -814,6 +814,7 @@ func (n *Node) apis() []rpc.API {
 const (
 	ntpTolerance = time.Second
 	RFC3339Nano  = "2006-01-02T15:04:05.999999999Z07:00"
+	retry        = 10
```
blukat29, line 817:
```suggestion
	ntpMaxRetry  = 10
```

node/node.go
```diff
@@ -840,14 +841,29 @@ func NtpCheckWithLocal(n *Node) error {
...
+		time.Sleep(ntpRetryTime)
+		remote, err = ntpclient.GetNetworkTime(url, portNum)
+		if remote != nil {
+			break
+		}
+		// remote == nil && err != nil && i >= retry
+		if i >= retry {
+			return err
+		}
+		ntpRetryTime = ntpRetryTime * 2
```
blukat29, line 856:
```suggestion
		ntpRetryTime = ntpRetryTime * 2
		time.Sleep(ntpRetryTime)
```

blukat29 (approved):

blukat29:
LGTM except minors

node/node.go
kjeom: @blukat29 the sleep at the head of loop is intended for immediate restart.

node/node.go
kjeom: Resolved it.

PR #1804, opened 2023-03-03
[API] Added current fork status getter API

hyunsooda:
## Proposed changes

The current Klaytn implementation only knows the current fork status through predefined comparison internally. This PR proposes explicitly showing the current fork-update status through an API getter call.

Using this API, users can easily query the fork status with negligible management costs. This function should be updated whenever a future fork is scheduled. 

Sample output:
```
> klay.forkStatus(20)
{
  EthTxType: true,
  Istanbul: true,
  KIP103: false,
  Kore: true,
  London: true,
  Magma: true,
  Mantle: false
}
```

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


kjeom:
How about just add the hard fork status info. to the `klay_clientVersion` API ?

hyunsooda:
Hi @kjeom, that seems like another appropriate position to place this change. However, the node package does not have an exported interface to access the required values, such as getting the current block number due to the well-defined orgnization. Any other suggestions? If this change is anticipated to be useless, another option is to close it.

aidan-kwon:
@hyunsooda How about returning true/false for each hardforks? All hardforks are not designed to be a squeal update. A chain may skip some hardforks such as KIP103Fork.

hyunsooda:
@aidan-kwon, Updated. Please give further feedback if you have.

aidan-kwon (dismissed):
Thanks! Please update this on the Klaytn Docs API page 
cc @kjeom

blukat29:
@hyunsooda Can we have explicit `false` for inactive forks? I'd like to see the list of forks. e.g.
```
klay.forkStatus(20)
{
  EthTxType: true,
  Istanbul: true,
  Kore: false,
  KIP103: false,
  London: true,
  Magma: true
}
```

hyunsooda:
@blukat29, Sure.

TL;DR
I believe that by leveraging the robust type system of Golang, specifically through the reflect package, it is possible to refactor this convoluted if-else code into a more advantageous solution. This approach eliminates the need to update the function in the future if the struct field members have specific rules, such as requiring the block number to have a suffix of Block. However, upon closer consideration, I realized that implementing such implicit rules might introduce unforeseen side effects, particularly when modifications are made to the ChainConfig structure.

api/api_public_blockchain.go
```diff
@@ -150,6 +150,53 @@ func (s *PublicBlockChainAPI) GetAccount(ctx context.Context, address common.Add
...
+	}
+	blockNumber := block.Number()
+	cfg := s.b.ChainConfig()
+	activated := make(map[string]bool)
+
+	if cfg.IsIstanbulForkEnabled(blockNumber) {
+		activated["Istanbul"] = true
+	} else {
+		activated["Istanbul"] = false
+	}
```
blukat29, line 166: How about this?

```go
activated := map[string]bool{
  "Istanbul": cfg.IsIstanbulForkEnabled(blockNumber),
  "London": cfg.IsLondonForkEnabled(blockNumber),
}
```

hyunsooda: Ah..?

aidan-kwon (approved):

blukat29 (approved):

aidan-kwon:
@kjeom @JayChoi1736 @nohkwak Please, take a look

JayChoi1736 (approved):

aidan-kwon:
@kjeom Can you review this?

kjeom:
note: we should f/u this API every hard fork.

@hyunsooda I recommend you to add an function that return that information in the `params/config.go` because it is easy to miss when we are on hardfork update

aidan-kwon:
> note: we should f/u this API every hard fork.
> 
> @hyunsooda I recommend you to add an function that return that information in the `params/config.go` because it is easy to miss when we are on hardfork update

@kjeom  It would be better to make a test case checking a remainder after parsing the API

hyunsooda:
@kjeom,

>  I recommend you to add an function that return that information in the params/config.go because it is easy to miss when we are on hardfork update

Do you mean that it's better to move the function in `param/config.go`? If not, could you elaborate more? I couldn't exactly get what your concern is.

@aidan-kwon, 

I think that rewriting this proposed API with the result of `Rules()`. And, the excpetional two cases (`EthTxType`, `KIP103`) can be manually added. Then, I think we don't need to worry about the missing fork update. While I was writting this, you merged this. How do you think?

Issue #1823, opened 2023-04-07
Introduce BLS library into codebase


blukat29:
**Is your request related to a problem? Please describe.**
Klaytn needs to bring in BLS algorithm for RANDAO, optimized consensus and EIP-2537.

BLS functionality can be split into two layers:
- BLS12-381 arithmetic: functions like `G1.Add(P,Q)`. An example usage is the [EIP-2537 precompiles](https://eips.ethereum.org/EIPS/eip-2537).
- BLS (simple or aggregate) signature: functions like `Sign()` and `AggregateVerify()`. Example usages are RANDAO and block signatures.

Klaytn should need both layers.

**Describe the solution you'd like**
Below are potential Go open sources we can use.

- **[`geth/crypto/bls12381`](https://github.com/ethereum/go-ethereum/tree/v1.11.5/crypto/bls12381)** package implements BLS12-381 arithmetic. It has been [stable since 2022](https://github.com/ethereum/go-ethereum/commits/v1.11.5/crypto/bls12381) and geth [implements EIP-2537 using this package](https://github.com/ethereum/go-ethereum/blob/v1.11.5/core/vm/contracts.go#L653). So I think it's a safe bet to use geth's crypto/bls12381 if Klaytn will support EIP-2537.
- **[`celo/crypto/bls12381`](https://github.com/celo-org/celo-blockchain/tree/v1.7.3/crypto/bls12381)** is similar to `geth/crypto/bls12381`, but has [diverted from geth's since 2021](https://github.com/celo-org/celo-blockchain/commits/master/crypto/bls12381).
- **[`celo-bls-go`](https://github.com/celo-org/celo-bls-go)** is another package [used by celo blockchain](https://github.com/celo-org/celo-blockchain/blob/v1.7.3/crypto/bls/bls.go) for its consensus. celo-bls-go [implements BLS12-377 curve](https://github.com/celo-org/celo-bls-snark-rs/). Note that Ethereum uses BLS12-381. 
- **[`prysm/crypto/bls`](https://github.com/prysmaticlabs/prysm/tree/v4.0.1/crypto/bls)** is a package within [prysm](https://github.com/prysmaticlabs/prysm) consensus layer (CL) client that implements BLS signature. It is a wrapper around the `supranational/blst` package. It is configured as (G1=pub,G2=sig) mode.
[...]

PR #1824, opened 2023-04-11
Add package crypto/bls

blukat29:
## Proposed changes

Add BLS signature package `github.com/klaytn/crypto/bls`
- Inspired by [prysmaticlabs/prysm/crypto/bls](https://github.com/prysmaticlabs/prysm/tree/develop/crypto/bls), This package wraps around the package [supranational/blst](https://github.com/supranational/blst)
- It uses "[miniminal-pubkey-size](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-variants)" variant, which the Ethereum uses.
  Public keys are points on G1 (48 bytes) and signatures are points on G2 (96 bytes).
- Optimized uncompress operations using in-memory cache
- See [bls_test.go](https://github.com/blukat29/klaytn/blob/bls-package/crypto/bls/bls_test.go) for usage

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- #1823 

## Further comments

**Performance benchmark**

Some notable results are (on Linux x86-64 16-core):
- `[]bytes48 -> AggregatePublicKey` for 100 items takes 7.1ms uncached, 0.3ms cached.
- `[]bytes96 -> AggregateSignature` for 100 items takes 9.0ms uncached, 0.3ms cached.
- `Sign` takes 0.5ms.
- `Verify` takes 1.3ms.

Full result here: [BLS.library.benchmark.2023.04.klaytn.230419.pdf](https://github.com/klaytn/klaytn/files/11271873/BLS.library.benchmark.2023.04.klaytn.230419.pdf)

**Notices**
- The bls12381 package will be added in another PR.
- supranational/blst contains cgo code. We may need to check the cross-compiler availability.


crypto/bls/blst/bench_test.go
```diff
@@ -0,0 +1,237 @@
...
+		tc.sigs[i] = sig
+		tc.sigbs[i] = sig.Marshal()
+	}
+	return tc
+}
+
+func runUncached(b *testing.B, name string, fn func()) {
+	b.Run(name, func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			resetCache()
```
ian0371, line 217: shouldn't we stop the timer during resetCache()?

blukat29: Fixed in e7bdb56e.
I re-ran the benchmark, and thankfully the result hasn't significantly changed.

crypto/bls/blst/public_key_test.go
```diff
@@ -0,0 +1,119 @@
...
+	bs[0] = testBadPublicKeyBytes // in_g2?
+	_, err = MultiplePublicKeysFromBytes(bs)
+	assert.Equal(t, types.ErrPublicKeyUnmarshal, err)
+}
+
+func TestAggregatePublicKeys(t *testing.T) {
+	L := benchAggregateLen
+	tc := generateBenchmarkMaterial(L)
+
+	// Correctness check is done in Sign() and Verify() tests
```
ian0371, line 88: This comment is written in these functions and thus skips their correctness check:
```
AggregatePublicKeysFromBytes (calls AggregatePublicKeys)
PublicKey
AggregateSignaturesFromBytes (calls AggregateSignatures)
```

Specifically, which test are you talking about?
- `blst.Sign()` and `blst.Verify()` do not call the aformentioned functions
- thus, `TestSignVerify()` in `blst/signature_test.go` wouldn't test them
- `TestAggregateVerify()` in `blst/signature_test.go` only uses the result of `AggregatePublicKeysFromBytes()`.

Did you mean `TestAggregateVerify()` in `bls_test.go` perhaps? It seems to be the only place that uses the return value of all those functions (not the error). If so, please update the comment.

I just want to make sure the correctness of those functions has been tested, because they're wrappers of blst and thus might behave incorrectly.

crypto/bls/blst/public_key.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, types.ErrEmptyArray
+	}
+
+	ps := make([]*blstPublicKey, len(pks))
+	for i, pk := range pks {
+		ps[i] = pk.(*publicKey).p
+	}
+
+	agg := new(blstAggregatePublicKey)
+	groupcheck := false // alreaay checked in *PublicKeyFromBytes()
```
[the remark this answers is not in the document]
blukat29: Fixed in 6e9cb529.

crypto/bls/blst/cache.go
```diff
@@ -0,0 +1,60 @@
...
+	signatureCache common.Cache // Signature Uncompress
+}
+
+var computeCache = &blstComputeCache{}
+
+func cacheKey(b []byte) common.CacheKey {
+	return crypto.Keccak256Hash(b)
+}
+
+func initCacheLocked() {
```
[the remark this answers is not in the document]
blukat29: Sure. Fixed in cf4c52e9.

crypto/bls/blst/public_key_test.go
blukat29: I meant TestAggregateVerify in signature_test.go which uses a known test vector.
Let me supplement those tests with more test vectors.

blukat29: Fixed in 7b91434.

ian0371 (approved):

crypto/bls/bls.go
```diff
@@ -0,0 +1,85 @@
...
+	return blst.AggregatePublicKeys(pks)
+}
+
+// AggregatePublicKeysFromBytes unmarshals and validates multiple BLS public key from bytes
+// and then aggregates them. Returns error if an empty slice is given.
+func AggregatePublicKeysFromBytes(bs [][]byte) (PublicKey, error) {
+	return blst.AggregatePublicKeysFromBytes(bs)
+}
+
+// AggregatePublicKeys aggregates multiple BLS signatures.
```
aidan-kwon, line 63:
```suggestion
// AggregateSignatures aggregates multiple BLS signatures.
```

crypto/bls/bls.go
```diff
@@ -0,0 +1,85 @@
...
+// AggregatePublicKeys aggregates multiple BLS public keys.
+// Assumes that all given public keys are previously validated.
+// Returns error if an empty slice is given.
+func AggregatePublicKeys(pks []PublicKey) (PublicKey, error) {
+	return blst.AggregatePublicKeys(pks)
+}
+
+// AggregatePublicKeysFromBytes unmarshals and validates multiple BLS public key from bytes
+// and then aggregates them. Returns error if an empty slice is given.
+func AggregatePublicKeysFromBytes(bs [][]byte) (PublicKey, error) {
```
aidan-kwon, line 59: Those function names are different from prysm code. Can I know the reason? 
I like your naming, but I am afraid that other developers may mis-use functions when they take code from prysm. Prysm has the same function name `AggregatePublicKeys`, but have a different meaning

blukat29: As you might have guessed, I renamed them because *PublicKeys and *Signatures have inconsistent prefix/suffix combinations. But I understand your concern.

I will change the names to be identical to prysm and then add a comment summarizing the interfaces. I should add NewAggregateSignatures as well then.

crypto/bls/blst/bench_test.go
```diff
@@ -0,0 +1,239 @@
...
+		apk, _ := AggregatePublicKeysFromBytes(tc.pkbs)
+		sig, _ := SignatureFromBytes(asigb)
+		Verify(sig, tc.msg, apk)
+	}
+	runUncached(b, "Uncached", fn)
+	runCached(b, "Cached", fn)
+}
+
+// FastAggregateVerify
+func BenchmarkFasAggregateVerify(b *testing.B) {
```
aidan-kwon, line 165:
```suggestion
func BenchmarkFastAggregateVerify(b *testing.B) {
```

crypto/bls/blst/public_key.go
```diff
@@ -0,0 +1,127 @@
...
+	p *blstPublicKey
+}
+
+func PublicKeyFromBytes(b []byte) (types.PublicKey, error) {
+	if len(b) != types.PublicKeyLength {
+		return nil, types.ErrPublicKeyLength(len(b))
+	}
+
+	if pk, ok := publicKeyCache.Get(cacheKey(b)); ok {
+		return pk.(*publicKey), nil
```
aidan-kwon, line 34: I am asking you for double-checking my concern. You didn't copy the return of the LRU cache because the cache' return is not free in eviction or purge situations, right?

crypto/bls/blst/public_key.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	p := new(blstPublicKey).Uncompress(b)
+	if p == nil || !p.KeyValidate() {
+		return nil, types.ErrPublicKeyUnmarshal
+	}
+
```
aidan-kwon, line 41: We missed `p.KeyValidate` when it compared to prysm. is it okay?

crypto/bls/bls.go
blukat29: I still need to add VerifyMultipleSignatures. I will fix typos at once.

crypto/bls/blst/public_key.go
blukat29: Yes, I thought the reference counter should increase when the pointer is returned.
However, it seems nice to add Copy() as a precaution.

crypto/bls/bls.go
blukat29: @aidan-kwon it turns out that `prysm NewAggregateSignatures()` is a [dummy constant](https://github.com/prysmaticlabs/prysm/blob/v4.0.2/crypto/bls/blst/signature.go#L177) and only used in tests. Therefore I did not include NewAggregateSignatures.

blukat29:
Updated:
- Add VerifyMultipleSignatures intended for deposit/registry contract validation use case
- Internal LRU cache stores Copy() of objects
- Run test vectors from https://github.com/ethereum/bls12-381-tests as-is
- Re-ran benchmarks

aidan-kwon (approved):

ian0371 (approved):

crypto/bls/blst/signature.go
```diff
@@ -0,0 +1,229 @@
...
+		return nil, types.ErrSignatureUnmarshal
+	}
+
+	s := &signature{p: p}
+	signatureCache.Add(cacheKey(b), s.Copy())
+	return s, nil
+}
+
+func MultipleSignaturesFromBytes(bs [][]byte) ([]types.Signature, error) {
+	if len(bs) == 0 {
```
jiseongnoh, line 53: shouldn't this be an error? is this an exception?

crypto/bls/blst/signature.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Compute on uncached elements
+	batchPs := new(blstSignature).BatchUncompress(batchBytes)
+	if batchPs == nil || len(batchPs) != len(batchBytes) {
+		return nil, types.ErrSignatureUnmarshal
+	}
+
+	// Merge cached and uncached elements
+	for i, outIdx := range batchIndices {
```
jiseongnoh, line 82: does outIdx start after the cached elements? i think sigs[] already has cached elements, i am just curious.

crypto/bls/blst/signature.go
```diff
@@ -0,0 +1,229 @@
...
+		return nil, types.ErrEmptyArray
+	}
+
+	ps := make([]*blstSignature, len(sigs))
+	for i, sig := range sigs {
+		ps[i] = sig.(*signature).p
+	}
+
+	agg := new(blstAggregateSignature)
+	groupcheck := false // alreaay checked in *SignatureFromBytes()
```
jiseongnoh, line 115: typo, already

jiseongnoh (approved):

crypto/bls/blst/signature.go
blukat29: Within the klaytn/bls/blst package,

- (a) functions converting Multiple bytes -> multiple objects: returns nil if given slice is empty
  - MultiplePublicKeysFromBytes
  - MultipleSignaturesFromBytes
- (a) functions converting Multiple bytes or objects -> aggregated object: obviously can't accept an empty slice
  - AggregatePublicKeys
  - AggregatePublicKeysFromBytes
  - AggregateSignatures
  - AggregateSignaturesFromBytes

It could be an error, but not necessary.

crypto/bls/blst/signature.go
blukat29: Did you ask if outIdx avoids cached elements? then yes. Please refer to the [range syntax](https://gobyexample.com/range).

Suppose [A,B,C,D] were given and A,C were previously cached. Then
batchBytes = [B,D]
batchIndices = [1,3]
In the for loop,
- i = 0, outIdx = 1
- i = 1, outIdx = 3

crypto/bls/blst/signature.go
blukat29: Thanks. I will fix it later.

crypto/bls/blst/signature.go
jiseongnoh: That's right. Thanks

PR #1766, opened 2023-01-25
Fix to resolve node loading failure after unexpected node termination

aidan-kwon:
### Root cause 
In the `loadLastState` method, `bc.currentHeader` can be updated after storing `bc.currentBlock`. This mismatch causes an expected result during the node initializing process. 

https://github.com/klaytn/klaytn/blob/5d7a03b7bb2d834393d26719e52896ed3eeb1a06/blockchain/blockchain.go#L468-L478

`Governance.UpdateParams` requires the head block number of the canonical chain, but it had used the head of header chain that is used for light client usually. This PR does
 1. Replace the use of `bc.currentHeader` to `bc.currentBlock`
 2. Add a parameter indicating a block number for `UpdateParams` method

## Types of changes

Please put an x in the boxes related to your change.

- [x] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1828, opened 2023-04-18
Add BlockchainContractCaller

blukat29:
## Proposed changes

Add `BlockchainContractCaller` that implements `bind.ContractCaller` based on live `BlockChain` instance. Note that `SimulatedBackend` creates a temporary BlockChain, so it's only useful for testing. However, `BlockchainContractCaller` wraps an existing BlockChain, so it can be used in block processing.

Replaced by `BlockchainContractCaller`

- [Kip103ContractCaller](https://github.com/klaytn/klaytn/blob/v1.10.2/consensus/istanbul/backend/kip103.go)
- [governance.contractCaller](https://github.com/klaytn/klaytn/blob/v1.10.2/governance/contract_connector.go)
- [addressBookConnector.getStakingInfoFromAddressBook](https://github.com/klaytn/klaytn/blob/v1.10.2/reward/addressbook_connector.go)

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


accounts/abi/bind/backends/blockchain.go
```diff
@@ -0,0 +1,142 @@
...
+	} else {
+		return state.GetCode(account), nil
+	}
+}
+
+// Executes a read-only function call with respect to the latest state.
+//
+// Returns call result in []byte.
+// Returns error when:
+// - a past (i.e. not latest) blockNumber is given
```
ian0371, line 82: why is this? does this mean GetStakingInfo or ParamsAt with past block cannot be handled?

2dvorak: It is a outdated comment, which should be updated.

The original implementation of BlockChainContractCaller did not have StateAt(). Now it has StateAt(), and past block can be handled. I've updated the comment accordingly.

Thanks!

accounts/abi/bind/backends/blockchain.go
```diff
@@ -0,0 +1,140 @@
...
+// Its intended purpose is reading system contracts during block processing.
+//
+// Note that SimulatedBackend creates a new temporary BlockChain for testing,
+// whereas BlockchainContractCaller uses an existing BlockChain with existing database.
+type BlockchainContractCaller struct {
+	bc blockChainForCaller
+}
+
+// This nil assignment ensures at compile time that BlockchainContractCaller implements bind.ContractCaller.
+var _ bind.ContractCaller = (*BlockchainContractCaller)(nil)
```
blukat29, line 62: Move this to blockchain_test.go please

accounts/abi/bind/backends/blockchain_test.go
```diff
@@ -0,0 +1,135 @@
...
+		testAddr:  {Balance: big.NewInt(10000000000)},
+		code1Addr: {Balance: big.NewInt(0), Code: code1Bytes},
+		code2Addr: {Balance: big.NewInt(0), Code: code2Bytes},
+	}
+
+	db := database.NewMemoryDBManager()
+	genesis := blockchain.Genesis{Config: config, Alloc: alloc}
+	genesis.MustCommit(db)
+
+	bc, _ := blockchain.NewBlockChain(db, nil, genesis.Config, gxhash.NewFaker(), vm.Config{})
```
blukat29, line 61: Please add GenerateChain here, then add test cases involving states at nonzero block numbers.

2dvorak: I missed those tests, thanks!

accounts/abi/bind/backends/blockchain.go
2dvorak: It seems that go won't include `_test.go` files when compiling the binary, however placing this nil assignment test in `blockchain.go` ensures this test to be done for both build and test.

blukat29: I don't think the test is necessary at build time. But since it's a zero-cost test, I am fine with it here.

accounts/abi/bind/backends/blockchain_test.go
```diff
@@ -0,0 +1,149 @@
...
+	assert.Nil(t, err)
+	assert.Equal(t, code1Bytes, code)
+
+	code, err = c.CodeAt(context.Background(), code1Addr, big.NewInt(10))
+	assert.Nil(t, err)
+	assert.Equal(t, code1Bytes, code)
+
+	// Non-code address
+	code, err = c.CodeAt(context.Background(), testAddr, nil)
+	assert.True(t, code == nil || err != nil)
```
yoomee1313, line 98: Shouldn't both conditions be satisfied?
```suggestion
	assert.True(t, code == nil && err != nil)
```

yoomee1313 (commented):
It looks good to me except for one minor thing.

accounts/abi/bind/backends/blockchain_test.go
2dvorak: You're right, however the err should be nil; because the given state - nil blockNumber leading to latest state - is valid and the address simply has no code.

I've updated the test codes to be more specific about the expected results. If it's still misleading or ambiguous, I'll try to add comments.

ian0371 (dismissed):

accounts/abi/bind/backends/blockchain_test.go
yoomee1313: Oh, err should be nil.. I didn't know that. Thank you.

yoomee1313 (approved):

accounts/abi/bind/backends/blockchain.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/klaytn/klaytn/blockchain/types"
+	"github.com/klaytn/klaytn/blockchain/vm"
+	"github.com/klaytn/klaytn/common"
+	"github.com/klaytn/klaytn/params"
+)
+
+// Maintain separate minimal interfaces of blockchain.BlockChain because ContractBackend are used
+// in various situations. BlockChain instances are often passed down as different interfaces such as
+// consensus.ChainReader, governance.blockChain, work.BlockChain.
+type blockChainForCaller interface {
```
kjeom, line 36: I think it's better to make it public to act as an interface.
```go
type BlockChainForCaller interface {
...
}
```

governance/contract.go
```diff
@@ -96,11 +98,25 @@ func (e *ContractEngine) contractGetAllParamsAt(num uint64) (*params.GovParamSet
...
+	caller := backends.NewBlockchainContractCaller(chain)
+	contract, _ := govcontract.NewGovParamCaller(addr, caller)
+
+	names, values, err := contract.GetAllParamsAt(nil, new(big.Int).SetUint64(num))
+	if err != nil {
+		logger.Warn("ContractEngine disabled: getAllParams call failed", "err", err)
+		return params.NewGovParamSet(), nil
+	}
+
+	if len(names) != len(values) {
```
jiseongnoh, line 110: Is it possible to have a name but no value?

2dvorak: It is impossible unless the GovParam contract is broken. If somehow the GovParam contract is broken, this safety check would prevent OOB in the following code.

We could try to panic without this check, however it would be impossible to fix the broken GovParam contract when the node panics and stops proposing blocks.

accounts/abi/bind/backends/blockchain.go
2dvorak: Done, PTAL.

kjeom (approved):

ian0371 (approved):

yoomee1313 (approved):

accounts/abi/bind/backends/blockchain.go
```diff
@@ -0,0 +1,140 @@
...
+	return blockchain.ApplyMessage(evm, msg)
+}
+
+func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block, *state.StateDB, error) {
+	var header *types.Header
+	if num == nil {
+		header = b.bc.CurrentHeader()
+	} else {
+		header = b.bc.GetHeaderByNumber(num.Uint64())
+	}
```
aidan-kwon, line 131: @2dvorak We better change this code to use CurrentBlock.Header instead of CurrentHeader. The currentHeader may have a different value. For more detail, see https://github.com/klaytn/klaytn/pull/1766

reward/staking_manager.go
```diff
@@ -195,6 +213,71 @@ func updateStakingInfo(blockNum uint64) (*StakingInfo, error) {
...
+		}
+	}
+
+	// validate parsed node information
+	if len(nodeIds) != len(stakingAddrs) ||
+		len(nodeIds) != len(rewardAddrs) ||
+		common.EmptyAddress(pocAddr) ||
+		common.EmptyAddress(kirAddr) {
+		// This is an expected behavior when the addressBook contract is not activated yet.
+		logger.Info("The addressBook is not yet activated. Use empty stakingInfo", "reason", err)
```
aidan-kwon, line 274: No `err` exist for this context. And, this code is not reachable if AddressBook is not activated. Don't we need to return an error for this case?

aidan-kwon: https://github.com/klaytn/klaytn/blob/2dc0b6578b6def6e80238351397170e92d747b0d/reward/addressbook_connector.go#L155-L160

aidan-kwon: In the previous code, an error log was printed and `newEmptyStakingInfo(blockNum)` was returned

accounts/abi/bind/backends/blockchain.go
2dvorak: @blukat29 handled this in #1882

reward/staking_manager.go
2dvorak: @blukat29 handled this in #1882

accounts/abi/bind/backends/blockchain.go
2dvorak: @aidan-kwon This will be handled in #1893

PR #1829, opened 2023-04-18
node: call CreateSnapshot explicitly when node starts

yoomee1313:
## Proposed changes
- gov.changeSet
  - gov.changeSet is one of the governance state cache: the vote data which is not stored in governance db yet.
  - It is a memory value. So, if the node shut down before it is flushed into gov db at governance block, it is deleted.
  - but it would restored well when the node restarts during the first `snapshot` call.
- However, after https://github.com/klaytn/klaytn/pull/1706 is applied, missing gov.changeSet problem appears
  - if the first `snapshot` is called by getValidator, it creates the snapshot caches without restoring gov.changeSet (because the snapshot called by getValidator is not able to write governance data)
  -  the next `snapshot` call skip the restoration part always whether it is able to write governance data or not because there's already snapshot cache.
  - and it leads to a verification error due to the mismatch between gov.changeSet and blockHeader.governanceData.
- So, the solution is calling `snapshot` explictly before sync starts to ensure the first `snapshot` call restores gov.changeSet always.
  - currentSet or idxCache are restored during intializeCache, but changeSet does not.

dev
- vote at block 34 and restart immediately
- at the next epoch start block (e.g. 60), it returned verification error
```
ERROR[05/05,09:01:42 Z] [30] Verification Error                        len(receivedChangeSet)=1 len(changeSet)=0
ERROR[05/05,09:01:42 Z] [5]
########## BAD BLOCK #########
...
```
PR
- vote at block 34 and restart immediately
- at the next epoch start block (e.g. 60), it handles governance data properly with out error.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -2021,6 +2021,13 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		if bc.engine.CreateSnapshot(bc, block.NumberU64(), block.Hash(), nil) != nil {
 			return i, events, coalescedLogs, err
 		}
+
+		// update governance parameters
+		if istanbul, ok := bc.engine.(consensus.Istanbul); ok {
+			if err = istanbul.UpdateParam(block.NumberU64()); err != nil {
+				return i, events, coalescedLogs, err
+			}
+		}
 	}
 	// Append a single chain head event if we've progressed the chain
 	if lastCanon != nil && bc.CurrentBlock().Hash() == lastCanon.Hash() {
--- a/node/cn/backend.go
+++ b/node/cn/backend.go
@@ -329,6 +329,14 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 		reward.NewStakingManager(cn.blockchain, governance, cn.chainDB)
 	}
 
+	// Governance states which are not yet applied to the db remains at in-memory storage
+	// It disappears during the node restart, so restoration is needed before the sync starts
+	// By calling CreateSnapshot, it restores the gov state snapshots and apply the votes in it
+	// Particularly, the gov.changeSet is also restored here.
+	if err := cn.Engine().CreateSnapshot(cn.blockchain, cn.blockchain.CurrentBlock().NumberU64(), cn.blockchain.CurrentBlock().Hash(), nil); err != nil {
+		logger.Error("CreateSnapshot failed", "err", err)
+	}
+
 	// set worker
 	if config.WorkerDisable {
 		cn.miner = work.NewFakeWorker()
--- a/consensus/clique/clique.go
+++ b/consensus/clique/clique.go
@@ -359,7 +359,7 @@ func (c *Clique) verifyCascadingFields(chain consensus.ChainReader, header *type
 	return c.verifySeal(chain, header, parents)
 }
 
-// CreateSnapshot does not return a snapshot but creates a new snapshot at a given point in time.
+// CreateSnapshot does not return a snapshot but creates a new snapshot if not exists at a given point in time
 func (c *Clique) CreateSnapshot(chain consensus.ChainReader, number uint64, hash common.Hash, parents []*types.Header) error {
 	_, err := c.snapshot(chain, number, hash, parents)
 	return err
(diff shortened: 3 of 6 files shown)


blukat29 (approved):

aidan-kwon (commented):
I'm sorry, but I cannot understand some points. Can you explain more? 

> if no such snapshot function is called until the next governance block is processed,

It seems that a node calls `snapshot()` on every block to verifyHeader. 

> this problem is caused by https://github.com/klaytn/klaytn/pull/1706

The PR disables "write" in some cases, but you are saying that there is an issue on "read" (restoration of vote data). So, I am confused.

node/cn/backend.go
```diff
@@ -329,6 +329,11 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 		reward.NewStakingManager(cn.blockchain, governance, cn.chainDB)
 	}
 
+	// Call CreateSnapshot to restore the votes since last checkpoint
+	if err := cn.engine.CreateSnapshot(cn.blockchain, cn.blockchain.CurrentBlock().NumberU64(), cn.blockchain.CurrentBlock().Hash(), nil); err != nil {
```
aidan-kwon, line 333: Can we change the place of the added code line? 
It is a little strange to me to create a governance snapshot and update govParam here. 
Because we already load the latest gov param at the above code lines, and the restored governance data is used to configure `blockchain`. If we update the governance-related data here again, then it can be misunderstood like "the governance data above was not correct"

https://github.com/klaytn/klaytn/blob/f3605ce7f24131ca23e89772497bad5a6740f71b/node/cn/backend.go#L270-L285

aidan-kwon: perhaps, restoring changeSet could be simple and clear

yoomee1313:
> It seems that a node calls snapshot() on every block to verifyHeader.

In a syncing situation, a snapshot is called, governance state data is writable, so governance state data is written(e.g. changeSet, currentSet). But the problem is that if the snapshot already exists, it will be hit and returned directly from the cache, and the code that writes the governance state data is not executed. (When I say that the snapshot already exists, I am assuming that the code has been run once when creating the snapshot.)
but the problem is, when the first snapshot call is non-writable at governance state data, the assumption the code has been run once is broken.

On second thought, I realized that it is good to change the implementation, because the fundamental solution is to decouple the governance state data recovery (specifically changeSet) from the snapshot call.
* implement changeSet recovery logic in `initializeCache`. (but if not possible, i will go back to original solution).

node/cn/backend.go
yoomee1313: @aidan-kwon At first I thought it would be easy to restore a changeSet, but my thought wrong. There are a lot of things to consider. `apply` in `Snapshot` function already considered all the things. I moved the UpdateParam outside of the createSnapshot.

So, calling `CreateSnapshot` at start point of the node makes sense to me in next two aspects
* restore the snapshots of the blocks from latestGovernanceState to currentBlock
* restore the in-memory gov states (e.g. gov.changeSet) (not gov.currentSet.. its restoration place is initializeCache)

Can I go with the original code?

blockchain/blockchain.go
```diff
@@ -2021,6 +2021,13 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		if bc.engine.CreateSnapshot(bc, block.NumberU64(), block.Hash(), nil) != nil {
 			return i, events, coalescedLogs, err
 		}
+
+		// update governance parameters
+		if istanbul, ok := bc.engine.(consensus.Istanbul); ok {
```
yoomee1313, line 2026: @ian0371 Could you review that calling UpdateParam only when we use istanbul engine is okay?

ian0371: lgtm

yoomee1313:
Dear reviewers, please review this PR one more time. I'll go as https://github.com/klaytn/klaytn/pull/1829#discussion_r1201600057.

blukat29 (approved):

ian0371 (approved):

aidan-kwon (approved):

aidan-kwon (approved):

aidan-kwon (approved):

PR #1845, opened 2023-05-17
return a specific error during validating the fee payer

kjeom:
## Proposed changes

- fixed to return a specific error when the fee payer validation failed.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/tx_pool.go
+++ b/blockchain/tx_pool.go
@@ -753,7 +753,7 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 	// Make sure the transaction is signed properly
 	gasFrom, err := tx.ValidateSender(pool.signer, pool.currentState, pool.currentBlockNumber)
 	if err != nil {
-		return err
+		return types.ErrSender(err)
 	}
 	from := tx.ValidatedSender()
 
@@ -769,7 +769,7 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 		// balance check for fee-delegated tx
 		gasFeePayer, err = tx.ValidateFeePayer(pool.signer, pool.currentState, pool.currentBlockNumber)
 		if err != nil {
-			return err
+			return types.ErrFeePayer(err)
 		}
 		feePayer := tx.ValidatedFeePayer()
 		feePayerBalance := pool.getBalance(feePayer)
--- a/blockchain/tx_pool.go
+++ b/blockchain/tx_pool.go
@@ -769,7 +769,7 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 		// balance check for fee-delegated tx
 		gasFeePayer, err = tx.ValidateFeePayer(pool.signer, pool.currentState, pool.currentBlockNumber)
 		if err != nil {
-			return ErrInvalidFeePayer
+			return err
 		}
 		feePayer := tx.ValidatedFeePayer()
 		feePayerBalance := pool.getBalance(feePayer)
--- a/blockchain/types/transaction.go
+++ b/blockchain/types/transaction.go
@@ -26,6 +26,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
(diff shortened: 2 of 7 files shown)


blukat29 (approved):

blukat29 (dismissed):
Please fix the tests, otherwise LGTM.

blockchain/tx_pool.go
```diff
@@ -769,7 +769,7 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 		// balance check for fee-delegated tx
 		gasFeePayer, err = tx.ValidateFeePayer(pool.signer, pool.currentState, pool.currentBlockNumber)
 		if err != nil {
-			return ErrInvalidFeePayer
+			return err
```
aidan-kwon, line 772: This change provides more detailed error messages to users, but it may confuse them. A user will receive the same error message for an invalid sender's signature and an invalid feePayer's signature. For example, "invalid chain id for signer" message can be returned for both invalid sender and feePayer.

aidan-kwon: If you want to get a more detailed error message, we may add an additional messages such as "from fee payer" at the end of `err`.

blukat29: To preserve original error messges, we could do:
```go
// error.go
func ErrInvalidSender(reason error) error {
    return fmt.Errorf("invalid sender: %s", reason)
}
func ErrInvalidFeePayer(reason error) error {
    return fmt.Errorf("invalid fee payer: %s", reason)
}
```
And use ErrInvalidSender at line 756, ErrInvalidFeePayer at line 772.

kjeom: Is it only for fee delegated transactions? I can't find the invalid chain id error in `tx.ValidateFeePayer()`

kjeom: I understood it. Thanks for the suggestion. I think @blukat29's suggestion is better in this case.

tests/tx_validation_test.go
```diff
@@ -663,7 +663,7 @@ func TestValidationInvalidSig(t *testing.T) {
 
 				// For block tx validation test
 				if expectedErr == blockchain.ErrInvalidFeePayer {
-					expectedErr = types.ErrInvalidSigFeePayer
+					expectedErr = types.ErrFeePayer(types.ErrInvalidSigFeePayer)
 				}
```
blukat29, line 667: This if statement is unnecessary.

blockchain/types/transaction.go
```diff
@@ -58,6 +59,13 @@ func deriveSigner(V *big.Int) Signer {
 	return LatestSignerForChainID(deriveChainId(V))
 }
 
+func ErrSender(err error) error {
+	return fmt.Errorf("sender: %s", err)
+}
+func ErrFeePayer(err error) error {
+	return fmt.Errorf("fee payer: %s", err)
+}
```
blukat29: Can we keep the original text "invalid sender"? I know that some inner errors say "invalid" again. But for the error messages to show up in Google search and docs, they should remain consistent.

kjeom: Sure, resolved it.

tests/tx_validation_test.go
kjeom: Thanks

kjeom: @blukat29 I have checked you need the condition because the expectedErr also can be for sender at
```go
tx, expectedErr := invalidCase.fn(t, txType, reservoir, contract, signer)
```

aidan-kwon (approved):

blukat29 (approved):

aidan-kwon:
@JayChoi1736  FYI, it may fail some of RPC tests requiring a specific error message

JayChoi1736:
> @JayChoi1736 FYI, it may fail some of RPC tests requiring a specific error message

i checked that the case testing theses does not exist in rpc-tester. i will update it by adding the corresponding test case.

JayChoi1736 (approved):

PR #1846, opened 2023-05-17
Refactor DiskDB trie accessors

blukat29:
## Proposed changes

- This PR contains several refactoring changes related to state trie accessors.

### Change 1. Cleanup DiskDB trie accessors

| Category             | Before                                                                                                                                    | After                                                                                                                                                                     |
|----------------------|-------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Contract code        | ReadCode(Hash)<br>ReadCodeWithPrefix(Hash)<br>HasCode(Hash)<br>HasCodeWithPrefix(Hash)<br>WriteCode(Hash,[]byte)<br>-<br>DeleteCode(Hash) | ReadCode(Hash)<br>ReadCodeWithPrefix(Hash)<br>HasCode(Hash)<br>HasCodeWithPrefix(Hash)<br>WriteCode(Hash,[]byte)<br>PutCodeToBatch(Batch,Hash,[]byte)<br>DeleteCode(Hash) |
| Trie node            | ReadCachedTrieNode(Hash)<br>ReadStateTrieNode([]byte)<br>HasStateTrieNode([]byte)<br>-<br>-<br>-                                          | ReadTrieNode(Hash)<br>-<br>HasTrieNode(Hash)<br>WriteTrieNode(Hash,[]byte)<br>PutTrieNodeToBatch(Batch,Hash,[]byte)<br>DeleteTrieNode(Hash)                               |
| Secure trie preimage | ReadCachedTrieNodePreimage([]byte)<br>ReadPreimage(Hash)<br>WritePreimages(map[Hash][]byte)                                               | -<br>ReadPreimage(Hash)<br>WritePreimages(map[Hash][]byte)                                                                                                                |

### Change 2. No direct access to the underlying Key-Value database

- Make sure all codes use the above Read~, Write~ functions instead of low-level DB.Get(), Batcher.Put(), etc.
- One exception is in TrieSync.Commit where too much change is required.

### Change 3. Use TrieNodeKey before calling DB

- Use `TrieNodeKey()` before calling DB.Get(), Batcher.Put(), etc.
- Currently, `TrieNodeKey(hash) == hash.Bytes`. Additional logic may be introduced later.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Further comments

Later, `TrieNodeKey()` may process ExtHash and legacy Hash to support state pruning.

### DIFF
--- a/datasync/downloader/downloader_test.go
+++ b/datasync/downloader/downloader_test.go
@@ -149,7 +149,7 @@ func newTester() *downloadTester {
 		peerMissingStates: make(map[string]map[common.Hash]bool),
 	}
 	tester.stateDb = localdb
-	tester.stateDb.GetMemDB().Put(genesis.Root().Bytes(), []byte{0x00})
+	tester.stateDb.WriteTrieNode(genesis.Root(), []byte{0x00})
 
 	tester.downloader = New(FullSync, tester.stateDb, statedb.NewSyncBloom(1, tester.stateDb.GetMemDB()), new(event.TypeMux), tester, nil, tester.dropPeer, uint64(istanbul.WeightedRandom))
 
@@ -348,7 +348,7 @@ func (dl *downloadTester) CurrentBlock() *types.Block {
 
 	for i := len(dl.ownHashes) - 1; i >= 0; i-- {
 		if block := dl.ownBlocks[dl.ownHashes[i]]; block != nil {
-			if _, err := dl.stateDb.GetMemDB().Get(block.Root().Bytes()); err == nil {
+			if has, _ := dl.stateDb.HasTrieNode(block.Root()); has {
 				return block
 			}
 		}
@@ -430,15 +430,15 @@ func (dl *downloadTester) InsertChain(blocks types.Blocks) (int, error) {
 	for i, block := range blocks {
 		if parent, ok := dl.ownBlocks[block.ParentHash()]; !ok {
 			return i, fmt.Errorf("InsertChain: unknown parent at position %d / %d", i, len(blocks))
-		} else if _, err := dl.stateDb.GetMemDB().Get(parent.Root().Bytes()); err != nil {
-			return i, fmt.Errorf("InsertChain: unknown parent state %x: %v", parent.Root(), err)
+		} else if has, _ := dl.stateDb.HasTrieNode(parent.Root()); !has {
+			return i, fmt.Errorf("InsertChain: unknown parent state %x", parent.Root())
 		}
 		if _, ok := dl.ownHeaders[block.Hash()]; !ok {
 			dl.ownHashes = append(dl.ownHashes, block.Hash())
 			dl.ownHeaders[block.Hash()] = block.Header()
 		}
 		dl.ownBlocks[block.Hash()] = block
-		dl.stateDb.GetMemDB().Put(block.Root().Bytes(), []byte{0x00})
+		dl.stateDb.WriteTrieNode(block.Root(), []byte{0x00})
 		dl.ownChainTd[block.Hash()] = new(big.Int).Add(dl.ownChainTd[block.ParentHash()], block.BlockScore())
 	}
 	return len(blocks), nil
@@ -734,7 +734,7 @@ func (dlp *downloadTesterPeer) RequestNodeData(hashes []common.Hash) error {
 
 	results := make([][]byte, 0, len(hashes))
 	for _, hash := range hashes {
-		if data, err := dlp.dl.peerDb.GetMemDB().Get(hash.Bytes()); err == nil {
+		if data, err := dlp.dl.peerDb.ReadTrieNode(hash); err == nil {
 			if !dlp.dl.peerMissingStates[dlp.id][hash] {
 				results = append(results, data)
 			}
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -162,7 +162,10 @@ type DBManager interface {
 	HasCodeWithPrefixFromOld(hash common.Hash) bool
 	ReadPreimageFromOld(hash common.Hash) []byte
 
+	// Write StateTrie
+	WriteTrieNode(hash common.Hash, node []byte) error
 	PutTrieNodeToBatch(batch Batch, hash common.Hash, node []byte) error
+	DeleteTrieNode(hash common.Hash) error
 	WritePreimages(number uint64, preimages map[common.Hash][]byte) error
 
 	// from accessors_indexes.go
@@ -1840,6 +1843,18 @@ func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
 	return data
 }
 
+func (dbm *databaseManager) WriteTrieNode(hash common.Hash, node []byte) error {
+	dbm.lockInMigration.RLock()
+	defer dbm.lockInMigration.RUnlock()
+
+	if dbm.inMigration {
+		if err := dbm.getDatabase(StateTrieMigrationDB).Put(hash[:], node); err != nil {
+			return err
+		}
+	}
+	return dbm.getDatabase(StateTrieDB).Put(hash[:], node)
+}
+
 func (dbm *databaseManager) PutTrieNodeToBatch(batch Batch, hash common.Hash, node []byte) error {
 	return batch.Put(hash[:], node)
 }
@@ -1862,6 +1877,10 @@ func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.H
 	preimageCounter.Inc(int64(len(preimages)))
 	preimageHitCounter.Inc(int64(len(preimages)))
 	return nil
+}
+
+func (dbm *databaseManager) DeleteTrieNode(hash common.Hash) error {
+	return dbm.getDatabase(StateTrieDB).Delete(hash[:])
 }
 
 // ReadTxLookupEntry retrieves the positional metadata associated with a transaction
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -163,9 +163,9 @@ type DBManager interface {
 	ReadPreimageFromOld(hash common.Hash) []byte
 
 	// Write StateTrie
-	WriteTrieNode(hash common.Hash, node []byte) error
-	PutTrieNodeToBatch(batch Batch, hash common.Hash, node []byte) error
-	DeleteTrieNode(hash common.Hash) error
+	WriteTrieNode(hash common.Hash, node []byte)
+	PutTrieNodeToBatch(batch Batch, hash common.Hash, node []byte)
+	DeleteTrieNode(hash common.Hash)
 	WritePreimages(number uint64, preimages map[common.Hash][]byte)
 
 	// from accessors_indexes.go
@@ -1843,20 +1843,24 @@ func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
 	return data
 }
 
-func (dbm *databaseManager) WriteTrieNode(hash common.Hash, node []byte) error {
+func (dbm *databaseManager) WriteTrieNode(hash common.Hash, node []byte) {
 	dbm.lockInMigration.RLock()
 	defer dbm.lockInMigration.RUnlock()
 
 	if dbm.inMigration {
 		if err := dbm.getDatabase(StateTrieMigrationDB).Put(TrieNodeKey(hash), node); err != nil {
-			return err
+			logger.Crit("Failed to store trie node", "err", err)
 		}
 	}
-	return dbm.getDatabase(StateTrieDB).Put(TrieNodeKey(hash), node)
+	if err := dbm.getDatabase(StateTrieDB).Put(TrieNodeKey(hash), node); err != nil {
+		logger.Crit("Failed to store trie node", "err", err)
+	}
 }
 
-func (dbm *databaseManager) PutTrieNodeToBatch(batch Batch, hash common.Hash, node []byte) error {
-	return batch.Put(TrieNodeKey(hash), node)
+func (dbm *databaseManager) PutTrieNodeToBatch(batch Batch, hash common.Hash, node []byte) {
+	if err := batch.Put(TrieNodeKey(hash), node); err != nil {
+		logger.Crit("Failed to store trie node", "err", err)
+	}
 }
 
 // WritePreimages writes the provided set of preimages to the database. `number` is the
@@ -1878,8 +1882,10 @@ func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.H
 	preimageHitCounter.Inc(int64(len(preimages)))
 }
 
-func (dbm *databaseManager) DeleteTrieNode(hash common.Hash) error {
-	return dbm.getDatabase(StateTrieDB).Delete(TrieNodeKey(hash))
+func (dbm *databaseManager) DeleteTrieNode(hash common.Hash) {
+	if err := dbm.getDatabase(StateTrieDB).Delete(TrieNodeKey(hash)); err != nil {
+		logger.Crit("Failed to delete trie node", "err", err)
+	}
 }
 
 // ReadTxLookupEntry retrieves the positional metadata associated with a transaction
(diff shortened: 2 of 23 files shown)


blukat29:
Found some test codes that directly access to the underlying Key-Value database.Fixed with [86f8973](https://github.com/klaytn/klaytn/pull/1846/commits/86f897397ca9236d1fa05a07a1a07aa3599394b3).

storage/database/db_manager.go
```diff
@@ -140,34 +140,33 @@ type DBManager interface {
 	ReadCode(hash common.Hash) []byte
 	ReadCodeWithPrefix(hash common.Hash) []byte
 	WriteCode(hash common.Hash, code []byte)
+	PutCodeToBatch(batch Batch, hash common.Hash, code []byte)
```
ethan-kr, line 143: PutCodeToBatch?
WriteCodeToBatch?

storage/statedb/database.go
```diff
@@ -829,42 +815,15 @@ func (db *Database) Cap(limit common.StorageSize) error {
...
-	}
-
-	// Write batch ready, unlock for readers during persistence
-	if _, err := database.WriteBatches(preimagesBatch); err != nil {
-		logger.Error("Failed to write preimages to disk", "err", err)
-		return err
-	}
-
-	return nil
+	return db.diskDB.WritePreimages(0, db.preimages)
```
ethan-kr, line 818: Good~~

ethan-kr (dismissed):

storage/database/db_manager.go
blukat29: Yes. I did sway between the two. But I decided to go with PutCodeToBatch.

Take "Body" schema as an example,
- klaytn 1.10.2 DBManager has both `WriteBody(...)` and `PutBodyToBatch(batch, ...)`
- [geth 1.11.5 eth/rawdb](https://github.com/ethereum/go-ethereum/blob/v1.11.5/core/rawdb/accessors_chain.go) has only `WriteBody(KVwriter, ...)` because they somehow treat LevelDB and Batcher equally.

To follow the current convention, PutCodeToBatch is a better fit.

yoomee1313 (approved):
Overall, it looks good to me.

storage/statedb/database.go
```diff
@@ -769,7 +751,11 @@ func (db *Database) Cap(limit common.StorageSize) error {
 		// Fetch the oldest referenced node and push into the batch
 		node := db.nodes[oldest]
 		enc := node.rlp()
-		if err := database.PutAndWriteBatchesOverThreshold(batch, oldest[:], enc); err != nil {
```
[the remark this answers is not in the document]
blukat29: Good Idea. It's better to discourage general-purpose batcher.Put([]byte), and prefer specialized PutXToBatch. Updated in [617dc9e](https://github.com/klaytn/klaytn/pull/1846/commits/617dc9eae2117f5b017dd036dc995a5ec16e3fcf).

storage/database/db_manager.go
```diff
@@ -1885,20 +1843,44 @@ func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
...
+	return batch.Put(TrieNodeKey(hash), node)
+}
+
 // WritePreimages writes the provided set of preimages to the database. `number` is the
 // current block number, and is used for debug messages only.
-func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.Hash][]byte) {
+func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.Hash][]byte) error {
 	batch := dbm.NewBatch(StateTrieDB)
 	for hash, preimage := range preimages {
 		if err := batch.Put(preimageKey(hash), preimage); err != nil {
```
jeongkyun-oh, line 1867: Don't we have to add `PutPreimageToBatch`?

storage/database/db_manager.go
```diff
@@ -1885,20 +1843,44 @@ func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
...
-			logger.Crit("Failed to store trie preimage", "err", err)
+			return err
```
jeongkyun-oh, line 1868: Any reason returning error here? Doesn't it have to handle the error in caller?

storage/database/db_manager.go
jeongkyun-oh: This `Put` method is a bit awkward to me. There is no other put method for batch in dbmanager except for this. On the other hand, I understood your point as well. Also, shouldn't we change `batch.Write` as well to something like `WriteBatches` method?

storage/statedb/database.go
```diff
@@ -605,26 +596,17 @@ func (db *Database) DoesExistNodeInPersistent(hash common.Hash) bool {
...
-}
-
-// secureKey returns the database key for the preimage of key (as a newly
-// allocated byte-slice)
-func secureKey(hash common.Hash) []byte {
-	buf := make([]byte, secureKeyLength)
-	copy(buf, secureKeyPrefix)
-	copy(buf[secureKeyPrefixLength:], hash[:])
-	return buf
+	return db.diskDB.ReadPreimage(hash)
```
jeongkyun-oh, line 609: is this okay to change key?

storage/database/db_manager.go
blukat29: If there is such a use case, then yes.
But when I looked up the preimage-related code, they all simply use WritePreimages to write multiple preimages.

storage/database/db_manager.go
blukat29: Why return an error?
- There were two use cases
  - call WritePreimage() and not check error because logger.Crit() inside
  - manually use Batcher.Put and check error.
- I intended to keep only the second case because returning an error is better than silently dying with Crit().

Error handling
- There are total 3 call sites of WritePreimages. I made sure the callers check the error.
- At blockchain/blockchain.go:1556, toss up the error
- At blockchain/blockchain.go:1657, toss up the error
- At storage/statedb/database.go:818, writeBatchPreimages() has been returning the error before this PR.

storage/statedb/database.go
blukat29: Actually, the secureKeyPrefix is equal to preimagePrefix, which is "secure-key-".

storage/database/db_manager.go
jeongkyun-oh: Oh, I missed the error handling logic in `blockchain.go`. Thanks.

storage/statedb/database.go
jeongkyun-oh: I got it. Thanks.

storage/database/db_manager.go
jeongkyun-oh: Got it. Thanks.

jeongkyun-oh (approved):

storage/database/db_manager.go
blukat29: There were a few `Put..ToBatch` methods before this PR: PutBodyToBatch, PutReceiptsToBatch, etc. I'm not sure about their origin because they don't exist in geth. Anyway, if we're to rename `Put..ToBatch` to something else, it should be another PR.

For replacing `batch.Write` to `WriteBatches(batch)`, I attempted but reverted because there are so many. Again, it should be another PR on its own.

storage/database/db_manager.go
aidan-kwon, line 1868: Most of databaseManager methods are implemented to raise panic when it fails to write/delete something, and it returns an empty value when it fails to read something. I think the initial developer think the writing failure should be detected by a node operator, and don't want to miss any error handling that causes a critical database integrity issue. 
Is there any reason you made this exceptional case?

aidan-kwon (dismissed):
Looks good to me except for a comment

storage/database/db_manager.go
blukat29: @aidan-kwon, please refer to https://github.com/klaytn/klaytn/pull/1846#discussion_r1206227026.
the reason was to preserve the original behavior of statedb.Database.writeBatchPreimages().

However, it's not a good sign two people asked about the exact same line.

I quickly skimmed [db_manager.go](https://github.com/klaytn/klaytn/blob/v1.10.2/storage/database/db_manager.go) and discovered that some Klaytn-specific methods return errors. It's worth discussing whether we're going to use errorless interfaces for all DBManager methods.

jeongkyun-oh (approved):

aidan-kwon (approved):

blukat29:
@jeongkyun-oh, @aidan-kwon, and I in a conversation, decided to:
- Revert WritePreimage() to logger.Crit, over returning an error.
- New data-modifying functions - WriteTrieNode, PutTrieNodeToBatch, DeleteTrieNode - logger.Crits over returning an error.
- Use the PutXxToBatch convention (at least for this PR)
- Other DBManager interfaces to be fixed for consistency, perhaps in another PR. 

@ian0371 @hyunsooda @2dvorak @yoomee1313 PTAL.

ethan-kr (approved):

ian0371 (approved):

yoomee1313 (approved):

storage/database/db_manager.go
```diff
@@ -1885,6 +1843,26 @@ func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
...
+	defer dbm.lockInMigration.RUnlock()
+
+	if dbm.inMigration {
+		if err := dbm.getDatabase(StateTrieMigrationDB).Put(TrieNodeKey(hash), node); err != nil {
+			logger.Crit("Failed to store trie node", "err", err)
+		}
+	}
+	if err := dbm.getDatabase(StateTrieDB).Put(TrieNodeKey(hash), node); err != nil {
+		logger.Crit("Failed to store trie node", "err", err)
+	}
```
hyunsooda, line 1857: This new changes (two writes to two different DB) is the same behavior with the code `dstDb.GetMemDB().Put()`?

datasync/downloader/downloader_test.go
```diff
@@ -348,7 +348,7 @@ func (dl *downloadTester) CurrentBlock() *types.Block {
 
 	for i := len(dl.ownHashes) - 1; i >= 0; i-- {
 		if block := dl.ownBlocks[dl.ownHashes[i]]; block != nil {
-			if _, err := dl.stateDb.GetMemDB().Get(block.Root().Bytes()); err == nil {
```
hyunsooda, line 351: Is it the same behavior `dbm.getDatabase(StateTrieDB) == GetMemDB()`?

storage/database/db_manager.go
```diff
@@ -1901,6 +1882,12 @@ func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.H
 	preimageHitCounter.Inc(int64(len(preimages)))
 }
 
+func (dbm *databaseManager) DeleteTrieNode(hash common.Hash) {
+	if err := dbm.getDatabase(StateTrieDB).Delete(TrieNodeKey(hash)); err != nil {
```
hyunsooda, line 1886: 1. Does not need a write lock?
2. The updated `WriteTrieNode` function stores the provided data in two databases, whereas deletion only impacts a single database. Is the consistency maintained in this process?

datasync/downloader/downloader_test.go
blukat29: Yes, but only if we're using Memory DB. See `NewMemoryDBManager()` and `getDatabase()`.

storage/database/db_manager.go
blukat29: 1. I believe respective DB implementations have a write lock inside. (e.g. NewLevelDB relies on [syndtr/leveldb](https://github.com/syndtr/goleveldb/blob/master/leveldb/db_write.go#L331) and it has internal write lock). The `database.Database` interface states that the Delete method must be thread-safe  (https://github.com/klaytn/klaytn/blob/v1.10.2/storage/database/interface.go#L70)
2. I simply followed the behavior of `DeleteCode`, which is another use-case of StateTrieDB. Currently, DeleteTrieNode is only used in tests, where there is only one MemDB.

storage/database/db_manager.go
blukat29: The `GetMemDB()` with `NewMemoryDBManager()` is only used in testing, where any `getDatabase(*)` call returns the same MemDB. If we limit the discussion to testing, then yes, it is equivalent to `GetMemDB().Put()`.

storage/database/db_manager.go
```diff
@@ -1885,6 +1843,26 @@ func (dbm *databaseManager) ReadPreimageFromOld(hash common.Hash) []byte {
 	return data
 }
 
+func (dbm *databaseManager) WriteTrieNode(hash common.Hash, node []byte) {
+	dbm.lockInMigration.RLock()
+	defer dbm.lockInMigration.RUnlock()
```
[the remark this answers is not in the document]
blukat29: `dbm.inMigration` is not modified in WriteTrieNode. Is there a reason we need `Lock()` instead of `RLock()`?

datasync/downloader/downloader_test.go
hyunsooda: Thanks.

storage/database/db_manager.go
hyunsooda: Thanks.

storage/database/db_manager.go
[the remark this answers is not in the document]
hyunsooda: I pointed this by `Put()` behavior, which is write behavior. If both `ReadTrieNode()` and `WriteTireNode()` are called simultaneously, then corruption may occur. No?

storage/database/db_manager.go
```diff
@@ -1744,59 +1749,33 @@ func (dbm *databaseManager) DeleteCode(hash common.Hash) {
...
-		if val, err := dbm.GetStateTrieMigrationDB().Get(key); err == nil {
-			return val, nil
-		}
-	}
-	return dbm.ReadStateTrieNodeFromOld(key)
-}
-
-func (dbm *databaseManager) HasStateTrieNode(key []byte) (bool, error) {
-	val, err := dbm.ReadStateTrieNode(key)
+func (dbm *databaseManager) HasTrieNode(hash common.Hash) (bool, error) {
```
2dvorak, line 1772: Wouldn't it be more appropriate to return only bool?
If we need the reason (error) too, we can call `ReadTrieNode` instead.

storage/database/db_manager.go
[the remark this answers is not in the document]
blukat29: Oh, that should be handled inside concrete Database implementations. I don't think that is DBManager's responsibility.

storage/database/db_manager.go
blukat29: I agree HasTrieNode need not return an error.
But I decided to limit the impact of this PR as in [this comment](https://github.com/klaytn/klaytn/pull/1846#issuecomment-1576007705).

Already existing functions (though slightly changed) keep the original interfaces (i.e. ReadTrieNode and HasTrieNode)
and newly introduced functions use the error-less interface (i.e. WriteTrieNode, DeleteTrieNode, PutTrieNodeToBatch).

2dvorak: OK, thanks.

storage/database/db_manager.go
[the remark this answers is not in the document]
hyunsooda: Ah, stupid. Thanks.

storage/database/db_manager.go
blukat29: Fair point.
However, I was pursuing another consistency with respect to WriteCode and DeleteCode, where we Put() to two different databases and Delete() to one database. I applied the same principle to WriteTrieNode and DeleteTrieNode.

2dvorak (approved):

2dvorak (approved):

storage/database/db_manager.go
hyunsooda: Got it. thanks.

hyunsooda (approved):

Issue #1841, opened 2023-05-09
Invalid memory address or nil pointer dereference


Andrew-Pohl:
**Describe the bug**
We noticed that occasionally eth_calls on an RPC endpoint was returning incorrect data inconsistently (when compared with other nodes we run). When looking through the logs we noticed an instance of the log line:

"[48] Unexpected error while reading cached trie node"

shortly followed by an invalid memory address or nil pointer dereference (I assume these are related) 

stack trace below:

"May 09 18:32:17 klaytn ken[2021685]: ERROR[05/09,18:32:17 Z] [39] RPC method eth_getBlockByNumber crashed: runtime error: invalid memory address or nil pointer dereference
May 09 18:32:17 klaytn ken[2021685]: goroutine 3681 [running]:
May 09 18:32:17 klaytn ken[2021685]: [github.com/klaytn/klaytn/networks/rpc.(*callback).call.func1(0xc04e273120](http://github.com/klaytn/klaytn/networks/rpc.(*callback).call.func1(0xc04e273120), 0x14, 0xc03cf44d78)
May 09 18:32:17 klaytn ken[2021685]:         /ext-go/1/src/github.com/klaytn/klaytn/networks/rpc/service.go:211 +0xba
May 09 18:32:17 klaytn ken[2021685]: panic(0x1b42720, 0x2ea61f0)
May 09 18:32:17 klaytn ken[2021685]:         /usr/local/go/src/runtime/panic.go:969 +0x1b9
May 09 18:32:17 klaytn ken[2021685]: [github.com/klaytn/klaytn/work.(*worker).pendingBlock(0xc0238d4160](http://github.com/klaytn/klaytn/work.(*worker).pendingBlock(0xc0238d4160), 0x0)
May 09 18:32:17 klaytn ken[2021685]:         /ext-go/1/src/github.com/klaytn/klaytn/work/worker.go:230 +0x10e
May 09 18:32:17 klaytn ken[2021685]: [github.com/klaytn/klaytn/work.(*Miner).PendingBlock(0xc02c2cfe50](http://github.com/klaytn/klaytn/work.(*Miner).PendingBlock(0xc02c2cfe50), 0x167b8000000000)
May 09 18:32:17 klaytn ken[2021685]:         /ext-go/1/src/github.com/klaytn/klaytn/work/work.go:219 +0x2f
[...]

PR #1847, opened 2023-05-17
fix incomplete pending block handling logic regarding API requests

### LINKED ISSUE #1842: "eth" namespace APIs can't handle "pending" block correctly.
Klaytn has separated the role of nodes into various types; block generation node (CN) and API serving node (EN), and others. "pending" block (self-generated block but not confirmed by other nodes) as an API request parameter is meaningless. Therefore, "klay" namespace APIs handle "pending" block parameter as "latest" block parameter. 

However, it seems that some "eth" namespace APIs are trying to handle "pending" block and it causes an error on the endpoint node. We need to investigate how "eth" namespace APIs handle "pending" blocks and replace them to use "latest" block instead.

aidan-kwon:
## Proposed changes

This PR contains two different fixes regarding "pending" block requests. 

1) panic because of the absence of pending block
Panic was reported in #1841 regarding API servicing with "pending" block parameter. The panic can be triggered when a pending block has never been created yet but a user requests API with "pending" parameter. 

2) occasional API error with  "invalid signature length" error message
During the synchronization, a node uses a worker's snapshot to return a pending block. The block is created with the worker's `current` data in `updateSnapshot()` method. Since EN doesn't create a pending block correctly, however, the `current.header` doesn't have enough information including a miner's signature. Therefore, `eth_getBlockByNumber` and other API fail to recover a block proposer's address and return an error message. This PR set the previous block header for a pending block of EN/PN to support the API correctly. 

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

Resolve #1842

### DIFF
--- a/work/worker.go
+++ b/work/worker.go
@@ -207,6 +207,9 @@ func (self *worker) pending() (*types.Block, *state.StateDB) {
 		// return a snapshot to avoid contention on currentMu mutex
 		self.snapshotMu.RLock()
 		defer self.snapshotMu.RUnlock()
+		if self.snapshotState == nil {
+			return nil, nil
+		}
 		return self.snapshotBlock, self.snapshotState.Copy()
 	}
 
@@ -227,6 +230,9 @@ func (self *worker) pendingBlock() *types.Block {
 
 	self.currentMu.Lock()
 	defer self.currentMu.Unlock()
+	if self.current == nil {
+		return nil
+	}
 	return self.current.Block
 }
 
--- a/work/worker.go
+++ b/work/worker.go
@@ -471,7 +471,9 @@ func (self *worker) makeCurrent(parent *types.Block, header *types.Header) error
 	}
 	work := NewTask(self.config, types.MakeSigner(self.config, header.Number), stateDB, header)
 	if self.nodetype != common.CONSENSUSNODE {
+		// set the current block and header as pending block and header to support APIs requesting a pending block.
 		work.Block = parent
+		work.header = parent.Header()
 	}
 
 	// Keep track of transactions which return errors so they can be removed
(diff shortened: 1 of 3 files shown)


work/worker.go
```diff
@@ -471,7 +477,9 @@ func (self *worker) makeCurrent(parent *types.Block, header *types.Header) error
 	}
 	work := NewTask(self.config, types.MakeSigner(self.config, header.Number), stateDB, header)
 	if self.nodetype != common.CONSENSUSNODE {
+		// set the current block and header as pending block and header to support APIs requesting a pending block.
 		work.Block = parent
+		work.header = parent.Header()
```
kjeom, line 482: I didn't find any negative effect when setting the previous block header.

aidan-kwon: @kjeom You mean that you can't find the current problem related to block header, right?

1. Currently, an incomplete block header is used for `makeCurrent` and the method set the header to `worker.current.header`.
https://github.com/klaytn/klaytn/blob/6833919f7de6acda298cd3c9ff779daf5cdbd93c/work/worker.go#L529-L543


3. `worker.snapshotBlock` consists of `worker.current.header` and others
https://github.com/klaytn/klaytn/blob/6833919f7de6acda298cd3c9ff779daf5cdbd93c/work/worker.go#L616-L620

4. `worker.snapshotBlock` can be returned as a pending block 
https://github.com/klaytn/klaytn/blob/6833919f7de6acda298cd3c9ff779daf5cdbd93c/work/worker.go#L220-L225

5. In the JSON marshaling logic, the pending block's header is used to calculate a block proposer address 
https://github.com/klaytn/klaytn/blob/6833919f7de6acda298cd3c9ff779daf5cdbd93c/api/api_ethereum.go#L1463-L1474

Therefore, an incomplete block header assigned to `worker.current.header` can cause panic. Since EN/PN doesn't finish the pending block generation process, an alternative header is needed for `worker.current.header`

kjeom: @aidan-kwon Oh, sorry for confusing. I just noted what I have finished for review.
Anyway thanks for explaining it in detail.

blukat29 (approved):

PR #1855, opened 2023-05-31
Implemented RocksDB

jeongkyun-oh:
## Proposed changes

- implemented RocksDB in order to support concurrent reads from multiple process
- RocksDB supports read-only instance with extra ability to catch-up with primary instance. This is done by `TryCatchup` method.
- The chaindata directory is separated to share the chaindata only, not the other unnecessary data such as `klay.ipc`, `kend.pid`.
- `GetDBProperty` API is added
- The RocksDB options, metrics and testcases are added.
- In order to build with rocksdb, `USE_ROCKSDB` environment variable should be set in addition to rocksdb installation properly.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


storage/database/batch.go
```diff
@@ -38,6 +38,9 @@ type Batch interface {
 	// Reset resets the Batch for reuse.
 	Reset()
 
+	// Destroy deallocate WriteBatch object of RocksDB.
+	Destroy()
```
blukat29, line 42: Can we call it Batch.Release() just like Iterator.Release()?

blockchain/blockchain.go
```diff
@@ -658,6 +658,9 @@ func (bc *BlockChain) FastSyncCommitHead(hash common.Hash) error {
 // CurrentBlock retrieves the current head block of the canonical chain. The
 // block is retrieved from the blockchain's internal cache.
 func (bc *BlockChain) CurrentBlock() *types.Block {
+	if bc.db.GetDBConfig().DBType == database.RocksDB && bc.db.GetDBConfig().RocksDBConfig.Secondary {
+		bc.loadLastState()
```
blukat29, line 662: Tell me if I understood - secondary nodes do not insert blocks themselves; so they are not notified of new blocks. Thus `bc.currentBlock` is kept untouched. To know the current state in a timely manner, every `CurrentBlock` call queries the database.

If so, then why update here? I guess because `CurrentBlock()` is used in most of the RPCs? I somewhat agree with the choice. But code-wise it looks arbitrary that only CurrentBlock updates and other getters (such as GetBlock) do not.

How about having a background loop that periodically loadLastState, to not depend on a specific method's side effect.

jeongkyun-oh: Yes, you are right. Still `CurrentBlock` is called periodically somewhere while processing blocks, but I agreed with your point. Let me find better place to update last state, or I will take your advice.

blukat29: Thank you.
Another concern occurs to me though. loadLastState is a little heavy task involving several database reads. Having to loadLastState on every getter call might add some DB pressure. Though I'm just speculating because I don't know much about RocksDB client-side caching or remote read performance.

jeongkyun-oh: @blukat29 I have updated this. The logic is borrowed from KES logic. Instead of subscribing current block from redis-cache, this reads current block from rocksdb per 1 second (would be better parameterized). DB reads only twice for `HeadBlockHash` and `Block` per a second so that I think it would be fine. It replaces current block along with updating assoicated variables and metrics. Also, resetting txpool enables to handle transactions, too. However, it won't update subscription data (e.g. tx log) at the momoment.

Please take a look and leave your thought. Thanks.

blukat29: @jeongkyun-oh Looks good to me. Reading per 1 second (or any reasonable interval) will be good.
My previous concern was that reading per every CurrentBlock call would cause hundreds or thousands of reads per sec. Using the loop clears my concern.

blockchain/blockchain.go
```diff
@@ -2151,6 +2156,47 @@ func (bc *BlockChain) CloseBlockSubscriptionLoop() {
...
+	refresher := time.NewTicker(1 * time.Second)
+	defer refresher.Stop()
+
+	for {
+		select {
+		case <-refresher.C:
+			// Restore the last known head block
+			head := bc.db.ReadHeadBlockHash()
+			if head == (common.Hash{}) {
+				logger.Warn("Failed to read head block hash")
```
aidan-kwon, line 2170: Isn't `continue` required in the next line?

node/cn/api.go
```diff
@@ -283,6 +284,11 @@ func NewPublicDebugAPI(cn *CN) *PublicDebugAPI {
 	return &PublicDebugAPI{cn: cn}
 }
 
+func (api *PublicDebugAPI) GetRocksDBProperty(dt database.DBEntryType, name string) string {
+	dbm := api.cn.ChainDB()
+	return dbm.GetProperty(dt, name)
```
aidan-kwon, line 289: How about adding a more general API such as `GetDBProperty`? Because Klaytn already has a similar API called `ChaindbProperty` and it provides properties of levelDB.

storage/database/sharded_database.go
```diff
@@ -428,6 +428,15 @@ func (db *shardedDB) Meter(prefix string) {
 	}
 }
 
+func (db *shardedDB) GetProperty(name string) string {
+	var buf bytes.Buffer
+	for index, shard := range db.shards {
+		buf.WriteString(fmt.Sprintf("shard %d:\n", index))
+		buf.WriteString(shard.GetProperty(name))
```
aidan-kwon, line 435: The Shard index and its property will be written in different lines. Is that intended? For me,  
`buf.WriteString(fmt.Sprintf("shard %d: %s\n", index, shard.GetProperty(name)))` looks more intuitive.

storage/database/rocksdb_database.go
```diff
@@ -0,0 +1,399 @@
...
+	dat, err := db.db.GetBytes(db.ro, key)
+	if dat == nil {
+		return nil, dataNotFoundErr
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return dat, nil
+}
```
aidan-kwon, line 201: Don't we need to call `db.db.TryCatchUpWithPrimary()` for this? The method isn't used anywhere, though.

aidan-kwon (commented):
Please note that the current RocksDB implementation has some limitations for endpoint clustering usage. It should be handled in the next update. 
1. The secondary cannot handle subscription APIs as described in the following `TODO` comment.
https://github.com/klaytn/klaytn/blob/c46219e889579984b861af3e1521e876181cc508/blockchain/blockchain.go#L2192
The following function is one solution used by KES for the same issue. 
https://github.com/klaytn/klaytn/blob/228dc06eceed121d0a36de59f0d2112fbf52b4f2/blockchain/blockchain.go#L2095
2. Since the refreshing time of `CurrentBlockUpdateLoop` is 1 second, endpoint users may experience inconsistency among API returns.

blockchain/blockchain.go
jeongkyun-oh: If head block hash is nil, then it will retrieve hash from backup hash. So, it has to check the further code.

node/cn/api.go
jeongkyun-oh: I will leave the `ChaindbProperty` and `ChaindbCompaction`, but add new `GetDBProperty` for rocksdb and leveldb.

storage/database/rocksdb_database.go
jeongkyun-oh: There is `Get` method. `get` method is in there. It won't be searched if you use IDE because of build tag I guess.

jeongkyun-oh:
> Please note that the current RocksDB implementation has some limitations for endpoint clustering usage. It should be handled in the next update.
> 
> 1. The secondary cannot handle subscription APIs as described in the following `TODO` comment.
>    https://github.com/klaytn/klaytn/blob/c46219e889579984b861af3e1521e876181cc508/blockchain/blockchain.go#L2192
>    
>    The following function is one solution used by KES for the same issue.
>    https://github.com/klaytn/klaytn/blob/228dc06eceed121d0a36de59f0d2112fbf52b4f2/blockchain/blockchain.go#L2095
> 2. Since the refreshing time of `CurrentBlockUpdateLoop` is 1 second, endpoint users may experience inconsistency among API returns.

aidan-kwon (approved):

aidan-kwon:
Reviewers, PTAL

blukat29 (dismissed):

cmd/utils/flags.go
```diff
@@ -106,6 +106,12 @@ var (
 		Usage:  "Data directory for the databases and keystore. This value is only used in local DB.",
 		EnvVar: "KLAYTN_DATADIR",
 	}
+	ChainDataDirFlag = DirectoryFlag{
+		Name:   "chaindatadir",
+		Value:  DirectoryString{""},
+		Usage:  "Data directory for chaindata. If this is not specified, chaindata is stored in datadir",
```
kjeom, line 112: If this flag and rocksdb.secondary are set, does the node use only RocksDB?
A bit confused for me. If it is true, I think it's better to make a mode flag like KES.

Overall this PR, I can see only checking whether the RocksDB is used or not rather than checking something like a mode for the auto-scaling.

jeongkyun-oh: @kjeom I think we can consider `ChainDataDir` flag to make data separately apart from other data (e.g. nodes, klay.ipc...). No need to consider rocksdb for this flag. And what do you mean by mode for the auto-scaling?

kjeom: Auto-scaling means what we named scalable endpoint.
Anyway now I understood the primary, read-only and secondary are options only for RocksDB.
I confused it as it is concerned with the primary and secondary for Klaytn.
Thanks

kjeom (dismissed):

kjeom:
@ethan-kr Please take a look

JayChoi1736:
because in service chain or other private chain , "ken init" will generate automatically generate chaindata with given genesis.json in datadir.
Therefore, i think we need to put ChainDataDirFlag in InitCommand in cmd/utils/nodecmd/chaincmd.go

jeongkyun-oh:
> 

@JayChoi1736 Good point!! Thanks for reviewing. Apart from `init` command, I think we need to add this flags to several places regarding to `DataDir` flag. I will add this in another PR.

aidan-kwon:
@ethan-kr Pleas take a look

JayChoi1736 (approved):

aidan-kwon (approved):

blukat29 (approved):

JayChoi1736 (approved):

kjeom (approved):

yoomee1313 (approved):

PR #1871, opened 2023-06-26
KIP-111 Prune obsolete trie nodes

blukat29:
## Proposed changes

See [KIP-111](https://github.com/ethan-kr/kips/blob/main/KIPs/kip-111.md) for the background.

Usage:

```bash
# init custom genesis and sync
kcn init --state.live-pruning genesis.json
kcn   # after init, --state.live-pruning is optional

# cypress sync
ken --state.live-pruning  # first run, when database is clean.
ken   # from the second run, --state.live-pruning is optional

# temporarily disable live pruning
ken --state.live-pruning-retention 0
```

Changes:

- New database schema
  - (MiscDB) `PruningEnabled: ["PruningEnabled"] -> true` is a database-stored flag that tells the trie nodes are in prunable format (i.e. with ExtHash).
  - (MiscDB) `PruningMark: ["Pruning-" + blockNum + ExtHash] -> true` marks an obsolete trie node that's not yet deleted.

- New CLI flags
  - `kcn init --state.live-pruning`
    - if specified, write PruningEnabled to database.
  - `kcn --state.live-pruning`
    - if specified and database is clean, write PruningEnabled to database.
    - After the first writing PruningEnabled, this flag becomes optional.
  - `kcn --state.live-pruning-retention <172800>`
    - Number of blocks to keep obsolete nodes before deleting. Defaults to 2 days (172800).
    - Set 0 to disable live pruning.
    - Ignored if the database has pruning disabled.

- New struct fields
	```
	cn.Config.LivePruning
	cn.Config.LivePruningRetention
	blockchain.CacheConfig.LivePruningRetention
	statedb.TrieOpts.PruningBlockNumber uint64        // replaces  TrieOpts.Pruning bool
	```

- Pruning actions
  - `Trie.insert` and `Trie.delete` marks obsolete nodes as PruningMarks, only if:
    - PruningEnabled in database AND
    - The Trie is created with `PruningBlockNumber != 0` (i.e. through `PrunableStateAt()`)
  - `blockchain.insertChain` uses `PrunableStateAt()`
  - `worker.makeCurrent` uses `PrunableStateAt()`
  - `blockchain.writeStateTrie` asynchronously deletes trie nodes according to PruningMarks.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -251,8 +251,8 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 		cacheConfig:        cacheConfig,
 		db:                 db,
 		triegc:             prque.New(),
-		chBlock:            make(chan gcBlock, 1000),
-		chPrune:            make(chan uint64, 1000),
+		chBlock:            make(chan gcBlock, 2048), // downloader.maxResultsProcess
+		chPrune:            make(chan uint64, 2048),  // downloader.maxResultsProcess
 		stateCache:         state.NewDatabaseWithNewCache(db, cacheConfig.TrieNodeCacheConfig),
 		quit:               make(chan struct{}),
 		futureBlocks:       futureBlocks,
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -107,9 +107,10 @@ const (
 )
 
 const (
-	DefaultTriesInMemory = 128
-	DefaultBlockInterval = 128
-	MaxPrefetchTxs       = 20000
+	DefaultTriesInMemory        = 128
+	DefaultBlockInterval        = 128
+	DefaultLivePruningRetention = 172800
+	MaxPrefetchTxs              = 20000
 
 	// BlockChainVersion ensures that an incompatible database forces a resync from scratch.
 	// Changelog:
@@ -127,6 +128,7 @@ type CacheConfig struct {
 	CacheSize            int                          // Size of in-memory cache of a trie (MiB) to flush matured singleton trie nodes to disk
 	BlockInterval        uint                         // Block interval to flush the trie. Each interval state trie will be flushed into disk
 	TriesInMemory        uint64                       // Maximum number of recent state tries according to its block number
+	LivePruningRetention uint64                       // Number of blocks before trie nodes in pruning marks to be deleted. If zero, obsolete nodes are not deleted.
 	SenderTxHashIndexing bool                         // Enables saving senderTxHash to txHash mapping information to database and cache
 	TrieNodeCacheConfig  *statedb.TrieNodeCacheConfig // Configures trie node cache
 	SnapshotCacheSize    int                          // Memory allowance (MB) to use for caching snapshot entries in memory
@@ -161,6 +163,7 @@ type BlockChain struct {
 	snaps   *snapshot.Tree     // Snapshot tree for fast trie leaf access
 	triegc  *prque.Prque       // Priority queue mapping block numbers to tries to gc
 	chBlock chan gcBlock       // chPushBlockGCPrque is a channel for delivering the gc item to gc loop.
+	chPrune chan uint64        // chPrune is a channel for delivering the current block number for pruning loop.
 
 	hc            *HeaderChain
 	rmLogsFeed    event.Feed
@@ -224,13 +227,14 @@ type prefetchTx struct {
 func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, engine consensus.Engine, vmConfig vm.Config) (*BlockChain, error) {
 	if cacheConfig == nil {
 		cacheConfig = &CacheConfig{
-			ArchiveMode:         false,
-			CacheSize:           512,
-			BlockInterval:       DefaultBlockInterval,
-			TriesInMemory:       DefaultTriesInMemory,
-			TrieNodeCacheConfig: statedb.GetEmptyTrieNodeCacheConfig(),
-			SnapshotCacheSize:   512,
-			SnapshotAsyncGen:    true,
+			ArchiveMode:          false,
+			CacheSize:            512,
+			BlockInterval:        DefaultBlockInterval,
+			TriesInMemory:        DefaultTriesInMemory,
+			LivePruningRetention: DefaultLivePruningRetention,
+			TrieNodeCacheConfig:  statedb.GetEmptyTrieNodeCacheConfig(),
+			SnapshotCacheSize:    512,
+			SnapshotAsyncGen:     true,
 		}
 	}
 
@@ -248,6 +252,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 		db:                 db,
 		triegc:             prque.New(),
 		chBlock:            make(chan gcBlock, 1000),
+		chPrune:            make(chan uint64, 1000),
 		stateCache:         state.NewDatabaseWithNewCache(db, cacheConfig.TrieNodeCacheConfig),
 		quit:               make(chan struct{}),
 		futureBlocks:       futureBlocks,
@@ -353,6 +358,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 	// Take ownership of this particular state
 	go bc.update()
 	bc.gcCachedNodeLoop()
+	bc.pruneTrieNodeLoop()
 	bc.restartStateMigration()
 
 	if cacheConfig.TrieNodeCacheConfig.DumpPeriodically() {
@@ -685,6 +691,18 @@ func (bc *BlockChain) State() (*state.StateDB, error) {
 // StateAt returns a new mutable state based on a particular point in time.
 func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
 	return state.New(root, bc.stateCache, bc.snaps, nil)
+}
+
+// PrunableStateAt returns a new mutable state based on a particular point in time.
+// If live pruning is enabled on the databse, and num is nonzero, then trie will mark obsolete nodes for pruning.
+func (bc *BlockChain) PrunableStateAt(root common.Hash, num uint64) (*state.StateDB, error) {
+	if bc.db.ReadPruningEnabled() && bc.cacheConfig.LivePruningRetention != 0 {
+		return state.New(root, bc.stateCache, bc.snaps, &statedb.TrieOpts{
+			PruningBlockNumber: num,
+		})
+	} else {
+		return bc.StateAt(root)
+	}
 }
 
 // StateAtWithPersistent returns a new mutable state based on a particular point in time with persistent trie nodes.
@@ -1335,8 +1353,8 @@ func (bc *BlockChain) writeStateTrie(block *types.Block, state *state.StateDB) e
 					logger.Error("Error from trieDB.Cap by state migration", "err", err)
 				}
 			}
-
 			bc.lastCommittedBlock = block.NumberU64()
+			bc.chPrune <- block.NumberU64()
 		}
 
 		bc.chBlock <- gcBlock{root, block.NumberU64()}
@@ -1390,6 +1408,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 					cnt++
 				}
 				logger.Debug("GC cached node", "currentBlk", blkNum, "chosenBlk", chosen, "deferenceCnt", cnt)
+			case <-bc.quit:
+				return
+			}
+		}
+	}()
+}
+
+func (bc *BlockChain) pruneTrieNodeLoop() {
+	// ReadPruningMarks(1, limit) is very slow because it iterates over the most of MiscDB.
+	// ReadPruningMarks(start, limit) is much faster because it only iterates a small range.
+	startNum := uint64(1)
+
+	bc.wg.Add(1)
+	go func() {
+		defer bc.wg.Done()
+		for {
+			select {
+			case num := <-bc.chPrune:
+				if !bc.db.ReadPruningEnabled() || bc.cacheConfig.LivePruningRetention == 0 {
+					continue
+				}
+				if num <= bc.cacheConfig.LivePruningRetention {
+					continue
+				}
+				limit := num - bc.cacheConfig.LivePruningRetention // Prune [1, latest - retention]
+
+				startTime := time.Now()
+				marks := bc.db.ReadPruningMarks(startNum, limit+1)
+				bc.db.PruneTrieNodes(marks)
+				bc.db.DeletePruningMarks(marks)
+
+				logger.Info("Pruned trie nodes", "number", num, "start", startNum, "limit", limit,
+					"count", len(marks), "elapsed", time.Since(startTime))
+
+				startNum = limit + 1
 			case <-bc.quit:
 				return
 			}
@@ -1918,7 +1971,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 			return i, events, coalescedLogs, err
 		}
 
-		stateDB, err := bc.StateAt(parent.Root())
+		stateDB, err := bc.PrunableStateAt(parent.Root(), parent.NumberU64())
 		if err != nil {
 			return i, events, coalescedLogs, err
 		}
--- a/cmd/utils/nodecmd/chaincmd.go
+++ b/cmd/utils/nodecmd/chaincmd.go
@@ -133,6 +133,7 @@ func initGenesis(ctx *cli.Context) error {
 	singleDB := ctx.GlobalIsSet(utils.SingleDBFlag.Name)
 	numStateTrieShards := ctx.GlobalUint(utils.NumStateTrieShardsFlag.Name)
 	overwriteGenesis := ctx.GlobalBool(utils.OverwriteGenesisFlag.Name)
+	livePruning := ctx.GlobalBool(utils.LivePruningFlag.Name)
 
 	dbtype := database.DBType(ctx.GlobalString(utils.DbTypeFlag.Name)).ToValid()
 	if len(dbtype) == 0 {
@@ -175,9 +176,9 @@ func initGenesis(ctx *cli.Context) error {
 		}
 
 		// Write the live pruning flag to database
-		if ctx.Bool(utils.LivePruningFlag.Name) {
+		if livePruning {
 			chainDB.WritePruningEnabled()
-			logger.Info("Live pruning enabled")
+			logger.Info("Enabling live pruning")
 		}
 
 		logger.Info("Successfully wrote genesis state", "database", name, "hash", hash.String())
--- a/cmd/utils/nodecmd/chaincmd.go
+++ b/cmd/utils/nodecmd/chaincmd.go
@@ -58,6 +58,7 @@ var (
 			utils.LevelDBCompressionTypeFlag,
 			utils.DataDirFlag,
 			utils.OverwriteGenesisFlag,
+			utils.LivePruningFlag,
 		},
 		Category: "BLOCKCHAIN COMMANDS",
 		Description: `
@@ -171,6 +172,12 @@ func initGenesis(ctx *cli.Context) error {
 		gov := governance.NewMixedEngineNoInit(genesis.Config, chainDB)
 		if err := gov.WriteGovernance(0, govSet, governance.NewGovernanceSet()); err != nil {
 			logger.Crit("Failed to write governance items", "err", err)
+		}
+
+		// Write the live pruning flag to database
+		if ctx.Bool(utils.LivePruningFlag.Name) {
+			chainDB.WritePruningEnabled()
+			logger.Info("Live pruning enabled")
 		}
 
 		logger.Info("Successfully wrote genesis state", "database", name, "hash", hash.String())
--- a/node/cn/backend.go
+++ b/node/cn/backend.go
@@ -271,6 +271,20 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 	}
 	bc.SetCanonicalBlock(config.StartBlockNumber)
 
+	// Write the live pruning flag to database if the node is started for the first time
+	if config.LivePruning && !chainDB.ReadPruningEnabled() {
+		if bc.CurrentBlock().NumberU64() > 0 {
+			return nil, errors.New("cannot enable live pruning after chain has advanced")
+		}
+		chainDB.WritePruningEnabled()
+		logger.Info("Enabling live pruning")
+	}
+	// Live pruning is enabled according to the flag in database
+	// regardless of the command line flag --state.live-pruning
+	if chainDB.ReadPruningEnabled() {
+		logger.Info("Live pruning is enabled")
+	}
+
 	cn.blockchain = bc
 	governance.SetBlockchain(cn.blockchain)
 	if err := governance.UpdateParams(cn.blockchain.CurrentBlock().NumberU64()); err != nil {
--- a/node/cn/backend.go
+++ b/node/cn/backend.go
@@ -281,7 +281,8 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 	}
 	// Live pruning is enabled according to the flag in database
 	// regardless of the command line flag --state.live-pruning
-	if chainDB.ReadPruningEnabled() {
+	// But live pruning is disabled when --state.live-pruning-retention 0
+	if chainDB.ReadPruningEnabled() && config.LivePruningRetention != 0 {
 		logger.Info("Live pruning is enabled")
 	}
 
--- a/node/cn/backend.go
+++ b/node/cn/backend.go
@@ -253,9 +253,15 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 	var (
 		vmConfig    = config.getVMConfig()
 		cacheConfig = &blockchain.CacheConfig{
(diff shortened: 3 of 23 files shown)


blukat29:
Cypress chain import performance test
- Used `admin_importChain` with Cypress blocks 0 to 50,000,000
- chaindata size overall 70% down = statetrie 87% down, miscdb 200% up
- insertion time overall 30% down = processTxs 50% down, others 20~30% up

chaindata_size
insertion_time

blockchain/blockchain.go
```diff
@@ -248,6 +252,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig
 		db:                 db,
 		triegc:             prque.New(),
 		chBlock:            make(chan gcBlock, 1000),
+		chPrune:            make(chan uint64, 1000),
```
aidan-kwon, line 255: The default block chunk size from downloader is 2048. If the channel size is 2048, this channel will not be blocked during the block chunk processing.

node/cn/backend.go
```diff
@@ -265,6 +271,20 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
...
+	if config.LivePruning && !chainDB.ReadPruningEnabled() {
+		if bc.CurrentBlock().NumberU64() > 0 {
+			return nil, errors.New("cannot enable live pruning after chain has advanced")
+		}
+		chainDB.WritePruningEnabled()
+		logger.Info("Enabling live pruning")
+	}
+	// Live pruning is enabled according to the flag in database
+	// regardless of the command line flag --state.live-pruning
+	if chainDB.ReadPruningEnabled() {
```
aidan-kwon, line 284: We may check whether `LivePruningRetention` is zero or not because live pruning doesn't accept zero retention setting.

storage/database/db_manager.go
```diff
@@ -1863,6 +1873,12 @@ func (dbm *databaseManager) PutTrieNodeToBatch(batch Batch, hash common.ExtHash,
 	}
 }
 
+func (dbm *databaseManager) DeleteTrieNode(hash common.ExtHash) {
```
aidan-kwon, line 1876:
```suggestion
// DeleteTrieNode deletes a trie node having a specific hash. It is used only for testing. 
func (dbm *databaseManager) DeleteTrieNode(hash common.ExtHash) {
```

blockchain/blockchain.go
```diff
@@ -1397,6 +1415,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 	}()
 }
 
+func (bc *BlockChain) pruneTrieNodeLoop() {
+	// ReadPruningMarks(1, limit) is very slow because it iterates over the most of MiscDB.
```
aidan-kwon, line 1419: Is this an acceptable speed? Don't we need to apply more fast retrieval methods such as binary search between [1, latest]?

blockchain/blockchain.go
blukat29: Both `chBlock` and `chPrune` are used when not in archive mode, meaning the channel is stuffed every other 128 blocks. But I will increase them to 2048 anyway because that's not too big.

node/cn/backend.go
blukat29: Thanks. Fixed in 05a54f1.

storage/database/db_manager.go
blukat29: Thanks. Fixed in 05a54f1.

storage/database/db_manager.go
```diff
@@ -1882,9 +1898,87 @@ func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.H
...
-func (dbm *databaseManager) DeleteTrieNode(hash common.ExtHash) {
-	if err := dbm.getDatabase(StateTrieDB).Delete(TrieNodeKey(hash)); err != nil {
-		logger.Crit("Failed to delete trie node", "err", err)
+func (dbm *databaseManager) ReadPruningEnabled() bool {
+	ok, _ := dbm.getDatabase(MiscDB).Has(pruningEnabledKey)
+	return ok
+}
+
+func (dbm *databaseManager) WritePruningEnabled() {
+	if err := dbm.getDatabase(MiscDB).Put(pruningEnabledKey, []byte("42")); err != nil {
```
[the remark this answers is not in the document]
blukat29: It's a meaningless nonempty value. I followed WriteSnapshotDisabled (https://github.com/klaytn/klaytn/blob/v1.10.2/storage/database/db_manager.go#L2196). Same in go-ethereum.

blockchain/blockchain.go
blukat29: Acceptable to me. Before having `startNum` it took around 3 seconds per loop and most of the time was spent in ReadPruningMarks. But now it only happens at the early stage of the program.

blukat29:
@aidan-kwon Fixed the potential concurrency issue regarding `insertPruningMark()`. The treatment is something similar to what `insertPreimage()` does. I think the performance impact is not too serious because `db.lock` is acquired only at `Trie.Commit()`. `Trie.Commit()` happens [1 + number of contract accounts changed] times per block.

blockchain/blockchain.go
blukat29: Update: Tested with a node with an existing database. The first pruning loop right after the program started took 1m17s. After that, it takes around 50ms every 128 blocks.

Considering that the pruning loop runs in the background, I think 1 minute is tolerable.
```
INFO[07/25,06:45:00 Z] [5] Pruned trie nodes                         number=49000064 start=1 limit=48827264 count=12795 elapsed=1m17.616352458s
INFO[07/25,06:45:00 Z] [5] Pruned trie nodes                         number=49000192 start=48827265 limit=48827392 count=17332 elapsed=55.195561ms
INFO[07/25,06:45:00 Z] [5] Pruned trie nodes                         number=49000320 start=48827393 limit=48827520 count=15679 elapsed=63.148902ms
INFO[07/25,06:45:00 Z] [5] Pruned trie nodes                         number=49000448 start=48827521 limit=48827648 count=17534 elapsed=46.980492ms
```

blockchain/blockchain.go
```diff
@@ -107,9 +107,10 @@ const (
 )
 
 const (
-	DefaultTriesInMemory = 128
-	DefaultBlockInterval = 128
-	MaxPrefetchTxs       = 20000
+	DefaultTriesInMemory        = 128
+	DefaultBlockInterval        = 128
+	DefaultLivePruningRetention = 172800
```
jeongkyun-oh, line 112: For my curiosity, where does this number come from?

blockchain/blockchain.go
```diff
@@ -247,7 +251,8 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig
 		cacheConfig:        cacheConfig,
 		db:                 db,
 		triegc:             prque.New(),
-		chBlock:            make(chan gcBlock, 1000),
+		chBlock:            make(chan gcBlock, 2048), // downloader.maxResultsProcess
+		chPrune:            make(chan uint64, 2048),  // downloader.maxResultsProcess
```
jeongkyun-oh, line 255: Are these channels related to downloader?

blockchain/blockchain.go
jeongkyun-oh: Oh, 86400 * 2 ! it would be nice to leave a comment for this, but never mind.

blockchain/blockchain.go
blukat29: @jeongkyun-oh See https://github.com/klaytn/klaytn/pull/1871#discussion_r1272376894

blockchain/blockchain.go
```diff
@@ -1397,6 +1415,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
...
+	// ReadPruningMarks(1, limit) is very slow because it iterates over the most of MiscDB.
+	// ReadPruningMarks(start, limit) is much faster because it only iterates a small range.
+	startNum := uint64(1)
+
+	bc.wg.Add(1)
+	go func() {
+		defer bc.wg.Done()
+		for {
+			select {
+			case num := <-bc.chPrune:
```
jeongkyun-oh, line 1428: Do we have to get notified on every block? How about subscribing `ChainHeadEvent` which is a latest block event after inserting blocks?

blockchain/blockchain.go
aidan-kwon: @blukat29 Thanks for the update. 1 min at the first booting sounds acceptable to me as well. For a better understanding, be aware that heavy disk read will be conducted at that time.

aidan-kwon (approved):

ian0371 (approved):

blockchain/blockchain.go
jeongkyun-oh: Well, it  doesn't really matter if it is not possible that `bc.chPrune` is full. And it won't be full since the logics are only related to db read/write only. Let me resolve this.

jeongkyun-oh (approved):

aidan-kwon:
@2dvorak @hyunsooda @yoomee1313 PTAL

blockchain/blockchain.go
blukat29: FYI, `bc.chPrune` is notified only when `isCommitTrieRequired(bc, block.NumberU64())` - in every other 128 blocks.

node/cn/backend.go
```diff
@@ -265,6 +271,21 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
...
+		}
+		chainDB.WritePruningEnabled()
+		logger.Info("Enabling live pruning")
+	}
+	// Live pruning is enabled according to the flag in database
+	// regardless of the command line flag --state.live-pruning
+	// But live pruning is disabled when --state.live-pruning-retention 0
+	if chainDB.ReadPruningEnabled() && config.LivePruningRetention != 0 {
+		logger.Info("Live pruning is enabled")
+	}
```
2dvorak, line 287: @blukat29 We could consider logging when live pruning is disabled because of `--state.live-pruning-retention` set to 0 although `--state.live-pruning` was set?

Also, two consecutive logs with almost the same meaning could be confusing, although their internal semantic differs.
```
Enabling live pruning
Live pruning is enabled
```

These are not major concerns (all log-related), so I'm approving this PR. Thanks

2dvorak (approved):

blockchain/blockchain.go
jeongkyun-oh: Oh, okay. Thanks.

node/cn/backend.go
blukat29: I will refine in another PR
```
Writing live pruning flag to database
Live pruning is enabled
Live pruning is disabled (no flag in database)
Live pruning is disbaled (retention set to zero)
```

yoomee1313 (approved):

blockchain/blockchain.go
```diff
@@ -1397,6 +1415,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 	}()
 }
 
+func (bc *BlockChain) pruneTrieNodeLoop() {
+	// ReadPruningMarks(1, limit) is very slow because it iterates over the most of MiscDB.
+	// ReadPruningMarks(start, limit) is much faster because it only iterates a small range.
+	startNum := uint64(1)
```
hyunsooda, line 1421: This is a minor comment.

By replacing the hardcoded value with a command line flag, we can preserve certain ranges. This straightforward modification allows for the retention of specific old states that are required. Implementing pruning in conjunction with archive mode would further enhance the effectiveness of the process. Compared to the standard practice of committing every 128 blocks (default interval)

blockchain/blockchain.go
```diff
@@ -1397,6 +1415,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
...
+
+	bc.wg.Add(1)
```
hyunsooda, line 1423: There's no `Wait()` call. What it has an effect without this,?

blockchain/blockchain.go
```diff
@@ -1397,6 +1415,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
...
+				if !bc.db.ReadPruningEnabled() || bc.cacheConfig.LivePruningRetention == 0 {
+					continue
+				}
+				if num <= bc.cacheConfig.LivePruningRetention {
+					continue
+				}
+				limit := num - bc.cacheConfig.LivePruningRetention // Prune [1, latest - retention]
+
+				startTime := time.Now()
+				marks := bc.db.ReadPruningMarks(startNum, limit+1)
```
hyunsooda, line 1438: The tasks below can be invoked only when effective marks are available. Then, `len(makrs) > 0`  can wrap them (L1439-1443). How does it look?

blockchain/blockchain.go
```diff
@@ -1397,6 +1415,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
...
+				bc.db.PruneTrieNodes(marks)
+				bc.db.DeletePruningMarks(marks)
+
+				logger.Info("Pruned trie nodes", "number", num, "start", startNum, "limit", limit,
+					"count", len(marks), "elapsed", time.Since(startTime))
```
hyunsooda, line 1443:
```suggestion
				logger.Info("Pruned trie nodes", "from", startNum, "to", limit,
					"count", len(marks), "elapsed", time.Since(startTime))
```
Does it feel more informative to you? (Additionally, the `number` seems not informative in this log)

cmd/utils/nodecmd/chaincmd.go
```diff
@@ -173,6 +175,12 @@ func initGenesis(ctx *cli.Context) error {
 			logger.Crit("Failed to write governance items", "err", err)
 		}
 
+		// Write the live pruning flag to database
+		if livePruning {
+			chainDB.WritePruningEnabled()
+			logger.Info("Enabling live pruning")
+		}
```
hyunsooda, line 182: Pruning-enable store would be done at https://github.com/klaytn/klaytn/pull/1871/files#diff-de2bd0674bf71cd6e3d92cb5621577831308e4cc990a15933e5e853f3358e4e0R275-R281

Is it not duplicated?

storage/database/db_manager.go
```diff
@@ -168,6 +168,16 @@ type DBManager interface {
 	DeleteTrieNode(hash common.ExtHash)
 	WritePreimages(number uint64, preimages map[common.Hash][]byte)
 
+	// Trie pruning
+	ReadPruningEnabled() bool
+	WritePruningEnabled()
+	DeletePruningEnabled()
```
hyunsooda, line 174: A comment seems required that says this function must not be explicitly called. If not, the pruning process would not be available by https://github.com/klaytn/klaytn/pull/1871/files#diff-de2bd0674bf71cd6e3d92cb5621577831308e4cc990a15933e5e853f3358e4e0R277. Is it correct?

storage/database/db_manager.go
```diff
@@ -1882,9 +1899,87 @@ func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.H
...
+		logger.Crit("Failed to batch delete pruning mark", "err", err)
+	}
+}
+
+// PruneTrieNodes deletes the trie nodes according to the provided set of pruning marks.
+func (dbm *databaseManager) PruneTrieNodes(marks []PruningMark) {
+	batch := dbm.NewBatch(StateTrieDB)
+	for _, mark := range marks {
+		if err := batch.Delete(TrieNodeKey(mark.Hash)); err != nil {
+			logger.Crit("Failed to prune trie node", "err", err)
```
hyunsooda, line 1975: Instead of logging and exiting, using the Error level appears to be a more dependable approach. Although the database pruning operation failed for unknown reasons, it is my belief that it did not produce any side effects on other operations, such as block validation, RPC, etc. I would like to know your thoughts on this matter.

storage/database/schema.go
```diff
@@ -270,3 +276,31 @@ func TrieNodeKey(hash common.ExtHash) []byte {
...
+type PruningMark struct {
+	Number uint64
+	Hash   common.ExtHash
+}
+
+// TriePruningMarkKey = prefix + number + hash
+// Block number comes first to sort the entries by block numbers.
+// Later we can iterate through the marks and extract any given block number range.
+func pruningMarkKey(mark PruningMark) []byte {
+	bNumber := make([]byte, 8)
```
hyunsooda, line 289: Considerably used the magic number 8 . How about declaring a dedicated constant?

storage/statedb/database.go
```diff
@@ -751,12 +761,14 @@ func (db *Database) Cap(limit common.StorageSize) error {
...
+	numPreimages := 0
 	if flushPreimages {
-		if err := db.writeBatchPreimages(); err != nil {
-			db.lock.RUnlock()
-			return err
-		}
+		db.diskDB.WritePreimages(0, db.preimages)
+		numPreimages = len(db.preimages)
 	}
+	db.diskDB.WritePruningMarks(db.pruningMarks)
```
hyunsooda, line 769: `(64B + 39B) * # of dirty nodes` are continuously and unconditionally stored in MiscDB. 
Good for future use of pruning becuase these provide the exact obsolete node hash and delete them after lookup.
Bad for uncertain about future-use of pruning becauase these are continuously take disk space. The [provided numbers](https://github.com/klaytn/klaytn/pull/1871#issuecomment-1641223838) seem not to show this disk usage. How do you anticipate it?

hyunsooda (commented):
Learned a good implementation. For the minor question, How did you draw [the chart](https://github.com/klaytn/klaytn/pull/1871#issuecomment-1641223838)? Did you just simply lookup the directory size

blockchain/blockchain.go
blukat29: There is https://github.com/klaytn/klaytn/blob/320554640a13fbadf12d42db0a0f91c0edc14386/blockchain/blockchain.go#L1037C14-L1037C14.

storage/statedb/database.go
blukat29: When live pruning is disabled, no pruning marks are written. https://github.com/klaytn/klaytn/blob/320554640a13fbadf12d42db0a0f91c0edc14386/storage/statedb/trie.go#L605-L611

blockchain/blockchain.go
```diff
@@ -1397,6 +1415,41 @@ func (bc *BlockChain) gcCachedNodeLoop() {
...
+
+	bc.wg.Add(1)
+	go func() {
+		defer bc.wg.Done()
+		for {
+			select {
+			case num := <-bc.chPrune:
+				if !bc.db.ReadPruningEnabled() || bc.cacheConfig.LivePruningRetention == 0 {
+					continue
+				}
```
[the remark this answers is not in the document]
blukat29: Then I can move the conditional to following two sites:
- `bc.chPrune <- block.NumberU64()`
- `bc.pruneTrieNodeLoop()`

blockchain/blockchain.go
blukat29: That could be a useful feature - to preserve only a certain range of blocks - but we need to consider how this feature can be used before implementing it. Thanks.

blockchain/blockchain.go
blukat29: I think performance impact is small. I'd keep it for brevity.

blockchain/blockchain.go
blukat29: I left it because "Pruned trie nodes" log is printed out-of-order with respect to block insertion logs.

```
INFO[07/06,07:48:08 Z] [5] Inserted a new block                      number=9999871 hash=d2089b…c99ab6 txs=0     gas=0          elapsed=2.526ms      processTxs=1.307µs   finalize=712.298µs validateState=69.476µs  totalWrite=1.203515ms   trieWrite=983.898
INFO[07/06,07:48:09 Z] [49] Persisted trie from memory database       blockNum=9999872 updated nodes=6003   updated nodes size=2.64mB   time=65.080295ms  gcnodes=4632   gcsize=2.40mB   gctime=14.170282ms  livenodes=4735   livesize=2.58mB   preimages=1051
INFO[07/06,07:48:09 Z] [5] Inserted a new block                      number=9999872 hash=fbdef9…9208da txs=1     gas=57117      elapsed=69.023ms     processTxs=344.698µs finalize=987.294µs validateState=95.961µs  totalWrite=67.023094ms  trieWrite=66.7293
INFO[07/06,07:48:09 Z] [5] Inserted a new block                      number=9999873 hash=37990f…af3163 txs=2     gas=788464     elapsed=8.285ms      processTxs=2.123ms   finalize=2.001ms   validateState=121.781µs totalWrite=3.292306ms   trieWrite=2.84218
INFO[07/06,07:48:09 Z] [5] Inserted a new block                      number=9999874 hash=70ac3a…31be14 txs=1     gas=741619     elapsed=7.836ms      processTxs=2.290ms   finalize=1.815ms   validateState=111.841µs totalWrite=2.736272ms   trieWrite=2.46042
INFO[07/06,07:48:09 Z] [5] Pruned trie nodes                         number=9999872 start=9826945 limit=9827072 count=12894  elapsed=30.478143ms
INFO[07/06,07:48:09 Z] [5] Inserted a new block                      number=9999875 hash=065177…63dc39 txs=1     gas=57117      elapsed=12.240ms     processTxs=442.191µs finalize=1.295ms   validateState=106.877µs totalWrite=9.544038ms   trieWrite=1.26109
INFO[07/06,07:48:09 Z] [5] Inserted a new block                      number=9999876 hash=6251db…fa4e69 txs=0     gas=0          elapsed=2.754ms      processTxs=1.341µs   finalize=812.607µs validateState=69.993µs  totalWrite=1.271557ms   trieWrite=1.02655
INFO[07/06,07:48:09 Z] [5] Inserted a new block                      number=9999877 hash=bfafae…2820de txs=2     gas=114192     elapsed=4.168ms      processTxs=720.532µs finalize=1.129ms   validateState=100.133µs totalWrite=1.527832ms   trieWrite=1.29307

```

cmd/utils/nodecmd/chaincmd.go
blukat29: It is to support the first use-case

```
kcn init --state.live-pruning genesis.json
kcn 
```

storage/database/db_manager.go
blukat29: I agree that PruneTrieNodes failure is not critical. However, it's unlikely that only pruning-related writes fail and others succeed; and making such an exception would degrade the consistency of the database module. Thanks.

blockchain/blockchain.go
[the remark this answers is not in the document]
blukat29: See https://github.com/klaytn/klaytn/pull/1914.

storage/database/db_manager.go
```diff
@@ -1882,9 +1899,87 @@ func (dbm *databaseManager) WritePreimages(number uint64, preimages map[common.H
...
+	if err := batch.Write(); err != nil {
+		logger.Crit("Failed to batch write pruning mark", "err", err)
+	}
+}
+
+// ReadPruningMarks reads the pruning marks in the block number range [startNumber, endNumber).
+func (dbm *databaseManager) ReadPruningMarks(startNumber, endNumber uint64) []PruningMark {
+	prefix := pruningMarkPrefix
+	startKey := pruningMarkKey(PruningMark{startNumber, common.ExtHash{}})
+	it := dbm.getDatabase(MiscDB).NewIterator(prefix, startKey[len(prefix):])
```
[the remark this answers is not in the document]
blukat29: See https://github.com/klaytn/klaytn/pull/1914.

storage/database/db_manager.go
blukat29: See https://github.com/klaytn/klaytn/pull/1914.

storage/database/schema.go
blukat29: I tried to find an appropriate built-in constant for uint64 byte size but to no avail. There is an unaccepted feature request but not in standard packages: https://github.com/golang/go/issues/29982#issuecomment-572199921. Even golang PutUint64 uses hardcoded constant  https://github.com/golang/go/blob/go1.20.6/src/encoding/binary/binary.go#L184. Therefore, I shall keep it literal number 8.

node/cn/backend.go
```diff
@@ -265,6 +271,21 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 	}
 	bc.SetCanonicalBlock(config.StartBlockNumber)
 
+	// Write the live pruning flag to database if the node is started for the first time
+	if config.LivePruning && !chainDB.ReadPruningEnabled() {
+		if bc.CurrentBlock().NumberU64() > 0 {
+			return nil, errors.New("cannot enable live pruning after chain has advanced")
```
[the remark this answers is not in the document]
blukat29: Turning the live pruning on and off does seem possible. But this hypothetical feature has some consistency risks. Suppose a database is built without pruning, where all trie nodes are legacy nodes (i.e. no extension). If we enable the pruning, then obsolete legacy nodes will be deleted (some nodes under block 10000 in your example). Note that we added extensions to safely delete trie nodes, and in contrast, legacy nodes are not safely removable because they may be referenced from multiple tries. So I don't think it is easy to implement such a feature.

storage/database/schema.go
hyunsooda: `const size = unsafe.Sizeof(uint64(0))` was not fit it? This alternative seems exactly the same behavior. I guess it seems not to fit your favorite, shall we skip it?

blukat29: Oh, unsafe should do the trick. But I think the variable name `bNumber` should give the necessary hint. Yes, let's skip it.

PR #1891, opened 2023-07-21
Implement Vrank metrics and logs

ian0371:
## Proposed changes

We collect all commits between two preprepares.
```
collect      start                                     stop
---------------|--------------|------------|-------------|
state:     preprepared      prepared     commit      preprepared
```

Existing implementation discards "late commit messages" that we receive after turning into the "commit" state.
To receive those commit messages, we must define an independent data structure, hence `Vrank struct`.

#### Metrics
- target: the time elapsed between the preprepared time (start time) and the followings:
  - the first commit arrival time
  - the [2f+1]-th commit arrival time
  - the average arrival times from the first commit to the [2f+1]-th commit
  - the last commit before stop

See further comments for an examplery metric output.

#### Logs
We assess each validator in the committee by how fast it sent its commit message:
- early: received before threshold
- late: received after threshold
- not received

The threshold that divides early and late commits is defined as follows:
- threshold =  min( [2f+1]-th commit arrival time, preprepared time + 300ms )

This is logged as `bitmap`.
In addition, the arrival times of late commits are logged as `late`.

See further comments for log interpretation.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments
Example metric output:
```
klaytn_vrank_avg_commit_within_quorum 1.944361e+06
klaytn_vrank_first_commit 1.520958e+06
klaytn_vrank_last_commit 6.891958e+06
klaytn_vrank_quorum_commit 2.290458e+06
```

Example log:
```
INFO[07/21,22:12:16 +09] [25|consensus/istanbul/core/vrank.go:118]  VRank     bitmap=01 late=[2]
```
It means that the fourth validator in the committee is late, and that its arrival time is 2ms.
Interpretation of bitmap: 01 => 0x01 => 0b00_00_00_01 => fourth 2-bit is `vrankArrivedLate`, others are `vrankArrivedEarly`.

Q. Why compress the log?
A. We wanted to reduce the log size as much as possible and reduce verbosity. If the threshold is [2f+1]-th commit arrival time, the size of `late` is `f`. If the threshold is preprepared time + 300ms, the size of `late` is more than `f`.

Q. Why not use existing variables such as `consensusTimestamp`, or `currentView`?
A. They are reset at `startNewRound()`, which is called almost immediately after turning into the "commit" state.

Q. Why threshold is 300ms?
A. In an ideal network, a validator is supposed to receive 2f+1 number of commits within 300ms.

### DIFF
--- a/consensus/istanbul/core/commit.go
+++ b/consensus/istanbul/core/commit.go
@@ -78,8 +78,8 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 		return errInvalidMessage
 	}
 
-	if Vrank != nil {
-		Vrank.AddCommit(commit, src)
+	if vrank != nil {
+		vrank.AddCommit(commit, src)
 	}
 
 	// logger.Error("receive handle commit","num", commit.View.Sequence)
@@ -154,7 +154,7 @@ func (c *core) acceptCommit(msg *message, src istanbul.Validator) error {
 	}
 
 	if c.current.Commits.Size() == 1 {
-		vrankFirstCommitArrivalTimeGauge.Update(int64(Vrank.TimeSinceStart()))
+		vrankFirstCommitArrivalTimeGauge.Update(int64(vrank.TimeSinceStart()))
 	}
 
 	return nil
--- a/consensus/istanbul/core/commit.go
+++ b/consensus/istanbul/core/commit.go
@@ -78,6 +78,10 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 		return errInvalidMessage
 	}
 
+	if Vrank != nil {
+		Vrank.AddCommit(commit, src)
+	}
+
 	// logger.Error("receive handle commit","num", commit.View.Sequence)
 	if err := c.checkMessage(msgCommit, commit.View); err != nil {
 		// logger.Error("### istanbul/commit.go checkMessage","num",commit.View.Sequence,"err",err)
@@ -147,6 +151,10 @@ func (c *core) acceptCommit(msg *message, src istanbul.Validator) error {
 	if err := c.current.Commits.Add(msg); err != nil {
 		logger.Error("Failed to record commit message", "msg", msg, "err", err)
 		return err
+	}
+
+	if c.current.Commits.Size() == 1 {
+		vrankFirstCommitArrivalTimeGauge.Update(int64(Vrank.TimeSinceStart()))
 	}
 
 	return nil
--- a/consensus/istanbul/core/commit.go
+++ b/consensus/istanbul/core/commit.go
@@ -153,7 +153,7 @@ func (c *core) acceptCommit(msg *message, src istanbul.Validator) error {
 		return err
 	}
 
-	if c.current.Commits.Size() == 1 {
+	if c.current.Commits.Size() == 1 && vrank != nil {
 		vrankFirstCommitArrivalTimeGauge.Update(int64(vrank.TimeSinceStart()))
 	}
 
--- a/consensus/istanbul/core/vrank.go
+++ b/consensus/istanbul/core/vrank.go
@@ -97,8 +97,10 @@ func (v *Vrank) HandleCommitted(blockNum *big.Int) {
 	for _, v := range v.commitArrivalTimeMap {
 		sum += int64(v)
 	}
-	avg := sum / int64(len(v.commitArrivalTimeMap))
-	vrankAvgCommitArrivalTimeWithinQuorumGauge.Update(avg)
+	if len(v.commitArrivalTimeMap) != 0 {
+		avg := sum / int64(len(v.commitArrivalTimeMap))
+		vrankAvgCommitArrivalTimeWithinQuorumGauge.Update(avg)
+	}
 }
 
 // Log logs accumulated data in a compressed form
--- a/consensus/istanbul/core/vrank.go
+++ b/consensus/istanbul/core/vrank.go
@@ -43,7 +43,7 @@ var (
 	vrankAvgCommitArrivalTimeWithinQuorumGauge = metrics.NewRegisteredGauge("vrank/avg_commit_within_quorum", nil)
 	vrankLastCommitArrivalTimeGauge            = metrics.NewRegisteredGauge("vrank/last_commit", nil)
 
-	vrankDefaultThreshold = "300ms"
+	vrankDefaultThreshold = "300ms" // the time to receive 2f+1 commits in an ideal network
 
 	vrank *Vrank
 )
--- a/consensus/istanbul/core/vrank.go
+++ b/consensus/istanbul/core/vrank.go
@@ -183,7 +183,7 @@ func compress(arr []int) []byte {
 	}
 
 	// pad zero to make len(arr)%4 == 0
-	for i := 0; i < 4-len(arr)%4; i++ {
+	for len(arr)%4 != 0 {
 		arr = append(arr, 0)
 	}
 
--- a/consensus/istanbul/core/vrank.go
+++ b/consensus/istanbul/core/vrank.go
@@ -101,23 +101,30 @@ func (v *Vrank) HandleCommitted(blockNum *big.Int) {
 	}
 }
 
-// Log logs accumulated data in a compressed form
-func (v *Vrank) Log() (string, []string) {
-	var (
-		serialized = serialize(v.committee, v.commitArrivalTimeMap)
-		assessed   = assessBatch(serialized, v.threshold)
-		compressed = compress(assessed)
-		bitmap     = hex.EncodeToString(compressed)
+func (v *Vrank) Bitmap() string {
+	serialized := serialize(v.committee, v.commitArrivalTimeMap)
+	assessed := assessBatch(serialized, v.threshold)
+	compressed := compress(assessed)
+	return hex.EncodeToString(compressed)
+}
 
-		lateCommits = make([]time.Duration, 0)
-		lastCommit  = time.Duration(0)
-	)
-
+func (v *Vrank) LateCommits() []time.Duration {
+	serialized := serialize(v.committee, v.commitArrivalTimeMap)
+	lateCommits := make([]time.Duration, 0)
 	for _, t := range serialized {
 		if assess(t, v.threshold) == vrankArrivedLate {
 			lateCommits = append(lateCommits, t)
 		}
 	}
+	return lateCommits
+}
+
+// Log logs accumulated data in a compressed form
+func (v *Vrank) Log() {
+	var (
+		lastCommit  = time.Duration(0)
+		lateCommits = v.LateCommits()
+	)
 
 	// lastCommit = max(lateCommits)
 	for _, t := range lateCommits {
@@ -129,14 +136,11 @@ func (v *Vrank) Log() (string, []string) {
 		vrankLastCommitArrivalTimeGauge.Update(int64(lastCommit))
 	}
 
-	lateCommitsStrArr := encodeDurationBatch(lateCommits)
-
 	logger.Info("VRank", "seq", v.view.Sequence.Int64(),
 		"round", v.view.Round.Int64(),
-		"bitmap", bitmap,
-		"late", lateCommitsStrArr,
+		"bitmap", v.Bitmap(),
+		"late", encodeDurationBatch(lateCommits),
 	)
-	return bitmap, lateCommitsStrArr
 }
 
 func (v *Vrank) isTargetCommit(msg *istanbul.Subject, src istanbul.Validator) bool {
@@ -221,7 +225,7 @@ func compress(arr []uint8) []byte {
 }
 
 // encodeDuration encodes given duration into string
-// The returned string is at most 3 bytes
+// The returned string is at most 4 bytes
 func encodeDuration(d time.Duration) string {
 	if d > 10*time.Second {
 		return fmt.Sprintf("%.0fs", d.Seconds())
--- a/consensus/istanbul/core/vrank.go
+++ b/consensus/istanbul/core/vrank.go
@@ -93,11 +93,11 @@ func (v *Vrank) HandleCommitted(blockNum *big.Int) {
 	}
 
 	vrankQuorumCommitArrivalTimeGauge.Update(int64(committedTime))
-	sum := int64(0)
-	for _, v := range v.commitArrivalTimeMap {
-		sum += int64(v)
-	}
 	if len(v.commitArrivalTimeMap) != 0 {
+		sum := int64(0)
+		for _, v := range v.commitArrivalTimeMap {
+			sum += int64(v)
+		}
 		avg := sum / int64(len(v.commitArrivalTimeMap))
 		vrankAvgCommitArrivalTimeWithinQuorumGauge.Update(avg)
 	}
--- a/consensus/istanbul/core/vrank.go
+++ b/consensus/istanbul/core/vrank.go
@@ -73,13 +73,11 @@ func (v *Vrank) TimeSinceStart() time.Duration {
 	return time.Now().Sub(v.startTime)
 }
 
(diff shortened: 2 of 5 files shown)


consensus/istanbul/core/vrank_test.go
```diff
@@ -0,0 +1,128 @@
...
+	committee := []istanbul.Validator{}
+	for _, addr := range addrs {
+		committee = append(committee, validator.New(addr))
+	}
+	return committee
+}
+
+func TestVrank(t *testing.T) {
+	var (
+		N         = 6
```
jiseongnoh, line 26: Given 3f+2 total validators, the quorum must be 2f+2. With six validators (3f+2, where f=1), the quorum should be four to maintain system integrity even with a single faulty validator.

consensus/istanbul/core/commit.go
```diff
@@ -78,6 +78,10 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 		return errInvalidMessage
 	}
 
+	if vrank != nil {
+		vrank.AddCommit(commit, src)
```
hyeonLewis, line 82: How about moving `AddCommit` after checking the commit? (After `valSet.CheckInSubList`) It might be an invalid commit since we only check `View` in `isTargetCommit`.

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,217 @@
...
+func assessBatch(ts []time.Duration, threshold time.Duration) []int {
+	ret := make([]int, 0)
+	for _, t := range ts {
+		ret = append(ret, assess(t, threshold))
+	}
+	return ret
+}
+
+// serialize serializes arrivalTime hashmap into array.
+// It needs to be deterministic, so the order of array is equal to that of sorted committee.
```
hyeonLewis, line 155: What does it mean? I'm trying to understand why we should sort the committee array.

consensus/istanbul/core/vrank_test.go
ian0371: I know but I wanted to test with a random quorum. If you think this is misleading, I'll change the quorum to four.

consensus/istanbul/core/commit.go
ian0371: We need to receive "late commits" in vrank. "late commits" can be filtered in `checkMessage()` call, so it needs to be up here.

An unwanted scenario:
- as soon as 2f+1 commits are received, the state becomes committed
- soon after, `startNewRound()` is called to increase the current view
- then, `checkMessage()` filters "late commits" because they have the previous view

consensus/istanbul/core/vrank.go
ian0371: It's to ensure that all CNs have the same committee order. It makes the parsing logic easier; we can simply sort the rpc output `klay.getCommittee()` to know which the validator position in the array.

ian0371: The comment has been also updated accordingly

hyeonLewis: Now I got the point. Thanks!

hyeonLewis: Now I got the point. Thanks!

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,218 @@
...
+			lastCommit = t
+		}
+	}
+	if lastCommit != time.Duration(0) {
+		vrankLastCommitArrivalTimeGauge.Update(int64(lastCommit))
+	}
+
+	lateCommitsStrArr := encodeDurationBatch(lateCommits)
+
+	logger.Info("VRank", "bitmap", bitmap, "late", lateCommitsStrArr)
```
blukat29, line 118: To make parsing simpler, it may also print the View.

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,218 @@
...
+	if v.threshold > committedTime {
+		v.threshold = committedTime
+	}
+
+	vrankQuorumCommitArrivalTimeGauge.Update(int64(committedTime))
+	sum := int64(0)
+	for _, v := range v.commitArrivalTimeMap {
+		sum += int64(v)
+	}
+	avg := sum / int64(len(v.commitArrivalTimeMap))
```
blukat29, line 84: Please, either add a comment below or check for zero. integer divide-by-zero is a panic.

```suggestion
	// commitArrivalTimeMap cannot be empty because transition into Committed state requires at least one commit message.
	avg := sum / int64(len(v.commitArrivalTimeMap))
```

consensus/istanbul/core/vrank.go
ian0371: good idea. added 46116f32

consensus/istanbul/core/vrank_test.go
ian0371: fixed in 913085d9

consensus/istanbul/core/vrank.go
ian0371: nice catch; I prefer safeness. fixed in 9df7a8c8

blukat29 (dismissed):

hyeonLewis (dismissed):

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+	if lastCommit != time.Duration(0) {
+		vrankLastCommitArrivalTimeGauge.Update(int64(lastCommit))
+	}
+
+	lateCommitsStrArr := encodeDurationBatch(lateCommits)
+
+	logger.Info("VRank", "seq", v.view.Sequence.Int64(),
+		"round", v.view.Round.Int64(),
+		"bitmap", bitmap,
+		"late", lateCommitsStrArr,
```
hyunsooda, line 139: I see that the bitmap serves as a compressed log of messages, detailing the status of individuals being early, late, or not sent, and it follows the order of the committee. This appears to be intelligent logging for storage save. However, I have some reservations regarding the absence of a convenient method to retrieve the complete set of vrank metrics in the specified sequence. Without such a method, developers might put additional work to make an intactly paired metrics.

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	_, ok := v.commitArrivalTimeMap[src.Address()]
+	if ok {
+		return false
+	}
+	return true
+}
+
+// assess determines if given time is early, late, or not arrived
+func assess(t, threshold time.Duration) int {
```
hyunsooda, line 156: How about changing the type `int` to `uint8`? Then, `vrankNotArrivedPlaceholder` should be changed accordingly.

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+
+	if t > threshold {
+		return vrankArrivedLate
+	} else {
+		return vrankArrivedEarly
+	}
+}
+
+func assessBatch(ts []time.Duration, threshold time.Duration) []int {
+	ret := make([]int, 0)
```
hyunsooda, line 169: Size initialization can be done with the length of `ts`

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+	logger.Info("VRank", "seq", v.view.Sequence.Int64(),
+		"round", v.view.Round.Int64(),
+		"bitmap", bitmap,
+		"late", lateCommitsStrArr,
+	)
+	return bitmap, lateCommitsStrArr
+}
+
+func (v *Vrank) isTargetCommit(msg *istanbul.Subject, src istanbul.Validator) bool {
+	if msg.View.Cmp(&v.view) != 0 {
```
yoomee1313, line 145: vrank.AddCommit is called before checkMessage is called, so msg.View can be nil, right? Then, this line will be panic if msg.View is nil.

consensus/istanbul/core/vrank_test.go
```diff
@@ -0,0 +1,144 @@
...
+	sort.Sort(committee)
+	for i := 0; i < quorum; i++ {
+		vrank.AddCommit(msg, committee[i])
+		expectedAssessList = append(expectedAssessList, vrankArrivedEarly)
+	}
+	vrank.HandleCommitted(view.Sequence)
+	for i := quorum; i < N; i++ {
+		t := vrank.AddCommit(msg, committee[i])
+		expectedAssessList = append(expectedAssessList, vrankArrivedLate)
+		expectedLateCommits = append(expectedLateCommits, t)
```
yoomee1313, line 63: What about removing the return type of AddCommit and change the test code like this? It looks better to me.
```
		vrank.AddCommit(msg, committee[i])
		expectedAssessList = append(expectedAssessList, vrankArrivedLate)
		expectedLateCommits = append(expectedLateCommits, vrank.commitArrivalTimeMap[committee[i].Address()])
```

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+var (
+	// VRank metrics
+	vrankFirstCommitArrivalTimeGauge           = metrics.NewRegisteredGauge("vrank/first_commit", nil)
+	vrankQuorumCommitArrivalTimeGauge          = metrics.NewRegisteredGauge("vrank/quorum_commit", nil)
+	vrankAvgCommitArrivalTimeWithinQuorumGauge = metrics.NewRegisteredGauge("vrank/avg_commit_within_quorum", nil)
+	vrankLastCommitArrivalTimeGauge            = metrics.NewRegisteredGauge("vrank/last_commit", nil)
+
+	vrankDefaultThreshold = "300ms"
```
yoomee1313, line 46: It would be more helpful if you add a comment about what vrankDefaultThreshold is.

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+	return ret
+}
+
+// encodeDuration encodes given duration into string
+// The returned string is at most 3 bytes
+func encodeDuration(d time.Duration) string {
+	if d > 10*time.Second {
+		return fmt.Sprintf("%.0fs", d.Seconds())
+	} else if d > time.Second {
+		return fmt.Sprintf("%.1fs", d.Seconds())
```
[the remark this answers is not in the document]
ian0371: Without L225-226, the return value is 5 bytes (e.g., "10.5s"), which is too long.

consensus/istanbul/core/vrank.go
ian0371: fixed in 28665aea

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+		d &= 0b11
+		return byte(a<<6 | b<<4 | c<<2 | d<<0)
+	}
+
+	// pad zero to make len(arr)%4 == 0
+	for len(arr)%4 != 0 {
+		arr = append(arr, 0)
+	}
+
+	ret := make([]byte, 0)
```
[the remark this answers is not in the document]
ian0371: fixed in 28665aea

consensus/istanbul/core/vrank.go
ian0371: fixed in c34ac7bb. there's no need to change `vrankNotArrivedPlaceholder` though

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+	}
+	return serialized
+}
+
+// compress compresses data into 2-bit bitmap
+// e.g., [1, 0, 2] => [0b01_00_10_00]
+func compress(arr []int) []byte {
+	zip := func(a, b, c, d int) byte {
```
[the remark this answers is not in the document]
ian0371: I didn't think any code will use this function. I'm not sure even if this will be used in the future other than logging.

consensus/istanbul/core/vrank.go
jiseongnoh: While your concerns regarding bitmap data retrieval are valid, the bitmap approach is more efficient for data storage. This is beneficial during our temporary use in setting an appropriate VRank threshold on the fluctuating network conditions. 
We will develop a tool for bitmap interpretation, which will address data extraction challenges.

ian0371: We'll make a log parser. However, I'm not sure if the parser needs to be publicly open. The purpose of this PR is solely for our experiments for the reputation system. Could you elaborate on cases where other developers need to use/parse vrank logs?

consensus/istanbul/core/vrank.go
ian0371: nice catch! fixed in f46f35f6

consensus/istanbul/core/vrank_test.go
ian0371: Right, I reluctantly added the return type of AddCommit just because of this. updated in 23e6513a

consensus/istanbul/core/vrank.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+// Log logs accumulated data in a compressed form
+func (v *Vrank) Log() (string, []string) {
+	var (
+		serialized = serialize(v.committee, v.commitArrivalTimeMap)
+		assessed   = assessBatch(serialized, v.threshold)
+		compressed = compress(assessed)
+		bitmap     = hex.EncodeToString(compressed)
```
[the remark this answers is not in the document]
ian0371: fixed in 62bf430d

consensus/istanbul/core/vrank.go
ian0371: added in e4036a59

yoomee1313: Thank you.

consensus/istanbul/core/vrank.go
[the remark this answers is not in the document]
hyunsooda: Because I was concerned if it consumes unnecessary overhead, compared to a case that defines to the outer function. If the function does not capture anything, however, I guess it does not may create another overhead if the compiler is smart by analyzing that there's no capturing, so it may decide that it is safe to move the function to the outer scope.

Anyway, my context was not visibility, but the performance aspect.

consensus/istanbul/core/vrank.go
hyunsooda: I just thought that a bitmap does sorely not represent the metric fully and intactly itself. Thus I thought some of assistance, matching validators, exact timestamp, etc may need for debugging. 

Rather than making an invert function that returns full metric from a bitmap, I thought that API would be more convenient and robust because the log may be truncated for various reasons (e.g., disk save), or hard to find the exact message.

consensus/istanbul/core/vrank.go
hyunsooda: I'm sorry. I omitted to mention that defines a new type for it (e.g., type Bitmap []uint8). You may consider it later.

hyunsooda (approved):

yoomee1313 (dismissed):

blukat29 (dismissed):

hyeonLewis (dismissed):

jiseongnoh (approved):

consensus/istanbul/core/commit.go
```diff
@@ -149,5 +153,9 @@ func (c *core) acceptCommit(msg *message, src istanbul.Validator) error {
 		return err
 	}
 
+	if c.current.Commits.Size() == 1 {
+		vrankFirstCommitArrivalTimeGauge.Update(int64(vrank.TimeSinceStart()))
```
aidan-kwon, line 157: This code doesn't do `nil` checking for vrank. is it okay?

ian0371: fixed in 4b9963cf

ian0371: Also checked all `vrank` references. thanks!

aidan-kwon (approved):

hyeonLewis (approved):

PR #1897, opened 2023-07-26
Change Log Level for "Mining too far in the future" to DEBUG

### LINKED ISSUE #1885: Change Log Level for "Mining too far in the future" to DEBUG
**Is your request related to a problem? Please describe.**
The log message "Mining too far in the future wait=???ms" message is less valuable for our normal monitoring activities.

**Describe the solution you'd like**
I suggest changing the log level of this message from INFO to DEBUG. 
This would make the message no longer visible in standard output, but still accessible in debug logs for analysis.

Additionally, we may create a new log for tracking instances where the start of block mining is delayed beyond its expected time. This may be useful for monitoring and optimizing system performance.

**Describe alternatives you've considered**
This new log message could replace the existing message, providing information about the late blocks.

**Additional context**
Before implementing these changes, it would be necessary to gather feedback to ensure I don't disrupt anyone's monitoring process. 
If you are using this log message, or if you have feedback regarding the proposed new log for tracking delayed mining starts, please share it. Thank you.

ian0371:
## Proposed changes

Closes #1885.

- decrease log level of "Mining too far in the future"
- add counterpart log

Example log:
```
INFO[07/26,12:25:02 +09] [51|work/worker.go:544] Mining start for new block is later than expected  nextBlockNum=7 delay=723.393ms       parentBlockTimestamp=1690341901 nextBlockTimestamp=1690341902
```

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/work/worker.go
+++ b/work/worker.go
@@ -533,11 +533,18 @@ func (self *worker) commitNewWork() {
 		// wait for a while and get a new timestamp
 		if tstart.Before(ideal) {
 			wait := ideal.Sub(tstart)
-			logger.Info("Mining too far in the future", "wait", common.PrettyDuration(wait))
+			logger.Debug("Mining too far in the future", "wait", common.PrettyDuration(wait))
 			time.Sleep(wait)
 
 			tstart = time.Now()    // refresh for metrics
 			tstamp = tstart.Unix() // refresh for block timestamp
+		} else {
+			logger.Info("Mining start for new block is later than expected",
+				"nextBlockNum", nextBlockNum,
+				"delay", tstamp-parent.Time().Int64()-1,
+				"previousBlockTimestamp", parent.Time().Int64(),
+				"nextBlockTimestamp", tstamp,
+			)
 		}
 	}
 
--- a/work/worker.go
+++ b/work/worker.go
@@ -528,7 +528,8 @@ func (self *worker) commitNewWork() {
 	tstart := time.Now()
 	tstamp := tstart.Unix()
 	if self.nodetype == common.CONSENSUSNODE {
-		ideal := time.Unix(parent.Time().Int64()+params.BlockGenerationInterval, 0)
+		parentTimestamp := parent.Time().Int64()
+		ideal := time.Unix(parentTimestamp+params.BlockGenerationInterval, 0)
 		// If a timestamp of this block is faster than the ideal timestamp,
 		// wait for a while and get a new timestamp
 		if tstart.Before(ideal) {
@@ -538,11 +539,11 @@ func (self *worker) commitNewWork() {
 
 			tstart = time.Now()    // refresh for metrics
 			tstamp = tstart.Unix() // refresh for block timestamp
-		} else {
+		} else if tstart.After(ideal) {
 			logger.Info("Mining start for new block is later than expected",
 				"nextBlockNum", nextBlockNum,
-				"delay", tstamp-parent.Time().Int64()-1,
-				"previousBlockTimestamp", parent.Time().Int64(),
+				"delay", tstart.Sub(ideal),
+				"parentBlockTimestamp", parentTimestamp,
 				"nextBlockTimestamp", tstamp,
 			)
 		}


JayChoi1736 (approved):

aidan-kwon (approved):

work/worker.go
```diff
@@ -533,11 +533,18 @@ func (self *worker) commitNewWork() {
...
 			wait := ideal.Sub(tstart)
-			logger.Info("Mining too far in the future", "wait", common.PrettyDuration(wait))
+			logger.Debug("Mining too far in the future", "wait", common.PrettyDuration(wait))
 			time.Sleep(wait)
 
 			tstart = time.Now()    // refresh for metrics
 			tstamp = tstart.Unix() // refresh for block timestamp
+		} else {
+			logger.Info("Mining start for new block is later than expected",
+				"nextBlockNum", nextBlockNum,
```
jiseongnoh, line 543: How about newBlockNum?

jiseongnoh (approved):

work/worker.go
```diff
@@ -533,11 +533,18 @@ func (self *worker) commitNewWork() {
 		// wait for a while and get a new timestamp
 		if tstart.Before(ideal) {
 			wait := ideal.Sub(tstart)
-			logger.Info("Mining too far in the future", "wait", common.PrettyDuration(wait))
+			logger.Debug("Mining too far in the future", "wait", common.PrettyDuration(wait))
 			time.Sleep(wait)
 
 			tstart = time.Now()    // refresh for metrics
 			tstamp = tstart.Unix() // refresh for block timestamp
+		} else {
```
hyunsooda, line 541: Doesn't it `else if tstart.After(deal)`? Ideally, if this log is for a time that is far from the ideal time, push log it as suggested below, if not (ideal time), do not log. Correct?

work/worker.go
```diff
@@ -533,11 +533,18 @@ func (self *worker) commitNewWork() {
...
+			logger.Info("Mining start for new block is later than expected",
+				"nextBlockNum", nextBlockNum,
+				"delay", tstamp-parent.Time().Int64()-1,
```
hyunsooda, line 544: 1. I couldn't get why subtract 1 is added here. Could you explain it?
2. How about declaring `parentTime` in L531 and replacing L531,544,545?

work/worker.go
jiseongnoh, line 541: what about 
`} else if tstart.After(ideal) {`?

work/worker.go
jiseongnoh: Since blocks are generated every second, each new block has a one-second delay. We subtract 1 from the total delay to focus on any extra delay beyond this one second to find the difference between the ideal and actual start of the next block.

jiseongnoh:
```suggestion
//  subtract 1 from the total delay to focus on any extra delay beyond this one second 
// to find the difference between the ideal and actual start of the next block.
				"delay", tstamp-parent.Time().Int64()-1,
```

work/worker.go
ian0371: You're right. Although `tstart==ideal` is extremely difficult because they have to match the nanoseconds. However, I'll add the condition to be more explicit.

ian0371: `tstart==ideal` is extremely difficult because they have to match the nanoseconds. However, I'll add the condition to be more explicit.

work/worker.go
hyunsooda: Thank you

hyunsooda (approved):

work/worker.go
ian0371: Instead, I renamed previous to parent

work/worker.go
ian0371: fixed in 4a0d746d

ian0371: fixed in 4a0d746d

work/worker.go
```diff
@@ -528,16 +528,24 @@ func (self *worker) commitNewWork() {
...
 		// wait for a while and get a new timestamp
 		if tstart.Before(ideal) {
 			wait := ideal.Sub(tstart)
-			logger.Info("Mining too far in the future", "wait", common.PrettyDuration(wait))
+			logger.Debug("Mining too far in the future", "wait", common.PrettyDuration(wait))
 			time.Sleep(wait)
 
 			tstart = time.Now()    // refresh for metrics
 			tstamp = tstart.Unix() // refresh for block timestamp
+		} else if tstart.After(ideal) {
```
aidan-kwon, line 542: Why don't you use `else` instead of `else if`? It misses an equal case

aidan-kwon: oh, it was intended behavior. I understand now

aidan-kwon (approved):

jiseongnoh (approved):

blukat29 (approved):

aidan-kwon (approved):

PR #1911, opened 2023-07-28
Klaytn composite accountKey aware recoverTransaction and recoverMessage

kjeom:
### Why these APIs are added or modified?
- It actually comes from SDK's requirements.
- We should offer the recovery mechanism for transactions or messages signed by accounts with composite accountKey Klaytn composite accountKey types.
- It can be implemented at the SDK level, but the SDK has to call the getAccountKey anyway (i.e. klay_getAccount RPC) to find the accountKey information.
- As a result, I think it would be better to offer these functions as Klaytn core JSON-RPC APIs

### API to be changed
- ecRecover
  - Klaytn's signMessage signs a message with `"\x19Klaytn signed message"` prefix. But many ethereum based wallet sign it with "\x19Ethereum signed message" and It bothers developers to implement recover logic personally.
  - So it is to be extended to recover both prefixes and then return true when at least one verification passes.
  - I think It doesn't matter with the replay attack since someone who has a private key of course can control other chain  that is using ecdsa mechanism.

### APIs to be added
- recoverTransaction
  - Receives
    - `an RLP encoded transaction` that includes signatures and
    - `blockNumber` for getting a specific accountKey state
  - Returns recovered `address`
  - If the sender address of the tx has composite accountKey type, it will return the sender address, not the addresses extracted from signatures.
  - If the recovery failed, it returns `0x0 address`.
- recoverMessage
  - Receives
    - `address`, `blockNumber` for getting a specific accountKey state
    - `data`, `signature`: same with ecRecover
  - Returns extracted address from the signature
    - need to discuss what to return:
      - option1 - the recovered address from the recovered public key
      - option2 - the recovered public key, to avoid confusion with the account address
  - Use-case: suppose Person A wants to sign a message with a Kaikas wallet,
    - if there is a multisig accountKey that has 3 private keys and 'A' has one of them
    - 'A' signed a message with the private key and then request it to the recoverMessage.
    - recoverMessage is going to check whether the multisig accountKey includes the key of 'A' and then it will pass.
    - As a result, recoverMessage returns the address extracted from the signature

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -573,7 +573,8 @@ func (s *PublicTransactionPoolAPI) RecoverTransaction(ctx context.Context, encod
 }
 
 func (s *PublicTransactionPoolAPI) RecoverMessage(
-	ctx context.Context, address common.Address, data, sig hexutil.Bytes, blockNumber rpc.BlockNumber) (common.Address, error) {
+	ctx context.Context, address common.Address, data, sig hexutil.Bytes, blockNumber rpc.BlockNumber,
+) (common.Address, error) {
 	pubkey, err := klayEthEcRecover(data, sig)
 	if err != nil {
 		return common.Address{}, err
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -561,7 +561,12 @@ func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, e
 		return common.Address{}, err
 	}
 
-	signer := types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(blockNumber.Uint64()))
+	var signer types.Signer
+	if blockNumber == rpc.LatestBlockNumber || blockNumber == rpc.PendingBlockNumber {
+		signer = types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(s.b.CurrentBlock().Number().Uint64()))
+	} else {
+		signer = types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(blockNumber.Uint64()))
+	}
 	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
 	if err != nil {
 		return common.Address{}, err
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -561,17 +561,19 @@ func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, e
 		return common.Address{}, err
 	}
 
-	var signer types.Signer
+	var bn uint64
 	if blockNumber == rpc.LatestBlockNumber || blockNumber == rpc.PendingBlockNumber {
-		signer = types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(s.b.CurrentBlock().Number().Uint64()))
+		bn = s.b.CurrentBlock().NumberU64()
 	} else {
-		signer = types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(blockNumber.Uint64()))
+		bn = blockNumber.Uint64()
 	}
+
+	signer := types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(bn))
 	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
 	if err != nil {
 		return common.Address{}, err
 	}
-	_, err = tx.ValidateSender(signer, state, blockNumber.Uint64())
+	_, err = tx.ValidateSender(signer, state, bn)
 	if err != nil {
 		return common.Address{}, err
 	}
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -553,7 +553,7 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
 	return common.Hash{}, fmt.Errorf("Transaction %#x not found", matchTx.Hash())
 }
 
-// RecoverTransaction recovers the sender address from a signed raw transaction.
+// RecoverFromTransaction recovers the sender address from a signed raw transaction.
 // The signature is validated against the sender account's key configuration at the given block number.
 func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, encodedTx hexutil.Bytes, blockNumber rpc.BlockNumber) (common.Address, error) {
 	tx := new(types.Transaction)
@@ -573,7 +573,7 @@ func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, e
 	return signer.Sender(tx)
 }
 
-// RecoverMessage validates that the message is signed by one of the keys in the given account.
+// RecoverFromMessage validates that the message is signed by one of the keys in the given account.
 // Returns the recovered signer address, which may be different from the account address.
 func (s *PublicTransactionPoolAPI) RecoverFromMessage(
 	ctx context.Context, address common.Address, data, sig hexutil.Bytes, blockNumber rpc.BlockNumber,
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -584,7 +584,10 @@ func (s *PublicTransactionPoolAPI) RecoverMessage(
 	if err != nil {
 		return common.Address{}, err
 	}
-	if state.GetKey(address).IsContainedKey(pubkey) {
+	key := state.GetKey(address)
+	// If the key is the AccountKeyTypeLegacy, confirmation will be delegated to the user
+	// by returning the recovered address
+	if key.ValidateMember(pubkey, address) {
 		return crypto.PubkeyToAddress(*pubkey), nil
 	}
 	return common.Address{}, nil
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -555,7 +555,7 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
 
 // RecoverTransaction recovers the sender address from a signed raw transaction.
 // The signature is validated against the sender account's key configuration at the given block number.
-func (s *PublicTransactionPoolAPI) RecoverTransaction(ctx context.Context, encodedTx hexutil.Bytes, blockNumber rpc.BlockNumber) (common.Address, error) {
+func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, encodedTx hexutil.Bytes, blockNumber rpc.BlockNumber) (common.Address, error) {
 	tx := new(types.Transaction)
 	if err := rlp.DecodeBytes(encodedTx, tx); err != nil {
 		return common.Address{}, err
@@ -575,7 +575,7 @@ func (s *PublicTransactionPoolAPI) RecoverTransaction(ctx context.Context, encod
 
 // RecoverMessage validates that the message is signed by one of the keys in the given account.
 // Returns the recovered signer address, which may be different from the account address.
-func (s *PublicTransactionPoolAPI) RecoverMessage(
+func (s *PublicTransactionPoolAPI) RecoverFromMessage(
 	ctx context.Context, address common.Address, data, sig hexutil.Bytes, blockNumber rpc.BlockNumber,
 ) (common.Address, error) {
 	pubkey, err := klayEthEcRecover(data, sig)
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -561,7 +561,7 @@ func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, e
 		return common.Address{}, err
 	}
 
-	signer := types.MakeSigner(s.b.ChainConfig(), s.b.CurrentBlock().Number())
+	signer := types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(blockNumber.Uint64()))
 	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
 	if err != nil {
 		return common.Address{}, err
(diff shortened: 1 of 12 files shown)


blockchain/types/accountkey/account_key.go
```diff
@@ -87,6 +87,8 @@ type AccountKey interface {
 	// IsCompositeType returns true if the account type is a composite type.
 	// Composite types are AccountKeyRoleBased and AccountKeyRoleBasedRLPBytes.
 	IsCompositeType() bool
+
+	IsContainedKey(recoveredKey *ecdsa.PublicKey) bool
```
blukat29, line 91: Let's use a more generalized method description.
```suggestion
    // ContainsPubkey returns true if the account type contains the given public key
    // in any part of its AccountKey.
	ContainsPubkey(pub *ecdsa.PublicKey) bool
```

blockchain/types/accountkey/account_key_legacy.go
```diff
@@ -45,6 +45,11 @@ func (a *AccountKeyLegacy) IsCompositeType() bool {
 	return false
 }
 
+// AccountKeyLegacy cannot check the public key, just return address of the public key argument always
+func (a *AccountKeyLegacy) IsContainedKey(recoveredKey *ecdsa.PublicKey) bool {
+	return true
+}
```
blukat29, line 51: 1. Unlike the comment, it just returns a bool.
2. It's awkward for ContainsPubkey(pub) is true for any pubkey. I think it should be false

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,40 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+	pubkey, err := klayEthEcRecover(data, sig)
+	if err != nil {
+		return common.Address{}, err
+	}
+
+	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
+	if err != nil {
+		return common.Address{}, err
+	}
+	if state.GetKey(address).IsContainedKey(pubkey) {
```
blukat29, line 587: Related to my comment at AccountKeyLegacy.IsContaindKey(),
```suggestion
	if state.GetKey(address).Type() == AccountKeyTypeLegacy || state.GetKey(address).IsContainedKey(pubkey)  {
```

blockchain/types/accountkey/account_key_legacy.go
kjeom: 1. Thanks, It should be fixed to return boolean, not address.
2. Right, it's weird a bit But It should return `true` so that the user can check it. If we return false and then `nil address` in the recoverMessage, the user don't have a chance to check.
How do you think the following comments?
```go
// AccountKeyLegacy cannot check the public key, just return true always so that the user can verify it
func (a *AccountKeyLegacy) IsContainedPubkey(recoveredKey *ecdsa.PublicKey) bool {
	return true
}
```

blockchain/types/accountkey/account_key.go
kjeom: How do you think `IsContainedPubkey`? I also want to remove 'Is' prefix but we already have the `IsCompositeType` function returning boolean.

api/api_public_transaction_pool.go
kjeom: Okay, I got your point.
Returning false in the legacy account is also weird since correct pubkey also return false.
Anyway it looks better to denote explicitly in the `recoverMessage`.

blukat29: I see. Returning false is also weird. I came up with an unambiguous method.
We could add `ValidateMember()` alongside `Validate()` which are very similar, but ValidateMember is satisfied with just one public key.

```go
type AccountKey interface {
	Validate(currentBlockNumber uint64, r RoleType, recoveredKeys []*ecdsa.PublicKey, from common.Address) bool
	ValidateMember(blockNum uint64, recoveredKey *ecdsa.PublicKey, from common.Address) bool
}
```

kjeom: Reflected [002ffc5](https://github.com/klaytn/klaytn/pull/1911/commits/002ffc540b33cf96e5ae0981aa28f37c4e43f556)
This review seems to include all your points.

blukat29: Thanks! That looks nice.

blockchain/types/accountkey/account_key_legacy.go
```diff
@@ -45,6 +45,11 @@ func (a *AccountKeyLegacy) IsCompositeType() bool {
 	return false
 }
 
+// AccountKeyLegacy cannot check the public key, just return false and then it will be controled by the caller.
```
blukat29, line 48: Update the comment to match the function please.

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,43 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
 
 	return common.Hash{}, fmt.Errorf("Transaction %#x not found", matchTx.Hash())
 }
+
+// This API doesn't allow rueqesters to use block hash according to ValidateSender function's parameter
```
blukat29, line 556:
```suggestion
// RecoverTransaction recovers the sender address from a signed raw transaction.
// The signature is validated against the sender account's key configuration at the given block number.
```

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,43 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+	if err != nil {
+		return common.Address{}, err
+	}
+	_, err = tx.ValidateSender(signer, state, blockNumber.Uint64())
+	if err != nil {
+		return common.Address{}, err
+	}
+	return signer.Sender(tx)
+}
+
```
blukat29, line 574:
```suggestion
// RecoverMessage validates that the message is signed by one of the keys in the given account.
// Returns the recovered signer address, which may be different from the account address.
```

blukat29 (approved):

JayChoi1736 (dismissed):

aidan-kwon (dismissed):
LGTM except for a minor comment

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,46 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+// RecoverMessage validates that the message is signed by one of the keys in the given account.
+// Returns the recovered signer address, which may be different from the account address.
+func (s *PublicTransactionPoolAPI) RecoverMessage(
```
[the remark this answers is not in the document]
kjeom: @aidan-kwon I made it return `boolean` first. but I thought there is a case the `pubkeyToAddress` or `pubkey` return value. Given a multisig key that includes 3keys and 3 persons have each key, dapp want to check how many persons signed the message. Returning boolean couldn't resolve this case.
In other words, It is to divide the duplicate sign.

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,46 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+	if err != nil {
+		return common.Address{}, err
+	}
+	key := state.GetKey(address)
+	// If the key is the AccountKeyTypeLegacy, confirmation will be delegated to the user
+	// by returning the recovered address
+	if key.ValidateMember(pubkey, address) {
+		return crypto.PubkeyToAddress(*pubkey), nil
+	}
+	return common.Address{}, nil
```
ian0371, line 596: shouldn't we return an error?

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,46 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+
+	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
+	if err != nil {
+		return common.Address{}, err
+	}
+	key := state.GetKey(address)
+	// If the key is the AccountKeyTypeLegacy, confirmation will be delegated to the user
+	// by returning the recovered address
+	if key.ValidateMember(pubkey, address) {
+		return crypto.PubkeyToAddress(*pubkey), nil
```
ian0371, line 594: I think option 1 is better. I think `ecrecover`-ed address doesn't imply anything.

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,46 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+		return common.Address{}, err
+	}
+
+	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
+	if err != nil {
+		return common.Address{}, err
+	}
+	key := state.GetKey(address)
+	// If the key is the AccountKeyTypeLegacy, confirmation will be delegated to the user
+	// by returning the recovered address
```
blukat29, line 592: Since AccountKeyLegacy.ValidateMember does check the recovered address == account address, this comment is not needed anymore.

api/api_public_transaction_pool.go
kjeom: What you mean option 1?

api/api_public_transaction_pool.go
kjeom: You are right. looks better to return an error.

api/api_public_transaction_pool.go
kjeom: resolved fba0b15

api/api_public_transaction_pool.go
ian0371: in the pr description

```
need to discuss what to return:
option1 - the address
option2 - extracted public key from the signature to avoid confusion with the account address
```

blukat29 (approved):

api/api_public_transaction_pool.go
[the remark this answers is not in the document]
aidan-kwon: @kjeom How about changing the method name to `RecoverAddr`?

api/api_public_transaction_pool.go
blukat29: option 1 refers to 'the recovered address' and option 2 refers to the 'recovered public key'. Both are from the signature. Me and @kjeom had a discussion about this API, and we concluded that returning the recovered address/pubkey is meaningful in some context - where a webapp has to collect signatures from distinct individuals of their shared account, so the webapp has to know which key signed the message.

ian0371: I see. Please clarify the difference of the address parameter and the returned address in the API docs.

ian0371 (approved):

api/api_public_transaction_pool.go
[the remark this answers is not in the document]
kjeom: If it's not a legacy key, I'm afraid it's confusing because it's not a real address. That's why I also considered returning the pubkey format. How do you think about `RecoverAddr`? @nohkwak @JayChoi1736 @blukat29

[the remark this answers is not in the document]
blukat29: `RecoverAddr` is ambiguous in that both RecoverTransaction and RecoverMessages return an address.
`RecoverMessage` might be problematic because in some Klaytn code "message" refers to "transaction". 
`RecoverSignature` or `RecoverSignedMessage` in the first comment looks more accurate.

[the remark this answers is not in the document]
aidan-kwon: The method recovers an address from a message and a signature. I think `recoverAddr`, `recoverPubkey`, `recoverSigner`, or `recoverAddrFromMessage` are grammatically better.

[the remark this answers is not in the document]
kjeom: After discussion, we have renamed each as recoverFromTransaction and recoverFromMessage.

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,45 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+// RecoverTransaction recovers the sender address from a signed raw transaction.
+// The signature is validated against the sender account's key configuration at the given block number.
+func (s *PublicTransactionPoolAPI) RecoverTransaction(ctx context.Context, encodedTx hexutil.Bytes, blockNumber rpc.BlockNumber) (common.Address, error) {
+	tx := new(types.Transaction)
+	if err := rlp.DecodeBytes(encodedTx, tx); err != nil {
+		return common.Address{}, err
+	}
+
+	signer := types.MakeSigner(s.b.ChainConfig(), s.b.CurrentBlock().Number())
+	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
```
hyunsooda, line 565: The state retrieved at L565 corresponds to the input variable blockNumber. However, at L564, a current block number is being passed. This introduces an inconsistency in the selections of status and signer. Consequently, there's a risk that an incorrect signer might erroneously invoke the Sender() function at L574.

kjeom: Thanks

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,45 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
 
 	return common.Hash{}, fmt.Errorf("Transaction %#x not found", matchTx.Hash())
 }
+
+// RecoverTransaction recovers the sender address from a signed raw transaction.
```
aidan-kwon, line 556:
```suggestion
// RecoverFromTransaction recovers the sender address from a signed raw transaction.
```

api/api_public_transaction_pool.go
```diff
@@ -552,3 +552,45 @@ func resend(s *PublicTransactionPoolAPI, ctx context.Context, sendArgs NewTxArgs
...
+		return common.Address{}, err
+	}
+	_, err = tx.ValidateSender(signer, state, blockNumber.Uint64())
+	if err != nil {
+		return common.Address{}, err
+	}
+	return signer.Sender(tx)
+}
+
+// RecoverMessage validates that the message is signed by one of the keys in the given account.
```
aidan-kwon, line 576:
```suggestion
// RecoverFromMessage validates that the message is signed by one of the keys in the given account.
```

aidan-kwon (dismissed):

JayChoi1736 (dismissed):

blukat29 (approved):

kjeom:
@ian0371 @hyunsooda please review

ian0371 (approved):

hyunsooda (approved):

aidan-kwon (approved):

blukat29 (approved):

jeongkyun-oh (approved):

JayChoi1736 (approved):

nohkwak (approved):

PR #1913, opened 2023-07-31
added missing rocksdb configuration

jeongkyun-oh:
## Proposed changes

- Missed rocksdb configuration on several subcommand including `init`, `snapshot`, `db-migration` apart from #1900.
- Also, there is no destination db flag when migrating from a db to rocksdb so those flags were added.
- Each dbentry cache size for rocksdb is rebalanced as like leveldb

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


kjeom:
@jeongkyun-oh When these flags are missed?

cmd/utils/nodecmd/snapshot.go
```diff
@@ -109,6 +123,13 @@ reading all nodes and logging the missing nodes.
 				utils.LevelDBCompressionTypeFlag,
 				utils.DataDirFlag,
 				utils.ChainDataDirFlag,
+				utils.RocksDBSecondaryFlag,
+				utils.RocksDBCacheSizeFlag,
+				utils.RocksDBDumpMallocStatFlag,
+				utils.RocksDBFilterPolicyFlag,
+				utils.RocksDBCompressionTypeFlag,
+				utils.RocksDBBottommostCompressionTypeFlag,
+				utils.RocksDBDisableMetricsFlag,
```
kjeom, line 132: Could the flags be different for each command? If it isn't, It would be better to make a global variable like "snapshotFlags"

jeongkyun-oh: @kjeom Agreed. Instead of `snapshotFlags`, how about using `DBFlags`? There is already `DBMigrationSrcFlags` including the same flags related to database. So, renaming `DBMigrationSrcFlags` to `DBFlags` is also necessary. What do you think?

kjeom: Hmm.. DBMigrationSrcFlags looks a little bit different with DBMigrationDstFlags. Can we make DB related all the things DBFlags? I think it's our best.

jeongkyun-oh: @kjeom Wait. It is not the same. I introduced snapshot flag group. seems a bit duplicated, but it looks more clear.

kjeom (approved):

ian0371 (approved):

cmd/utils/nodecmd/migrationcmd.go
```diff
@@ -138,6 +148,16 @@ func createDBConfigForMigration(ctx *cli.Context) (*database.DBConfig, *database
...
+
+		RocksDBConfig: &database.RocksDBConfig{
+			CacheSize:                 ctx.Uint64(utils.DstRocksDBCacheSizeFlag.Name),
+			DumpMallocStat:            ctx.Bool(utils.DstRocksDBDumpMallocStatFlag.Name),
+			DisableMetrics:            ctx.Bool(utils.DstRocksDBDisableMetricsFlag.Name),
+			Secondary:                 ctx.Bool(utils.DstRocksDBSecondaryFlag.Name),
+			CompressionType:           ctx.String(utils.DstRocksDBCompressionTypeFlag.Name),
+			BottommostCompressionType: ctx.String(utils.DstRocksDBBottommostCompressionTypeFlag.Name),
+			FilterPolicy:              ctx.String(utils.DstRocksDBFilterPolicyFlag.Name),
+		},
```
ethan-kr, line 160: Code similar to the one above is being repeated in the same pattern in many places. In the future, it is hoped that the db_manager side will set the default value as a flag value and be refactored so that only the options to be specially set are entered separately.

ethan-kr (approved):
Please refer to the comments

PR #1914, opened 2023-07-31
Live pruning is available in archive mode

blukat29:
## Proposed changes

- Enable live pruning in archive mode
  - Currently only worked in full mode. Now it also works in archive mode.
- Address some leftover comments in https://github.com/klaytn/klaytn/pull/1871.
  - Add comments to DeletePruningEnabled() that it is for testing
  - Add missing it.Release() in ReadPruningMarks()
  - Refactor bc.IsLivePruningRequired()
  - When live pruning is disabled, do not start the goroutine in pruneTrieNodeLoop

## Types of changes

- [x] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1916, opened 2023-08-02
Disable state migration when live pruning is enabled

blukat29:
## Proposed changes

- `admin_startStateMigration` RPC returns an error if the database has live pruning enabled.
  - State migration does not work with the ExtHash trie database.
  - State migration is unnecessary when Live Pruning is used.
- Fix `ken --state.live-pruning` flag to work
  - Reverts some change from https://github.com/klaytn/klaytn/pull/1914
  - When `bc.pruneTrieNodeLoop()` is executed, the pruning flag might not be in database just yet. For example in `cn.New()`, `NewBlockChain()` is first called and then `WritePruningEnabled()`.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1428,11 +1428,6 @@ func (bc *BlockChain) gcCachedNodeLoop() {
 }
 
 func (bc *BlockChain) pruneTrieNodeLoop() {
-	// If live pruning is disabled, do not bother starting a goroutine.
-	if !bc.IsLivePruningRequired() {
-		return
-	}
-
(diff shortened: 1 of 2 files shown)


hyunsooda (approved):
An idle goroutine would not matter at all as the data channel (`chPrune`) would not be filled on the non-pruning option.

jeongkyun-oh (approved):

ian0371 (approved):

kjeom (approved):

PR #1923, opened 2023-08-07
Revert "Replace Kip103ContractCaller by BlockchainContractCaller"

blukat29:
## Proposed changes

This reverts commit 15fb12bd2c70c2e6124c75053b96a6b8acc1dcb4.

BlockchainContractCaller as-is was not appropriate to substitute Kip103ContractCaller.
- Kip103ContractCaller, in its implementation, increments the `caller.From` nonce in `CallContract`
- BlockchainContractCaller does not
- This difference caused the Baobab sync test failure.

This PR revives Kip103ContractCaller as a quick fix.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -29,7 +29,6 @@ import (
 	"time"
 
 	lru "github.com/hashicorp/golang-lru"
-	"github.com/klaytn/klaytn/accounts/abi/bind/backends"
 	"github.com/klaytn/klaytn/blockchain/state"
 	"github.com/klaytn/klaytn/blockchain/types"
 	"github.com/klaytn/klaytn/common"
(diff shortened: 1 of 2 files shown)


hyunsooda (commented):
I have two questions:

1. I'm struggling to understand the discrepancy between this and old versions' code.
I got that there's an inconsistent state between `Finalize()` call's one and the contract caller's one. The last block's state (contract caller's state) was used for the KIP-103 contract read. Here, even the current transactions created different state compared to the last block's state, the read operation of KIP-103 with the state at last block seems not correlated. So, I'm now confusing how two different states are created here. The `RebalanceTreasury()` function modifies the currently being modified state and read the contract storage with the last block's state. The final state should not be different if the reading the KIP-103 contract does not have effect by the current state and I think there's no any  correlation. If no, the `memo` logging must be different. 

2. If we consider a scenario where no transactions occur at the KIP-103 activation block number, could this error potentially go unnoticed or be overlooked to some extent?

ian0371 (approved):

blukat29:
@hyunsooda Good question.

There was a bug, a side effect, in `Kip103ContractCaller` where the nonce of `0x0` account was incremented during contract calls. BlockchainContractCaller does not have the bug. That's why the `memo` and account balances were all correct yet sync test failed.

aidan-kwon (approved):

yoomee1313 (approved):

jiseongnoh (approved):

PR #1924, opened 2023-08-07
KIP-103 code uses PendingContractCaller with intermediate state

blukat29:
## Proposed changes

Alternative fix to #1923.
- accounts/abi/bind/backends/blockchain.go
  BlockchainContractCaller implements PendingContractCaller
- consensus/istanbul/backend/engine.go
  Pass `header` and `state` to the PendingContractCaller as `pendingHeader` and `pendingState`.
- consensus/istanbul/backend/kip103.go
  Use `CallOpts{Pending: true}` at contract call sites

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Further comments

### DIFF
--- a/accounts/abi/bind/backends/blockchain.go
+++ b/accounts/abi/bind/backends/blockchain.go
@@ -56,7 +56,9 @@ type BlockChainForCaller interface {
 // Note that SimulatedBackend creates a new temporary BlockChain for testing,
 // whereas BlockchainContractCaller uses an existing BlockChain with existing database.
 type BlockchainContractCaller struct {
-	bc BlockChainForCaller
+	bc            BlockChainForCaller
+	pendingHeader *types.Header
+	pendingState  *state.StateDB
 }
 
 // This nil assignment ensures at compile time that BlockchainContractCaller implements bind.ContractCaller.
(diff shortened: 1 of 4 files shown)


accounts/abi/bind/backends/blockchain.go
```diff
@@ -90,7 +109,22 @@ func (b *BlockchainContractCaller) CallContract(ctx context.Context, call klaytn
 		return nil, err
 	}
 
-	res, err := b.callContract(call, block, state)
+	res, err := b.callContract(call, block.Header(), state)
```
jiseongnoh, line 112: Is 'block'.Number and 'header'.number different? Still, b.callContract() can get the same header number by calling 'block'.Header() inside the function, can't it? If not, please explain.

blukat29: Yes, they are the same.

But `callContract()` had to receive Header because...
`PendingCallContract()` had only Header and no Block because...
`NewBlockchainPendingContractCaller()` had only Header and no Block because...
`Finalize()` had only `header` and no `block` [consensus/istanbul/backend/engine.go](https://github.com/klaytn/klaytn/pull/1924/files#diff-44abfea3bdbe16ce301a955732b485d1f0af2cc6c0c4cee3fbce0feb6ce051b0).

jiseongnoh: type BlockchainContractCaller struct {
	bc            BlockChainForCaller
	pendingHeader *types.Header
	pendingState  *state.StateDB
}

I understand that the bc does not have a block body.
However, the variable name confuses me, because bc looks like a blockchain structure. wonderting why BlockChainForCaller does not have a block body.
type BlockChainForCaller interface {
	// Required by NewEVMContext
	blockchain.ChainContext

blukat29: It does have CurrentBlock, but no pending block.
Regarding KIP-103, the pending block is produced in Finalize(), which does not yet exist.
We cannot use `bc.CurrentBlock()` because `bc.CurrentBlock().Number + 1 == pendingHeader.Number`.

jiseongnoh: I see. Thank you for the explanation.

blukat29:
Closing in favor of #1923.

PR #1937, opened 2023-08-09
bugfix: recoverFromMessage always returns klayEcRecover

kjeom:
## Proposed changes

- changes
  - existing ecrecover only covers klayEcrecover (`\x19Klaytn Signed Message` prefix case)
- fix
  - after checking both klay and eth ecrecover, return the recovered address only if either matches

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/api/api_private_account.go
+++ b/api/api_private_account.go
@@ -485,31 +485,37 @@ func (s *PrivateAccountAPI) Sign(ctx context.Context, data hexutil.Bytes, addr c
 //
 // https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal_ecRecover
 func (s *PrivateAccountAPI) EcRecover(ctx context.Context, data, sig hexutil.Bytes) (common.Address, error) {
-	pubkey, err := klayEthEcRecover(data, sig)
+	if len(sig) != crypto.SignatureLength {
+		return common.Address{}, fmt.Errorf("signature must be 65 bytes long")
+	}
+	if sig[crypto.RecoveryIDOffset] != 27 && sig[crypto.RecoveryIDOffset] != 28 {
+		return common.Address{}, fmt.Errorf("invalid Klaytn signature (V is not 27 or 28)")
+	}
+
+	// Transform yellow paper V from 27/28 to 0/1
+	sig[crypto.RecoveryIDOffset] -= 27
+
+	pubkey, err := klayEcRecover(data, sig)
 	if err != nil {
 		return common.Address{}, err
 	}
 	return crypto.PubkeyToAddress(*pubkey), nil
 }
 
-func klayEthEcRecover(data, sig hexutil.Bytes) (*ecdsa.PublicKey, error) {
-	if len(sig) != crypto.SignatureLength {
-		return nil, fmt.Errorf("signature must be 65 bytes long")
+func klayEcRecover(data, sig hexutil.Bytes) (*ecdsa.PublicKey, error) {
+	rpk, err := crypto.SigToPub(signHash(data), sig)
+	if err == nil {
+		return rpk, nil
 	}
-	if sig[crypto.RecoveryIDOffset] != 27 && sig[crypto.RecoveryIDOffset] != 28 {
-		return nil, fmt.Errorf("invalid Klaytn signature (V is not 27 or 28)")
-	}
-	sig[crypto.RecoveryIDOffset] -= 27 // Transform yellow paper V from 27/28 to 0/1
+	return nil, err // klayErr has more priority
+}
 
-	klayRpk, klayErr := crypto.SigToPub(signHash(data), sig)
-	if klayErr == nil {
-		return klayRpk, nil
+func ethEcRecover(data, sig hexutil.Bytes) (*ecdsa.PublicKey, error) {
+	rpk, err := crypto.SigToPub(ethSignHash(data), sig)
+	if err == nil {
+		return rpk, nil
 	}
-	ethRpk, ethErr := crypto.SigToPub(ethSignHash(data), sig)
-	if ethErr == nil {
-		return ethRpk, nil
-	}
-	return nil, klayErr // klayErr has more priority
+	return nil, err
 }
 
 // SignAndSendTransaction was renamed to SendTransaction. This method is deprecated
--- a/api/api_private_account.go
+++ b/api/api_private_account.go
@@ -503,19 +503,11 @@ func (s *PrivateAccountAPI) EcRecover(ctx context.Context, data, sig hexutil.Byt
 }
 
 func klayEcRecover(data, sig hexutil.Bytes) (*ecdsa.PublicKey, error) {
-	rpk, err := crypto.SigToPub(signHash(data), sig)
-	if err == nil {
-		return rpk, nil
-	}
-	return nil, err // klayErr has more priority
+	return crypto.SigToPub(signHash(data), sig)
 }
 
 func ethEcRecover(data, sig hexutil.Bytes) (*ecdsa.PublicKey, error) {
-	rpk, err := crypto.SigToPub(ethSignHash(data), sig)
-	if err == nil {
-		return rpk, nil
-	}
-	return nil, err
+	return crypto.SigToPub(ethSignHash(data), sig)
 }
 
 // SignAndSendTransaction was renamed to SendTransaction. This method is deprecated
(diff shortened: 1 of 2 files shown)


api/api_public_transaction_pool.go
```diff
@@ -596,8 +610,11 @@ func (s *PublicTransactionPoolAPI) RecoverFromMessage(
 	}
 
 	key := state.GetKey(address)
-	if key.ValidateMember(pubkey, address) {
-		return crypto.PubkeyToAddress(*pubkey), nil
+	if key.ValidateMember(klayRpk, address) {
+		return crypto.PubkeyToAddress(*klayRpk), nil
+	}
+	if key.ValidateMember(ethRpk, address) {
+		return crypto.PubkeyToAddress(*ethRpk), nil
```
blukat29, line 617: I do agree that `klayEcRecover` and `ethEcRecover` realistically won't fail. But just in case only one of them succeeds,

```suggestion
	var recoverErr error
	if pubkey, err := klayEcRecover(data, sig); err == nil {
		if key.ValidateMember(pubkey, address) {
			return crypto.PubkeyToAddress(*pubkey), nil
		}
	} else {
		recoverErr = err
	}
	if pubkey, err := ethEcRecover(data, sig); err == nil {
		if key.ValidateMember(pubkey, address) {
			return crypto.PubkeyToAddress(*pubkey), nil
		}
	} else {
		recoverErr = err
	}

	if recoverErr != nil {
		return common.Address{}, recoverErr
	} else {
		return common.Address{}, errors.New("Invalid signature")
	}
```

api/api_private_account.go
```diff
@@ -485,31 +485,37 @@ func (s *PrivateAccountAPI) Sign(ctx context.Context, data hexutil.Bytes, addr c
...
-func klayEthEcRecover(data, sig hexutil.Bytes) (*ecdsa.PublicKey, error) {
-	if len(sig) != crypto.SignatureLength {
-		return nil, fmt.Errorf("signature must be 65 bytes long")
-	}
-	if sig[crypto.RecoveryIDOffset] != 27 && sig[crypto.RecoveryIDOffset] != 28 {
-		return nil, fmt.Errorf("invalid Klaytn signature (V is not 27 or 28)")
+func klayEcRecover(data, sig hexutil.Bytes) (*ecdsa.PublicKey, error) {
+	rpk, err := crypto.SigToPub(signHash(data), sig)
+	if err == nil {
+		return rpk, nil
```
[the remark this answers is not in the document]
kjeom: clean code, thanks

blukat29 (approved):

api/api_private_account.go
```diff
@@ -485,31 +485,29 @@ func (s *PrivateAccountAPI) Sign(ctx context.Context, data hexutil.Bytes, addr c
...
+	}
+
+	// Transform yellow paper V from 27/28 to 0/1
+	sig[crypto.RecoveryIDOffset] -= 27
+
+	pubkey, err := klayEcRecover(data, sig)
 	if err != nil {
 		return common.Address{}, err
 	}
 	return crypto.PubkeyToAddress(*pubkey), nil
```
hyunsooda, line 502: The version of ether seems not have a chance to be invoked. Is it the correct new one? then could you explain why it was changed again?

api/api_private_account.go
```diff
@@ -485,31 +485,29 @@ func (s *PrivateAccountAPI) Sign(ctx context.Context, data hexutil.Bytes, addr c
 //
 // https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal_ecRecover
 func (s *PrivateAccountAPI) EcRecover(ctx context.Context, data, sig hexutil.Bytes) (common.Address, error) {
-	pubkey, err := klayEthEcRecover(data, sig)
+	if len(sig) != crypto.SignatureLength {
```
hyunsooda, line 488: How about wrapping L488 to L498 to a function? then L588-596 can be short

api/api_public_transaction_pool.go
```diff
@@ -585,19 +585,40 @@ func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, e
...
+		if key.ValidateMember(pubkey, address) {
+			return crypto.PubkeyToAddress(*pubkey), nil
+		}
+	} else {
+		recoverErr = err
+	}
+	if recoverErr != nil {
+		return common.Address{}, recoverErr
+	} else {
+		return common.Address{}, errors.New("Invalid signature")
```
hyunsooda, line 622: The `else` seems redundant. Just one returning `return recoverErr` seems enough. Could you verify it again?

api/api_public_transaction_pool.go
```diff
@@ -585,19 +585,40 @@ func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, e
...
+		}
+	} else {
+		recoverErr = err
+	}
+	if pubkey, err := ethEcRecover(data, sig); err == nil {
+		if key.ValidateMember(pubkey, address) {
+			return crypto.PubkeyToAddress(*pubkey), nil
+		}
+	} else {
+		recoverErr = err
```
hyunsooda, line 617: `recoverErr` seems to be overwritten by the error of ether version. It seems the reverse compared to the previous proposed one.

hyunsooda (commented):
I read the difference and the description twice, but could not get where the significant changes were made. Could you explain them directly?

api/api_private_account.go
kjeom: When I tested, klayEcRecover function always return the `pubkey` regardless of klay or ether prefix.
It's becuase we cannot check if it is signed with klay or eth prefix in this function. So I just removed the eth context in ecrecover as it is.

api/api_public_transaction_pool.go
kjeom: You are right, but I think It depends on our choice. Do you think we need additional error type when `klayErr and ethErr != nil`?

api/api_private_account.go
kjeom: I made the function first. We need an additional function that cover followings
```go
// Transform yellow paper V from 27/28 to 0/1
sig[crypto.RecoveryIDOffset] -= 27
```
We can name it "transformRecoverIdOffset". But I don't want to add some util property function in this file.
So I just wrote that code twice.

api/api_public_transaction_pool.go
kjeom: If recoverErr == nil, I want to add some error explicitly. You can see https://github.com/klaytn/klaytn/pull/1911#discussion_r1284147124

aidan-kwon, line 622: @kjeom The current logic is a little confusing to me. Can we refactor them like this?
```suggestion
	if pubkey, err := klayEcRecover(data, sig); err != nil {
		if pubkey, errEth := ethEcRecover(data, sig); errEth != nil {
			return common.Address{}, errEth
			// or return common.Address{}, err
		}
	} 
	if !key.ValidateMember(pubkey, address) {
		return common.Address{}, errors.New("Invalid signature")
	}	
	return crypto.PubkeyToAddress(*pubkey), nil
```

api/api_private_account.go
```diff
@@ -485,31 +485,29 @@ func (s *PrivateAccountAPI) Sign(ctx context.Context, data hexutil.Bytes, addr c
...
+		return common.Address{}, fmt.Errorf("signature must be 65 bytes long")
+	}
+	if sig[crypto.RecoveryIDOffset] != 27 && sig[crypto.RecoveryIDOffset] != 28 {
+		return common.Address{}, fmt.Errorf("invalid Klaytn signature (V is not 27 or 28)")
+	}
+
+	// Transform yellow paper V from 27/28 to 0/1
+	sig[crypto.RecoveryIDOffset] -= 27
+
+	pubkey, err := klayEcRecover(data, sig)
```
aidan-kwon, line 498: Why this API was reverted back not to support `ethEcRecover`?

api/api_public_transaction_pool.go
kjeom: @aidan-kwon No, it's different logic. klayEcRecover will not return error according to eth or klay signed prefix.

api/api_private_account.go
kjeom: We cannot identify if the signature has signed with klay or eth prefix without the signer's address.
Even though a user signed message with eth prefix, it will return something. that's why we can verify only klay prefixed signature with this function.

aidan-kwon: Thanks. I am gonna leave this comment open for other reviewer.

aidan-kwon (approved):

hyunsooda (approved):

nohkwak (approved):

PR #1944, opened 2023-08-21
Add BlockChainContractBackend

2dvorak:
## Proposed changes

- This PR follows up #1828
- Adds `BlockchainContractBackend` that implements:
   - [`bind.ContractTransactor`](https://github.com/klaytn/klaytn/blob/v1.11.0/accounts/abi/bind/backend.go#L73)
   - [`bind.ContractFilterer`](https://github.com/klaytn/klaytn/blob/v1.11.0/accounts/abi/bind/backend.go#L95)
   - [`bind.ContractBackend`](https://github.com/klaytn/klaytn/blob/v1.11.0/accounts/abi/bind/backend.go#L114)
   - [`bind.DeployBackend`](https://github.com/klaytn/klaytn/blob/v1.11.0/accounts/abi/bind/backend.go#L108)
- Replaced by  `BlockchainContractBackend`:
   - [`sc.LocalBackend`](https://github.com/klaytn/klaytn/blob/v1.11.0/node/sc/local_backend.go)
   - [`tests.testKip103TxTransactor`](https://github.com/klaytn/klaytn/blob/v1.11.0/tests/kip103_test.go)
- `BlockchainContractBackend` will be utilized in future implementations that deals with contracts.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +144,197 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+	// TODO-Klaytn this is not pending code but latest code
+	state, err := b.bc.State()
+	if err != nil {
+		return nil, err
+	}
+	return state.GetCode(account), nil
+}
+
+func (b *BlockchainContractTransactor) PendingNonceAt(ctx context.Context, account common.Address) (uint64, error) {
+	if b.txPool != nil {
```
hyunsooda, line 170: Under what conditions can `b.txPool` be nil? If the NewBlockchainContractTransactor already includes checks for both bc and txPool, it might render these checks unnecessary. Therefore, a single statement like return b.txPool.GetPendingNonce(account), nil could suffice in my view.

accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +144,197 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+		}
+		return res.Failed(), res, nil
+	}
+
+	estimated, err := blockchain.DoEstimateGas(ctx, call.Gas, 0, call.Value, call.GasPrice, balance, executable)
+	if err != nil {
+		return 0, err
+	} else {
+		return uint64(estimated), nil
+	}
```
hyunsooda, line 220: Why not `return uint64(esmitated), err`?

accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +144,197 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+	return b.txPool.AddLocal(tx)
+}
+
+func (b *BlockchainContractTransactor) ChainID(ctx context.Context) (*big.Int, error) {
+	return b.bc.Config().ChainID, nil
+}
+
+type BlockchainContractFilterer struct {
+	bc BlockChainForCaller
+	es *filters.EventSystem
```
hyunsooda, line 236: `events` seems more intuitive rather than `es`

accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +144,197 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+	if _, state, err := b.getBlockAndState(blockNumber); err != nil {
+		return nil, err
+	} else {
+		return state.GetBalance(account), nil
+	}
+}
+
+func (b *BlockchainContractBackend) CurrentBlockNumber(ctx context.Context) (uint64, error) {
+	var bc BlockchainContractCaller = b.BlockchainContractCaller
+	return bc.bc.CurrentBlock().NumberU64(), nil
```
hyunsooda, line 339:
```suggestion
	return b.BlockchainContractCaller.bc.CurrentBlock().NumberU64(), nil
```

accounts/abi/bind/backends/blockchain_test.go
```diff
@@ -147,3 +154,235 @@ func TestBlockchainCallContract(t *testing.T) {
...
+	bc.InsertChain(blocks)
+
+	block = bc.GetBlockByNumber(11)
+	if block == nil {
+		t.Errorf("could not get block at height 1")
+	}
+
+	if signedTx.Hash() != block.Transactions()[0].Hash() {
+		t.Errorf("did not commit sent transaction. expected hash %v got hash %v", block.Transactions()[0].Hash(), signedTx.Hash())
+	}
```
hyunsooda, line 258:
```suggestion
        assert.Equal(t, block.Header().EmptyReceipts(), false)
	}
```

accounts/abi/bind/backends/blockchain_test.go
```diff
@@ -147,3 +154,235 @@ func TestBlockchainCallContract(t *testing.T) {
...
+		Addresses: []common.Address{code1Addr},
+	})
+	assert.Nil(t, err)
+	assert.Equal(t, 2, len(logs))
+
+	// Non-code address
+	logs, err = c.FilterLogs(context.Background(), klaytn.FilterQuery{
+		FromBlock: big.NewInt(0),
+		ToBlock:   big.NewInt(11),
+		Addresses: []common.Address{testAddr},
```
hyunsooda, line 338: How about to change with code2Addr and the comment (L334) as // No receipt logs exist for code1Addr`

node/sc/local_backend.go
```diff
@@ -66,277 +43,6 @@ func checkCtx(ctx context.Context) error {
...
-
-// ChainID can return the chain ID of the chain.
-func (lb *LocalBackend) ChainID(ctx context.Context) (*big.Int, error) {
-	if err := checkCtx(ctx); err != nil {
-		return nil, err
-	}
-	return lb.config.ChainID, nil
-}
-
-func (lb *LocalBackend) TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
```
hyunsooda, line 258: Why was the decision made to omit the migration of this function? Its absence could potentially lead to a crash during bridge deployment, which wasn't addressed in the unit tests.

accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +144,197 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+	state, err := b.bc.State()
+	if err != nil {
+		return 0, err
+	}
+	return state.GetNonce(account), nil
+}
+
+func (b *BlockchainContractTransactor) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
+	if b.bc.Config().IsMagmaForkEnabled(b.bc.CurrentHeader().Number) {
+		return new(big.Int).SetUint64(b.bc.Config().Governance.KIP71.UpperBoundBaseFee), nil
```
blukat29, line 183: I think `[latest block BaseFee] * 2` is more appropriate, as in our `eth_gasPrice` API ([code](https://github.com/klaytn/klaytn/blob/v1.11.0/node/cn/gasprice/gasprice.go#L99), [code](https://github.com/klaytn/klaytn/blob/v1.11.0/api/api_public_klay.go#L45))

Depending on the use cases,
- When the transaction must succeed, specify high GasPrice in `TransactOpts`
  - e.g. in service chain bridges where the operator balance should be enough.
- When we need realistic gas prices, rely on SuggestGasPrice
  - e.g. when we add a CLI command to publish the BLS public key to a system contract. In this case the sender's balance might have just enough balance to send transaction at normal BaseFee, but not UpperBoundBaseFee.

accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +144,197 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+	// Create a helper to check if a gas allowance results in an executable transaction
+	executable := func(gas uint64) (bool, *blockchain.ExecutionResult, error) {
+		call.Gas = gas
+
+		currentState, err := b.bc.State()
+		if err != nil {
+			return true, nil, nil
+		}
+		bc := NewBlockchainContractCaller(b.bc)
+		res, err := bc.callContract(call, b.bc.CurrentBlock(), currentState)
```
blukat29, line 205: Why didn't BlockchainContractTransactor inherit (i.e. embed) the BlockchainContractCaller?

In other words, is this possible?
```suggestion
		res, err := b.CallContract(context.Background(), call, nil)
```

accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +144,197 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+
+// This nil assignment ensures at compile time that BlockchainContractCaller implements bind.ContractCaller.
+var _ bind.ContractBackend = (*BlockchainContractBackend)(nil)
+
+func NewBlockchainContractBackend(bc BlockChainForCaller, es *filters.EventSystem, tp *blockchain.TxPool) *BlockchainContractBackend {
+	return &BlockchainContractBackend{
+		BlockchainContractCaller{bc},
+		BlockchainContractTransactor{bc, tp},
+		BlockchainContractFilterer{bc, es},
+	}
```
blukat29, line 326: Can we have one unified class `BlockchainContractBackend` that implements all methods?
i.e.

```
type BlockchainContractBackend struct {
	bc     BlockChainForCaller
	txPool *blockchain.TxPool
	events *filters.EventSystem
}

var _ bind.ContractCaller = (*BlockchainContractBackend)(nil)
var _ bind.ContractTransactor = (*BlockchainContractBackend)(nil)
var _ bind.ContractFilterer = (*BlockchainContractBackend)(nil)
var _ bind.ContractBackend = (*BlockchainContractBackend)(nil)
```

accounts/abi/bind/backends/blockchain.go
2dvorak: `txPool` should be provided from the caller of the `BlockchainContractTransactor` instance. If the caller didn't provide `txPool` for the corresponding blockchain backend (e.g., pass `nil` in the argument), the `txPool` would be `nil`.

I could force caller to pass non-nil argument for `txPool`, however there are some functions that does not depend on the `txPool` in `BlockchainContractTransactor`. So instead I added above checks in functions that uses `txPool`.

node/sc/local_backend.go
2dvorak: Because `bind.ContractTransactor` does not include `TransactionReceipt` and I tested value transfer across bridge without `TransactionReceipt` which worked fine.

Also, adding `TransactionReceipt` does not look trivial because `BlockChainForCaller` provides a limited set of functions.. I'll have to manually iterate block 0 ~ current and find if corresponding transaction exists.

accounts/abi/bind/backends/blockchain.go
2dvorak: Updated in c2b0f3d, thanks!

accounts/abi/bind/backends/blockchain.go
2dvorak: Updated in c2b0f3d, thanks!

accounts/abi/bind/backends/blockchain.go
2dvorak: Updated in c2b0f3d, thanks!

accounts/abi/bind/backends/blockchain_test.go
2dvorak: Updated in c2b0f3d, thanks!

accounts/abi/bind/backends/blockchain_test.go
2dvorak: Updated in c2b0f3d, thanks!

accounts/abi/bind/backends/blockchain.go
2dvorak: Yes, however we still need to call `callContract` (not `CallContract`) to get `*blockchain.ExecutionResult`..

Do you think it is more appropriate to embed `BlockchainContractCaller` in `BlockchainContractTransactor`? I implemented corresponding structs for `bind.Contract*` and there was no reason for not embedding `BlockchainContractCaller` in `BlockchainContractTransactor`. But I just found out that `bind.ContractTransactor` states that it `defines the methods needed to allow operating with contract on a write only basis` in this [comment](https://github.com/klaytn/klaytn/blob/v1.11.0/accounts/abi/bind/backend.go#L69).

accounts/abi/bind/backends/blockchain.go
2dvorak: I think we can.

What you're suggesting is to remove `BlockchainContractCaller`, `BlockchainContractTransactor`, and `BlockchainContractFilterer` and move all their methods to `BlockchainContractBackend`, right?

blukat29: @2dvorak Yes, you got it correctly.

accounts/abi/bind/backends/blockchain.go
hyunsooda: Relocation the check to the inside of `NewBlockchainContractBackend` does not clearly solve?

node/sc/local_backend.go
hyunsooda: There are two ways to bridge contract deployments.
1.  By contract deployment transaction through web3, ethers, caver, etc.
2. `subbridge.deployBridge()` RPC

There's no side effect with the first approach. The second one also works but it has different behavior internally.
As the localbackend does not have `TransactionReceipt` method, which is deleted in this PR, the [type assertion](https://github.com/klaytn/klaytn/blob/dev/node/sc/bridge_manager.go#L939C1-L943C3) fails here and do not call `WaitDeployed()`. Thus, I suggest that the refactoring should refrain from any underlying behaviors to be modified implicitly. 

>  I'll have to manually iterate block 0 ~ current and find if corresponding transaction exists.

The following code snippet works. I'll look into this further in detail later if you confirm.
```go
func (b *BlockchainContractCaller) TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
    bc := b.bc.(*blockchain.BlockChain) // Avoid including a conversion check at this point, as the BlockChain struct conversion must be done. Additionally, there is no anticipation of encountering errors in this function.
    receipt := bc.GetReceiptByTxHash(txHash)
    if receipt != nil {
        return receipt, nil
    }
    return nil, errors.New("receipt does not exist")
}
```

accounts/abi/bind/backends/blockchain.go
2dvorak: Yes, that could be one option.

I thought it is OK to allow calller to pass `nil` for `txPool` when calling `NewBlockchainContractTransactor`. Then, the caller could only use methods that does not depend on `txPool` (e.g., `EstimateGas`). I'm not sure how many cases would use `BlockchainContractTransactor` like that way.

I'll update to force `txPool` to be non-nil.

node/sc/local_backend.go
blukat29: I agree with @hyunsooda that the `BlockChainContractBackend` should implement `DeployBackend`. `WaitDeployed` and `WaitMined` is such a useful feature.

I understand that `BlockChainForCaller` has only a limited set of methods. However, the concrete instance will be `*blockchain.BlockChain` in most cases, if not all. So `b.bc.(*blockchain.BlockChain)` should work fine (at least when we use `WaitDeployed` and such)

The following method might look more panic-safe.
```go
    if bc, ok := b.bc.(*blockchain.BlockChain); ok {
    	// return receipt or error
    } else {
        return nil, errors.New("Not BlockChain")
    }
```

accounts/abi/bind/backends/blockchain.go
blukat29: If we unify the classes into one BlockchainContractBackend (see https://github.com/klaytn/klaytn/pull/1944#discussion_r1308460561), `b.txPool` may be nil as @2dvorak explained.

I'd say we keep the nil check.

accounts/abi/bind/backends/blockchain.go
2dvorak: Done in a5d598c, PTAL

node/sc/local_backend.go
2dvorak: @hyunsooda @blukat29 Thanks for the feedbacks, PTAL at a5d598c.

accounts/abi/bind/backends/blockchain.go
2dvorak: Done in a5d598c, PTAL

accounts/abi/bind/backends/blockchain.go
blukat29: @2dvorak Sure thing.

blukat29 (approved):

tests/kip103_test.go
```diff
@@ -114,13 +38,19 @@ func TestRebalanceTreasury_EOA(t *testing.T) {
 	}()
 
 	optsOwner := bind.NewKeyedTransactor(validator.Keys[0])
-	transactor := &testKip103TxTransactor{node: node}
-	targetBlockNum := new(big.Int).Add(node.BlockChain().CurrentBlock().Number(), big.NewInt(5))
+	es := filters.NewEventSystem(node.EventMux(), node.APIBackend, false)
+	transactor := backends.NewBlockchainContractBackend(node.BlockChain(), node.TxPool().(*blockchain.TxPool), es)
```
ian0371, line 42: Is `es` required here?

2dvorak:
@ian0371 Would it be beneficial to add `NewBlockchainContractCaller` and so on?

Pros
- It may help caller to decide which arguments to provide. If the caller needs `bind.ContractCaller` methods, then just use `NewBlockchainContractCaller(bc)` instead of `NewBlockchainContractBackend(bc, nil, nil)`. If we only have `NewBlockchainContractBackend`, the caller could have to investigate if he/she has to provide `txPool` or `events` to use `bind.ContractCaller` methods

Cons
- `NewBlockchainContract*` would all return `*BlockchainContractBackend` type, this may misleading
- Should we block nil arguments for `NewBlockchainContractBackend` when we have `NewBlockchainContractCaller`? Then how? (`NewBlockchainContractBackend` does not have error to return, then should we panic..?)

cc. @blukat29 @hyunsooda

tests/kip103_test.go
2dvorak: Nope, removed in 94b9f33.

accounts/abi/bind/backends/blockchain.go
```diff
@@ -141,3 +152,175 @@ func (b *BlockchainContractCaller) getBlockAndState(num *big.Int) (*types.Block,
...
+	// TODO-Klaytn this is not pending nonce but latest nonce
+	state, err := b.bc.State()
+	if err != nil {
+		return 0, err
+	}
+	return state.GetNonce(account), nil
+}
+
+func (b *BlockchainContractBackend) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
+	if b.bc.Config().IsMagmaForkEnabled(b.bc.CurrentHeader().Number) {
```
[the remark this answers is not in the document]
2dvorak: Good point, thanks. Updated in 94b9f33.

accounts/abi/bind/backends/blockchain.go
```diff
@@ -49,26 +52,34 @@ type BlockChainForCaller interface {
...
+	_ bind.ContractFilterer   = (*BlockchainContractBackend)(nil)
+	_ bind.DeployBackend      = (*BlockchainContractBackend)(nil)
+	_ bind.ContractBackend    = (*BlockchainContractBackend)(nil)
+)
 
-func NewBlockchainContractCaller(bc BlockChainForCaller) *BlockchainContractCaller {
-	return &BlockchainContractCaller{
-		bc: bc,
-	}
+func NewBlockchainContractBackend(bc BlockChainForCaller, tp *blockchain.TxPool, es *filters.EventSystem) *BlockchainContractBackend {
```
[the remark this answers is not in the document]
2dvorak: All cases are possible. The use cases could be:
- bc, nil, nil - `bind.ContractCaller`
- bc, tp, nil - `bind.ContractTransactor`
- bc, nil, es - `bind.ContractFilterer`
- bc, tp, es - `bind.ContractBackend`

Added comments in 94b9f33.

ian0371 (approved):
For now, let's leave it as a comment and ponder on this later.
I think blocking nil arguments will degrade usability.

blukat29 (approved):

hyunsooda (approved):

Issue #1850, opened 2023-05-22
Support the full features of AccessList tx type and consider the increment of CallStipend gas cost


aidan-kwon:
**Is your request related to a problem? Please describe.**
After adopting EIP2929, the gas cost for the initial account access increased. Sometimes, it causes a failure of existing contract behavior. One of the cases is [Klaytn Safe and gnosis wallet](https://github.com/safe-global/safe-contracts/blob/e870f514ad34cd9654c72174d6d4a839e3c6639f/contracts/base/FallbackManager.sol#L61-L82) and this kind of pattern is widely used recently. 

**Describe the solution you'd like**
To mitigate this issue (partially), we need to support the full features of TxTypeEthereumAccessList that pre-pay gas fee for account and storage as Ethereum already did. Indeed, we can consider the increment of CallStipend gas cost as well because Klaytn's native transaction type doesn't support the access list. 

- [ ] Support the full features of TxTypeEthereumAccessList that pre-pay gas fee 
- [ ] Consider the increment of CallStipend gas cost

PR #1705, opened 2022-11-16
vm: bringing gascost change eips

yoomee1313:
## Proposed changes

- This PR brings gas cost change eips to klaytn.
  - [EIP-3529: Reduction in refunds](https://eips.ethereum.org/EIPS/eip-3529)
  - [EIP-2929: Gas cost increases for state access opcodes](https://eips.ethereum.org/EIPS/eip-2929)
  - [EIP-2565: ModExp Gas Cost](https://eips.ethereum.org/EIPS/eip-2565)

- Please review this PR commit by commit.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [ ] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- #1677

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1955, opened 2023-09-04
Add support for AccessList tx

2dvorak:
## Proposed changes

- Although we accept `TxTypeEthereumAccessList` after EthTxType hardfork and EIP-2930 has been accounted after Kore hardfork(#1705), the provided access list is not properly handled (i.e., has no benefit using access list)
- This PR adds full support for address list

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- #1850

### DIFF
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -1411,19 +1411,19 @@ func TestAccessListTx(t *testing.T) {
 	config.Governance = params.GetDefaultGovernanceConfig()
 	config.Governance.KIP71.LowerBoundBaseFee = 0
 	var (
-		aa     = common.HexToAddress("0x000000000000000000000000000000000000aaaa")
-		engine = gxhash.NewFaker()
+		contractAddr = common.HexToAddress("0x000000000000000000000000000000000000aaaa")
+		engine       = gxhash.NewFaker()
+		signer       = types.LatestSigner(config)
 
 		// A sender who makes transactions, has some funds
-		key, _  = crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
-		address = crypto.PubkeyToAddress(key.PublicKey)
-		funds   = big.NewInt(1000000000000000)
-		gspec   = &Genesis{
+		senderKey, _ = crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
+		senderAddr   = crypto.PubkeyToAddress(senderKey.PublicKey)
+		senderNonce  = uint64(0)
+		gspec        = &Genesis{
 			Config: config,
 			Alloc: GenesisAlloc{
-				address: {Balance: funds},
-				// The address 0xAAAA sloads 0x00 and 0x01
-				aa: {
+				senderAddr: {Balance: big.NewInt(params.KLAY)},
+				contractAddr: { // SLOAD 0x00 and 0x01
 					Code: []byte{
 						byte(vm.PC),
 						byte(vm.PC),
@@ -1438,7 +1438,7 @@ func TestAccessListTx(t *testing.T) {
 		db       = database.NewMemoryDBManager()
 		block    = gspec.MustCommit(db)
 		accesses = types.AccessList{{
-			Address:     aa,
+			Address:     contractAddr,
 			StorageKeys: []common.Hash{{0}},
 		}}
 	)
@@ -1450,22 +1450,25 @@ func TestAccessListTx(t *testing.T) {
 	}
 	defer chain.Stop()
 
-	// Generate blocks
-	blocks, _ := GenerateChain(gspec.Config, block, engine, db, 1, func(i int, b *BlockGen) {
-		b.SetRewardbase(common.Address{1})
+	// helper function to insert a block with a transaction
+	insertBlockWithTx := func(list types.AccessList) *types.Block {
+		// Generate blocks
+		blocks, _ := GenerateChain(gspec.Config, block, engine, db, 1, func(i int, b *BlockGen) {
+			b.SetRewardbase(common.Address{1})
 
-		// One transaction to 0xAAAA, with access list
-		intrinsicGas, _ := types.IntrinsicGas([]byte{}, accesses, false, gspec.Config.Rules(big.NewInt(1)))
-		signer := types.LatestSigner(gspec.Config)
-		tx, _ := types.SignTx(types.NewMessage(address, &aa, 0, big.NewInt(0), 30000, big.NewInt(1), []byte{}, false, intrinsicGas, accesses), signer, key)
-		b.AddTx(tx)
-	})
-
-	if n, err := chain.InsertChain(blocks); err != nil {
-		t.Fatalf("block %d: failed to insert into chain: %v", n, err)
+			// One transaction to 0xAAAA
+			intrinsicGas, _ := types.IntrinsicGas([]byte{}, list, false, gspec.Config.Rules(block.Number()))
+			tx, _ := types.SignTx(types.NewMessage(senderAddr, &contractAddr, senderNonce, big.NewInt(0), 30000, big.NewInt(1), []byte{}, false, intrinsicGas, list), signer, senderKey)
+			b.AddTx(tx)
+		})
+		if n, err := chain.InsertChain(blocks); err != nil {
+			t.Fatalf("block %d: failed to insert into chain: %v", n, err)
+		}
+		senderNonce++
+		return chain.CurrentBlock()
 	}
 
-	block = chain.GetBlockByNumber(1)
+	block = insertBlockWithTx(accesses) // with AccessList
 
 	// Expected gas is intrinsic + 2 * pc + hot load + cold load, since only one load is in the access list
 	expected := params.TxGas + params.TxAccessListAddressGas + params.TxAccessListStorageKeyGas +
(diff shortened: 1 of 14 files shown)


consensus/istanbul/backend/kip103.go
```diff
@@ -42,8 +42,12 @@ func (caller *Kip103ContractCaller) CallContract(ctx context.Context, call klayt
 	// call.To: the target contract address will be assigned by `BoundContract`
 	// call.Value: nil value is acceptable for `types.NewMessage`
 	// call.Data: a proper value will be assigned by `BoundContract`
+	var accessList types.AccessList
+	if call.AccessList != nil {
+		accessList = *call.AccessList
+	}
 	msg := types.NewMessage(call.From, call.To, caller.state.GetNonce(call.From),
-		call.Value, gasLimit, gasPrice, call.Data, false, intrinsicGas)
+		call.Value, gasLimit, gasPrice, call.Data, false, intrinsicGas, accessList)
```
hyeonLewis, line 50: Do we need to pass `accessList` instead of `nil`?

blockchain/state_transition.go
```diff
@@ -335,7 +337,7 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 
 	rules := st.evm.ChainConfig().Rules(st.evm.Context.BlockNumber)
 	if rules.IsKore {
-		st.state.PrepareAccessList(rules, msg.ValidatedSender(), msg.ValidatedFeePayer(), st.evm.Coinbase, msg.To(), vm.ActivePrecompiles(rules))
+		st.state.PrepareAccessList(rules, msg.ValidatedSender(), msg.ValidatedFeePayer(), st.evm.Coinbase, msg.To(), vm.ActivePrecompiles(rules), msg.AccessList())
```
hyeonLewis, line 340: Will `msg.AccessList()` be always `nil`? Or doesn't it matter whether `nil` or not?

consensus/istanbul/backend/kip103.go
2dvorak: I'm not sure if caller of `Kip103ContractCaller` would specify access list when calling `CallContract`. But wouldn't it be OK to properly handle access list if it was specified?

blockchain/state_transition.go
2dvorak: @hyeonLewis PTAL

consensus/istanbul/backend/kip103.go
2dvorak: I agree that it would be better not to handle access list here because:
- it's a call so handling access list won't have any effect
- the caller of `Kip103ContractCaller` probably won't use access list

Updated, PTAL

blockchain/blockchain_test.go
```diff
@@ -1397,6 +1397,107 @@ func TestLargeReorgTrieGC(t *testing.T) {
...
+		// A sender who makes transactions, has some funds
+		key, _  = crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
+		address = crypto.PubkeyToAddress(key.PublicKey)
+		funds   = big.NewInt(1000000000000000)
+		gspec   = &Genesis{
+			Config: config,
+			Alloc: GenesisAlloc{
+				address: {Balance: funds},
+				// The address 0xAAAA sloads 0x00 and 0x01
+				aa: {
```
blukat29, line 1426:
```suggestion
		contractAddr = common.HexToAddress("0x000000000000000000000000000000000000aaaa")
		engine = gxhash.NewFaker()
		signer = types.LatestSigner(config)

		// A sender who makes transactions, has some funds
		senderKey, _  = crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
		senderAddr = crypto.PubkeyToAddress(senderKey.PublicKey)
		gspec   = &Genesis{
			Config: config,
			Alloc: GenesisAlloc{
				senderAddr: {Balance: big.NewInt(params.KLAY)},
				contractAddr: { // SLOAD 0x00 and SLOAD 0x01
```

blockchain/blockchain_test.go
```diff
@@ -1397,6 +1397,107 @@ func TestLargeReorgTrieGC(t *testing.T) {
...
+		}}
+	)
+
+	// Import the canonical chain
+	chain, err := NewBlockChain(db, nil, gspec.Config, engine, vm.Config{})
+	if err != nil {
+		t.Fatalf("failed to create tester chain: %v", err)
+	}
+	defer chain.Stop()
+
```
blukat29, line 1452:
```go
insertBlockWithTx = func(list *AccessList) {
    GenerateChain()
    InsertChain()
    return CurrentBlock()
}

block := insertBlockWithTx(accessList) // with AccessList
expectedGas := params.TxGas + ...
assert.Equal(t, expectedGas, block.GasUsed())

block := insertBlockWithTx(nil) // without AccessList
expectedGas := params.TxGas + ...
assert.Equal(t, expectedGas, block.GasUsed())
```

blockchain/state/statedb.go
```diff
@@ -1102,14 +1102,13 @@ func (s *StateDB) PrepareAccessList(rules params.Rules, sender, feepayer, coinba
...
-	//	for _, key := range el.StorageKeys {
-	//		s.AddSlotToAccessList(el.Address, key)
-	//	}
-	//}
+	for _, el := range list {
+		s.AddAddressToAccessList(el.Address)
+		for _, key := range el.StorageKeys {
+			s.AddSlotToAccessList(el.Address, key)
+		}
+	}
```
blukat29, line 1111: Can we check `rules.IsCancun` here instead of state_transition.go:346?
The similar `rules.IsShanghai` is also below here.

2dvorak: Good point, thanks.

BTW, should we move Kore check in `state_transition.go` to `statedb.go` too? It looks like geth does so.

blukat29: That's not bad. Please check other instances of `if IsKore { PrepareAccessList }` pattern.

blockchain/types/transaction.go
```diff
@@ -1024,16 +1024,21 @@ func (t *TransactionsByTimeAndNonce) Pop() {
...
 	transaction := &Transaction{
 		validatedIntrinsicGas: intrinsicGas,
 		validatedFeePayer:     from,
 		validatedSender:       from,
 		checkNonce:            checkNonce,
 	}
 
-	internalData := newTxInternalDataLegacyWithValues(nonce, to, amount, gasLimit, gasPrice, data)
-	transaction.setDecoded(internalData, 0)
+	if list != nil {
```
hyunsooda, line 1035: Hardfork condition seems required. What do you think?

blockchain/blockchain_test.go
```diff
@@ -1397,6 +1397,107 @@ func TestLargeReorgTrieGC(t *testing.T) {
...
+		vm.GasQuickStep*2 + params.WarmStorageReadCostEIP2929 + params.ColdSloadCostEIP2929
+	if block.GasUsed() != expected {
+		t.Fatalf("incorrect amount of gas spent: expected %d, got %d", expected, block.GasUsed())
+	}
+
+	// Generate blocks
+	blocks, _ = GenerateChain(gspec.Config, block, engine, db, 1, func(i int, b *BlockGen) {
+		b.SetRewardbase(common.Address{1})
+
+		// One transaction to 0xAAAA, with access list
```
hyunsooda:
```suggestion
		// One transaction to 0xAAAA, without access list
```

hyunsooda (commented):
Creating access list for dynamic storage keys for array or mapping is not easy to exploit 100%. The reference `eth_createAccessList` API seems perfectly fits for the access create as well as expected gasused because the listing access list should be careful to not pay more compared with naive tx. So, I think `eth_createAccessList` API needs to be implemented together. What do you think?

blockchain/blockchain_test.go
2dvorak: That comment is removed, thanks.

blockchain/blockchain_test.go
2dvorak: Done, thanks

blockchain/blockchain_test.go
2dvorak: Done, thanks

2dvorak:
> Creating access list for dynamic storage keys for array or mapping is not easy to exploit 100%. The reference `eth_createAccessList` API seems perfectly fits for the access create as well as expected gasused because the listing access list should be careful to not pay more compared with naive tx. So, I think `eth_createAccessList` API needs to be implemented together. What do you think?

@hyunsooda Good point, however `createAccessList` will be covered in a separate PR. I created a new issue regarding your point #1976

blockchain/types/transaction.go
2dvorak: I've looked into references to `NewMessage`, and it seems they are all related to calls (e.g., `klay_call` API).
Since calls do not affect states, hardfork condition could be omitted.

hyunsooda: Could you check it out again? I found one:

Call, EstimateGas (api_ethereum.go) -> EthDoCall (api_ethereum.go) -> args.ToMessage (api_ethereum.go) -> ToMessage (tx_args.go) -> NewMessage (transaction.go)

2dvorak: They are 'call's, right? The `eth_call` or `eth_estimateGas` could behave differently, however that won't change the states.
Could you help me with a concrete example for problematic case? I cannot come up with one.

hyunsooda: Aha, yes, it would not be required. I'm sorry I overlooked your first comment. Thank you.

hyunsooda (commented):
Did you deliberately choose to exclude the access list check in the `LocalBackend.callContract()` function within the SC package?

2dvorak:
> Did you deliberately choose to exclude the access list check in the `LocalBackend.callContract()` function within the SC package?

@hyunsooda No, the `LocalBackend` was replaced by `BlockchainContractBackend` in #1944.

hyunsooda (approved):
I forgot that I reviewed that PR.. Two noise comments on this PR. I'm sorry.

Looks fine and correctly works.

hyeonLewis (approved):

blukat29 (approved):

PR #1956, opened 2023-09-06
common: use crypto/rand library for random byte generation

yoomee1313:
## Proposed changes
- Database testcase has been failed frequently. The reason is because`insertRandomData` generates test key,value data for batchwrite and the key can be duplicated. If the key is duplicated, the old one in batch is overwritten. However, the old one is not deleted from the expect value for the test case so validation failed.
- I have changed the rand library from math/rand to crypto/rand to reduce the collision. macos has not been failed since the math/rand implementation was different from linux.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/common/bytes.go
+++ b/common/bytes.go
@@ -21,10 +21,10 @@
 package common
 
 import (
+	"crypto/rand"
 	"encoding/binary"
 	"encoding/hex"
-	"math/rand"
-	"time"
+	"io"
 )
 
 // ToHex returns the hex representation of b, prefixed with '0x'.
@@ -161,9 +161,11 @@ func TrimRightZeroes(s []byte) []byte {
 }
 
 func MakeRandomBytes(n int) []byte {
-	rand.Seed(time.Now().UTC().UnixNano())
 	s := make([]byte, n)
-	rand.Read(s)
+	_, err := io.ReadFull(rand.Reader, s)
+	if err != nil {
+		return nil
+	}
 	return s
 }
 
--- a/storage/database/database_test.go
+++ b/storage/database/database_test.go
@@ -394,10 +394,10 @@ func insertRandomData(db KeyValueWriter, prefix []byte, num int) (testDataSlice,
 	ret := testDataSlice{}
 	for i := 0; i < num; i++ {
 		key := common.MakeRandomBytes(32)
+		val := append(key, key...)
 		if len(prefix) > 0 {
 			key = append(prefix, key...)
 		}
-		val := common.MakeRandomBytes(100)
 		ret = append(ret, &testData{k: key, v: val})
 
 		if err := db.Put(key, val); err != nil {


blukat29 (approved):
Thanks!

common/bytes.go
```diff
@@ -161,9 +161,11 @@ func TrimRightZeroes(s []byte) []byte {
 }
 
 func MakeRandomBytes(n int) []byte {
-	rand.Seed(time.Now().UTC().UnixNano())
```
hyunsooda, line 164: To be duplicated, the UTC value must be the same. It implies there are at least two callers that call the function `MakeRandomBytes()` that have the same timing at the level of nanosecond. This case seems really rare. 

i.e.,
```
common.MakeRandomBytes() // call1
common.MakeRandomBytes() // call2
// the nanosecond of call1 and call2 must be the same to output the same result.
```

I'm intrigued by the feasibility of this phenomenon. Within my local environment, I've noticed that two consecutive calls to `time.Now().UTC().UnixNano()` yield distinct time intervals. Triggering this path appears to be quite challenging. Can you give me a case you found so that I can get it clear?

yoomee1313: I have tested in docker environement [klaytn/build_base/1.2-ubuntu18.04-go1.20.6-solc0.8.13](https://hub.docker.com/layers/klaytn/build_base/1.2-ubuntu18.04-go.1.20.6-solc0.8.13/images/sha256-cbb3c5bdfe7eb089bb6933c3f3bed2edf90dc05ccddb50ab4bde13c61646c5a5?context=explore) which OS is ubuntu18.04.
What was the OS of your local enviroment?

hyunsooda: I tested it in Ubuntu 18.04 as well, and still the output of `MakeRandomBytes` was not the same.

What does this means?
> hasn't failed unlike linux failed due to the different implementation of Rand.Seed

The input is clearly the current time, in nanosecond unit. Regardless, did you obtain identical outputs by invoking `MakeRandomBytes` consecutively twice, and did this change resolve the issue?

yoomee1313: @hyunsooda Sorry for the late. I see now. The UTC calls should be different. but what I meant was the random generator algorithm. It is well-known that crypto/random generates cryptographically secure random value, but math/random generates psuedo random value. Anyway, math/rand.Seed() is deprecated since go1.20. Check here: https://cs.opensource.google/go/go/+/go1.21.1:src/math/rand/rand.go;l=394. I could only change the code as it suggested, but I changed the library too.

Btw, I was unable to reproduce in macOS and I thought it's because the syscall it used is different from linux. I'm not sure so if there's any misunderstanding, please share me. Check here: https://pkg.go.dev/crypto/rand#Read

>Regardless, did you obtain identical outputs by invoking MakeRandomBytes consecutively twice, and did this change resolve the issue?
by changing the code not to use math/rand.Seed() solves the problem, but I changed the value too just in case..

hyunsooda: I just had a doubt that the error localization was correct. For now, if this change solves it, go with this, and let’s dig in later when the same problem pops up.

I quickly skimmed the implementation of `math.rand/Read` and `crypto.rand/Read` seems not to have a distinct implementation per OS.

hyunsooda (approved):

PR #1756, opened 2023-01-06
updated governance state after setHead

jeongkyun-oh:
## Proposed changes

- Describe your changes to communicate to the maintainers why we should accept this pull request.
- If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [ ] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1186, opened 2022-02-17
Add a logic to delete Governance, StakingInfo data from DB

hqjang-pepper:
## Proposed changes

- Currently there is no logic to delete data of governance and stakingInfo DB.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

https://github.com/klaytn/klaytn/issues/1165

## Further comments
Some things need to be checked
- updating actual governance block
- when write governance state, how can we load data of items?

PR #1966, opened 2023-09-12
Fixed corrupt governance state after rewinding chain

hyunsooda:
## Proposed changes

Added additional clear logics after  `debug.SetHead`.

### Problem Statement
The existing `setHead` implementation considered only for the block data such as receipts and transactions. This PR addresses governance (header and contract) and staking state.

### What PR Does
```
---------------------
setHead (remove transaction, receipt, and block)
remove governance database
---------------------
Post-task
  1. Initiailize snapshot cache
  2. Initiailize governance cache and last state block number
```

Compared to two related PRs (https://github.com/klaytn/klaytn/pull/1186, https://github.com/klaytn/klaytn/pull/1756), this PR does not bridge additional implementation to fix this problem. I just defined the post-task and these utilize the existing implementation.

### Cases
Case 1: header governance (+ cache)
  - Verification error fixed
  - Delete local database and memory cache

Case 2: contract governance
  - We don't need to put additional workaround because the contract-based governance state does not have a dependency on local MiscDB and in-memory cache.

Case 3: staking info DB (+ cache)
  - Delete local database and cache

Case 4: Other DBs (i.e., state DB, MiscDB, etc)
  - There seems not a strong motivation to remove existing state DB by the `setHead` call.

### Bug fix

The original codebase contained a bug where a discrepancy arises between the blockchain state and the headerchain state when lost state data appears because blockchain goes further delete operation exceeding the origin target number which is the origin block number to be rewound. Fixed in this PR together.

### Test
[Prepration]
```
epoch = 10
# of validators = 4
```

[Execution]
1. Vote the following item at the beginning of the network. `governance.vote("reward.mintingamount","123")`
2. `klay.getHeader(5)` contains non-empty vote data
3. `klay.getHeader(10)` contains non-empty governance data
4. Let's rewind 1,2,3,4,5,6,7,8,9,10
    ex) `debug.setHead(7)`
5. Wait to be synced from connected peers
    - Before this PR: A bad block appears at the time of verifying block number 10, which contains non-empty governance data.
    - After this PR: Fixed
6. Catch up the latest block

[Secanrio test]
(Preparation is the same with above)

1. Vote at block 33
2. Terminate a node at block 36
3. Restart the node and wait to sync latest block number
4. Failed to verify block 40 that contains non-empty governnace data
5. call `klay.setHead(30)`.
6. Successfully catchup the latest block number
7. (additional) call `klay.setHead()` with `10, 20, 30, 33, 35, 40, n` works.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


governance/default.go
```diff
@@ -1027,6 +1028,14 @@ func (gov *Governance) UnmarshalJSON(b []byte) error {
...
 }
 
+func (gov *Governance) InitGovCache() {
+	gov.initializeCache(gov.ChainConfig)
+}
+
+func (gov *Governance) InitLastGovStateBlkNum() {
+	atomic.StoreUint64(&gov.lastGovernanceStateBlock, 0)
+}
+
```
blukat29, line 1038: Can we merge both functions? In the current PR, they always go together.

storage/database/db_manager.go
```diff
@@ -2697,6 +2689,30 @@ func (dbm *databaseManager) WriteGovernanceIdx(num uint64) error {
 	return db.Put(governanceHistoryKey, data)
 }
 
+// DeleteGovernanceIdxRange leaves governanceHistory from 0 to `num`
```
blukat29, line 2692:
```suggestion
// DeleteGovernanceIdxRange deletes governanceHistory beyond `num`
```

blukat29: But since `delFn` and `DeleteGovernance` is called at every block,
This function might peek at the last element of the idxHistory and delete on exact match.

governance/default.go
hyunsooda: I deliberately and thoughtfully kept them separate, considering the possibility of future use.  Do you believe merging would be a better option

storage/database/db_manager.go
hyunsooda: The governance index is read and stored in the unit of the list by the existing implementation. So, the function `deleteGovernanceIdxRange` stores an index list from 0 to `num`. To peek at one element, the existing implementation should be rewritten (not changed in this PR) at read and store by a pair of number and index.

blukat29: Oh, by peeking at the last element, I simply meant something like
```go
end := len(idxHistory)
if idxHistory[end-1] == num {
  data, err := json.Marshal(idxHistory[0:end-1])
  db.Put(governanceHistoryKey, data)
}
```
No need to change the database schema.

governance/default.go
blukat29: If you expect future use, then I'm not against.

storage/database/db_manager.go
hyunsooda: Do you mean the change provided can make a shortcut, which is an alternative to the following loop?

```go
for i := 0; i < len(idxHistory); i++ {
    if idxHistory[i] > num {
        break
    }
    newSlice = append(newSlice, idxHistory[i])
}
```

blukat29: I wanted to make `DeleteGovernance` analogous to `DeleteBody` and `DeleteReceipts` in that they delete the elements strictly at `num`.

Your loop indeed handles every element at once and is therefore efficient, if called only once.
But `DeleteGovernance` is called multiple times in `bc.hc.SetHead`, because of the loop that calls `delFn` for every block number. The time complexity here is `len(idxHistory) * [number of blocks rewinded]`.
That's why I suggested making `DeleteGovernanceIdxRange` an O(1) algorithm without the loop.

storage/database/db_manager.go
```diff
@@ -2697,6 +2689,30 @@ func (dbm *databaseManager) WriteGovernanceIdx(num uint64) error {
 	return db.Put(governanceHistoryKey, data)
 }
 
+// DeleteGovernanceIdxRange deletes governanceHistory beyond `num`
```
ian0371, line 2692:
```suggestion
// deleteGovernanceIdxRange deletes governanceHistory beyond `num`
```

storage/database/db_manager.go
ian0371: In case the suggestion is accepted, the name should be revised as well: maybe `deleteGovernanceIdx`

node/cn/api_backend_test.go
```diff
@@ -658,3 +681,125 @@ func TestCNAPIBackend_IsSenderTxHashIndexingEnabled(t *testing.T) {
...
+		expHeadFastBlock:   6,
+		expHeadBlock:       6,
+	})
+}
+
+func testCfg(epoch uint64) *params.ChainConfig {
+	config := params.TestChainConfig.Copy()
+	config.Governance = params.GetDefaultGovernanceConfig()
+	config.Istanbul = params.GetDefaultIstanbulConfig()
+	config.Istanbul.Epoch = 5
```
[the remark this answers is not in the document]
hyunsooda: Thanks.

storage/database/db_manager.go
hyunsooda: Changed. If the function name feels verbose or not appropriate, please suggest an alternative one.

hyunsooda:
As this change deletes memory cache and local database only, resync resolve it natrually. Not changed process and communication at consensus and governance layers, thus no hardfork seems required.

blukat29 (approved):

hyunsooda:
@blukat29, Updated the description of how to test. PTAL.

2dvorak (approved):

hyunsooda:
Given feedback, the updated version removes snapshot and staking info database and memory cache. Also, added an additional fix for the previously existing bug. The description updated. PTAL. Thank you very much.

blockchain/blockchain.go
```diff
@@ -627,10 +636,31 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
...
+		epoch           = bc.Config().Istanbul.Epoch
+		twoEpochs       = epoch * 2
+		beforeTwoEpochs uint64
+	)
+	if twoEpochs > curBlkNum {
+		beforeTwoEpochs = 1
+	} else {
+		beforeTwoEpochs = curBlkNum - twoEpochs
+	}
+	for i := curBlkNum; i >= beforeTwoEpochs; i-- {
```
ian0371, line 651: I'm wondering why the for loop's condition is `i >= beforeTwoEpochs`.

Assume there's a vote block `V` (that contains vote data) and a governance block `G` (that contains governance data).
For `setHeadBeyondRoot(x)`,
- if `x < V` or `x >= G`: no problem
- if `V <= x < G`: bad block at `G`

I think this logic addresses the second condition.
Then, shouldn't the condition of this for loop be `i >= V`?
If it's hard to fetch `V`, I'm thinking `x - ( x % epoch)` would do (which will be less than `V`).

blockchain/headerchain.go
```diff
@@ -407,8 +407,15 @@ func (hc *HeaderChain) SetHead(head uint64, updateFn UpdateHeadBlocksCallback, d
 		//
 		// Update head first(head fast block, head full block) before deleting the data.
 		if updateFn != nil {
-			if err := updateFn(parent); err != nil {
-				return err
+			latestBlkNum, err := updateFn(parent)
+			if err != nil {
+				// rewound blockchain to genesis block
+				head = 0
```
ian0371, line 413: I'm not sure if this is ok. The previous logic stops at error, while the new logic resets to genesis at error. Could you tell me why this logic is required?

storage/database/db_manager.go
```diff
@@ -2697,6 +2698,25 @@ func (dbm *databaseManager) WriteGovernanceIdx(num uint64) error {
 	return db.Put(governanceHistoryKey, data)
 }
 
+// deleteGovernanceBeyondWithExactMatch deletes governanceHistory beyond `num`
```
ian0371, line 2701: How about this?
```suggestion
// deleteLastGovernance deletes the last governanceIdx only if it is equal to `num`
```

blockchain/headerchain.go
hyunsooda: My apologies, it should return an error. No need to delete anything further at this point. Updated.

storage/database/db_manager.go
hyunsooda: Thanks updated.

blockchain/blockchain.go
hyunsooda: A smarter comment. Thank you. Updated. Thank you for clearly pinpointing the problem that PR addresses.

ian0371 (approved):

hyunsooda:
@yoomee1313, I added the scenario test you provided in the description.

Also, during the scenario test, I found a bug, which is not related to the sethead implementation. Rather than fixing it together here, allow me to upload a dedicated one separately. Thanks.

yoomee1313:
Thanks @hyunsooda. I didn't know there was another issue..

Apart from that, I have tested next scenario and it works well with this PR.
Steps.
1. Block 15: vote
2. Block 30: governance block
3. Block 34: call setHead(20)

Result.
* dev
bad block has been occured when processing governance block(30)
```
ERROR[10/06,14:14:25 +08] [30] Verification Error                        len(receivedChangeSet)=1 len(changeSet)=0
ERROR[10/06,14:14:25 +08] [30] Verification Error                        key=reward.ratio received=30/40/30 have=nil receivedType=string haveType=nil
WARN[10/06,14:14:25 +08] [48] Failed to load old bad blocks             error="data is not found with the given key"
ERROR[10/06,14:14:25 +08] [5] ########## BAD BLOCK ######### Chain config: {ChainID: 940625
```

* PR
bad block disappeared

yoomee1313 (approved):

blukat29 (approved):

PR #1968, opened 2023-09-14
System contract registry accessors

blukat29:
## Proposed changes

- Add simple Registry contract mockup
- Actual Registry.sol will be added later.
- Add accessors to the Registry system contract.
  - `AllocRegistry` creates an initial storage state from the given contents (i.e. records). Intended to be used to generate genesis allocation and the InstallRegistry function.
  - `InstallRegistry` installs the code and initial storage to the given StateDB. Intended to be used at hardforks.
  - `ReadRegistryAddr` calls the `getActiveAddr` function.

### Use cases

- Install at genesis
  ```go
  genesisAlloc[RegistryAddr] = {
      Code: RegistryCode
      Storage: AllocRegistry(...)
  }
  ```
- Install at nonzero hardfork block number
  ```go
  // post-transaction state modifications
  func Finalize() {
      if num + 1 == config.CancunForkNumber { // If next block is fork block
          InstallRegistry(state, ...)
      }
  }
  ```
- Read registry
  ```go
  if config.IsCancunForkEnabled(num) {
      addr = ReadRegistryActiveAddr(chain, AddressBookName, num)
  } else {
      addr = addressBookAddress
  }
  ```

### Future works

- Call `AllocRegistry()` in `homi setup` if CancunForkNumber == 0
- Call `InstallRegistry()` in `Finalize()` if CancunForkNumber > 0
- Add end-to-end test in `tests` directory

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Further comments

**Why calculate solidity slots in `AllocRegistry`?**
I chose between the two approaches, (1) calculate slot index (as implemented in this PR) (2) execute a Mock contract and extract resulting storage (as in `registry_test.go:80:TestAllocRegistry`)
Option (1) was viable because slot calculation is well-documented. Option (2) was concerning because in the process of executing a Mock contract, a new BlockChain instance is created and that might interfere with global variables or logging systems.

**Why fill storage states at all?**
At hardfork, we have two options: (a) state.SetCode and state.SetStorage (b) state.SetCode and run `register()` transactions on the state. But at genesis (e.g. test chain), we only have option (a). The storage states have to be statically determined when injecting the bytecode.

### DIFF
--- a/blockchain/system/registry.go
+++ b/blockchain/system/registry.go
@@ -1,0 +1,49 @@
+// Copyright 2023 The klaytn Authors
+// This file is part of the klaytn library.
+//
+// The klaytn library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The klaytn library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+package system
+
+import (
+	"context"
+	"math/big"
+
+	"github.com/klaytn/klaytn/accounts/abi/bind"
+	"github.com/klaytn/klaytn/blockchain/state"
+	"github.com/klaytn/klaytn/common"
+	contracts "github.com/klaytn/klaytn/contracts/system_contracts"
+)
+
+func InstallRegistry(state *state.StateDB) error {
+	return state.SetCode(RegistryAddr, RegistryCode)
+}
+
+func ReadRegistryActiveAddr(backend bind.ContractCaller, name string, num *big.Int) (common.Address, error) {
+	caller, err := contracts.NewIRegistryCaller(RegistryAddr, backend)
+	if err != nil {
+		return common.Address{}, err
+	}
+
+	code, err := backend.CodeAt(context.Background(), RegistryAddr, nil)
+	if err != nil {
+		return common.Address{}, err
+	} else if code == nil {
+		return common.Address{}, ErrRegistryNotInstalled
+	}
+
+	opts := &bind.CallOpts{BlockNumber: num}
+	addr, err := caller.GetActiveContract(opts, name)
+	return addr, err
+}
--- a/blockchain/system/registry.go
+++ b/blockchain/system/registry.go
@@ -26,6 +26,55 @@ import (
 	contracts "github.com/klaytn/klaytn/contracts/system_contracts"
 )
 
+type RegistryRecord struct {
+	Addr       common.Address
+	Activation *big.Int
+}
+
+type AllocRegistryInit struct {
+	Records map[string][]RegistryRecord
+	Names   []string
+	Owner   common.Address
+}
+
+// Create storage state from the given initial values.
+// The storage slots are calculated according to the solidity layout rule.
+// https://docs.soliditylang.org/en/v1.8.20/internals/layout_in_storage.html
+func AllocRegistry(init *AllocRegistryInit) map[common.Hash]common.Hash {
+	storage := make(map[common.Hash]common.Hash)
+
+	// slot[0]: mapping(string => Record[]) records;
+	// - records[x].length @ Hash(x, 0)
+	// - records[x][i].addr @ Hash(Hash(x, 0)) + (2*i)
+	// - records[x][i].activation @ Hash(Hash(x, 0)) + (2*i + 1)
+	for name, records := range init.Records {
+		arraySlot := calcMappingSlot(0, name)
+		storage[arraySlot] = lpad32(len(records))
+
+		for i, record := range records {
+			addrSlot := calcArraySlot(arraySlot, 2, i, 0)
+			activationSlot := calcArraySlot(arraySlot, 2, i, 1)
+
+			storage[addrSlot] = lpad32(record.Addr)
+			storage[activationSlot] = lpad32(record.Activation)
+		}
+	}
+
+	// slot[1]: string[] names;
+	// - names.length @ 1
+	// - names[i] @ Hash(1) + i
+	storage[lpad32(1)] = lpad32(len(init.Names))
+	for i, name := range init.Names {
+		nameSlot := calcArraySlot(1, 1, i, 0)
+		storage[nameSlot] = encodeShortString(name)
+	}
+
+	// slot[2]: address _owner;
+	storage[lpad32(2)] = lpad32(init.Owner)
+
+	return storage
+}
+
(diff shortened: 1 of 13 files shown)


blockchain/system/registry.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return common.Address{}, err
+	}
+
+	code, err := backend.CodeAt(context.Background(), RegistryAddr, nil)
+	if err != nil {
+		return common.Address{}, err
+	} else if code == nil {
+		return common.Address{}, ErrRegistryNotInstalled
+	}
```
hyunsooda, line 44:
```suggestion
	if err != nil {
		return common.Address{}, err
	}
	if code == nil {
		return common.Address{}, ErrRegistryNotInstalled
	}
```

blockchain/system/registry.go
```diff
@@ -0,0 +1,49 @@
...
+
+	opts := &bind.CallOpts{BlockNumber: num}
+	addr, err := caller.GetActiveContract(opts, name)
+	return addr, err
```
hyunsooda, line 48:
```suggestion
	return caller.GetActiveContract(opts, name)
```

hyunsooda (approved):

blockchain/system/registry.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/klaytn/klaytn/common"
+	contracts "github.com/klaytn/klaytn/contracts/system_contracts"
+)
+
+func InstallRegistry(state *state.StateDB) error {
+	return state.SetCode(RegistryAddr, RegistryCode)
+}
+
+func ReadRegistryActiveAddr(backend bind.ContractCaller, name string, num *big.Int) (common.Address, error) {
+	caller, err := contracts.NewIRegistryCaller(RegistryAddr, backend)
```
hyeonLewis, line 34:
```suggestion
	caller, err := contracts.NewRegistryCaller(RegistryAddr, backend)
```

blockchain/system/registry.go
blukat29: Fixed in ba70e229.

blockchain/system/registry.go
blukat29: Fixed in ba70e229.

hyeonLewis (dismissed):

hyeonLewis (commented):
Is added `AllocRegistry` for directly allocating pre-deployed system contracts to the state when deploying the Registry contract?

hyeonLewis:
> Is added `AllocRegistry` for directly allocating pre-deployed system contracts to the state when deploying the Registry contract?

Confirmed. Let me test this in the hardhat as well. Thanks.

ian0371 (dismissed):
I'm a little concerned about the correctness of `calc~~Slot`, however for `TestAllocRegistry` can cover a portion of cases.
When adding a system contract that uses the functions, a test like `TestAllocRegistry` must be implemented.

blukat29:
> I'm a little concerned about the correctness of `calc~~Slot`, however for `TestAllocRegistry` can cover a portion of cases. When adding a system contract that uses the functions, a test like `TestAllocRegistry` must be implemented.

I agree. We'll need thorough testing regarding `InstallRegistry` after I modify `engine.Finalize()` to actually inject the code.

hyeonLewis (dismissed):
Let's add more tests for `AllocRegistry` after finalizing `Registry` contract and `engine.Finalize()`.

ian0371 (approved):

blockchain/system/registry.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Create storage state from the given initial values.
+// The storage slots are calculated according to the solidity layout rule.
+// https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html
+func AllocRegistry(init *AllocRegistryInit) map[common.Hash]common.Hash {
+	storage := make(map[common.Hash]common.Hash)
+	if init == nil {
+		return storage
+	}
```
hyunsooda, line 47:
```suggestion
        if init == nil {
		return nil
	}
	storage := make(map[common.Hash]common.Hash)
```

hyunsooda (commented):
Rather than relying solely on precalculations, I'm curious about your motivation for engaging with storage at such a granular level. I think smart contract transactions would work too in the hardfork block number. Are upcoming system contracts set to follow this approach, and could it potentially impact scalability negatively?

Learned a storage layout. Thanks.

blukat29:
@hyunsooda 

> I think smart contract transactions would work too in the hardfork block number.

Yes that works, but only if the hardfork number > 0. Block-processing codes will start relying on the Registry from the hardfork number. If the hardfork number is a nonzero number N, then the Registry must exist *before* block N. Therefore the `InstallRegistry` is applied during the finalization of block N-1. However, if the hardfork number = 0, then codes immediately start reading the Registry from block 1, without proper initial data.

To encompass both cases, the initial data must be populated without using transactions. The initial data must be either (1) statically specified as storage slots or (2) hardcoded in the bytecode.

The hardcoding option (2) has a flaw. "Hardcoding" initial global variables is non-trivial because initial assignments happen in the constructor, which is skipped if you inject just bytecode.
```solidity
contract C {
    uint256 num = 1234; // num is still 0 after state.SetCode
    function getNum() external returns(uint256) {
        return num;
    }
}
```
The viable option is to add a fallback logic in the getter.
```solidity
contract C {
    uint256 num;
    function getNum() external returns(uint256) {
        return num > 0 ? num : 1234;
    }
}
```
This additional logic permanently increases the cost of getter calls. This cost is not an issue in the Klaytn node, but it is an issue in other smart contracts that refer to the Registry. It's also clumsy to dynamically create the hardcoded source code

> Are upcoming system contracts set to follow this approach

In Cypress and Baobab, I don't think such an acrobatics is needed anymore. If you see the [KIP-149 draft](https://github.com/hyeonLewis/kips/blob/kip149/KIPs/kip-149.md), a system contract's initial data should be prepared using transactions before relevant hardfork block.

In test chains, yes the static storage slot calculation is needed. For instance, a test chain with [KIP-113](https://github.com/klaytn/kips/tree/main/KIPs/kip-113.md) BLS registry must have the initial data from the genesis. In this case `homi` should create BLS keys and fill in the genesis.json accordingly.

> could it potentially impact scalability negatively?

I'm not sure what kind of scalability you are referring to. the AllocRegistry is a one-time operation and it shouldn't be slow.

blockchain/system/registry.go
blukat29: Fixed

hyunsooda:
@blukat29, Thanks for the kind description.

> Are upcoming system contracts set to follow this approach, and could it potentially impact scalability negatively?

I meant upcoming system contracts follow this kind of pipeline at the time of their deployment. I noticed that this manual setup is required only for a contract that requires state initialization for the time being the smart contract deploys. In that case, most of the upcoming system contracts would not need this kind of work. I misunderstood that this is a new standard preparation for the upcoming system contract, but noticed, it is not actually.

hyeonLewis (approved):

ian0371 (approved):
Please check in the future if 0x401 is ok to use

hyunsooda (approved):

PR #1969, opened 2023-09-15
Add config for Cancun hardfork

2dvorak:
## Proposed changes

- This PR adds config for Cancun hardfork

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/cmd/homi/setup/cmd.go
+++ b/cmd/homi/setup/cmd.go
@@ -125,9 +125,9 @@ var HomiFlags = []cli.Flag{
 	altsrc.NewInt64Flag(magmaCompatibleBlockNumberFlag),
 	altsrc.NewInt64Flag(koreCompatibleBlockNumberFlag),
 	altsrc.NewInt64Flag(shanghaiCompatibleBlockNumberFlag),
+	altsrc.NewInt64Flag(cancunCompatibleBlockNumberFlag),
 	altsrc.NewInt64Flag(kip103CompatibleBlockNumberFlag),
 	altsrc.NewStringFlag(kip103ContractAddressFlag),
-	altsrc.NewInt64Flag(cancunCompatibleBlockNumberFlag),
 }
 
 var SetupCommand = &cli.Command{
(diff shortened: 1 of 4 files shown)


params/config.go
```diff
@@ -49,6 +49,7 @@ var (
 		KoreCompatibleBlock:      big.NewInt(119750400),
 		Kip103CompatibleBlock:    big.NewInt(119750400),
 		ShanghaiCompatibleBlock:  big.NewInt(135456000),
+		CancunCompatibleBlock:    big.NewInt(0xffffffff),
```
yoomee1313, line 52: As far as I know, nil is okay. What about using nil?
```
...(nil = no fork, 0 already on Cancun)
```

cmd/homi/setup/flags.go
```diff
@@ -494,4 +494,10 @@ var (
 		Name:  "kip103-contract-address",
 		Usage: "kip103 contract address",
 	}
+
+	cancunCompatibleBlockNumberFlag = &cli.Int64Flag{
+		Name:  "cancun-compatible-blocknumber",
+		Usage: "cancunCompatible blockNumber",
+		Value: 0,
+	}
```
yoomee1313, line 502: This is my personal question I had before. If the Value of the developing hardfork Flag is zero, then private network created by homi will automatically activate cancunCompatible hardfork. Do you know if it is a intended behavior or not?

2dvorak: I have no idea if it's intended or our devs followed the previous hardforks' conventions, like I did.

IMHO, it would be OK to leave it as it is because builders that are not familiar with Klaytn would have hard time configuring correct hardfork numbers on their own. Our docs and posts are mostly focused on the latest version and it might be confusing for them to find correct hardfork for their need and setup required hardforks (our hardforks have dependencies).

If we want to change that, I could be handled in a separate PR because it would require to change other hardfork values which may not be in the scope of this PR. However, if you have other opinions, I'm happy to discuss it with you in this conversation.

yoomee1313: Oh I mean that cancunCompatible hardfork is still on development, and I'm curious if the hardfork items on development can be activated in dev branch.

2dvorak: Ah, that is another point.

Both options seems OK, if a user is compiling `homi` on the dev branch,
- probably interested in features under development, so enable those features by default
- probably familiar with Klaytn, so make the user configure hardfork number by his/her own

Given that both options are possible, I'd like to leave it as it is because
- we have been doing like that
- we don't have to add another PR for this when all features are developed and the hardfork is ready

params/config.go
2dvorak: Done in 0b8adebe, thanks.

cmd/homi/setup/flags.go
yoomee1313: Great, thank you. It solves the curiosity.

cmd/homi/setup/cmd.go
```diff
@@ -127,6 +127,7 @@ var HomiFlags = []cli.Flag{
 	altsrc.NewInt64Flag(shanghaiCompatibleBlockNumberFlag),
 	altsrc.NewInt64Flag(kip103CompatibleBlockNumberFlag),
 	altsrc.NewStringFlag(kip103ContractAddressFlag),
+	altsrc.NewInt64Flag(cancunCompatibleBlockNumberFlag),
```
blukat29, line 130: It's unrelated to Cancun hardfork, but can you reorder to kip103number - kip103address - shanghai - cancun throughout the codebase?
in cmd.go:128, cmd.go:685, flags.go:481, params/config.go:285, params/config.go:300, params/config.go:353

hyunsooda (approved):

hyeonLewis (approved):

blukat29 (approved):

PR #1983, opened 2023-10-02
Fix WASM build error when importing common package

hackartists:
# JS building failure
Some packages fails when building `js` because no alternatives for `js` with the below command.

```
GOOS=js GOARCH=wasm go build
```

## `common` package
- `IsTty` function is not defined when building WASM with the below options
  - Enable `js` tag to the function using `appengine` tag.

## `crypto` package
- `signature_cgo.go` is excluded in `js` build.
   - `signature_nocgo.go` is ported from `go-ethereum`  repo.

### DIFF
--- a/crypto/signature_nocgo.go
+++ b/crypto/signature_nocgo.go
@@ -1,0 +1,156 @@
+// Modifications Copyright 2018 The klaytn Authors
+// Copyright 2017 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+//
+// This file is derived from crypto/signature_cgo.go (2018/06/04).
+// Modified and improved for the klaytn development.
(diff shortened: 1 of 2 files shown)


crypto/signature_nocgo.go
```diff
@@ -0,0 +1,156 @@
+// Modifications Copyright 2018 The klaytn Authors
```
hyeonLewis, line 1: Maybe we need to update to `2023`?

aidan-kwon (approved):
Long time no see @hackartists :) Thank you for your contribution.
Your request doesn't seem to be harmful to existing code. As you probably know, however, Klaytn has not been tested much in the WASM environment. If you need something more for your project, please share and ask us what you need.

crypto/signature_nocgo.go
hackartists: @hyeonLewis I updated it.

aidan-kwon (approved):

hyeonLewis (approved):

aidan-kwon:
@hackartists I think we need more discussion. I will contact you through another channel soon.

crypto/signature_nocgo.go
```diff
@@ -0,0 +1,156 @@
...
+package crypto
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"errors"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	btc_ecdsa "github.com/btcsuite/btcd/btcec/v2/ecdsa"
```
blukat29, line 33: @hackartists, Please run `go get github.com/btcsuite/btcd/btcec/v2` and push the updated `go.mod` and `go.sum` files.

hyunsooda:
Hi @hackartists,

Instead of modifying `terminal_appengine.go`, it would be more suitable to include the following build condition in `log/test_util.go`.
```
//go:build !js
// +build !js
```

crypto/signature_nocgo.go
aidan-kwon: @blukat29 let's merge this and update packages in the other PR

blukat29 (approved):

PR #1984, opened 2023-10-04
[Governance] Vote reflecting condition adjusted

hyunsooda:
## Proposed changes

Considering the following voting scenario:

0. preparation: chain's epoch = 10
1. Vote at block 33
2. Terminate a node at block 36
3. Restart the node and wait to sync latest block number
4. Failed to verify block 40 that contains non-empty governnace data

The root cause of this failure was the condition that compares if a current block number is greater than `gov.lastGovernanceStateBlock`.  The above scenario has a block that contains non-empty vote data at block number 33. On rebooting the node, the node rewinds the istanbul snapshot through `snap.apply()` function with 33 blocks (1 ~ 33). However, the previously stored `gov.lastGovernanceStateBlock` was 33 and the block number that executes `snap.apply()` is also 33. Then, the condition `blockNum > atomic.LoadUint64(&gov.lastGovernanceStateBlock)` evaluates to false and leading to not calling `gov.ReflectVotes()` which updates the changeset.

This PR fixes the condition `>` to `>=` and resolves the problem. 

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/governance/handler.go
+++ b/governance/handler.go
@@ -492,7 +492,6 @@ func (gov *Governance) addNewVote(valset istanbul.ValidatorSet, votes []Governan
 			(governanceMode == params.GovernanceMode_Ballot && currentVotes > valset.TotalVotingPower()/2) {
 			switch GovernanceKeyMap[gVote.Key] {
 			case params.AddValidator:
-				// reward.GetStakingInfo()
 				if addr, ok := gVote.Value.(common.Address); ok {
 					valset.AddValidator(addr)
 				} else {
--- a/governance/handler.go
+++ b/governance/handler.go
@@ -514,7 +514,7 @@ func (gov *Governance) addNewVote(valset istanbul.ValidatorSet, votes []Governan
 				atomic.StoreUint64(&istanbul.DefaultConfig.Timeout, timeout)
 				fallthrough
 			default:
-				if writable && blockNum > atomic.LoadUint64(&gov.lastGovernanceStateBlock) {
+				if writable && blockNum >= atomic.LoadUint64(&gov.lastGovernanceStateBlock) {
 					logger.Info("Reflecting parameter vote", "num", blockNum, "key", gVote.Key, "value", gVote.Value)
 					gov.ReflectVotes(*gVote)
 				}


ian0371 (commented):
Could you check the GovernanceState DB upon the node restart by adding a log at `ReadGovernanceState()`? IMO the `changeSet` is supposed to be restored correctly.
- process block 33: `WriteGovernanceState(33, true)` must have written `changeSet`. (checkpoint interval is a factor of 33, right?)
- note restart at block 36: `ReadGovernanceState()` should have restored `changeSet`.

In which case `ReflectVotes()` need not be called. Although I'm not sure if my assumption is legit.

hyunsooda:
@ian0371,

`checkpointerInterval` is not a factor of 33. This is a constant value of 1024, which is not configurable. So, `WriteGovernanceState()` is not invoked at the time of validating block number 33 which has non-empty vote data.

Actually, the storing governance state is currently connected with the time of `checkpointerInterval`(Istanbul snapshot checkpointer interval). I noticed it seems not an appropriate condition to write the state to DB because they do not have a correlation.  Even though this is another topic of this PR, it caused an error today.

ian0371:
hmm... then how come `lastGovernanceStateBlock` is 33? It can be only set via either `WriteGovernance()` or `UnmarshalJSON()`. I think we have an inconsistency bug where `lastGovernanceStateBlock` is incorrectly set. It should have been 0 so the branch that this PR addresses must have been executed. Please tell me if I'm wrong. Thanks.

hyunsooda:
[This line](https://github.com/klaytn/klaytn/blob/dev/governance/handler.go#L393) caused this error. `RemoveVote()` calls `WriteGovernanceState()`. This call is not guarded with the condition that `snapshot()` does. Thus, the time of writing data to DB were diferernt between two calls. 

The code following fixed this discrepency and resolved the issue without the change of this PR.
```
if number%1024 == 0 && g.CanWriteGovernanceState(number) { // currently, checkpoint is private variable in the pacakge of `backend`
    g.WriteGovernanceState(number, false)
}
```

Please review the following list:
1. Is the initial change in this PR still valid, and should we keep or revert it?
2. We should focus on the crucial patch to align the discrepancy. This change should be made after the merge of [PR #1966](https://github.com/klaytn/klaytn/pull/1966), which refactors the `checkpointinterval` constant for better importability across packages.
3. Introduce a new interval number for the governance state store to eliminate dependencies on other contexts, such as the Istanbul snapshot. Please refer to this for more details:

> Actually, the governance state store is currently connected with the time of checkpointerInterval(Istanbul snapshot checkpointer interval). I noticed it seems not an appropriate condition to write the state to DB because they do not have a correlation. Even though this is another topic of this PR, it caused an error today.

ian0371:
Now I get the big picture:

Environment
- checkpoint blocks `c1`, `c2`
- node `A`'s VoteData block `V`
- node `A`'s reboot block `R`
- epoch block `e1`, `e2`
- When `c1 < e1 < V < R < c2, e2`, node `A` reports bad block at `e2`; other nodes are fine

Bug
- Node `A` receives block `V`, then writes `changeSet` at `V` to GovState DB; [link](https://github.com/klaytn/klaytn/blob/cd09595fe98c768526089d080731e40787b15dcb/governance/default.go#L537). This only happens to node `A`.
- Upon reboot, `changeSet` is correctly restored; `lastGovernanceStateBlock` is `V`; [link](https://github.com/klaytn/klaytn/blob/cd09595fe98c768526089d080731e40787b15dcb/governance/default.go#L444)
- `snapshot.apply()` is called with the range `[c1, R]`
	- Applying `e1`: `changeSet` is cleared via `ClearVotes(e1)`; [link](https://github.com/klaytn/klaytn/blob/cd09595fe98c768526089d080731e40787b15dcb/consensus/istanbul/backend/snapshot.go#L180)
	- Applying `V`: `HandleGovernanceVote()` ignores `addNewVote` because the branch is false (`blockNum = lastGovernanceStateBlock = V`); [link](https://github.com/klaytn/klaytn/blob/cd09595fe98c768526089d080731e40787b15dcb/governance/handler.go#L517)
- Bad block occurs at `e2` because node `A` receives `GovernanceData` when `changeSet` is empty

Given this situation, we are left with two options:
- Accept this PR
- Remove `WriteGovernanceState()` from `RemoveVote()`

This PR fixes the condition so `changeSet` when applying `V` is restored.
However, it seems that if there are two votes, the prior vote will be ignored even after this PR resulting in a bad block (even though highly unlikely in practice).

The fundamental approach could be to remove `WriteGovernanceState()` from `RemoveVote()`, so that GovState DB is always updated at the checkpoint.
Even though they should be ideally decoupled, the current implementation seems to strongly couple the checkpoint interval with GovState write interval.

I think we should choose the one that has less side effect, which makes this PR more appealing. I'd like to hear what you think, as well as others' opinions. Thanks.

hyunsooda:
Thanks for the detail.

Given your two options, I agree that the second one is more appropriate. Also, the first option can be considered together as it is still logically valid. So, both options can be applied together in my view.

What do you think about reducing the interval value to 1? It doesn't seem to make overhead against write disk latency too much.

ian0371:
> the first option can be considered together as it is still logically valid

Technically, the previous logic seems correct as the governance state at `lastGovernanceStateBlock` has been reflected. However, it seems `ReflectVotes()` is idempotent, so I think this PR is ok. The second option might have a potential side effect, so I don't think applying both options is a good idea.

> What do you think about reducing the interval value to 1?

Snapshot DB will take up x1024 space, which I'm not sure is ok.

ian0371 (approved):

yoomee1313 (approved):

hyunsooda:
@ian0371, Thanks got it.

PR #1989, opened 2023-10-10
[Log] Trie memory logging refined

hyunsooda:
## Proposed changes

1. Added trie memory size configuration log at the time of node start
2. Appended retention value to the GC log.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -262,6 +262,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 		stopStateMigration: make(chan struct{}),
 		prefetchTxCh:       make(chan prefetchTx, MaxPrefetchTxs),
 	}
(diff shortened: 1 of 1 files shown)


blukat29:
Printing the same value all the time unnecessarily increases the log size.

Why do you need to print config variables when they are available in `kcnd.conf`? What is your use case?

hyunsooda:
The initial configuration is sourced from `k*nd.conf`, but the runtime value may be uncertain due to frequent node restarts and configuration resets. Occasionally, operators may need to retrieve the dynamically set value. There are two options for this: using a getter API or logging. However, in the case of logging, some logs may be trimmed by system management like `logrotate`, causing the loss of initial configuration values. So, added the initial configuration to that logs.

The logging approach was my bad now. I believe using an API getter (possibly a setter) would be a more suitable approach than the current method. Do you think that this complement does not need to be provided?

hyunsooda:
Continued at #1996.

PR #1987, opened 2023-10-05
Install Registry before Cancun fork block

blukat29:
## Proposed changes

- Install Registry.sol code and initial state at CancunForkBlockParent (`num = cancun - 1`).
- TODO: determine how to pass initial owner account.

## Types of changes

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

- [ ] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

PR #1990, opened 2023-10-10
KIP-113 BLS PublicKey Registry accessors

blukat29:
## Proposed changes

- Add KIP113 contract mockup (see [KIP-113](https://github.com/klaytn/kips/blob/main/KIPs/kip-113.md))
- Actual KIP113.sol won't be added because the concrete implementation is out of the Klaytn node's scope.
- Add accessors to the KIP113 system contract.
  - `ReadKip113All()` calls the `getAllInfo()` function, and filters out records with invalid public key or pop values.
  - `AllocKip113()` calculates the storage layout of KIP113Mock contract, for testing.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Further comments

Since #1987 conflicts with `contracts/system_contracts/all.go`, This PR will be merged after that.

### DIFF
--- a/blockchain/system/kip113.go
+++ b/blockchain/system/kip113.go
@@ -1,0 +1,83 @@
+// Copyright 2023 The klaytn Authors
+// This file is part of the klaytn library.
+//
+// The klaytn library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The klaytn library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+package system
+
+import (
+	"math/big"
+
+	"github.com/klaytn/klaytn/accounts/abi/bind"
+	"github.com/klaytn/klaytn/common"
+	contracts "github.com/klaytn/klaytn/contracts/system_contracts"
+	"github.com/klaytn/klaytn/crypto/bls"
+)
+
+type BlsPublicKeyInfo struct {
+	PublicKey []byte
+	Pop       []byte
+}
+
+type BlsPublicKeyInfos map[common.Address]BlsPublicKeyInfo
+
+func (info BlsPublicKeyInfo) Verify() error {
+	pk, err := bls.PublicKeyFromBytes(info.PublicKey)
+	if err != nil {
+		return err
+	}
+
+	sig, err := bls.SignatureFromBytes(info.Pop)
+	if err != nil {
+		return err
+	}
+
+	if !bls.PopVerify(pk, sig) {
+		return ErrKip113BadPop
+	}
+
+	return nil
+}
+
+func ReadKip113All(backend bind.ContractCaller, contractAddr common.Address, num *big.Int) (BlsPublicKeyInfos, error) {
+	caller, err := contracts.NewIKIP113Caller(contractAddr, backend)
+	if err != nil {
+		return nil, err
+	}
+
+	opts := &bind.CallOpts{BlockNumber: num}
+	ret, err := caller.GetAllInfo(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(ret.AddrList) != len(ret.PubkeyList) {
+		return nil, ErrKip113BadResult
+	}
+
+	infos := make(BlsPublicKeyInfos)
+	for i := 0; i < len(ret.AddrList); i++ {
+		addr := ret.AddrList[i]
+		info := BlsPublicKeyInfo{
+			PublicKey: ret.PubkeyList[i].PublicKey,
+			Pop:       ret.PubkeyList[i].Pop,
+		}
+
+		if err := info.Verify(); err == nil {
+			infos[addr] = info
+		}
+	}
+
+	return infos, err
+}
--- a/blockchain/system/kip113.go
+++ b/blockchain/system/kip113.go
@@ -30,7 +30,69 @@ type BlsPublicKeyInfo struct {
 	Pop       []byte
 }
 
+type AllocKIP113Init struct {
+	// Owner common.Address
+	Infos map[common.Address]BlsPublicKeyInfo
+	Addrs []common.Address
+}
+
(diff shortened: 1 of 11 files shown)


blukat29:
Rebased to latest dev.

blukat29:
I have read the CLA Document and I hereby sign the CLA

hyeonLewis:
I have read the CLA Document and I hereby sign the CLA

blockchain/system/kip113.go
```diff
@@ -0,0 +1,147 @@
...
+	PublicKey []byte
+	Pop       []byte
+}
+
+type (
+	BlsPublicKeyInfos map[common.Address]BlsPublicKeyInfo
+	AllocKip113Init   = BlsPublicKeyInfos
+)
+
+func AllocKip113(init AllocKip113Init) map[common.Hash]common.Hash {
```
hyunsooda, line 40: For the dynamic type (`bytes` and `string`), refactoring may be considered so that the generic calcuator address all of inputs  such as

```go
func calcDynTyp(baseSlot interface{}, b []byte) (common.Hash, map[common.Hash]common.Hash) {
    var (
        WORD_SIZE = 32
        ret       = make(map[common.Hash]common.Hash)
        l         = len(b)
        dynLen    = lpad32(WORD_SIZE)
    )
    if l >= WORD_SIZE {
        dynLen = lpad32(l*2 + 1)
        done := 0

        for i := WORD_SIZE; i <= l; i += WORD_SIZE {
            key := calcStructSlot(baseSlot, i/WORD_SIZE-1)
            val := common.BytesToHash(b[i/WORD_SIZE-1 : i])
            ret[key] = val
            done += WORD_SIZE
        }
        if l-done > 0 {
            padded := append(b[done:], make([]byte, l-done)...)
            key := calcStructSlot(baseSlot, len(ret))
            val := common.BytesToHash(padded)
            ret[key] = val
        }
    } else {
        var val common.Hash
        copy(val[:], b)

        val[WORD_SIZE-1] = byte(l * 2)
        ret[baseSlot.(common.Hash)] = val
    }
    return dynLen, ret
}
```

With this, the inner loop can be shortened with consistent format
```go
dynLen, ret := calcDynTyp(pupKeySlot, info.PublicKey)
storage[pupKeySlot] = dynLen
for k, v := range ret {
    storage[k] = v
}

dynLen, ret = calcDynTyp(popKeySlot, info.Pop)
storage[popKeySlot] = dynLen
for k, v := range ret {
    storage[k] = v
}
```

blockchain/system/kip113.go
```diff
@@ -0,0 +1,147 @@
...
+	// - infos[x].pop.length    @ Hash(x, 0) + 1
+	//   - infos[x].pop[:32]    @ Hash(Hash(x, 0) + 1) + 0
+	//   - infos[x].pop[32:64]  @ Hash(Hash(x, 0) + 1) + 1
+	//   - infos[x].pop[64:]    @ Hash(Hash(x, 0) + 1) + 2
+	for addr, info := range init {
+		// The below slot calculation assumes 48-byte and 96-byte Solidity `bytes` values.
+		if len(info.PublicKey) != 48 || len(info.Pop) != 96 {
+			logger.Crit("Invalid AllocKip113Init")
+		}
+		pupKeySlot := calcMappingSlot(0, addr)
```
hyunsooda, line 59: typo

blockchain/system/kip113.go
```diff
@@ -0,0 +1,147 @@
...
+
+	infos := make(BlsPublicKeyInfos)
+	for i := 0; i < len(ret.NodeIdList); i++ {
+		addr := ret.NodeIdList[i]
+		info := BlsPublicKeyInfo{
+			PublicKey: ret.PubkeyList[i].PublicKey,
+			Pop:       ret.PubkeyList[i].Pop,
+		}
+
+		if err := info.Verify(); err == nil {
```
hyunsooda, line 141:
```suggestion
		if info.Verify() == nil {
```

blockchain/system/kip113.go
blukat29: @hyunsooda Nice. I will try.

blukat29: @hyunsooda Implemented 6830206 PTAL.

blockchain/system/kip113.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	if !bls.PopVerify(pk, sig) {
+		return ErrKip113BadPop
+	}
+
+	return nil
+}
+
+func ReadKip113All(backend bind.ContractCaller, contractAddr common.Address, num *big.Int) (BlsPublicKeyInfos, error) {
```
hyeonLewis, line 109: How about adding a function `ReadKip113(... nodeId common.Address)`, which returns `BlsPublicKeyInfo` only for target `nodeId`? It would be helpful for future use.

blukat29: But the IKIP113 only has the function to query all.
The decision was made in this order: (1) Let's make SBR forward compatible with CandidateList, so we can keep using ReadKip113* later (2) CandidateList has no getters using nodeId. Its getter uses candidateIdx. (3) Let's not use nodeId in the KIP-113 interface altogether, hence https://github.com/klaytn/kips/pull/151.

Maybe we could cache the full list and return them one by one, but caching on-chain data may complicate some APIs such as setHead, and traceBlock. Because ReadKip113All is expected to be called a few times per block, I decided not to provide such a feature.

hyeonLewis: Ok, I got the point. Thanks.

blockchain/system/registry.go
```diff
@@ -65,10 +65,12 @@ func AllocRegistry(init *params.RegistryConfig) map[common.Hash]common.Hash {
 	// slot[1]: string[] names;
 	// - names.length @ 1
 	// - names[i] @ Hash(1) + i
-	storage[lpad32(1)] = lpad32(len(names)) // names.length
+	storage[lpad32(1)] = lpad32(len(names))
 	for i, name := range names {
-		nameSlot := calcArraySlot(1, 1, i, 0)
-		storage[nameSlot] = encodeShortString(name) // names[i]
+		nameSlot := calcArraySlot(1, 1, i, 0) // Hash(1) + 1*i + 0
+		for k, v := range allocDynamicData(nameSlot, []byte(name)) {
```
ian0371, line 71: this looks like a loop, but `allocDynamicData` always returns a map of size 1.
How about changing `allocDynamicData()` to return `(Hash slot, Hash val)`

ian0371 (approved):

hyunsooda (approved):

hyeonLewis (approved):

blukat29:
@yoomee1313 PTAL

yoomee1313 (approved):

PR #2013, opened 2023-10-30
Setting reliable archive mode EN

sjnam:
## Proposed changes

- Full EN can only store the latest block states (e.g. latest-128 block ~ latest block)
- Querying latest block state of archive EN is not efficient compared to query latest block state of full EN
- You can add the command line flag of an EN `--upstream-en archive-en-rpc-url` to test. For example, `--upstream-en http://localhost:8557`

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Further comments


networks/rpc/handler.go
```diff
@@ -427,6 +428,23 @@ func (h *handler) handleSubscribe(cp *callProc, msg *jsonrpcMessage) *jsonrpcMes
 func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *callback, args []reflect.Value) *jsonrpcMessage {
 	result, err := callb.call(ctx, msg.Method, args)
 	if err != nil {
+		if _, ok := err.(*statedb.MissingNodeError); ok {
```
blukat29, line 431:
```go
if UpstreamArchiveEN != "" && shouldRequestUpstream(result, err) {
    return requestUpstream(ctx, msg, args)
}
```

```go
func shouldRequestUpstream(result, err) bool {
    switch err.(type) {
    case *statedb.MissingNodeError:
        return true
    default:
        return false
    }
}

func requestUpstream(ctx, msg, args) *jsonrpcMessage {
}
```

sjnam: @blukat29 
`UpstreamArchiveEN != ""` is checked at `url.Parse(rawurl)` of `DialContext` method.
https://github.com/sjnam/klaytn/blob/reliable-archive-en/networks/rpc/client.go#L176-L180
But, it's good idea to explicitly check for `UpstreamArchiveEN != ""`.

blukat29: @sjnam I see. But if we dial when `UpstreamArchiveEN == ""`, then (depending on the implementation) the outer RPC returns "URL parse error", not "missing node error". So I think the explicit check is necessary.

blukat29 (approved):

networks/rpc/handler.go
```diff
@@ -427,6 +428,9 @@ func (h *handler) handleSubscribe(cp *callProc, msg *jsonrpcMessage) *jsonrpcMes
 func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *callback, args []reflect.Value) *jsonrpcMessage {
 	result, err := callb.call(ctx, msg.Method, args)
 	if err != nil {
+		if UpstreamArchiveEN != "" && shouldRequestUpstream(err) {
```
hyunsooda, line 431: LGTM. Regarding this feature, let me ask three usage questions.

1. The single URL may be extended to the list of URL. Was there an issue regarding this?
2. The URL (list or single) seems not modifiable at runtime. If the initially set node becomes unreachable, the routing node should be restarted to change that value. It seems to degrade the usage.
3. Any request can be relayed as well as the state lack. Is this an obvious limitation by the consideration of big load to the upstream node?

hyunsooda: Got it thanks.

hyunsooda (dismissed):

networks/rpc/handler.go
yoomee1313: @sjnam can you add todo about those suggestions? Also, not mentioned here, additional suggestion is to make a new rpcErrorResponse struct for it.

blukat29 (approved):

blukat29:
@sjnam Please rebase against latest dev.

yoomee1313 (approved):

Issue #2023, opened 2023-11-07
Quorum Calculation Issue


jiseongnoh:
**Describe the bug**
The current quorum calculation method is 2F+1 where F is the maximum number of byzantine nodes. However, this formula does not hold for the network with 4x+1 validators, which might cause chain splits due to the incorrect quorum size. The correct formula should be Ceil(2 * valSet.Size())/3, to comply with the BFT principles.

**How to reproduce**
please refer to https://github.com/klaytn/klaytn/blob/dev/consensus/istanbul/core/handler_test.go#L619

**Expected behavior**

**Attachments**

**Environment (please complete the following information)**

**Additional context**
A hard fork is required to implement the correct quorum calculation.

PR #2026, opened 2023-11-07
Fix Quorum

jiseongnoh:
## Proposed changes
This PR addresses a consensus instability issue in networks with 4x+1 validators. It updates the quorum calculation from the previous 2F+1 to math.Ceil(float64(2*valSet.Size())/3), aligning with QBFT requirements for enhanced network stability and Byzantine fault tolerance.

Closes https://github.com/klaytn/klaytn/issues/2023.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -357,7 +357,7 @@ func (sb *backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
 	}
 
 	// The length of validSeal should be larger than number of faulty node + 1
-	if validSeal < istanbulCore.RequiredMessageCount(snap.ValSet, chain.Config().IsCancunForkEnabled(header.Number)) {
+	if validSeal <= 2*snap.ValSet.F() {
 		return errInvalidCommittedSeals
 	}
 
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -357,7 +357,7 @@ func (sb *backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
 	}
 
 	// The length of validSeal should be larger than number of faulty node + 1
-	if validSeal <= 2*snap.ValSet.F() {
+	if validSeal < istanbulCore.RequiredMessageCount(snap.ValSet, chain.Config().IsCancunForkEnabled(header.Number)) {
 		return errInvalidCommittedSeals
 	}
 
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -436,7 +436,14 @@ func PrepareCommittedSeal(hash common.Hash) []byte {
 
 // Minimum required number of consensus messages to proceed
 func RequiredMessageCount(valSet istanbul.ValidatorSet, isCancunHardforkEnabled bool) int {
+	// Quorum size adjustment for 4x+1 validator counts to maintain BFT consensus stability
 	if isCancunHardforkEnabled {
+		// For less than 4 validators, quorum size equals validator count.
+		if valSet.Size() < 4 {
+			return int(valSet.Size())
+		}
+		// Adopted QBFT quorum implementation
+		// https://github.com/Consensys/quorum/blob/master/consensus/istanbul/qbft/core/core.go#L312
 		return int(math.Ceil(float64(2*valSet.Size()) / 3))
 	}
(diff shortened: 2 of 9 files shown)


consensus/istanbul/backend/engine.go
```diff
@@ -357,7 +357,7 @@ func (sb *backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
 	}
 
 	// The length of validSeal should be larger than number of faulty node + 1
-	if validSeal <= 2*snap.ValSet.F() {
+	if validSeal < istanbulCore.RequiredMessageCount(snap.ValSet, chain.Config().IsCancunForkEnabled(header.Number)) {
```
hyeonLewis, line 360: I think `2F()` can't be directly changed to `RequiredMessageCount` since it differs for small validators' cases. Is it okay?

hyeonLewis: Oh, it was updated. But I think we still need to consider `n=6` case.

consensus/istanbul/core/core.go
```diff
@@ -428,7 +435,10 @@ func PrepareCommittedSeal(hash common.Hash) []byte {
 }
 
 // Minimum required number of consensus messages to proceed
-func requiredMessageCount(valSet istanbul.ValidatorSet) int {
+func RequiredMessageCount(valSet istanbul.ValidatorSet, isAbcEnabled bool) int {
+	if isAbcEnabled {
+		return int(math.Ceil(float64(2*valSet.Size()) / 3))
```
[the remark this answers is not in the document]
jiseongnoh: Thank you. reflected at e22d882c7dbfb34eb13c8f15e55aed181f38d0f9

consensus/istanbul/backend/engine.go
jiseongnoh: Let's try a non hard-fork-required one

blukat29 (approved):

hyeonLewis (approved):

consensus/istanbul/core/core.go
```diff
@@ -428,20 +428,18 @@ func PrepareCommittedSeal(hash common.Hash) []byte {
 }
 
 // Minimum required number of consensus messages to proceed
-func requiredMessageCount(valSet istanbul.ValidatorSet) int {
+func RequiredMessageCount(valSet istanbul.ValidatorSet) int {
```
hyunsooda, line 431: Is this counting change safe without HF? 
Assuming there are a total of `2F+1` nodes operational within the network, specifically denoting that `5` nodes are active among a total of `8` nodes.

In this scenario, where half of the nodes are responsible for calculating the necessary count of messages using the formula `2*valSet.F() + 1`, thereby anticipating `6` messages, while the other half calculates it using `int(math.Ceil(float64(2*size) / 3))`, expecting `5` messages.

Up to this point, the chain has been able to progress continuously. However, there emerges an issue wherein half of the updated nodes may not facilitate the forwarding of the chain. Does this not pose a potential risk or detriment associated with this change?

jiseongnoh: The update to the quorum calculation without a hard fork is designed to be low-risk.

1. Cypress Network Stability: In the case of the Cypress mainnet, there are 31 committee nodes in operation, which aligns with the 3F+1 configuration. The quorum count will be consistent with both 2F+1 or Ceil(2N/3) formula. Therefore, the quorum calculation change will not pose a risk.
2. ServiceChain Network Resilience: ServiceChains with different numbers of validators might potentially experience a problem under extreme conditions. However, the problem will be solved as long as most nodes update the binary. This is acceptable considering the complexities of a hard fork.

In short, this method involves a minor relaxation of the BFT principle, a trade-off between an uninterrupted network and a quick update process.

jiseongnoh: It is also noteworthy that under normal operating conditions, the number of validators participating in consensus typically surpass the quorum required by both the 2F+1 and Ceil(2*N/3) formulas. This surplus of committed seals ensures that discrepancies in different quorum calculation will not compromise the consensus process.

ian0371 (approved):

consensus/istanbul/core/core.go
hyunsooda: Got it thanks.

hyunsooda (approved):

PR #2059, opened 2023-12-06
Install Registry at Randao hardfork

ian0371:
## Proposed changes

To facilitate the activation of Randao hardfork and Registry installation at the same block, this PR installs the Registry contact at `randao hardfork` (previously, at `randao hardfork - 1`).
It necessitates reading KIP113 address from ChainConfig for the verification of the Randao hardfork block because during the verification, the Registry does not exist at the latest block.

In short, previously,
```
GetBlsPubkey(num, ...):
	fetch kip113Addr from registry at state{num}

VerifyRandao(header, ...):
	GetBlsPubkey(header.Number - 1)
```

In this PR,
```
GetBlsPubkey(num, ...):
	if num == hf: fetch kip113Addr from ChainConfig
	if num > hf: fetch kip113Addr from registry at state{num-1}
 
VerifyRandao(header, ...):
	GetBlsPubkey(header.Number)
```

Note that the actual change is in `getAllCached`, not `GetBlsPubkey`.

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

## Further comments

Tested that hardfork block passes in the expected scenario and unexpected scenario.
In expected scenario, `randaoRegistry` field in ChainConfig is set before the hardfork pass.
In unexpected scenario, `randaoRegistry` field in ChainConfig is not set at the hardfork, so the operator updates the `randaoCompatibleBlock` to postpone the hardfork.

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -546,7 +546,7 @@ func (sb *backend) Finalize(chain consensus.ChainReader, header *types.Header, s
 
 	// The Registry contract must be immediately available from the fork block.
 	// So it is installed at block (RandaoCompatibleBlock - 1) which is before the fork block.
-	if chain.Config().IsRandaoForkBlockParent(header.Number) {
+	if chain.Config().IsRandaoForkBlock(header.Number) {
 		system.InstallRegistry(state, chain.Config().RandaoRegistry)
 	}
 
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -547,7 +547,10 @@ func (sb *backend) Finalize(chain consensus.ChainReader, header *types.Header, s
 	// The Registry contract must be immediately available from the fork block.
 	// So it is installed at block (RandaoCompatibleBlock - 1) which is before the fork block.
 	if chain.Config().IsRandaoForkBlock(header.Number) {
-		system.InstallRegistry(state, chain.Config().RandaoRegistry)
+		err := system.InstallRegistry(state, chain.Config().RandaoRegistry)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	header.Root = state.IntermediateRoot(true)
--- a/consensus/istanbul/backend/randao.go
+++ b/consensus/istanbul/backend/randao.go
@@ -61,9 +61,15 @@ func (p *ChainBlsPubkeyProvider) getAllCached(chain consensus.ChainReader, num *
 	}
 
 	backend := backends.NewBlockchainContractBackend(chain, nil, nil)
+	if common.Big0.Cmp(num) == 0 {
+		return nil, errors.New("num cannot be zero")
+	}
 	parentNum := new(big.Int).Sub(num, common.Big1)
 
 	var kip113Addr common.Address
+	// Because the system contract Registry is installed at Finalize() of RandaoForkBlock,
+	// it is not possible to read KIP113 address from the Registry at RandaoForkBlock.
+	// Hence the ChainConfig fallback.
 	if chain.Config().IsRandaoForkBlock(num) {
 		var ok bool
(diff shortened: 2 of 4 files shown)


consensus/istanbul/backend/randao.go
```diff
@@ -58,12 +61,26 @@ func (p *ChainBlsPubkeyProvider) getAllCached(chain consensus.ChainReader, num *
 	}
 
 	backend := backends.NewBlockchainContractBackend(chain, nil, nil)
-	kip113Addr, err := system.ReadRegistryActiveAddr(backend, system.Kip113Name, num)
-	if err != nil {
-		return nil, err
+	parentNum := new(big.Int).Sub(num, common.Big1)
```
blukat29, line 64: Check for num == zero please

consensus/istanbul/backend/engine.go
```diff
@@ -546,7 +546,7 @@ func (sb *backend) Finalize(chain consensus.ChainReader, header *types.Header, s
 
 	// The Registry contract must be immediately available from the fork block.
 	// So it is installed at block (RandaoCompatibleBlock - 1) which is before the fork block.
-	if chain.Config().IsRandaoForkBlockParent(header.Number) {
+	if chain.Config().IsRandaoForkBlock(header.Number) {
 		system.InstallRegistry(state, chain.Config().RandaoRegistry)
```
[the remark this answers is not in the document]
ian0371: fixed in c35af5d5

consensus/istanbul/backend/randao.go
```diff
@@ -58,12 +61,26 @@ func (p *ChainBlsPubkeyProvider) getAllCached(chain consensus.ChainReader, num *
...
+
+	var kip113Addr common.Address
+	if chain.Config().IsRandaoForkBlock(num) {
```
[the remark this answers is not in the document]
ian0371: fixed in 5bc42b6e

consensus/istanbul/backend/randao.go
ian0371: fixed in 5bc42b6e

blukat29 (approved):

consensus/istanbul/backend/randao.go
```diff
@@ -58,12 +61,32 @@ func (p *ChainBlsPubkeyProvider) getAllCached(chain consensus.ChainReader, num *
...
+			return nil, errors.New("KIP113 address not set in ChainConfig")
+		}
+	} else if chain.Config().IsRandaoForkEnabled(num) {
+		var err error
+		kip113Addr, err = system.ReadRegistryActiveAddr(backend, system.Kip113Name, parentNum)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		return nil, errors.New("Cannot read KIP113 address from registry before Randao fork")
```
aidan-kwon, line 86: `ChainBlsPubkeyProvider` should be used only after Randao HF, Randao-disabled condition is not needed 

Prerequisites of Randao Hardfork 
- BLS contract (KIP113) should be deployed
- Validators information should be registered on the BLS contract
- Randao registry should be specified with the KIP113 contract address 

```suggestion
	} else {
		var err error
		kip113Addr, err = system.ReadRegistryActiveAddr(backend, system.Kip113Name, parentNum)
		if err != nil {
			return nil, err
		}
```

aidan-kwon (approved):
There is an minor comment, but it is good to go

consensus/istanbul/backend/randao.go
ian0371: You're right. I wasn't sure what would happen when this function is invoked before Randao.

aidan-kwon: If the function is called before Randao HF, it will fail at `system.ReadRegistryActiveAddr()` because the registry is registered at Randao HF. It is not an important issue, though.

PR #2064, opened 2023-12-12
[API] Snapshot delete operation flag added

hyunsooda:
## Proposed changes

This update introduces a dedicated flag, allowing the rewind of a chain specifically triggered by an explicit API call. In the current setHead functionality, deleting the snapshot database around one epoch could be time-consuming (5-10 minutes per epoch on the Cypress network with a 604800 epoch value). The new flag addresses this by selectively deleting the snapshot database only when the call path originates from an explicit API call, not degrading the user experience and potential operator frustration during instantiation.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -301,7 +301,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 		if diskRoot != (common.Hash{}) {
 			logger.Warn("Head state missing, repairing", "number", head.Number(), "hash", head.Hash(), "snaproot", diskRoot)
 
-			snapDisk, err := bc.setHeadBeyondRoot(head.NumberU64(), diskRoot, true)
+			snapDisk, err := bc.setHeadBeyondRoot(head.NumberU64(), diskRoot, true, false)
 			if err != nil {
 				return nil, err
 			}
@@ -314,7 +314,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 			// Dangling block without a state associated, init from scratch
 			logger.Warn("Head state missing, repairing chain",
 				"number", head.NumberU64(), "hash", head.Hash().String())
-			if _, err := bc.setHeadBeyondRoot(head.NumberU64(), common.Hash{}, true); err != nil {
+			if _, err := bc.setHeadBeyondRoot(head.NumberU64(), common.Hash{}, true, false); err != nil {
 				return nil, err
 			}
 		}
@@ -327,7 +327,7 @@ func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig 
 			// make sure the headerByNumber (if present) is in our current canonical chain
 			if headerByNumber != nil && headerByNumber.Hash() == header.Hash() {
 				logger.Error("Found bad hash, rewinding chain", "number", header.Number, "hash", header.ParentHash)
-				bc.SetHead(header.Number.Uint64() - 1)
+				bc.SetHead(header.Number.Uint64()-1, false)
 				logger.Error("Chain rewind was successful, resuming normal operation")
 			}
 		}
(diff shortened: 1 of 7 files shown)


blockchain/blockchain.go
```diff
@@ -645,25 +645,28 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
...
-		votingEpoch = 1
-	}
-	// Delete the snapshot state beyond the block number of the previous epoch on the right
-	for i := curBlkNum; i >= votingEpoch; i-- {
-		if params.IsCheckpointInterval(i) {
-			// delete from sethead number to previous two epoch block nums
-			// to handle a block that contains non-empty vote data to make sure
-			// the `HandleGovernanceVote()` cannot be skipped
-			bc.db.DeleteIstanbulSnapshot(bc.GetBlockByNumber(i).Hash())
+	if deleteSnpashot {
```
ian0371, line 648: How about putting this logic in the "else" branch below? Would it have the same effect?

https://github.com/klaytn/klaytn/blob/b62c83401a40934fade47f406993e23b792bb778/blockchain/blockchain.go#L635-L646

hyunsooda: This flag was to distinigush the callsite between them. If it's absent, every callsite of `SetHead` would delete snapshot data for further epoch. To remove this bad UX, I think the API call path only should trigger it.

ian0371: Initially, I thought `deleteSnapshot` is always the inverse of `repair`, because
- `SetHead` from API: `repair=false`, `deleteSnapshot=true`,
- otherwise: `repair=true`, `deleteSnapshot=false`.

In my understanding, `repair` flag indicates whether or not to delete block bodies and receipts. So, when we're deleting them, I thought snapshot may be deleted as well.

I didn't notice at the time of writing the first comment, but I've found some counterexamples where block bodies are deleted but snapshot is not (i.e., `repair=false` and `deleteSnapshot=false`):
- https://github.com/hyunsooda/klaytn/blob/sethead-snapshot-delete-flag/blockchain/blockchain.go#L330
- https://github.com/hyunsooda/klaytn/blob/sethead-snapshot-delete-flag/blockchain/blockchain.go#L790

I just wanted to make sure if such behavior is okay. Please correct me if I'm wrong. Thanks

hyunsooda: Indeed, the desired outcome of the snapshot deletion was initiated solely through the API call. As previously highlighted, this removes the time wasted on regenerating deleted snapshot data. Given that header-based voting is infrequent and `sethead` calls are frequent, this implementation avoids boring delays. Should the necessity for a snapshot deletion arise, the operator can promptly initiate the process by making a simple API call.

ian0371: I wanted to ask if we can reuse `repair` flag to achieve the goal, instead of introducing new `deleteSnapshot` flag. It seems to me the both flags serve the same purpose.

hyunsooda: No, it seems not identical. All three calls unconditionally remove snpashot because of `repair=false` in your suggestion.
- https://github.com/hyunsooda/klaytn/blob/sethead-snapshot-delete-flag/blockchain/blockchain.go#L330
- https://github.com/hyunsooda/klaytn/blob/sethead-snapshot-delete-flag/blockchain/blockchain.go#L790
- https://github.com/klaytn/klaytn/pull/2064/files#diff-de2bd0674bf71cd6e3d92cb5621577831308e4cc990a15933e5e853f3358e4e0R323

ian0371: Aren't they rare cases? It seems that first is when bad hash exists, second is when reset to genesis, and third is when genesis config is not compatible. In such error cases, I wanted to ask which is ok: to leave or not to leave the snapshots.

ian0371: Because block bodies are removed in such error cases, I thought it may be ok to delete snapshots as well. I may be wrong so I'm asking the question :)

hyunsooda: I realize your suggestion appears identical, so deleting everything in such cases seems reasonable. I also need to correct my comment: `sethead` is rare, but `repairing` is frequent (`repair=true`). The snapshot removal handler was triggered unconditionally, and I've addressed it with your suggestion. PTAL. Thank you.

ian0371 (approved):

hyeonLewis (approved):

PR #2074, opened 2023-12-19
[Snapshot] Snapshot recovery handler added

hyunsooda:
## Proposed changes

If the required snapshot data is absent in the cache or database due to a `sethead` operation or manual deletion, snapshot data may not be accessible. This PR introduces a snapshot recovery handler, which stores all snapshot data within the specified header range. The handler is activated under two conditions: (1) when the node is first loaded and (2) during API serving. If no corresponding snapshot is found in both cases, the handler is triggered.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,52 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
 
+	sb.regen(chain, headers)
+
 	sb.recents.Add(snap.Hash, snap)
 	return snap, err
 }
 
+// regen commits snapshot data to database
+func (sb *backend) regen(chain consensus.ChainReader, headers []*types.Header) {
+	if !sb.isRestoring.Load() && len(headers) > 1 {
```
hyeonLewis, line 932: The loop condition check might be helpful in reducing a useless loop.
```
last := headers[len(headers)-1].Number
if last - (last % CheckpointInterval) >= headers[0].Number {
    // Start loop
}
```

hyunsooda: Good suggestion. Let me gather further opinion for this. What do you @ian0371  think about this shortcut?

ian0371: LGTM. Could you add the condition to the caller side?

```
// if checkpoints are in `headers`, store their snapshots
if last - (last % CheckpointInterval) >= headers[0].Number {
    sb.regen(chain, headers)
}
```

hyunsooda: How about to add into inner function? If not, the caller seems to get dirty like
```
if len(headers) > 1 {  
    from := headers[0].Number
    to := headers[len(headers)-1].Number

    if to.Uint64()-(to.Uint64()%uint64(params.CheckpointInterval)) >= from.Uint64() {
        sb.regen(chain, headers)
    }
}
```

versus (please put a special attention to the comparator)
```
if to.Uint64()-(to.Uint64()%uint64(params.CheckpointInterval)) < from.Uint64() {
    return
}
```


Of course, the place of guard in call side seems to stand out for readers, but let me ask where it's okay to go with.

hyeonLewis: It's a bit annoying to get the `from` and `to` values redundantly, but it seems better to do it on the caller side since this is an execution condition for the `regen` itself. It also helps not to call `regen` in most cases (`len(headers) < 2`)

ian0371: Then please add a comment above the function implementation indicating when this function becomes active.

hyunsooda: Added. PTAL.

consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,66 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+func (sb *backend) regen(chain consensus.ChainReader, headers []*types.Header) {
+	if !sb.isRestoringSnapshots.Load() && len(headers) > 1 {
+		var (
+			from        = headers[0].Number
+			to          = headers[len(headers)-1].Number
+			start       = time.Now()
+			commitTried = false
+		)
+
+		// Shortcut: No missing snaoshot data to be processed.
```
hyeonLewis, line 949:
```suggestion
		// Shortcut: No missing snapshot data to be processed.
```

consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,66 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+//       			   | header1, .. headerN |
+// Not triggered: (Guaranteed SI* was committed before )
+// |   ^                          ^                          ^                          ^  ...|
+//     SI                 SI*(last snapshot)                 SI                         SI
+//       			                 | header1, .. headerN |
+func (sb *backend) regen(chain consensus.ChainReader, headers []*types.Header) {
+	if !sb.isRestoringSnapshots.Load() && len(headers) > 1 {
+		var (
+			from        = headers[0].Number
+			to          = headers[len(headers)-1].Number
```
hyeonLewis, line 944: Isn't this better?
```suggestion
			from        = headers[0].Number.Uint64()
			to          = headers[len(headers)-1].Number.Uint64()
			// Fix below from, to accordingly
```

hyunsooda: Thanks! refactored.

consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,67 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
 	}
 
+	sb.regen(chain, headers)
+
 	sb.recents.Add(snap.Hash, snap)
 	return snap, err
 }
 
+// regen commits snapshot data to database
+// If the last header number is larger than the last snapshot number, regenerate handler gets triggered.
```
ian0371, line 931:
```suggestion
// regen is triggered if there is any checkpoint block in the `headers`.
// For each checkpoint block, this function verifies the existence of its snapshot in DB and stores one if missing.
```

I'm not sure how to read the diagram.. Does SI mean snapshot interval? Does it mean the last snapshot is in between header1 and headerN?

hyunsooda: SI = snapshot interval
last snapshot = latest snapshot interval number for the given header range (i.e., `to-(to%uint64(params.CheckpointInterval))`) 

```
 Triggered:
 |   ^                          ^                          ^                          ^  ...|
     SI                 SI*(last snapshot)                 SI                         SI
       			   | header1, .. headerN |
 Not triggered: (Guaranteed SI* was committed before )
 |   ^                          ^                          ^                          ^  ...|
     SI                 SI*(last snapshot)                 SI                         SI
                                  | header1, .. headerN |
```

Still uninterpretable?

ian0371: The last line was missing. Now i get it.

hyunsooda: Updated.

ian0371 (dismissed):

hyeonLewis (dismissed):

consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,69 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+ |   ^                          ^                          ^                          ^  ...|
+     SI                 SI*(last snapshot)                 SI                         SI
+       			   | header1, .. headerN |
+ Not triggered: (Guaranteed SI* was committed before )
+ |   ^                          ^                          ^                          ^  ...|
+     SI                 SI*(last snapshot)                 SI                         SI
+	                            | header1, .. headerN |
+*/
+func (sb *backend) regen(chain consensus.ChainReader, headers []*types.Header) {
+	if !sb.isRestoringSnapshots.Load() && len(headers) > 1 {
```
aidan-kwon, line 944: Minor comments related to the style. 
1. Reducing the call depth 
```
if sb.isRestoringSnapshots.Load() || len(headers) < 1 {  
    return 
}
```
BTW, is it okay to pass `len(headers) == 1` case in the original code? 

2. More safe operation for the atomic value change
`sb.isRestoringSnapshots.Load()` -> `sb.isRestoringSnapshots.CompareAndSwap(false, true)`

consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,69 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+
+		// Shortcut: No missing snapshot data to be processed.
+		if to-(to%uint64(params.CheckpointInterval)) < from {
+			return
+		}
+
+		sb.isRestoringSnapshots.Store(true)
+		defer func() {
+			sb.isRestoringSnapshots.Store(false)
+		}()
```
aidan-kwon, line 960: It would be better to print a log message informing the start of snapshot restoring

consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,69 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+				hh = header.Hash()
+			)
+			if params.IsCheckpointInterval(hn) {
+				snap, err := sb.snapshot(chain, hn, hh, nil, false)
+				if err != nil {
+					logger.Warn("[Snapshot] Snapshot restoring failed", "len(headers)", len(headers), "from", from, "to", to, "headerNumber", hn)
+					continue
+				}
+				// Store snapshot data if it was not committed before
+				if loadSnap, _ := sb.db.ReadIstanbulSnapshot(hh); loadSnap == nil {
```
yoomee1313, line 973: Can we check ReadIstanbulSnapshot first? `Snap` doesn't need to be generated if it was committed before.

hyunsooda: Okay, I'll work on that along with the feedback in the thread above.

consensus/istanbul/backend/engine.go
hyunsooda: Considering entry message should not be printed excessively, that message also should be printed [under the condition](https://github.com/klaytn/klaytn/pull/2074/files#diff-44abfea3bdbe16ce301a955732b485d1f0af2cc6c0c4cee3fbce0feb6ce051b0R981). Putting the initiating message seems redundant and not effective under this condition.

consensus/istanbul/backend/engine.go
hyunsooda: Thanks for the good suggestions.

First, I'll take your first suggestion. Once the suggestion is updated, the residual task will be as follows:
Whether to change `Store` to `CompareAndSwap`.
```
sb.isRestoringSnapshots.Store(true) // sb.isRestoringSnapshots.CompareAndSwap(false, true)
		defer func() {
			sb.isRestoringSnapshots.Store(false) // sb.isRestoringSnapshots.CompareAndSwap(true, false)
		}()
```
I think the existing code seems more intuitive and reliable.

Additionally, I couldn't understand what this implies.

> BTW, is it okay to pass len(headers) == 1 case in the original code?

Could you give a detail? BTW, I installed a guard from proceeding when the length of header is one because it was already handled before `regen` handler is called.

ian0371: I noticed before that load-store can have race conditions, but ignored because it's practically impossible to trigger.
But now that it's mentioned, safer design would be better.

compare-and-swap lock goes like this (it doesn't need load or store):

```go
lock := uint32(0)
if sb.isRestoringSnapshots.CompareAndSwap(false, true) {
  // enter critical section
}
```

Note that previous atomic load-store can have race conditions if two threads are running in parallel.
```
T1 - load and check
T2 - load and check
T1 - store
T2 - store
both threads enter critical section
```

hyunsooda: Possibly make sense. Thank you. Before making changes, let me ask one concern from your comment. Once the critical section executed, restoring `true` to `false` seems still required. Doesnt' it?

ian0371: Yep, you're right

hyunsooda: @ian0371 Updated. PTAL.
@aidan-kwon, I kept `len(headers) <= 1`. PTAL.

consensus/istanbul/backend/engine.go
hyunsooda: Updated. PTAL.

consensus/istanbul/backend/engine.go
```diff
@@ -921,10 +921,73 @@ func (sb *backend) snapshot(chain consensus.ChainReader, number uint64, hash com
...
+*/
+func (sb *backend) regen(chain consensus.ChainReader, headers []*types.Header) {
+	// Prevent nested call. Ignore header length one
+	// because it was handled before the `regen` called.
+	defer func() {
+		sb.isRestoringSnapshots.CompareAndSwap(true, false)
+	}()
+	if !sb.isRestoringSnapshots.CompareAndSwap(false, true) || len(headers) <= 1 {
+		return
+	}
```
ian0371, line 951: How about this?

```suggestion
	if !sb.isRestoringSnapshots.CompareAndSwap(false, true) || len(headers) <= 1 {
		return
	}
	defer func() {
		sb.isRestoringSnapshots.Store(true)
	}()
```

hyunsooda: You had a typo `sb.isRestoringSnapshots.Store(true)`? not `false`?

ian0371: oops it's a typo, `false` is right

hyunsooda: Updated.

ian0371 (approved):

hyeonLewis (approved):

blukat29:
@hyunsooda Please fix for linter or rebase against the latest dev.

Issue #2081, opened 2023-12-27
Result of `klay_estimateComputationCost`


jiOnederfull:
**Is your request related to a problem? Please describe.**
I set `--opcode-computation-cost-limit=200000000` then `klay_estimateComputationCost` is changed from `100000028` to `200000235`. I'm confused because it feels like klay_estimateComputationCost changes based on the opcode-computation-cost-limit setting.

**Describe the solution you'd like**
It would be better if klay_estimateComputationCost returned the actual value regardless of the opcode-computation-cost-limit or indicated if the computation exceeded the allowable limit.

hyunsooda:
Hi @jiOnederfull, Thanks for the issue posting.

I tend to favor the current behavior. Say I have a huge tx and want to test where it fits in which a specific range of computation cost allowed.

Try1. The tx failed at default computation cost (CC) `100000000`, and `klay_estimateComputationCost` returned `>100000000`
Try2. The tx failed at the modified CC `200000000` and `klay_estimateComputationCost` returned `>200000000`
Try3. The tx success at the modified CC `300000000` and `klay_estimateComputationCost` returned `>=200000000 x <=300000000`. 

If a developer gets the same value at Try2, he/she may think that this feature does not work. Please put additional input if there's another point of view you think. Thank you.

jiOnederfull:
Hi @hyunsooda, thanks for your reply.

So I believe that `klay_estimateComputationCost` should calculate and return the accurate value regardless of the `--opcode-computation-cost-limit`.

Let's assume the actual `klay_estimateComputationCost` is 210000000.

If we'd like to know this value, we would have to try up to try3 at the moment. Since, as far as I know, the current behavior of `klay_estimateComputationCost` is to stop the computation immediately and return the value if it exceeds the `--opcode-computation-cost-limit`. (So it will return 20000000+a(very small value).)

In that case, we would need to try all attempts you mentioned to obtain the accurate value.

If `klay_estimateComputationCost` returns the accurate value regardless of the `--opcode-computation-cost-limit`, we wouldn't need to attempt multiple tries.

hyunsooda:
Ah got it what you've been discouraged. In that case, setting `--opcode-computation-cost-limit=x`, where `x ~= infinite` would resolve it. Doesn'it?

For your last concern,
[...]

PR #2082, opened 2023-12-28
[API] Error response refined for the API `klay_estimateComputationCost`

hyunsooda:
## Proposed changes

Raised from #2081.

Added two values computation cost (CC) sum and cc limit.

For example, in a given environment with the setting `--opcode-computation-cost-limit=100000` the following error message returns.
```
Error: reached the opcode computation cost limit (opcodeComputationCostSum = 184311), (computationCostLimit = 100000)
```

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/api/api_public_blockchain.go
+++ b/api/api_public_blockchain.go
@@ -407,9 +407,8 @@ func (s *PublicBlockChainAPI) EstimateComputationCost(ctx context.Context, args 
 			return 0, err
 		}
 		rules := s.b.ChainConfig().Rules(block.Number())
-		errStr := fmt.Sprintf("%s (opcodeComputationCostSum = %d), (computationCostLimit = %d)",
+		errStr := fmt.Sprintf("%s (computationCostLimit = %d)",
 			vm.ErrOpcodeComputationCostLimitReached.Error(),
-			cc,
 			vm.GetComputationCost(rules, 0),
 		)
 		return cc, errors.New(errStr)
--- a/api/api_public_blockchain.go
+++ b/api/api_public_blockchain.go
@@ -398,8 +398,23 @@ func (s *PublicBlockChainAPI) EstimateComputationCost(ctx context.Context, args 
 	if rpcGasCap := s.b.RPCGasCap(); rpcGasCap != nil {
 		gasCap = rpcGasCap
 	}
-	_, computationCost, err := DoCall(ctx, s.b, args, blockNrOrHash, vm.Config{}, s.b.RPCEVMTimeout(), gasCap)
-	return (hexutil.Uint64)(computationCost), err
+	result, computationCost, err := DoCall(ctx, s.b, args, blockNrOrHash, vm.Config{}, s.b.RPCEVMTimeout(), gasCap)
+	cc := (hexutil.Uint64)(computationCost)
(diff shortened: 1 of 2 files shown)


api/api_public_blockchain.go
```diff
@@ -398,8 +398,23 @@ func (s *PublicBlockChainAPI) EstimateComputationCost(ctx context.Context, args
...
+	if err == nil && result.Unwrap() == vm.ErrOpcodeComputationCostLimitReached {
+		block, err := s.b.BlockByNumberOrHash(ctx, blockNrOrHash)
+		if err != nil {
+			return 0, err
+		}
+		rules := s.b.ChainConfig().Rules(block.Number())
+		errStr := fmt.Sprintf("%s (opcodeComputationCostSum = %d), (computationCostLimit = %d)",
+			vm.ErrOpcodeComputationCostLimitReached.Error(),
+			cc,
+			vm.GetComputationCost(rules, 0),
```
aidan-kwon, line 413: How about removing `opcodeComputationCostSum`? A user may think the real computation cost is the call is the same as `opcodeComputationCostSum`, but actually the call is stopped when it exceed the limit
```suggestion
		errStr := fmt.Sprintf("%s (computationCostLimit = %d)",
			vm.ErrOpcodeComputationCostLimitReached.Error(),
			vm.GetComputationCost(rules, 0),
```

api/api_public_blockchain.go
```diff
@@ -398,8 +398,23 @@ func (s *PublicBlockChainAPI) EstimateComputationCost(ctx context.Context, args
...
-	return (hexutil.Uint64)(computationCost), err
+	result, computationCost, err := DoCall(ctx, s.b, args, blockNrOrHash, vm.Config{}, s.b.RPCEVMTimeout(), gasCap)
+	cc := (hexutil.Uint64)(computationCost)
+
+	if err == nil && result.Unwrap() == vm.ErrOpcodeComputationCostLimitReached {
+		block, err := s.b.BlockByNumberOrHash(ctx, blockNrOrHash)
+		if err != nil {
+			return 0, err
+		}
+		rules := s.b.ChainConfig().Rules(block.Number())
```
aidan-kwon, line 409: When `blockNrOrHash` is a keyword like `"latest"`, the actual block number used in `DoCall()` can be different from the number used to get `rules`. The node can process more blocks during the execution of `DoCall()`.

It would not be an issue because the computation cost limit will not be changed frequently and this API is just for estimation. So, we can go with this code if providing consistency requires many changes. We can just type a comment, though.

api/api_public_blockchain.go
hyunsooda: Okay. updated.

api/api_public_blockchain.go
hyunsooda: For the first paragraph, if a proper block number is required, the user must pass the specific block number. Also, most of other APIs just take and proceed with the string parameters "latest", "pending", and "earlier". Could you give concrete view of your consideration? (why the problem matters only here )

blukat29:
Continued in #2086.

PR #2116, opened 2024-03-18
Fix gas tip related RPCs

2dvorak:
## Proposed changes

- This PR fixes some gas tip related RPCs to return correct values
   - Update `klay_gasPrice` to return `2 * BaseFee + SuggestedTip`
   - Update `klay_maxPriorityFeePerGas` to return `SuggestedTip`
   - Update `klay_feeHistory` to return correct values. For example, `gasUsedRatio` previously used `upperGasLimit`(`999999999999`) as denominator which is meaningless.
   - Update `SuggestTipCap` to suggest tip based on fee history
- See also
  - [KIP-162 Draft](https://github.com/blukat29/kips/blob/priority-fee-mechanism/KIPs/kip-162.md)
  - [Geth's API impl](https://github.com/ethereum/go-ethereum/blob/v1.13.14/internal/ethapi/api.go#L70-L90)
  - [Geth's GasPriceOracle impl](https://github.com/ethereum/go-ethereum/blob/v1.13.14/eth/gasprice/gasprice.go)

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist


hyunsooda (commented):
SGTM. Maybe you overlooked generating mock files for the newly defined test backend handler?

node/cn/gasprice/feehistory.go
```diff
@@ -101,8 +108,16 @@ func (oracle *Oracle) processBlock(bf *blockFees, percentiles []float64) {
 		bf.results.nextBaseFee = new(big.Int).SetUint64(params.ZeroBaseFee)
 	}
 
-	// There is no GasLimit in Klaytn, so it is enough to use pre-defined constant in api package as now.
-	bf.results.gasUsedRatio = float64(bf.header.GasUsed) / float64(params.UpperGasLimit)
+	// Use gasTarget as gasLimit for gasUsedRatio.
+	if chainconfig.IsMagmaForkEnabled(big.NewInt(int64(bf.blockNumber))) {
+		// TODO-Klaytn: Instead of using chainConfig, we should use governance set values.
+		if chainconfig.Governance.KIP71.GasTarget != 0 {
+			bf.results.gasUsedRatio = float64(bf.header.GasUsed) / float64(chainconfig.Governance.KIP71.GasTarget)
```
blukat29: 1. Please divide by MAX_BLOCK_GAS_USED_FOR_BASE_FEE as per https://github.com/blukat29/kips/blob/priority-fee-mechanism/KIPs/kip-162.md.
2. If the ratio exceeds 1, then set it to 1.

node/cn/gasprice/gasprice.go
```diff
@@ -83,16 +91,40 @@ func NewOracle(backend OracleBackend, params Config, txPool TxPool) *Oracle {
...
+	backend.SubscribeChainHeadEvent(headEvent)
+	go func() {
+		var lastHead common.Hash
+		for ev := range headEvent {
+			if ev.Block.ParentHash() != lastHead {
+				cache.Purge()
+			}
+			lastHead = ev.Block.Hash()
+		}
+	}()
```
blukat29: If the subscription is for cache invalidation, we can simply listen to SetHead instead. I suspect that geth uses subscription because of the chain fork (a.k.a. [reorg](https://www.alchemy.com/overviews/what-is-a-reorg)) possibility (see [geth commit](https://github.com/ethereum/go-ethereum/pull/23422/commits/27ef4000cf13c9f85a569c088d20868ef56a08a8)). But Klaytn has no reorg, so it's sufficient to invalidate cache only when SetHead.

We can modify https://github.com/klaytn/klaytn/blob/dev/node/cn/api_backend.go
```go
func (b *CNAPIBackend) SetHead(number uint64) error {
  return doSetHead(b.cn.blockchain, b.cn.engine, b.cn.governance, b.gpo, number)
}

func doSetHead(..) {
  gpo.PurgeCache()
}
```

2dvorak: Thanks for the suggestion, seems like it's much better approach.
Done in ba1ea0e.

node/cn/gasprice/feehistory.go
2dvorak: Done, thanks! 27fec24

node/cn/gasprice/gasprice.go
```diff
@@ -155,3 +186,133 @@ func (gpo *Oracle) SuggestTipCap(ctx context.Context) (*big.Int, error) {
...
+			number--
+		}
+		results = append(results, res.values...)
+	}
+	price := lastPrice
+	if len(results) > 0 {
+		slices.SortFunc(results, func(a, b *big.Int) int { return a.Cmp(b) })
+		price = results[(len(results)-1)*oracle.percentile/100]
+	}
+	if price.Cmp(maxPrice) > 0 {
```
hyunsooda: How about refining the `maxPrice` value by considering its upper bound basefee as the max price?

node/cn/gasprice/gasprice.go
```diff
@@ -155,3 +186,133 @@ func (gpo *Oracle) SuggestTipCap(ctx context.Context) (*big.Int, error) {
...
+
+type results struct {
+	values []*big.Int
+	err    error
+}
+
+// getBlockValues calculates the lowest transaction gas price in a given block
+// and sends it to the result channel. If the block is empty or all transactions
+// are sent by the miner itself(it doesn't make any sense to include this kind of
+// transaction prices for sampling), nil gasprice is returned.
```
hyunsooda: It returns a list of TX gas, not a single one. Does not need to be refined the description?

node/cn/gasprice/gasprice.go
2dvorak: @hyunsooda First, thanks for the suggestion.

1. Using `upperBoundBaseFee` for `maxPrice`
   The `maxPrice` is from geth, and it's for upper bound of the suggested priority fee (and suggested gasPrice before London HF). So it is not related to the upper bound of the dynamic basefee.

   However, we could use `upperBoundBaseFee` for the `maxPrice` - to limit the suggested priority fee never exceed the base fee?

2. Why use `maxPrice`?
   I've looked into geth code for reasons to introduce a upper bound for suggested priority fee, but could not find out. Seems like such concept was introduced a [long time ago](https://github.com/ethereum/go-ethereum/pull/990) when gas price oracle was first added, but couldn't find backgrounds for the upper limit.

   Instead of finding a suitable value or governance params for the `maxPrice`, why don't we just allow node operators to set `maxPrice` with cmd flags like [geth](https://github.com/ethereum/go-ethereum/blob/7bb3fb1481acbffd91afe19f802c29b1ae6ea60c/eth/gasprice/gasprice.go#L96) does? I couldn't find reasons to introduce `maxPrice` but also not having reasons not to use `maxPrice`. Since we are trying to enhance compatibility and improve UX for users familiar eth, we could bring such flag too.

blukat29: But the priority fee is unrelated to the base fee. It's the base fee that is upper bounded. Priority fee has no limitations. Having a constant maxPrice is good enough to filter out any arithmetical errors (e.g. divide by very small number, statistical outlier when there are very few transactions).

2dvorak: @blukat29 Sure, then I'll just leave `maxPrice` as constant value (500 ston) if @hyunsooda does not have any further comments.

In case we want to support node operators to change that value with cmd flag, we could make another PR for that change.

hyunsooda: Yes, they are not relevant, but I'm concerned about potential error cases where the suggested values return inaccurately due to the reliance on a hardcoded value, even when the correct answer should be higher

Having the same flag the Geth has sounded good to me, but it's okay, too, if you see that it's not mandatory.

blukat29: @hyunsooda @2dvorak I'm also good with adding the command line flags like in geth. gpo.blocks, gpo.percentile, gpo.maxprice. https://github.com/ethereum/go-ethereum/blob/v1.13.14/cmd/utils/flags.go#L791

hyunsooda (dismissed):

node/cn/gasprice/gasprice.go
2dvorak: OK, I'll add command line flags for GPO configs.

node/cn/gasprice/gasprice.go
2dvorak: Updated, PTAL.

node/cn/gasprice/gasprice.go
```diff
@@ -35,14 +35,15 @@ import (
 
 const sampleNumber = 3 // Number of transactions sampled in a block
 
-var maxPrice = big.NewInt(500 * params.Ston)
+var DefaultMaxPrice = big.NewInt(500 * params.Ston)
```
blukat29: Let's remove from here, and have the constant in GetDefaultConfig.

node/cn/gasprice/gasprice.go
```diff
@@ -252,8 +259,12 @@ func (oracle *Oracle) suggestTipCapUsingFeeHistory(ctx context.Context) (*big.In
 		slices.SortFunc(results, func(a, b *big.Int) int { return a.Cmp(b) })
 		price = results[(len(results)-1)*oracle.percentile/100]
 	}
-	if price.Cmp(maxPrice) > 0 {
-		price = new(big.Int).Set(maxPrice)
+	// NOTE: This upper bound for suggested gas tip can lead to suggesting insufficient gas tip,
+	//       however the possibility of gas tip exceeding 500 ston would be very low.
+	//       On the other hand, having no upper bount for gas tip can lead to suggesting
+	//       very high gas tip when there are only few transactions with unnecessarily high gas tip.
```
blukat29:
```suggestion
	// NOTE: This maximum suggested gas tip can lead to suggesting insufficient gas tip,
	//       however, the possibility of gas tip exceeding 500 ston would be very low given the block capacity of Klaytn.
	//       On the other hand, referencing the user-submitted transactions as-is can lead to suggesting
	//       very high gas tip when there are only a few transactions with unnecessarily high gas tip.
```

blukat29 (commented):
Once you wire the command line flags to GPO params, I am okay.

node/cn/gasprice/gasprice.go
2dvorak: @blukat29 Currently, we are using `DefaultMaxPrice` in `NewOracle()` to sanitize GPO config parameter `MaxPrice` if it is not properly set.  If `DefaultMaxPrice` is moved to `node/cn/config.go`. we cannot get that value in `NewOracle()` because of cyclic import. We should either 1) move this constant to `GetDefaultConfig` and not sanitize GPO config in `NewOracle` or 2) keep `DefaultMaxPrice` here.

2dvorak: @blukat29 Moved `DefaultMaxPrice` to params package, to remove cyclic import.
See 1ac015e

hyunsooda: How about removing `DefaultMaxPrice` and leave the constant at `GetDefaultConfig()`? If it is referenced multiple times, we may simply type long such as `GetDefaultConfig().GPO.MaxPrice`.

2dvorak: @hyunsooda Then we cannot reference that value in `node/cn/gasprice.go`, for example in `NewOracle()`, because of cyclic import. Do you want me to remove sanitizing `config.maxPrice` in `NewOracle()`?

hyunsooda: Okay, then the latest commit would be better.

hyunsooda (approved):

blukat29:
@2dvorak Please rebase against the latest dev to resolve test-linter.

blukat29 (dismissed):

hyunsooda (dismissed):

node/cn/gasprice/feehistory.go
```diff
@@ -228,6 +245,10 @@ func (oracle *Oracle) FeeHistory(
 		next    = oldestBlock
 		results = make(chan *blockFees, blocks)
 	)
+	percentileKey := make([]byte, 8*len(rewardPercentiles))
+	for i, p := range rewardPercentiles {
+		binary.LittleEndian.PutUint64(percentileKey[i*8:(i+1)*8], math.Float64bits(p))
+	}
 	for i := 0; i < maxBlockFetchers && i < blocks; i++ {
```
hyeonLewis, line 252: Is there any reason to keep `maxBlockFetchers == 1` which is `4` in Ethereum?

2dvorak: I don't see any. I've increased `maxBlockFetchers` to `4`

hyeonLewis (approved):

blukat29 (approved):

PR #2151, opened 2024-05-08
[KAIABridge] Added KAIABridge contracts

hyunsooda:
## Proposed changes

KAIABridge contracts and unit tests were added

@ian0371, @hyeonLewis  Could you please review this implementation?

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


hyeonLewis (commented):
@hyunsooda Do you manually bring `XXXUpgradable` since adding `contracts-upgradeable@5.0` package is broken due to missing peer dependency? They're currently importing lower version, which is strongly discouraged. I think we need to import `contracts-upgradeable@5.0` if we want to keep maintaining all contracts in klayn (kaia).

contracts/contracts/system_contracts/kaiabridge/EnumerableSetUpgradable.sol
```diff
@@ -0,0 +1,75 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the klaytn library. If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.24;
+
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+contract EnumerableSetUpgradable is Initializable {
```
hyeonLewis: Why did you use `contract` for `EnumerableSetUpgrdable`? You can directly use `openzeppelin-contracts-5.0/EnumerableSet.sol` since it's a library. I think implementing custom methods `getAll, getRange` in a wrapper library (if needed by many contracts) will be better.

contracts/contracts/system_contracts/kaiabridge/IGuardian.sol
```diff
@@ -0,0 +1,171 @@
...
+    /// @return Returns array of transaction IDs.
+    function getTransactionIds(uint256 from, uint256 to, bool pending, bool executed)
+        external
+        virtual
+        view
+        returns (uint256[] memory, uint256[] memory);
+
+
+    //////////////////// Storage variables ////////////////////
+    address[] public guardians;
```
hyeonLewis, line 163: Why don't we use `EnumerableSet.AddressSet` for `address[]` type? If they're `set` and need to be added/removed, then using `EnumerableSet` is more efficient.

contracts/contracts/system_contracts/kaiabridge/IBridge.sol
```diff
@@ -0,0 +1,286 @@
...
+    event ReleaseClaim(uint256 indexed seq, uint256 indexed time);
+
+    /// @dev Emitted when bridge is paused
+    event BridgePause(string indexed msg);
+
+    /// @dev Emitted when bridge is resumed
+    event BridgeResume(string indexed msg);
+
+    //////////////////// Exported functions ////////////////////
+    /// @dev A gateway function that triggers tokwn swap between FNSA and KLAY
```
ian0371:
```suggestion
    /// @dev A gateway function that triggers token swap between FNSA and KAIA
```

contracts/contracts/system_contracts/kaiabridge/Judge.sol
```diff
@@ -0,0 +1,340 @@
...
+        guardian = initGuardian;
+
+        // Fill the first index(0) with dummy tx
+        addTransaction(address(0), "", 0);
+
+        __Ownable_init();
+        __UUPSUpgradeable_init();
+    }
+
+    function _authorizeUpgrade(address newImplementation) internal virtual override onlyOwner {}
```
ian0371: Should this be `onlyGuardian`? Or is `owner` a separate entity who has the highest trust like guardian?

contracts/contracts/system_contracts/kaiabridge/Bridge.sol
```diff
@@ -0,0 +1,400 @@
...
+        // Allocate half of gas as available gas for the fallback code
+        (bool sent, ) = prov.receiver.call{
+            value: prov.amount,
+            gas: gasleft() / 2
+        }("");
+        if (!sent) {
+            if (revertOnFail) {
+                revert("PDT::Bridge: Failed to transfer amount of provision");
+            }
+            if (transferFail[prov.seq]++ > maxTryTransfer) {
```
ian0371: This allows users to try `maxTryTransfer + 2` claims at most. Was this your intention?
If not, to remove the provision at `maxTryTransfer + 1`-th `claim` call,
```suggestion
            transferFail[prov.seq]++;
            if (transferFail[prov.seq] > maxTryTransfer) {
```

blukat29:
Hello, could you either (1) rebase against the latest dev or (2) `git fetch --all && git cherry-pick 946bfe6`? This will fix the CI failure.

contracts/contracts/system_contracts/kaiabridge/EnumerableSetUpgradable.sol
hyunsooda: I wanted to wrap the contract as upgradable version. That's why I created a new dedicated contract to achieve upgradable.

contracts/contracts/system_contracts/kaiabridge/IGuardian.sol
hyunsooda: I thought the primitive array seems more intuitive and simple and seems better suited to covering the operations related to those types.

contracts/contracts/system_contracts/kaiabridge/Judge.sol
hyunsooda: Very good point. Currently, the owner only has the control to upgrade a contract. 
There are two options:
1. Use the EoA account as an owner which is controlled by multisig (not a multisig scheme in the bridge contract)
2. Use guardian authorization. However this is somewhat challenge because the contract calls this `__authorizeUpgrade()`, but the client side just only calls `upgrades.upgradeProxy()`.

Please advise me.

contracts/contracts/system_contracts/kaiabridge/Bridge.sol
hyunsooda: Your suggestion is the correct one. I fixed this in another repository. I'll update all together later.

contracts/contracts/system_contracts/kaiabridge/Judge.sol
ian0371: The upgradeable permission is so powerful that it can simply disable all on-chain access control. Thus, if you haven't discussed the existence of an entity which has the utmost trust higher than guardian, and haven't discussed how it should be managed, then I think guardian should have the upgradeable feature.

Guardian contract can directly invoke [Bridge.upgradeToAndCall(newImpl)](https://docs.openzeppelin.com/contracts/5.x/api/proxy#UUPSUpgradeable-upgradeToAndCall-address-bytes-), without using OZ's `upgrades` library. If you're worried about missing the validations of `upgrades.upgradeProxy()` (storage validation, etc.), I think you can [manually validate](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades#validate-implementation).

In addition, I just noticed that Guardian also has an owner. I believe Guardian's `_authorizeUpgrade()` should be `onlyGuardian` as well.

contracts/contracts/system_contracts/kaiabridge/EnumerableSetUpgradable.sol
hyeonLewis: Sorry for late reply.

When upgrading a contract, it's safe to replace library with additional function since pure library doesn't affect storage layout of contract at all. But if you expect to add some variables to library, then I think you need to add gap since it's inherited by other abstract contracts.

contracts/contracts/system_contracts/kaiabridge/IGuardian.sol
hyeonLewis: Yes, but I can't find operations better suited to array. 
For example, adding/removing elems from arrays is more efficient in EnumerableSet.

Could you please give me a simple example? Thanks.

contracts/test/KAIABridge/util.ts
```diff
@@ -0,0 +1,20 @@
+export const DAY = 24 * 60 * 60;
```
hyeonLewis: You can use `common/helpers.ts` for these helper functions.

```
setBlockTimestamp == addTime
getBalance == getBalance
getCurrentBlockTimestamp == nowTime
```

contracts/contracts/system_contracts/kaiabridge/Bridge.sol
```diff
@@ -0,0 +1,400 @@
...
+        uint64 cnt = 0;
+        uint64[] memory temp = new uint64[](n);
+
+        for (uint64 i=from; i<to; i++) {
+            if (isProvisioned(i)) {
+                temp[cnt++] = i;
+            }
+        }
+        // fitting
+        uint64[] memory provisionedRanges = new uint64[](cnt);
```
hyeonLewis, line 350: You can resize array directly for optimizations. (applicable to other functions)

```
uint64[] memory provisionedRanges = new uint64[](n);
uint64 cnt = 0;
for (uint64 i=from; i<to; i++) {
    if (isProvisioned(i)) {
        provisionedRanges[cnt++] = i;
     }
}

assembly {
    mstore(provisionedRanges, cnt)
}
```

contracts/contracts/system_contracts/kaiabridge/Operator.sol
```diff
@@ -0,0 +1,488 @@
...
+        public
+        override
+        returns (uint64)
+    {
+        require(data.length >= 4, "Calldata length must be larger than 4bytes");
+        uint64 seq = 0;
+        uint64 txID = 0;
+
+        try IBridge(bridge).bytes2Provision(data) returns (IBridge.ProvisionData memory provision) {
+            require(bridge == to, "PDT::Operator: Provision transaction must be targeted to knwon bridge contract address");
```
hyeonLewis:
```suggestion
            require(bridge == to, "PDT::Operator: Provision transaction must be targeted to known bridge contract address");
```

contracts/contracts/system_contracts/kaiabridge/Operator.sol
```diff
@@ -0,0 +1,488 @@
...
+
```
hyeonLewis, line 158: Don't we need to check `provision.seq > 0` here?

contracts/contracts/system_contracts/kaiabridge/Operator.sol
```diff
@@ -0,0 +1,488 @@
...
+        public
+        override
+        onlyGuardian
+        validRequirement(operators.length, _minOperatorRequiredConfirm)
+    {
+        minOperatorRequiredConfirm = _minOperatorRequiredConfirm;
+        emit RequirementChange(_minOperatorRequiredConfirm);
+    }
+
+    function submitTransaction(address to, bytes calldata data, uint256 uniqUserTxIndex)
```
hyeonLewis, line 146: Is it ok to allow `uniqUserTxIndex == 0`? I don't get the point of `submission2TxID`. If it's for searching the txId by `uniqUserTxIndex` for the user, then why don't we have something like `mapping(address => uint256) user2UniqIdx`?

contracts/contracts/system_contracts/kaiabridge/EnumerableSetUpgradable.sol
hyeonLewis: Sorry for late reply.

When upgrading a contract, it's safe to replace library with additional function since pure library doesn't affect storage layout of contract at all. But if you expect to add some variables to library, then I think it needs a gap since it's inherited by other abstract contracts.

contracts/contracts/system_contracts/kaiabridge/IGuardian.sol
hyeonLewis: Yes, I agree that it's more intuitive, but I think EnumerableSet would be more efficient for operations it needed. (e.g., adding/removing elems)

If it's a little tight to change the implementation now, I think it's fine as is, thank you.

contracts/contracts/system_contracts/kaiabridge/Bridge.sol
hyunsooda: Good suggestion. However, this will increase the bytecode size. Since these are used for getter purposes, how about leaving them as they are?

hyunsooda:
@ian0371, @hyeonLewis Updated based on your valuable comments. Please close if each issue is correctly patched

contracts/contracts/system_contracts/kaiabridge/Operator.sol
hyunsooda: The purpose of this storage is the correct one you found. If this explicit conversion is not necessary, the parameter value should be zero.

contracts/contracts/system_contracts/kaiabridge/Bridge.sol
hyeonLewis: Oh really? I thought it would decrease the bytecode size. Anyway, this is optimization, so let it be.

contracts/test/KAIABridge/util.ts
hyeonLewis: @hyunsooda Could you use `helpers.ts` instead of making duplicate `utils.ts`?

hyunsooda: Updated. PTAL.

ian0371 (approved):
please rebase

hyeonLewis (approved):

blukat29:
@hyunsooda Do not merge the dev. Please rebase against the latest dev.

PR #2160, opened 2024-05-13
Handle staking info since kaia fork in downloader.

hyeonLewis:
## Proposed changes

After kaia fork, the staking info for block `N` is determined at block `N-1`. Also, staking info is no longer stored in DB after kaia fork. This breaks current implementation of downloader to fetch staking info to verify headers (especially in snap sync). In this PR, all necessary staking info for block verification even after kaia fork will be requested to peer and stored in DB temporary. The stored staking info after kaia fork will be deleted after committing blocks to blockchain.

When kaia fork block is `N` and downloader tries to fetch staking info from `N-10` to `N+10` when staking interval is 4:
| Sync state | Fetched staking info | Stored staking info |
|--|--|--|
| While syncing | `N-10`, `N-6`, `N-2`, `N-1`, `N`, `N+1`, ... `N+9` | All 14 fetched |
| After sync | - |  `N-10`, `N-6`, `N-2` |

It enables snap sync which was disabled at #2154.

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

*Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.*

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/klaytn/klaytn/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/klaytn/klaytn)
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #2182, opened 2024-05-21
Revert "Handle staking info since kaia fork in downloader."

hyeonLewis:
Reverts klaytn/klaytn#2160

Initially, I mis-interpreted the testing results so I merged it. Sorry for the confusion.

During several snap sync tests, it was found that #2160 can't solve the block header verification issue correctly. The block header initially comes to `processHeaders`, which inserts the header into the header chain first. Then, the contents for fetched headers will be scheduled by `Schedule`, which includes staking info. So, the staking info needs to be fetched with headers and ready before `insertHeaderChain`. But the current implementation fetches staking info after `insertHeaderChain`, which means it does nothing. (It worked before kaia fork since it required staking info before staking update interval)

Additionally, it seems that `randao` also breaks snap sync since it requires state at a block to retrieve the BLS registry. https://github.com/klaytn/klaytn/blob/dev/consensus/istanbul/backend/randao.go#L63

To sum up, I concluded it would be better to consider all the cases carefully to fix the snap sync. If you have suggestions, please let me know. Thanks.

### DIFF
--- a/datasync/downloader/downloader.go
+++ b/datasync/downloader/downloader.go
@@ -396,18 +396,7 @@ func (d *Downloader) synchronise(id string, hash common.Hash, td *big.Int, mode 
 	if !atomic.CompareAndSwapInt32(&d.synchronising, 0, 1) {
 		return errBusy
 	}
-	defer func() {
-		atomic.StoreInt32(&d.synchronising, 0)
-		// Staking info for kaia block has been temporarily stored in DB for post-download VerifyHeader.
-		// Now that the download and sync are complete, remove staking info for kaia block from DB
-		// to save space. They might stay in the cache though.
-		for _, blockNum := range d.queue.stakingInfoStoredBlocks {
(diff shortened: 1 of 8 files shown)


blukat29:
I'm okay with turning off the snapsync again. It was our first plan wasn't it?
Let's sit back and try other solutions over time. Such as skipping header verification for the pivot block? which would require additional argument all the way through headerchain though.

blukat29 (approved):

hyeonLewis:
@blukat29 Thanks for your comment. Yes, disabling snap sync was our first plan. 

We actually need to skip all the blocks inserted via `insertHeaderChain`, not only pivot block. So I think we have two options:
1. Skip verifying headers, and verify it after fetching staking info.
2. Adjust fetching timing of staking info after Kaia fork.

As I mentioned, we must include the bls registry (+ temporal DB for BLS registry) fetcher.

ian0371 (approved):

